PR #1, opened 2012-06-08
Plugins

morevnaproject:
= Summary =
This modification allows to run custom python scripts directly from Synfig Studio menu. Each script takes .sif file as first argument and should modify its contents in some way. After script execution finished the file is automatically reopened back in Synfig Studio. 

= Rationale =
People often write some scripts to make useful things on Synfig (sif) files. The most of these scripts are written in python. But for ordinary users running custom scripts from terminal is tricky. With plugins feature user can install scripts as easy as they copy files and transparently run them in the same way as they use standard Synfig Studio commands. Also, runing scripts from menu is much faster than from terminal and it greatly improves the workflow for advanced users. Having this feature allows to easily add simple functions to Synfig Studio by writting scripts in python.

= Details =

All plugins are located in the "Plug-Ins" submenu of the canvas.

Plugins are read from two places:
- System location: INSTALL_PREFIX/share/synfig/plugins
- User location: USER_DIRECTORY/SYNFIG_CONFIGURATION_DIR/plugins
  Each plugin located in a separate subdirectory with unique name. The part of the name before first "-" symbol is used to set thegroup plugin belongs to (not implemented yet). The main information about plugin (plugins name and script to execute) is stored in the plugin.xml file. It's self-explanatory.

There is a sample plugin to add simple skeleton (stickman) into current document.

All scripts are interpreted with python. That means that we should bundle python into Windows build. For windows case Python binary is expected at INSTALL_PREFIX/python/python.exe.

New env variable SYNFIG_PYTHON_BINARY allows to set custom path to the python binary.

= ToDo List =
- Add support for localizing script names (using intltool)
- Group plugins into different submenus of "Plug-Ins" menu
- Add more plugins

P.S. Apologies if my code is crappy. ^__^"

### DIFF
--- a/synfig-studio/src/synfigapp/main.cpp
+++ b/synfig-studio/src/synfigapp/main.cpp
@@ -441,6 +441,5 @@ synfigapp::Main::get_user_app_directory()
 	}
 #endif
 	dir = dir+ETL_DIRECTORY_SEPARATOR+SYNFIG_USER_APP_DIR;
-	synfig::info("User app dir = %s", dir.c_str());
 	return dir;
 }


MarcWeber:
Fix regression in master: http://mawercer.de/tmp/0001-Fix-regression-in-running-plugins.patch
On irc I was told that plugins are not worth it - maybe there should be some additional demo plugins.
Anyway with this patch applied synfig master no longer crashes and actually shows the bones.

caryoscelus:
>It should be save to assume that adding .backup.sif is unique ..

Why would you think it's safe? What if user creates their own .backup.sif files?

Also commit message has "sfi" instead of "sif".

I can apply your patch with that bit reverted or you can fix it yourself.

Also, next time please open separate issue / pull request instead of writing into ancient thread.

MarcWeber:
@caryoscelus => the .backup files are placed  in ~/.config/synfig/tmp/.... I cannot imagine the user placing files there.
Just revert that bit. As you've said: There are more important things to work on. For instance undo history could be changed so that it allows "replacing the xml" so that undo button still works.

caryoscelus:
>the .backup files are placed in ~/.config/synfig/tmp/...

Unless `is_absolute_path(filename_original)` is true, whatever that means.

Anyway, pushed fixed commit in master.

PR #137, opened 2014-04-30
Renderer keyframe

genete:
This branch does the following things:

- Renames Timecode to Keyframe in classes and files, since the timecode is not being displayed on canvas anymore. It has its own widget.
- Allow to show the previous and next keyframes relative to the current time cursor as well as the keyframe behind the time cursor when there is.  
- The keyframe's name below time cursor is shown at the top center of canvas enclosed by << >> to enlighten it
- The previous and next keyframes are places on the top left and rigth zones in a relative position from the rendering area boundaries, to express how far or near is the previous or the next keyframe to where the time cursor is on.

Feel free to request corrections or simply don't merge this pull. I coded it in believe that it could help some users.


morevnaproject:
Closing as outdated.

PR #150, opened 2014-12-10
examples repacked .sfg

horkana:
examples/sfg subdirectory

inside each example .sfg is the same file with the extension .sif
all files compressed using Zip normal compression (deflate)

hopefully they can replace the existing examples in due course. 
(previous .sifz examples were compressed using Gzip compression, and inside some files did not have any file extension, some had the file extension .sif, and some had the extension .sifz)


ankit-kumar-dwivedi:
Is this still relevant @ice0 ?

morevnaproject:
In fact, I prefer examples to be stored uncompressed. PR #1023 takes care about that.

PR #161, opened 2015-01-11
Fixed gradient parsing in SVG module.

mrtrizer:
I had problems when I tried to import svg file, saved with Inkscape.
Inkscape uses another link format.
xlink:href="#linearGradient9319"
instead
href="#linearGradient9319"


morevnaproject:
Merged. Thank you!

PR #163, opened 2015-01-14
Spritesheet rendering improvement.

mrtrizer:
I want use synfig to make game-animation and need a spritesheet generator. 
So, I have added parameters dialog to the render window.
You can change several parameters:
- Offset X
- Offset Y
- Row count
- Columns count

New sprites are added into an existing file if it is. Otherwice new file will be created.


morevnaproject:
Merged. ^__^

PR #168, opened 2015-02-08
Fix 788 - Crash - Open File , click history

d-j-a-y:
Connect FileChooserDialog->signal_selection_changed() to a global function
be cause this ptr is not available from a static member (needed with sigc::mem_fun)


morevnaproject:
Merged.

PR #170, opened 2015-03-14
Move valuenodes into synfig/valuenodes folder

eldruin:
Enhancement #825 
Move all the valuenode_* files into synfig/valuenodes folder.
On install, these files will be installed under synfig/valuenodes so to include them from synfig-studio, for example, the paths now should be #include <synfig/valuenodes/valuenode_x.h>
Regarding the build system, the same technique using Makefile_insert like in synfig-studio has been used.


eldruin:
I have created a superseeding PR: https://github.com/synfig/synfig/pull/171

PR #171, opened 2015-03-14
File reorganization

eldruin:
Enhancement #825
Move all the valuenode_* files into synfig/valuenodes folder, layer_* into synfig/layers, etc
On install, these files will be installed under synfig/valuenodes, etc so to include them from synfig-studio, for example, the paths now should be, for example, synfig/valuenodes/valuenode_x.h
Regarding the build system, the same technique using Makefile_insert like in synfig-studio has been used.


morevnaproject:
I have made some fixes to that branch and pushed my changes here - https://github.com/morevnaproject/synfig/tree/file-reorganization
I'm afraid that this change will affect a lot of translated strings and we will end up with broken translations. So we have to postpone this change until the time when 1.0 will be released.

morevnaproject:
Merged to master branch. Thank you!

PR #175, opened 2015-03-31
Fix 568 : CRLF in lst file

d-j-a-y:
add FileSystem::safeGetline and use it in both listimporter and cairolistimporter


morevnaproject:
Merged. Thank you!

PR #176, opened 2015-04-02
Fix 754 : Onion skin broken ++

d-j-a-y:
* unhide future and past spin buttons ... and now working has expected.
(onion skin was not broken, just working with nothing to display!)

* add ONION_SKIN_FUTURE and ONION_SKIN_PAST #definition

* add "onion_skin_future" and "onion_skin_past" metadata entries
('full' onion skin state is now save&restored between working session)


morevnaproject:
Merged. Thank you!

PR #182, opened 2015-04-17
Fix 856 - 	Guide, can't remove dragging them back onto the canvas window...

d-j-a-y:
... rulers

* check content of pointer (not pointer)

NOTA : maybe this bug is also present in Renderer_Guides::event_vfunc

	// Calculate the position of the
// input device in canvas coordinates
// and the buttons
if(!event->button.axes)  // <<<--- HERE
{

https://github.com/synfig/synfig/blob/master/synfig-studio/src/gui/workarearenderer/renderer_guides.cpp#L133


morevnaproject:
Good catch!

PR #192, opened 2015-07-20
Fix #766: on and off states of a "Switch" node were linked by default

Permutatrix:
This is a fix for [this bug here](http://www.synfig.org/issues/thebuggenie/synfig/issues/766). When you made a "Switch" node, the "Link On" and "Link Off" parameters were linked by default, making it necessary to disconnect them before using it.


morevnaproject:
Merged. Will be delivered with version 1.0.2. Thank you!

PR #193, opened 2015-07-21
Make Time parameters convertible to Real

Permutatrix:
This patch simply adds the option to Convert a Time parameter to "Real". The value given is interpreted as a length of time in seconds.

As far as I know, it was only formerly possible to do this by converting to an Integer first, which has one obvious disadvantage... it's only accurate to the second.


morevnaproject:
Merged. Will be delivered with version 1.0.2. Thank you!

PR #194, opened 2015-07-21
Add time dilation for PasteCanvases/Groups

Permutatrix:
This patch adds a "Time Dilation" parameter to Group layers, allowing speed of playback to be adjusted just like offset.

It was already possible to achieve this effect by converting the time offset to "Linear" and taking into account the natural progression of time, but that solution always seemed clunky to me, and it can produce some weird output in the editing preview.

Also, I've found it to be useful to use an exported canvas's timeline to select a pose rather than as an actual time. By setting the Group layer's time dilation to zero, this can be done much more easily.

### DIFF
--- a/synfig-core/src/synfig/layers/layer_pastecanvas.cpp
+++ b/synfig-core/src/synfig/layers/layer_pastecanvas.cpp
@@ -81,6 +81,7 @@ public:
 Layer_PasteCanvas::Layer_PasteCanvas():
 	param_origin(Point()),
 	param_transformation(Transformation()),
+	param_time_dilation (Real(1)),
 	param_time_offset (Time(0)),
 	depth(0),
 	extra_reference(false)
@@ -136,6 +137,11 @@ Layer_PasteCanvas::get_param_vocab()const
 	ret.push_back(ParamDesc("canvas")
 		.set_local_name(_("Canvas"))
 		.set_description(_("Group content"))
+	);
+
+	ret.push_back(ParamDesc("time_dilation")
+		.set_local_name(_("Time Dilation"))
+		.set_description(_("Multiplier to speed up, slow down, freeze, or reverse time"))
 	);
 
 	ret.push_back(ParamDesc("time_offset")
@@ -204,6 +210,7 @@ Layer_PasteCanvas::set_param(const String & param, const ValueBase &value)
 		return true;
 	}
 #else
+	IMPORT_VALUE(param_time_dilation);
 	IMPORT_VALUE(param_time_offset);
 #endif
 
@@ -291,6 +298,7 @@ Layer_PasteCanvas::get_param(const String& param)const
 		synfig::ValueBase ret(canvas);
 		return ret;
 	}
+	EXPORT_VALUE(param_time_dilation);
 	EXPORT_VALUE(param_time_offset);
 	EXPORT_VALUE(param_children_lock);
 	EXPORT_VALUE(param_curr_time);
@@ -302,6 +310,7 @@ Layer_PasteCanvas::get_param(const String& param)const
 void
 Layer_PasteCanvas::set_time(IndependentContext context, Time time)const
 {
+	Real time_dilation=param_time_dilation.get(Real());
 	Time time_offset=param_time_offset.get(Time());
 
 	if(depth==MAX_DEPTH)return;depth_counter counter(depth);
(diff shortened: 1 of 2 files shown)


morevnaproject:
The improvement is on review. Thank you!

morevnaproject:
The improvement is merged into master branch. I have renamed "Time Dialation" to "Speed" since it's easier to understand for most audience. Thank you!

morevnaproject:
@Permutatrix Line 644 have division by zero - https://github.com/morevnaproject/synfig/commit/e9d9cb0819c84801501f4b1ae29723e204094fdf#diff-8fd40e8e111b4ae5d0d2e329bef11934R664
Would be nice to fix this. ^__^

Permutatrix:
Right... I'm a bit confused as to what that function is _for_, actually. It doesn't seem to be called anywhere at the moment, and the docstring isn't very helpful, so I'm not sure what the best behavior would be when time_dilation is zero. Also, I don't seem to have a way to test it.
So, I'll need a bit of clarification on the matter, but after that, it should be an easy fix.

morevnaproject:
This piece of code should update the display of waypoints according to the applied time offset (and now also the time_dilation). Unfortunately it doesn't works as expected (there is a bug) - the group doesn't displays the offset for waypoints. But if you assign a time_offset/time_dilation to group and then put it into another group, then you can see the waypoints are displayed shifted/stretched (when you have parent group selected).

In any case, let's think about expected behavior.  In minimal case, if time_dilation is zero, then it's OK to ignore it and just write `tp.set_time(tp.get_time() - time_offset);` The best solution would be to make  waypoints move together with time cursor. As if they are "sticked" to it.

P.S. It would be nice if you specify your real name for Git commits:
`git config --global user.name "Your Name"`

Just for mention in the list of contributors. ^__^

morevnaproject:
Oh, I have fixed the issue with time offset - https://github.com/synfig/synfig/commit/e9b5183ad18aa76924ef61b581920ef96c1e3958
Now if you apply time offset to Group Layer, then all its child waypoints (in the "Canvas" parameter) will be shifted.

I also made the commit to stretch waypoints according to defined speed - https://github.com/synfig/synfig/commit/a3e9cff193216d4b9c3b88f0c7495a2603c62fc4
Unfortunately, dragging stretched waypoints still glitches. It would be nice if you could solve this.

Permutatrix:
@morevnaproject Alright, I made a pull request to fix the dragging thing. Pinpointing and fixing that was a fun little exercise. :)

I have some input on the subject of waypoint behavior - I think I would make it so that no waypoints are displayed at all in the canvas parameter for a group with a speed of zero. After all, it _behaves_ as if there are no waypoints within the canvas; everything seems constant. I think that, if the zero-speed group was contained in another group that also contained layers with waypoints, having the zero-speed group's waypoints show up in the parent group's canvas parameter's timeline mixed in with the waypoints of the other layers would be confusing. Perhaps not as confusing as that sentence, but hopefully you get what I'm saying.
I would probably implement this behavior by pushing all the waypoints to the start or end of time, since that would be the most technically accurate portrayal of what's happening.

My real name is Max May, though I do generally go by Permutator when I'm online. Sorry for not including it in my latest commit; I will from now on. And thanks!

morevnaproject:
> Alright, I made a pull request to fix the dragging thing. Pinpointing and fixing that was a fun little exercise. :)

Great, thanks for the fix!

> I have some input on the subject of waypoint behavior - I think I would make it so that no waypoints are displayed at all in the canvas parameter for a group with a speed of zero. After all, it behaves as if there are no waypoints within the canvas; everything seems constant. I think that, if the zero-speed group was contained in another group that also contained layers with waypoints, having the zero-speed group's waypoints show up in the parent group's canvas parameter's timeline mixed in with the waypoints of the other layers would be confusing. Perhaps not as confusing as that sentence, but hopefully you get what I'm saying.

Yes, I understand. Actually, I really appreciate the way how you explain everything. ^__^
I agree, it's logical to hide the waypoints. I have made a commit for that - https://github.com/synfig/synfig/commit/c2b5bacb184e325ce71b9bd655ec50abd43a98d6

But if we can get behavior, as illustrated in this video - http://youtu.be/fiStRlFQ8wI - then it would be great! (check the moment when I have group "1" selected and move time cursor).

> My real name is Max May, though I do generally go by Permutator when I'm online. Sorry for not including it in my latest commit; I will from now on. And thanks!

No problem! I have put your name into credits - https://github.com/synfig/synfig/commit/cb4f88bd1197e03bd0901fdf72979333bb387002 ^__^

My name is Konstantin Dmitriev, BTW.

PR #195, opened 2015-07-22
Fix Cairo renderer sometimes handling exported canvases incorrectly

Permutatrix:
This fix makes PasteCanvas render the context below it _before_ setting the time and grow_value of its canvas in `accelerated_cairorender`. This way, if a layer sets the time and grow_value of an exported canvas while rendering, the settings won't overwrite those of a PasteCanvas higher up using the same canvas.

[This file](https://www.dropbox.com/s/5motgenwmapm3xq/walk.sifz?dl=0), for instance, was affected by the bug: two Groups using the same canvas were supposed to have different time offsets, but instead moved in sync.


morevnaproject:
Merged into master. Thank you!

PR #197, opened 2015-07-25
Fix dragged waypoints from time-dilated subgroups not working properly

Permutatrix:
Hopefully this will be the last of the trouble brought on by time dilation. When a group contained another group whose contents had waypoints, and the subgroup was time-dilated, dragging one of its waypoints on the canvas parameter of the supergroup wouldn't work correctly. Usually, it would simply do nothing, but occasionally, the wrong waypoint would be moved the wrong distance.

This fix makes dragging waypoints under those circumstances work as expected. You can even drag multiple waypoints from groups with different speeds at the same time.


morevnaproject:
I was about to describe the bug in detail right when you have posted the solution. Great work!

PR #198, opened 2015-07-26
Give Switch layer's "Active Layer Name" parameter a combo box

Permutatrix:
This patch makes it so that editing the "Active Layer Name" parameter of a Switch Group directly will show a combo box with the names of all of its sublayers. This is much more intuitive and convenient than typing a layer's name manually.

### DIFF
--- a/synfig-core/src/synfig/layers/layer_switch.cpp
+++ b/synfig-core/src/synfig/layers/layer_switch.cpp
@@ -90,6 +90,7 @@ Layer_Switch::get_param_vocab()const
 	ret.push_back(ParamDesc("layer_name")
 		.set_local_name(_("Active Layer Name"))
 		.set_description(_("Only layer with specified name are visible"))
+		.set_hint("sublayer_name")
 	);
 
 	return ret;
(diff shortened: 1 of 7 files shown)


jcome:
nice improvement.

morevnaproject:
Really cool. Merged into master. Thank you!

jcome:
Seems don't work properly if the inside layer is a group layer.

~ yu

2015-07-27 1:06 GMT+08:00 Konstantin Dmitriev <notifications@github.com>:

Permutatrix:
@jcome Hmm, it seems to me that it works fine. Group layers have empty names until you name them yourself; the "Group" or "[exported_id]" that appears in the layer tree is misleading. So if you make a new Group and put it in a Switch, it will only show up if you set the Active Layer to an empty string. This behavior isn't new, though it is a bit confusing, isn't it? Something should be done about that, I think.

jcome:
Thanks for the explanation, Any way, your improvement makes Synfig Studio
much more user friendly, this is one of the weaknesses we have to spend
long time to fix. Thank you so much.

~ yu

2015-07-30 19:40 GMT+08:00 Permutator <notifications@github.com>:

PR #199, opened 2015-08-09
Add "Reverse" value node

Permutatrix:
This patch adds a general-purpose value node for reversing things. It operates on every type I could figure out how to reverse: lists, strings, spline points and segments, gradients, width points, and dash items. When reversing a list, it also reverses the individual elements of that list, meaning that, for instance, a spline will look exactly the same after being reversed. There's also some special behavior to make DILists and WPLists work as expected; you can look at the code if you really want to get into the details.

I'll note that, though most value nodes try to construct themselves in a way that preserves the value that was there before, this one does not. I thought it would be convenient to be able to reverse any value just by converting it to a Reverse node and then back again.

I'm planning to make an "Append" value node next. Combined with the "Reverse" node, this will allow functionality equivalent to Inkscape's ["Fill Between Many"](http://masemj.deviantart.com/art/Tutorial-Inkscape-Ponyscape-Fill-Between-Strokes-431636734) path effect. (Yay, ponies! Seriously, though, that's the best explanation I could find.)


morevnaproject:
Merged now. Thank you for the contribution!
P.S. The idea with 'Append' looks cool.

d-j-a-y:
Hello .... just in case if you don't read the tracker .... 
http://www.synfig.org/issues/thebuggenie/synfig/issues/894

see'ya !

d-j-a-y:
(Enhancement) http://www.synfig.org/issues/thebuggenie/synfig/issues/895

morevnaproject:
Note about the part 1 of the issue #849: The handles appear grey because the Reverse convert type doesn't support reverse manipulations. Every convert type that doesn't support reverse manipulations turns handles into grey. So it's not really bug. We have many convert types without reverse manipulations.

genete:
Hi there!
Good contributions folks!
I've not compiled/test the "Reverse" value node yet but I guess that it
offers the following result:
Reverse of (item 1, item 2, item 3, item 4)= (item 4, item 3, item 2, item
1)
If I understood correctly, I wonder if it would be a good complement to
create a "Reverse" *action*. In many cases the user just want the reverse
result and would like to have the result in the current working list. It
would be a little hard to make a full reverse of an animated spline, due to
the interrelation of the items with the previous and the forth but I think
it would eliminate the inability of reverse manipulation the "Reversed"
value nodes.


Cheers!


2015-09-06 16:22 GMT+02:00 Konstantin Dmitriev <notifications@github.com>:

> Note about the part 1 of the issue #849: The handles appear grey because
> the Reverse convert type doesn't support reverse manipulations. Every
> convert type that doesn't support reverse manipulations turns handles into
> grey. So it's not really bug. We have many convert types without reverse
> manipulations.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/synfig/synfig/pull/199#issuecomment-138089537>.
>



-- 
Carlos

PR #203, opened 2015-09-06
fix48 - Cycles in "BLine Tangent" ValueNodes cause a crash 

d-j-a-y:
forbid recurvise connection

### DIFF
--- a/synfig-core/src/synfig/valuenode.cpp
+++ b/synfig-core/src/synfig/valuenode.cpp
@@ -400,6 +400,27 @@ ValueNode::get_description(bool show_exported_name)const
 	return ret;
 }
 
+bool
+ValueNode::is_descendant(ValueNode::Handle value_node_dest)
+{
+    if(!value_node_dest)
+        return false;
+    if(Handle(this) == value_node_dest)
+        return true;
+
+    //! loop through the parents of each node in current_nodes
+    set<Node*> node_parents(value_node_dest->parent_set);
+    ValueNode::Handle value_node_parent = NULL;
+    for (set<Node*>::iterator iter = node_parents.begin(); iter != node_parents.end(); iter++)
+    {
+        value_node_parent = ValueNode::Handle::cast_dynamic(*iter);
+        if(Handle(this) == value_node_parent)
+            break;
+    }
+
+    return value_node_dest->parent_count() ? is_descendant(value_node_parent) : false;
+}
+
 ValueNodeList::ValueNodeList():
 	placeholder_count_(0)
 {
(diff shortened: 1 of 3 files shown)


d-j-a-y:
maybe to hard ???
i mean ... maybe that forbid connect to often .... ?

morevnaproject:
Sorry, I do not understand what you written in the comment. ^__^ Do you mean this:
"Maybe too hard??? I mean... Maybe that forbid connection in more cases than needed?"
Am I understood correctly?

morevnaproject:
Unfortunately, the fix you provided does not solves the case when you have cycle involving multiple converts. I.e. try the following:

* create a new document
* draw a circle
* right-click the circle's 'amount' parameter and 'convert' to 'add'
* right-click again and 'export' as 'a'
* open the child panel, open 'valuebase nodes' and select 'a'
* expand the amount parameter in the 'params' panel
* right-click 'LHS' and **convert it into scale**
* expand the 'LHS' parameter
* right-click 'Link' parameter and you can see the "Connect" option is there
* click "Connect" -> crash

I think it is required to make deep search for cycles. In any case, it is good start - better cut some cases, than cut nothing! Let me know if you willing to continue working on the solution. If you will stop here, then I will merge as is.

d-j-a-y:
>Maybe that forbid connection in more cases than needed?"
>Am I understood correctly?

yes... but finally, that's not the case ;-)

>wiI think it is required to make deep search for cycles.

""ll look at the case you describe ...

morevnaproject:
> l'll look at the case you describe ... 
My case is just including one additional level of convert. Just consider a solution for arbitrary count of levels.

d-j-a-y:
I'm trying to get recursively the parent of a ValueDesc  .... but can't figure why **value_desc.get_parent_desc()** and  **value_desc.get_sub_parent_desc()** always give me a **NULL** ptr. 
Any idea ?

```cpp
bool is_value_node_descendant(ValueNode::Handle value_node, ValueDesc value_desc)
{
    if (value_desc.parent_is_value_node())
    {
        if(value_node == value_desc.get_parent_value_node())
            return true;

        ValueDesc vd_parent = value_desc.get_parent_desc();

        if(value_desc.get_parent_desc()) //! always null , same result with get_sub_parent_desc
            return is_value_node_descendant(value_node, value_desc.get_parent_desc());
    }

    return false;
}
```

d-j-a-y:
if think if have found how to iterate through parent ... (thank's to bones code)
```cpp
        // loop through the parents of each node in current_nodes
        set<Node*> node_parents(value_desc.get_parent_value_node()->parent_set);

        for (set<Node*>::iterator iter2 = node_parents.begin(); iter2 != node_parents.end(); iter2++)
        {
            bool check = false;
            ValueNode::Handle vn = ValueNode::Handle::cast_dynamic(*iter2);
            if(value_node == vn)
                check = true;

        }
```

d-j-a-y:
Fixed : Really forbid recurvise connection now !

morevnaproject:
Merged now. Thank you!

PR #204, opened 2015-09-06
Fix574 : Fix the fix ! (segfault right clicking a spline)

d-j-a-y:


morevnaproject:
Good catch! Thank you!

PR #205, opened 2015-09-09
Fixed typo in export dialog

Lechindianer:
While playing around with sprite sheet animation I stumbled across a little typo in the export dialog. It should read 'horizontal' instead of 'horisontal'.


morevnaproject:
Merged now. Thank you!

PR #207, opened 2015-10-11
Fix cppcheck reports:

serval2412:
[synfig-core/src/synfig/cairoimporter.cpp:141] -> [synfig-core/src/synfig/cairoimporter.cpp:144]: (error) Iterator 'iter' used after element has been erased.
[synfig-core/src/synfig/importer.cpp:141] -> [synfig-core/src/synfig/importer.cpp:144]: (error) Iterator 'iter' used after element has been erased.

There are still these:
[synfig-core/src/synfig/valuenode_bline.cpp:199]: (error) Invalid iterator 'prev' used.
[synfig-core/src/synfig/valuenode_bline.cpp:200]: (error) Invalid iterator 'prev' used.

[synfig-core/src/synfig/valuenode_bone.cpp:661]: (warning) Suspicious condition. The result of find() is an iterator, but it is not properly checked.
I noticed that "first" and "prev" aren't initialized

[synfig-studio/src/gui/widgets/widget_compselect.cpp:117]: (style) Variable 'iter' is assigned a value that is never used.
[synfig-studio/src/gui/trees/keyframetreestore.cpp:159]: (style) The function 'clear_iterator' is never used.
[synfig-studio/src/gui/trees/keyframetreestore.cpp:91]: (style) The function 'model_iter_2_keyframe_iter' is never used.

### DIFF
--- a/synfig-core/src/synfig/cairoimporter.cpp
+++ b/synfig-core/src/synfig/cairoimporter.cpp
@@ -141,6 +141,6 @@ CairoImporter::~CairoImporter()
 	for(iter=__open_cairoimporters->begin();iter!=__open_cairoimporters->end();++iter)
 		if(iter->second==this)
 		{
-			__open_cairoimporters->erase(iter);
+			iter = __open_cairoimporters->erase(iter);
 		}
 }
--- a/synfig-core/src/synfig/importer.cpp
+++ b/synfig-core/src/synfig/importer.cpp
@@ -141,6 +141,6 @@ Importer::~Importer()
 	for(iter=__open_importers->begin();iter!=__open_importers->end();++iter)
 		if(iter->second==this)
 		{
-			__open_importers->erase(iter);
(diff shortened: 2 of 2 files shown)


morevnaproject:
Hello Julien!
This fix gives me the following error on compilation:

```
importer.cpp: In destructor 'virtual synfig::Importer::~Importer()':
importer.cpp:144:9: error: no match for 'operator=' (operand types are 'std::map<synfig::FileSystem::Identifier, etl::loose_handle<synfig::Importer> >::iterator {aka std::_Rb_tree_iterator<std::pair<const synfig::FileSystem::Identifier, etl::loose_handle<synfig::Importer> > >}' and 'void')
    iter = __open_importers->erase(iter);
```

I am on Fedora 22. What is your build environment?

serval2412:
I haven't built it, my patch must be wrong then. Let's drop it.

PR #212, opened 2015-12-14
fix 250 - Add shortcut to hide/show all ducks 

d-j-a-y:
* New action "mask-none_ducks"
* Menu/View/Show Hide Handle/"Toggle None/Last visible Handles"
* ```<Alt><0>``` has toggle all handles.

http://www.synfig.org/issues/thebuggenie/synfig/issues/250


morevnaproject:
Merged. Thanks you!

PR #213, opened 2015-12-14
UI : Handle tooltip transformation widget - display Layer desc

d-j-a-y:
transfo widget

### DIFF
--- a/synfig-studio/src/gui/workarearenderer/renderer_ducks.cpp
+++ b/synfig-studio/src/gui/workarearenderer/renderer_ducks.cpp
@@ -89,6 +89,7 @@ struct ScreenDuck
 	ScreenDuck(): selected(), hover(), width(0), has_alternative(false) { }
 };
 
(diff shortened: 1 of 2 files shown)


d-j-a-y:
just to know> following that change, i'm working at dialog setup.cpp (preferences) accordingly to [jcome redesign work](http://www.synfig.org/forums/viewtopic.php?f=14&t=3648) ...

morevnaproject:
Good work! Merged the tooltips feature, together with preferences dialog.

morevnaproject:
Ooops, sorry, probably I was too fast. I think it would be better to apply this change into the "rendering" branch. The "rendering" branch will take place of "master" very soon.

d-j-a-y:
> I think it would be better to apply this change into the "rendering" branch. The "rendering" branch will take place of "master" very soon.
@morevnaproject Yep no prob', looking at what's going on rendering, i thought something like that will occur... 
I close the request has my "preference redesign" branch is upon that one.

PR #219, opened 2016-01-29
fix906 c++11 support

d-j-a-y:
* Add M4 macros : AX_CXX_COMPILE_STDCXX and AX_CXX_COMPILE_STDCXX_11
* Remplace isnan by std::isnan and isinf by std::isinf
* MacOSX : Suppress the isnan and isinf MacOSX quick fix.
* Fix "lvalue required as unary '&' operand" ( thank's to Ivan)
* autobuild - add fedora >= 22 : yum --> dnf
* fix sigc++ header : replace all single by global

NOTA : Under Fedora 23 mlt installation require extra repo


morevnaproject:
Great work, thanks!

d-j-a-y:
By the way ... i have b

morevnaproject:
That;s weird. I was building all linux packages for 1.0.x on wheezy. Did they change something radically?

d-j-a-y:
I have tried to explore ... and i suspect something around gcc update. We may have something weird in our sources because this kind of thing should have been reported else where, but i did'nt have found something revelant. 

What do you think about a post in gcc's gurus mailing list ?

morevnaproject:
First, let me try to rebuild master branch in my debian chroot.

PR #220, opened 2016-01-31
preference dial fixes

d-j-a-y:
* Preference : transfo tooltip - fix according yu comments 
* Preferences : document - fix according yu comments


morevnaproject:
Merged. Thank you!

PR #221, opened 2016-02-06
Synfig reloaded

eshikafe:
wxPython UI code base


morevnaproject:
Hello Austin!

First of all, I am highly appreciate any work on porting Synfig UI to Python. Thanks for taking on this.

Regarding this topic I have several concerns to discuss.

1. I can see many important files are deleted from the source tree. For example,  automake scripts (configure.ac, Makefile.am) and scripts from "autobuild" directory (we use them to build Synfig packages for different platforms). The "brushlib" directory is also moved from its original place. Those modifications break our current build.
2. As I have figured out from your pull request, you are trying to rebuild Synfig Studio UI using wxWidgets in Python. Maybe it would be easier to use PyGObject, which is allows to build full-featured Gtk3 applications in Python? - https://en.wikipedia.org/wiki/PyGTK#PyGObject

eshikafe:
Thanks Konstantin.

Sorry for the mix up. Please don't accept the pull requests. I will fix
this asap.
I had already thought of PyObject before selecting wxPython as the best
option for the following reasons:

1. Use a UI framework that uses the native platform's controls and
utilities as much as possible.
2. Avoid GTK+ on Windows (use wxMSW). GTK is best suited for Linux in my
opinion (use wxGTK).
3. Few dependencies, as much as possible, in order to run or build an
applicatio - avoid GTK+ on Windows.
4. Simple to install UI framework.

Regards,
Austin

eshikafe:
I think it is better I create a new repo for synfig-reloaded. This will make things much easier.

genete:
Hi!
Konstantin, I think that create a repository (synfig-wxPython) to pull from
the one from this guy is a good idea. What do you think?


2016-02-08 23:16 GMT+01:00 Austin Aigbe <notifications@github.com>:

> I think it is better I create a new repo for synfig-reloaded. This will
> make things much easier.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/synfig/synfig/pull/221#issuecomment-181586653>.
>



-- 
Carlos

morevnaproject:
Hello @genete and @eshikafe 

As far as I understand, Austin (eshikafe) also would like to do modifications to ETL and synfig-core. So I think creating additional repo (with its own copy of synfig-core) is not comfortable for collaboration.

I think the best way is to make wxPython project co-exist with current implementation. In this case we can easily merge all changes made by Austin into our master branch (because his work on removing ETL are also much appreciated). When the wxPython port will get to the mature status, we can safely remove C++ GUI components.

So, my suggestion for Austin:
1. Fork official repo - https://github.com/synfig/synfig
2. Create synfig-reloaded directory in the project root. So we will have "ETL", "synfig-core", "synfig-studio", "synfig-reloaded".
3. Put all python files into "synfig-reloaded" directory and do all work there. I think there is no need to copy .cpp files into "synfig-reloaded", since they are already exist in "synfig-studio" (and they might change).

Best Regards,
Konstantin

eshikafe:
I have created one already - https://github.com/eshikafe/synfig-reloaded

PR #223, opened 2016-02-10
fix 805 and 796 : guide rulers crash

d-j-a-y:
from [bob / windows](http://www.synfig.org/forums/viewtopic.php?f=15&t=9936&p=31046#p31045) !
> EDIT: My bad, everything works fine, all the guide work properly


d-j-a-y:
@morevnaproject  Could be pushed also to 1.0.x branch and the branch be released has new bug fix ?

PR #225, opened 2016-02-20
Fix previous commit d73f5315670

d-j-a-y:
studio::App::set_selected_canvas_view(this); is already done elsewhere.

* was causing many calls to canvasview::present


morevnaproject:
When I apply this patch the bug (inactive tools when closing one document) returns.

PR #228, opened 2016-09-17
Option to show/hide file toolbar

caryoscelus:
This is mostly to make onion skin controls usable on small screens (i choose file toolbar as least useful). Of course, it would be nice to have toolbar fully customizable, but even such optional hiding does the job.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -298,6 +298,7 @@ bool studio::App::restrict_radius_ducks=true;
 bool studio::App::resize_imported_images=false;
 bool studio::App::enable_experimental_features=false;
 bool studio::App::use_dark_theme=false;
+bool studio::App::show_file_toolbar=true;
 String studio::App::custom_filename_prefix(DEFAULT_FILENAME_PREFIX);
 int studio::App::preferred_x_size=480;
 int studio::App::preferred_y_size=270;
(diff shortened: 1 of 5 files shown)


morevnaproject:
There is already "View" -> "Toolbar" command exists. But for some reason it is inactive. Would be nice to implement it this way, instead of preferences option.

Also, it would be better to rename it to "View" -> "Show Toolbar" (in the same way as we have "View" -> "Show Menu"). Thanks!

caryoscelus:
Searched through code for that toolbar command, but apparently it simply isn't implemented. Also i suppose it was meant to show/hide the whole toolbar, not parts of it. But yeah, making it through view menu is more reasonable. The reason i did it this way is that i don't really know gtk, but wanted to make onion skin more accessible fast. I might look into proper implementation later.

jcome:
On Sat, Sep 17, 2016 at 9:39 PM, caryoscelus <notifications@github.com>
wrote:

> but apparently it simply isn't implemented.


As far as I remembered, it hides "global" toolbar instead of canvas view
toolbar. but the global toolbar was removed at a release.

it might be worth to mention the current toolbar is attaching to canvas
window, and Syfnig Studio is MDI, it means each canvas window has its
toolbar.

~ yu

morevnaproject:
Merged for now.

In the future it would be nice to have checkboxes replaced by modern Gtk3 switches (see illustration).
screen-7

caryoscelus:
Well, this should be easy ;)

PR #229, opened 2016-09-23
Add "Time" layer category

caryoscelus:
I am also thinking it would be nice to put other stuff out of "Other" as well, but can't think of categorization for them atm.


morevnaproject:
Merged. Спасибо! ^__^

PR #233, opened 2016-09-25
Fix for dark time widget

chrislondon:
In dark theme it's nearly impossible to see the time frame. This changes the background to white. Should work in both light and dark themes.

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -1213,7 +1213,6 @@ CanvasView::create_time_bar()
 	current_time_widget->set_size_request(0,-1); // request horizontal shrink
 	current_time_widget->set_width_chars(5);
 	current_time_widget->set_tooltip_text(_("Current time"));
-	current_time_widget->override_background_color(Gdk::RGBA("#FFFFFF"));
 	current_time_widget->show();
 
 	//Setup the FrameDial widget
@@ -2711,7 +2710,7 @@ CanvasView::on_time_changed()
 		get_canvas()->keyframe_list().find(time);
 		current_time_widget->override_color(Gdk::RGBA("#FF0000"));
 	}catch(...){
-		current_time_widget->override_color(Gdk::RGBA("#000000"));
+		current_time_widget->override_color(Gdk::RGBA(0));
 	}
 
 	if(get_time() != time_adjustment()->get_value())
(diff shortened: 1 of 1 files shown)


mray:
awww - no screenshots.

chrislondon:
Here's a picture of it. It also has custom icons (thanks jcome) but those aren't part of this commit.

white_background_time_frame

chrislondon:
Here's a picture of it before:

black-theme-synfig-studio

mray:
Thanks for providing before and after screenshots.
Why did you choose toinvert  black/white instead of continuing bright text on dark background?
Looks like the original problem was only that *dark* text on *dark* background does not work.

chrislondon:
The problem stems from:

synfig/synfig-studio/src/gui/canvasview.cpp:2708-2714
`current_time_widget->set_value(time);
try {
	get_canvas()->keyframe_list().find(time);
	current_time_widget->override_color(Gdk::RGBA("#FF0000"));
}catch(...){
	current_time_widget->override_color(Gdk::RGBA("#000000"));
}`

Because the text colors are being manually overriden it's ignore the dark theme setting.  Let me see if there's an easy way to fix that.

chrislondon:
Looks like by calling `current_time_widget->override_color(Gdk::RGBA(0));` instead of `current_time_widget->override_color(Gdk::RGBA('#000000'));` It will reset the color to the theme's specification. I like this solution.

chrislondon:
screenshot from 2016-09-26 09-49-32

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -2710,7 +2710,7 @@ CanvasView::on_time_changed()
 		get_canvas()->keyframe_list().find(time);
 		current_time_widget->override_color(Gdk::RGBA("#FF0000"));
 	}catch(...){
-		current_time_widget->override_color(Gdk::RGBA("#000000"));
```
chrislondon: Forces text to be black

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -2710,7 +2710,7 @@ CanvasView::on_time_changed()
...
+		current_time_widget->override_color(Gdk::RGBA(0));
```
chrislondon: Resets text color to system theme

caryoscelus:
First commit is overriden by second, so perhaps makes sense to leave only one. Otherwise, nice fix.

mray:
Awesome! This looks great. Unfortunately I have no clue what all the code lines mean and what you had to fight with. Thanks!

mray:
@chrislondon i don't want to stop your mood of squashing bugs: #236  
I even created one just for you!

morevnaproject:
Merged. Thanks to everyone!

PR #237, opened 2016-09-27
Fix objects being created with opacity slightly less than 1.0

### LINKED ISSUE #234: Amount is less than 1 when layer created with max opacity
Despite setting amount to max (1.0) when creating new layer (shape/outline/etc), the actual amount value is set to 0.999999990686774 (it's displayed as 1.0 in parameters due to rounding, of course). If changed to 1.0 it is saved properly.

caryoscelus:
Fixes #234


caryoscelus:
I guess this PR will be obsolete soon as i'm about to get rid of that horrible copypasted code in https://github.com/synfig/synfig/tree/master/synfig-studio/src/gui/states

morevnaproject:
I have merged it now. ^__^

PR #282, opened 2016-11-08
App::apply_gtk_settings - minor refactor

caryoscelus:
We don't need to pass use_dark because it is already present in App


morevnaproject:
Merged

PR #284, opened 2016-11-09
Fix zoomdial mouse control & revert to C++11

caryoscelus:
Ok, so i finally got back to that zoomdial issue. Here's a fix that allows using mouse after initial selection is done and i also reverted to c++11 (but not 14 as it was). @morevnaproject please test if it works for your build environment.


rodolforg:
Shouldn't we close this PR? @ice0 
The changes it proposes (reaction to mouse scroll) were already done in another way by blackwarthog.

PR #294, opened 2016-12-24
Valuenode registry

caryoscelus:
Replace manual valuenode type registering in `valuenode.cpp` with semi-automatic based on macros placed directly in relevant valuenode implementation file, which also handles get_name/get_local_name implementations.

Automatic registering is based on instantiating static dummy structures whose constructor registers appropriate class in `ValueNodeRegistry`

Additional changes in dynamic&static list valuenodes were required to unify their factory methods with the rest.

The only thing left is to test how the whole thing works across platforms (i.e. "does it work on windows?" cause i think unix-like systems have similar enough treatment of dynamic libs). According to info i found, there should be no problem however as all systems properly initialize static variables in shared libraries. Still would be nice if someone tests or reviews this.


caryoscelus:
Module support committed. So this is ready to merge. I'll leave it for a while in case someone wants to make a review though.

caryoscelus:
Fixed with https://github.com/synfig/synfig/pull/294/commits/19cf24fffcd13418f0ad033b7d3dbd52d3745df6 , should be fine for now.

PR #325, opened 2017-02-11
Module loading advanced configuration

caryoscelus:
This might require adding boost system&filesystem libs to automake build.

Instead of reading single synfig_modules.cfg files, synfig now
also checks files in synfig_modules.cfg.d/ directory (if found
in the same location as synfig_modules.cfg).

This simplifies stand-alone module installation as module
installation script can now simply install a new file instead
of editing existing.

### DIFF
--- a/synfig-core/src/modules/CMakeLists.txt
+++ b/synfig-core/src/modules/CMakeLists.txt
@@ -70,5 +70,5 @@ file(WRITE synfig_modules.cfg ${SYNFIG_MODULES_CONTENT})
 
 install(
     FILES synfig_modules.cfg
-    DESTINATION etc
+    DESTINATION etc/synfig
 )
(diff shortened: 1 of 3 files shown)


blackwarthog:
You may use functionality of FileSystemNative::instance() or Glib::Dir/Gio::File for scan directories and checking file existence.

caryoscelus:
Perhaps. Any reason to do so, though?

blackwarthog:
To avoid unneeded dependencies

caryoscelus:
That doesn't seem to be a good goal for me. Besides, we have these libs in synfig tool already.

blackwarthog:
Part of dependent code in synfig-tool are tiny and will be better to rewrite it in future.

caryoscelus:
Well, i'm going to add even more dependencies in the future.

blackwarthog:
This is the great beginning of another and epic tale. And our fairy tale about The Unneeded Dependencies went to The End. They lived long and happily, and died together on the same day.

PR #328, opened 2017-02-26
Fix dependencies for Debian Stretch/Sid

s-leroux:
As described on http://www.synfig.org/forums/viewtopic.php?f=13&t=10478

Linux Autobuild on Debian Strech/Sid has some issue with the libtiff-dev & libpng-dev dependency:

```
Package libtiff-dev is a virtual package provided by:
  libtiff5-dev 4.0.7-5
  libtiff4-dev 3.9.6-11+deb7u3
You should explicitly select one to install.
```

and

```
The following packages have unmet dependencies:
 libpng-dev : Conflicts: libpng12-0-dev
              Conflicts: libpng12-dev but 1.2.50-2+deb8u3 is to be installed
              Conflicts: libpng3-dev
              Recommends: libpng-tools but it is not going to be installed
 libpng12-dev : Conflicts: libpng-dev
```

In addition, the `libfftw3` was missing.

### DIFF
--- a/autobuild/synfigstudio-linux-build.sh
+++ b/autobuild/synfigstudio-linux-build.sh
@@ -869,7 +869,7 @@ get_dependencies()
 		libfreetype6-dev \
 		libfontconfig1-dev \
 		libxml2-dev \
-		libtiff4-dev \
+		libtiff5-dev \
 		libmlt-dev libmlt++-dev \
 		libjasper-dev \
 		x11proto-xext-dev libdirectfb-dev libxfixes-dev libxinerama-dev libxdamage-dev libxcomposite-dev libxcursor-dev libxft-dev libxrender-dev libxt-dev libxrandr-dev libxi-dev libxext-dev libx11-dev \
--- a/autobuild/synfigstudio-linux-build.sh
+++ b/autobuild/synfigstudio-linux-build.sh
@@ -863,12 +863,13 @@ get_dependencies()
 		libtool \
 		intltool \
 		gettext \
-		libpng12-dev \
+		libpng-dev \
+		libfftw3-dev \
 		fontconfig \
 		libfreetype6-dev \
 		libfontconfig1-dev \
 		libxml2-dev \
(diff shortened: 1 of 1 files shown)


caryoscelus:
Can you provide a clean patch without tons of .po changes?

s-leroux:
Sorry, I've just noticed that -- my mistake.

s-leroux:
@caryoscelus I removed the .po files by rewriting the branch history. I think the pull request is ok now.

caryoscelus:
Well, that part is ok, but otherwise it might break compatibility. Since i'm not using autobuild scripts and older debian versions, i can't check that.

I suppose it is (though maybe there are options to disable that) and they should be updated some times, but updating after every change (and especially in the same commit as unrelated change elsewhere) would clutter history.

blackwarthog:
I've problem at Debian/Jessie:

```
Note, selecting 'libltdl-dev' instead of 'libltdl3-dev'
Note, selecting 'libpng12-dev' instead of 'libpng-dev'
Package libtiff4-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libtiff5-dev:i386 libtiff5-dev

E: Package 'libtiff4-dev' has no installation candidate

```

I think it's typo, because libtiff4 was in wheezy and not included in jessie, stretch and sid. I've no libtiff4 installed in my system (libtiff5 only).
@s-leroux, may you check build script at sid with 'libtiff5-dev' at line 872?

s-leroux:
My system undergo several updates since Debian wheezy -- and for some reasons libtiff4 was still available in apt.

But you're right @blackwarthog, only libtiff5 is available on a clean Stetch/Sid system.
I've just tested it: Synfig compiles without any problem with libtiff5 and as far as I can tell, works flawlessly.

I've fixed the pull request accordingly.

blackwarthog:
Thank you!

Issue #421, opened 2017-09-09
shortcuts for pan,zoom and rotate canvas when using pen


lehtis:
**Synfig version & platform**: Synfig 1.2.0 Windows 10
(we're more likely to help if you're using latest development builds)

**Issue description**:
Feature Request:
When drawing and using Synfig with drawing pen, it would be handy pan,zoom,canvas rotate like Krita.
In Krita it's very handy system when drawing. 
space+pen=pan 
space+ctrl+pen=zoom 
space+shift+pen=rotate canvas.

bobbybee:
I'm (maybe) working on this

morevnaproject:
@bobbybee Would be very cool if you take this one!

Note to myself: This feature request conflicts with #414.

bobbybee:
@morevnaproject The same commands seem to work in Krita if you replace "space" with the pen button (this is what I use personally). That would not conflict with the other issue.

lehtis:
How about also possibility to edit hotkeys in preferences?

bobbybee:
Is, uh, is this a bug? :smile:

```
1889             if(canvas_view->get_smach().process_event(EventMouse(EVENT_WORKAREA_MOUSE_BUTTON_DOWN,BUTTON_MIDDLE,mouse_pos,pressure,modifier))==Smach::RESULT_OK)
1890 
1891             dragging=DRAG_WINDOW;
1892             drag_point=mouse_pos;
1893             signal_user_click(1)(mouse_pos);
```

(note lack of semicolon at the end of the if)

caryoscelus:
@bobbybee, link? Perhaps history will show whether that is misindent or missing if-clause.

bobbybee:
@caryoscelus https://github.com/synfig/synfig/blob/278ca7605a3121f53e62b8bcbd1260dabae810ec/synfig-studio/src/gui/workarea.cpp#L1889-L1893

caryoscelus:
Well, there you have it: https://github.com/synfig/synfig/commit/7fbbb48c38c6901291260c84d62f5323598fd351

Most likely the code isn't what it was supposed to be. However, i suppose it works because `drag_point` is probably harmless without `dragging` set. Not sure about `signal_user_click` though. I'd recommend putting those lines in braces and check if it works.

morevnaproject:
The fix #425 now merged into master. It does pan and scale with tablet pen button (the one that corresponds to middle mouse button). Rotation of canvas is not implemented in Synfig yet.

[...]

Issue #414, opened 2017-08-29
Pressing "Space" key should Play/Pause the animation preview


alexandervrs:
**Synfig version & platform**:
1.3.4 / Windows 10 64bit

**Issue description**:
Pressing the "Space" key while a frame is selected on the timeline should play/pause the animation. This behavior is seen in Flash and is very useful for quick preview using the keyboard

morevnaproject:
Some people ask to use Space for canvas pan/rotate operations.

alexandervrs:
I think it should be an editable Preference, some programs use Space to quickly pause/play while others use Enter.

morevnaproject:
Yes, editing shortcuts in preferences would be nice. Requested in #422.
At the moment you can check a file called "accelrc" in the configuration directory of Synfig.

Though, you might hit some issues related with current system of handling shortcuts - #303 #257 #403.

rodolforg:
@Keyikedalube  is this done?

Keyikedalube:
I didn't notice this issue existed.

Yes, this requested feature is already implemented in #1776 :)

PR #425, opened 2017-09-09
Implement zooming with pen

bobbybee:
This handles a subset of #421 

In particular, it does not implement rotation, since there is no canvas rotation implemented as it is (?). If/when that is changed, the stubs are here.

It also requires the pen button as opposed to the space bar, mimicking the existing pan keybinding. Fixing this (with a configuration option) would be somewhat more involved. It would require some refactoring. It would also conflict with #414 as noted. Perhaps that should be a separate patchset?

### DIFF
--- a/synfig-studio/src/gui/workarea.cpp
+++ b/synfig-studio/src/gui/workarea.cpp
@@ -1868,6 +1868,7 @@ WorkArea::on_drawing_area_event(GdkEvent *event)
 						return true;
 					}
 				}
+
 				// All else fails, try making a selection box
 				dragging=DRAG_BOX;
 				curr_point=drag_point=mouse_pos;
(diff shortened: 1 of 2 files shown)


bobbybee:
(Found a bug, don't merge)

bobbybee:
OK. This should match Krita's algorithm.

morevnaproject:
> there is no canvas rotation implemented as it is (?)

Yes, there are no canvas rotation implemented.

Tested the fix - works perfectly for me. Will merge it right away. Thank you! ^__^

morevnaproject:
Now shipped with Synfig 1.3.5 - https://morevnaproject.org/2017/12/25/synfig-studio-1-3-5/

bobbybee:
Hooray! :)

PR #439, opened 2017-10-01
Remove stub code (#if 0) with coan

alyssarosenzweig:
This patch was generated with the command:

    find . -type f -name '*.cpp' -exec ../Downloads/coan-6.0.1/src/coan source --replace --eval-wip \{} \

executed at the root of the synfig repository with the tool coan, which can be downloaded from [Sourceforge](http://sourceforge.net/projects/coan2) to check reproducability.

My machine is currently recompiling synfig to ensure there was no unintentional breakage; it may be a while until I'll have confidence that the patch is correct.


bobbybee:
Build finished; still seems to work :tada:

PR #443, opened 2017-10-05
Remove commented deadcode

alyssarosenzweig:
Extension of #439 to also remove commented out code with [comment-clean](https://rosenzweig.io/git/?p=comment-clean.git).

Separate pull request as comment-clean is not yet mature software (it was written specifically for synfig) so there may be problems. (Some were corrected in the commit of the series)

(I have not yet rebuilt synfig with this patchset -- do not merge yet)


caryoscelus:
I don't suppose there should problem with building if it only removes comments, but some of comments it remove are actually comments, such as https://github.com/synfig/synfig/pull/443/commits/41f11e4040d660ab6904b1de8421205434923981#diff-182e4f172f565fe501341235c044d27dL54

bobbybee:
@caryoscelus It looks like any comment starting with ! should be special-cased, yes?

synfig-studio/src/gui/docks/dockmanager.cpp
```diff
@@ -421,7 +419,6 @@ Gtk::Widget* DockManager::read_widget(std::string &x)
 	// paned
 	if (hor || vert)
 	{
-		// skip "[hor|" or "[vert|"
```
caryoscelus: Not saying that this kind of comment has a right to live, but it's not commented code either.

synfig-studio/src/gui/duck.h
```diff
@@ -96,15 +96,7 @@ class Duck : public etl::shared_object
 		TYPE_TANGENT				=	(1 <<  1),	//    2
 		TYPE_RADIUS					=	(1 <<  2),	//    4
 		TYPE_WIDTH					=	(1 <<  3),	//    8
-		TYPE_ANGLE					=	(1 <<  4),	//   16
```
caryoscelus: Hmm, why are these removed?

PR #474, opened 2017-12-12
Sort sets by name

### LINKED ISSUE #372: Implement a correct sorting in 'Sets'
### Description:
Synfig has a very nice thing called 'Sets'. You can add diffident control layers in it and easily access them from one place. I use it extremely. But layers inside a set are sorted randomly without any pattern: nor alphabetically nor by Z depth, which increase the time to find the needed layer. This is how it looks:
setssorting

See? Random order, but should be: R-1, R-2, R-3, etc.

### Feature request:
Implement a correct sorting algorithm in 'Sets' panel. It can sort by name or by Z depth. For me, alphabetical sorting would be ideal, but if it possible to implement different methods it would be great.

### To reproduce:
Open attached file (done in 1.3.3) and go to 'Window->Sets', you'll see that layers are sorted in a random order.
Note: Synfig can actually sort it right sometimes, so if it looks OK, restart the program several times until you see a screwed up sorting.
[BadSorting.zip](https://github.com/synfig/synfig/files/1035475/BadSorting.zip)

### Experienced on:
All versions since 1.0.2 on Windows 10. Didn't test on other platforms.

caryoscelus:
fixes #372

### DIFF
--- a/synfig-studio/src/gui/CMakeLists.txt
+++ b/synfig-studio/src/gui/CMakeLists.txt
@@ -2,7 +2,7 @@
 ## Configure
 ##
 
-set(STUDIO_VERSION "1.3.2")
+set(STUDIO_VERSION "1.3.5")
 set(SHOW_EXTRA_INFO OFF) # this needs some fixes to work
 configure_file(autorevision.h.cmake.in autorevision.h)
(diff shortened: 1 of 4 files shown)


morevnaproject:
Merged now. Thank you!

PR #478, opened 2017-12-16
Workaround to implement MagickCore's hidden InitializeExceptionInfo() method

mehw:
Since MagickCore version 7 deprecated methods have been removed. GetExceptionInfo() can no longer be used to initialize an exception info.

With the help of @caryoscelus I'm trying to find a way to implement MagickCore's hidden InitializeExceptionInfo() method in a manner which is compatible with version 7 and 6 of ImageMagick.

### DIFF
--- a/synfig-core/src/modules/mod_magickpp/trgt_magickpp.cpp
+++ b/synfig-core/src/modules/mod_magickpp/trgt_magickpp.cpp
@@ -54,6 +54,20 @@ SYNFIG_TARGET_SET_CVS_ID(magickpp_trgt,"$Id$");
 
 /* === M E T H O D S ======================================================= */
 
+/*
+ * WORKAROUND
+ *
+ * Mimics the hidden MagickCore's InitializeExceptionInfo() method.
+ */
+namespace MagickCore {
+static void InitializeExceptionInfo(ExceptionInfo *exception)
+{
+	MagickCore::ExceptionInfo* exceptionInfo = AcquireExceptionInfo();
+	*exception = *exceptionInfo; // copy assignment, ExceptionInfo is a struct
+	MagickCore::DestroyExceptionInfo(exceptionInfo);
+}
+}
+
 template <class Container>
 MagickCore::Image* copy_image_list(Container& container)
 {
@@ -61,7 +75,7 @@ MagickCore::Image* copy_image_list(Container& container)
 	MagickCore::Image* previous = 0;
 	MagickCore::Image* first = NULL;
 	MagickCore::ExceptionInfo exceptionInfo;
-	MagickCore::GetExceptionInfo(&exceptionInfo);
+	MagickCore::InitializeExceptionInfo(&exceptionInfo);
 	for (Iter iter = container.begin(); iter != container.end(); ++iter)
 	{
 		MagickCore::Image* current;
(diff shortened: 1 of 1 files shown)


caryoscelus:
Looks like you've mixed up DestroyExceptionInfo and InitializeExceptionInfo. Also, why define macro? If GetExceptionInfo is already removed, i'm pretty sure InitializeExceptionInfo long time enough to be present everywhere.

mehw:
Thanks, but [InitializeExceptionInfo()](https://imagemagick.org/api/MagickCore/exception_8c.html#aad9984b1b0d45c9fbac8f16a6192ef8f) is a private method (aka [MagickPrivate](https://imagemagick.org/api/MagickCore/method-attribute_8h.html#a07784d9b8e7a718c9b4f65368cf546be)) both in version 7 and 6, so it cannot be used.

As an alternative to the synfig-core/configure.ac patch and macro use, it is possible to always call `DestroyExceptionInfo()`, which is part of both version 7 and 6 of MagickCore, avoiding to call the deprecated `GetExceptionInfo()` of version 6.

`GetExceptionInfo()` and `DestroyExceptionInfo()` should be close enough.

Can you confirm?

caryoscelus:
>GetExceptionInfo() and DestroyExceptionInfo() should be close enough.

I don't see how they are, when they are doing pretty much opposite things: initialization and deinitialization.

> InitializeExceptionInfo() is a private method

Ah, i see. Then, perhaps, [AcquireExceptionInfo()](https://imagemagick.org/api/MagickCore/exception_8c_source.html#l00108) should be used. E.g. `auto exceptionInfo = AcquireExceptionInfo()`. But then it also have to be cleaned up later, i suppose.

Btw, more generally it seems to me that using C api is not a good thing. Magick++ api shuold be used instead.

mehw:
> Ah, i see. Then, perhaps, AcquireExceptionInfo() should be used. E.g. auto exceptionInfo = AcquireExceptionInfo(). But then it also have to be cleaned up later, i suppose.

What about calling `AcquireExceptionInfo()` to initialize a temporary exceptionInfo, then copy its values to the variable `exceptionInfo` (it is a structure), and lastly call `DestroyExceptionInfo()` to clean the temprary exceptionInfo?

caryoscelus:
I have no idea about how libmagick works, but i would assume that GetExceptionInfo was removed and InitializeExceptionInfo hidden for a reason (even if mostly aesthetic one). Official sample uses Acquire & later Destroy ExceptionInfo: http://www.imagemagick.org/script/magick-core.php

Also, do you actually use mod_magickpp? If all you want to do is to make things compile, it's easier to just disable building it. Otherwise it makes sense to follow the new MagickCore conventions of creating with Acquire and destroying with Destroy. Or, better yet, rewriting in proper C++ api.

mehw:
Thanks for the explanations!

> Also, do you actually use mod_magickpp? If all you want to do is to make things compile, it's easier to just disable building it.

I see what you mean, it is used just for animated GIFs, right? With my current setup mod_magickpp is compiled by default. I bumped into the `GetExceptionInfo()` compile error when I updated ImageMagick to version 7.

> it makes sense to follow the new MagickCore conventions of creating with Acquire and destroying with Destroy.

The current patch kinda follows your suggestion. It uses Acquire and Destroy as a mean to create a temporary initialized exception info, and copyes its values to `exceptionInfo`.

> Or, better yet, rewriting in proper C++ api.

I'll take a look into that... just for curiosity...
http://www.imagemagick.org/api/Magick++/index.html

mehw:
Just pushed a patch implementing `exceptionInfo` as a pointer. `AcquireExceptionInfo()` and `DestroyExceptionInfo()` now operates directly over `exceptionInfo`.

The code should be compatible with ImageMagick version 7 and 6.

caryoscelus:
Looks good, merged.

mehw:
Nice, thanks!

PR #496, opened 2018-02-22
Libxml++ 3.0 support

cyisfor:
I couldn't get synfig to compile since I've no idea where on earth libxml++-2.6 could be found, so I just made the changes to get it working with libxml++-3.0. That might break people's stuff, but I thought I'd put it out there, in case anyone else isn't using an obsolete version of libxml++ but wants to use synfig studio. Also fixed some deprecated stuff, and added a hack so the cmake could build in the source directory, which probably isn't strictly necessary long as the README specifies that you're required to build synfig-studio in a separate directory from the source (for build_images).


synfig-studio/images/CMakeLists.txt
```diff
@@ -250,7 +250,7 @@ foreach (IMAGE IN ITEMS ${IMAGES})
     add_custom_command(
         TARGET build_images POST_BUILD
         COMMAND synfig ${SRC}.sif --time ${TIME}f -w ${W} -h ${H}
-        COMMAND mv ${SRC}.png ${CMAKE_CURRENT_BINARY_DIR}/${IMAGE}.png
+        COMMAND ./move.sh ${SRC}.png ${CMAKE_CURRENT_BINARY_DIR}/${IMAGE}.png; exit 0
```
eldruin: I do not understand what this changes other than ignoring errors. Anyway I think it would be better to replace this with `cmake -E rename`.

synfig-studio/src/synfigapp/cvs.cpp
```diff
@@ -159,11 +161,11 @@ CVSInfo::calc_repository_info()
 			}
 			else
 			{
-				system(strprintf(
+				IGNORE_RESULT(system(strprintf(
```
eldruin: Why was this added? it does not do anything.

synfig-studio/src/gui/devicetracker.cpp
```diff
@@ -61,23 +61,42 @@ using namespace studio;
...
+	seat = gdk_display_get_default_seat(
+		gdk_display_manager_get_default_display(
+			gdk_display_manager_get()));
+	return gdk_seat_get_slaves(
+		seat,
+		GDK_SEAT_CAPABILITY_ALL_POINTING );
+}
+			
+static
+GList* get_devices(GdkDevice*& pointer) {
```
eldruin: It would be better to return a tuple/pair with the pointer together with the list. Then the function below without the argument is not necessary, as they do the same thing. The caller would just discard the pointer when it is not interested in using it.

synfig-studio/src/gui/devicetracker.cpp
```diff
@@ -160,34 +162,20 @@ DeviceTracker::save_preferences()
 					guint gdk_keyval = 0;
 					GdkModifierType gdk_modifiers = GdkModifierType();
 					gdk_device_get_key(gdk_device, j, &gdk_keyval, &gdk_modifiers);
-					keys[i].keyval = gdk_keyval;
-					keys[i].modifiers = gdk_modifiers;
+					keys[j].keyval = gdk_keyval;
```
eldruin: Good catch!

morevnaproject:
I agree with @eldruin  - we require to submit a separate PR for every feature/improvements. This PR addresses several issues. @cyisfor can you please separate them to independent PRs?

PR #689, opened 2018-12-18
Render done sound enabled by default on restore_default_settings

BobSynfig:
Set enabled by default on restore default settings


morevnaproject:
Good catch! Thank you!

Issue #666, opened 2018-10-31
Simplify building process by utilizing Conan dependency manager


morevnaproject:
I think we can integrate with Conan C/C++ package manager to provide all required  dependencies for building Synfig on any platform. 

https://conan.io/

Benefits:
- Developers can easily set up build environment on any platform (and theoretically any build toolchain).
- This will also simplify packaging process (Appimages, Win installers, OSX App) and will allow to implement automatization for this process via CI.

Previously, I tried to achieve the same using NIX package manager, but at the end it turned out not a best fit for this task (in particular, it is hard to deal with switching versions for libraries; cases when you  want to temporary switch library version for testing is hard to handle).

I am still in the process of investigating Conan, but I like what I see so far. This page is giving a good overview - https://docs.conan.io/en/latest/getting_started.html
I am especially like the integration with BinTray ecosystem.

rohitsanjay:
Hi,
I am Rohit from IIT Kharagpur. I would like to contribute to this project in Gsoc. Currently, I am going through the given doc( https://docs.conan.io/en/latest/getting_started.html). What are things to be done before submitting the proposal?

morevnaproject:
Hi @rohitsanjay! My name is Konstantin Dmitriev, I will be mentoring this project.
Can you please give me some information about your skills and background? Also, please let me know which operation systems you are capable to work with (Linux, Windows, OSX)?

morevnaproject:
About the work on this issue:

I suggest to make some preparation work, to proof your capabilities to work on this project.

Synfig depends on various libraries, that required for building (see https://synfig-docs-dev.readthedocs.io/en/latest/common/dependencies.html). Some of those libraries are already available as Conan packages, some are not. If some library is not available as Conan package, then we need to pack it.

So, as first step, we need to create a list of all required libraries. In this list we will mark which libraries are Conan-packed and which are not (this is going to be our TODO list and it will show the expected amount of work).

Let's start with a list of dependencies for synfig-core: please examine https://github.com/synfig/synfig/blob/master/synfig-core/src/CMakeLists.txt and make a list of required libraries.

Same work should be done for synfig-studio.

rohitsanjay:
@morevnaproject I am second-year computer science undergraduate at IIT Kharagpur, India. I have prior experience with C/C++ and python. I use both Ubuntu and Windows as per requirement.

rohitsanjay:
I have made a list and verified if they are Conan packed or not. The list is followed:

[...]

PR #694, opened 2018-12-30
Reorganize build system files

morevnaproject:
Hello! This is a part of ongoing work for bringing our build scripts into order. ^__^

The idea behind this rework is to bring all scripts for various build systems into logical schema, which is easy to maintain and easy to understand.

So, my idea is:

In general, build process is divided in 3 steps -

1. Installing dependencies / preparing environment
1. Building Synfig code
1. Packaging (optional, for deployment only)

Steps 1 and 3 are OS-specific.

I.e. for **step 1**: the way user installs dependencies depends on OS he uses __and__ the build environment he uses - Conan, Homebrew or MacPorts (in OSX), NIX, native package manager (in Linux), Cygwin or MSYS2 (in Windows). So, we will have a set of scripts for step 1, so user can choose from:
```
1-setup-conan.sh
1-setup-brew.sh [DONE]
1-setup-homebrew.sh
1-setup-ubuntu.sh
1-setup-fedora.sh
1-setup-msys2.sh
...
```

The  **step 2** is building. Here we will have one script, which is common for all operating systems. As soon as the environment/dependencies are configured properly, the build process i expected to be same for all systems (./configure, make install, ...).

I have just added 2 wrapper scripts for step 2 to allow building debug and non-debug variations:

```
2-build-debug.sh [DONE]
2-build-production.sh [DONE]
```

The instructions on usage for those scripts are described here https://synfig-docs-dev.readthedocs.io/en/latest/common/building.html#re-building-your-changes (they are remaining actual).

Finally, step 3 is executed when we need a wrap results of our build into a installer/package.
Here we will have choices for different operating systems:

```
3-package-linux-appimage.sh
3-package-osx-dmg.sh [DONE]
3-package-windows-nsis.sh
3-package-windows-zip.sh
```
So, this is it.

For end user (developer) the process is split into 3 easy steps - setup, building, packaging. At each step user chooses appropriate script for his case and executes it.

I.e. to build nsis installer package on windows using MSYS2 environment the sequence is:
```
./1-setup-msys2.sh
./2-build-production.sh
./3-package-windows-nsis.sh
```

In the future, if the migration to Conan build system will be successful (#666), I expect most of scripts from step 1 go away. This will reduce maintenance. For now, as the migration is still in progress and results are not clear yet, we will keep them that way. It is possible we will keep those scripts even after migration to Conan will be completed (if there will be such demand from the community).

Things to do before merge of this PR:
- [x] Figure out why CI failing to find "1-setup-brew.sh" script ([error message](https://travis-ci.com/morevnaproject/synfig/jobs/167516420)). @ice0 ?
- [x] Re-organize linux build script (```autobuild/synfigstudio-linux-build.sh```) according to described schema. Help appreciated.
- [ ] Update build instructions - https://synfig-docs-dev.readthedocs.io/en/latest/common/building.html

@blackwarthog @ice0 @BobSynfig - looking forward for your feedback. ^__^


morevnaproject:
@BobSynfig You right! Silly mistake. Thank you! ^__^

morevnaproject:
Okay, now we've got OSX build working. Tested on OSX 10.9 and OSX 10.10. Higher version should work as well.

morevnaproject:
Linux scripts are re-organized now.

morevnaproject:
With a short delay the build instructions are updated - https://synfig-docs-dev.readthedocs.io/en/latest/common/building.html

Now we have build instructions for Linux and OSX! Working on build system for Windows...

Issue #693, opened 2018-12-30
Remove "Cancelled by user" dialog that happens when the user tries to move an animated object


unternehmen:
**Synfig version & platform**:
1.3.10, Commit 064ca470469b3802f84a91d3419424067d7d6f5a (Dec 22 2018)
Debian GNU/Linux 9, x86_64

**Issue description**:
When you try to move an object whose position is animated, a dialog asks if you want to apply an offset to the animated parameters.  If you press "no," a second dialog appears which says "Set Layer Parameter (...): Origin: Cancelled by user" and looks like this:

![screenshot from 2018-12-29 21-22-59](https://user-images.githubusercontent.com/18196805/50543873-6d288600-0bb1-11e9-8bb9-82ceeff89c94.png)

It seems like this dialog interrupts the workflow of the user, and it does not provide new information since the user intentionally chooses not to apply the offset to the parameter.  I believe the user experience would be improved if this dialog box was not present.

unternehmen:
Thanks, I tried your branch, and the dialog no longer appears.  However seems to cause an issue.  if I press "no" on the offset dialog, it applies the offset anyway instead of canceling it.  I think this is because

    throw Error(_("Cancelled by user"));

was responsible for breaking out of Action::ValueDescSet::prepare(), but now the throw statement is gone so the function just continues and applies the offset.

I would like to work out a solution on my fork if that is okay with you. (I want to join development and this seems like an easy enough first issue :-)

unternehmen:
I have added a commit which fixes the issue.  Can someone double-check it? :+1:

BobSynfig:
Argh, I just pushed another way tricky way, yours looks a lot more clean ^^
https://github.com/BobSynfig/synfig/commit/3d5aa74ed39a15368921cbf89436a1fff7b34654

unternehmen:
Oops! :"-D

morevnaproject:
@unternehmen @BobSynfig Thank you for your contributions for solving this issue!

After examining your pull requests #695 and #696 I found that the silent error handling is already implemented.
Please check line 133 here:
https://github.com/synfig/synfig/blob/25d55580bb036fdba1c6559693fce5ad9ddbbad0/synfig-studio/src/synfigapp/action_system.cpp#L129-L141

So, I think replacing
```throw Error(_("Cancelled by user"));```
to
```throw Error(Error::TYPE_UNABLE);```
will be sufficient. ^__^

PR #695, opened 2018-12-30
Fix for #693 - (Remove "Cancelled by user")

BobSynfig:
Done :)


BobSynfig:
Not needed 
See solution by K. https://github.com/synfig/synfig/issues/693#issuecomment-450791387

Issue #626, opened 2018-09-22
Feature request: Give feedback when rendering is happening and complete


RichardJECooke:
**Synfig version & platform**:
Synfig 1.2.2.. Windows

**Issue description**:

File - Render - GIF - go.
Nothing seems to happen. But in my folder I can see a file is gradually getting bigger so it's working. But I have no idea when it's finished. Please add a a window or progress bar or toast or something to give user feedback here

wstaelens:
Yea for mp4, you just see a cmd prompt and it is complete when cmd closes… just a console.writeline like in blender would be great!

morevnaproject:
The fix is merged now! Will be shipped with next release of Synfig. Thank you!

Issue #383, opened 2017-06-30
Indication needed that rendering is in progress.


CoryAlbrecht:
There's no indication that rendering is is being done unless you happen to have started synfigstudio from a terminal and can watch output to stdout/stderr go past.

There needs to be a way to know when a render is complete when launched from synfigstudio.

Synfig Studio 1.0.2 on Ubuntu 16.04.2 GNOME

caryoscelus:
Check status panel under drawing area. At least in recent versions there are messages "Rendering.." and "File rendered successfully" in it.

CoryAlbrecht:
OK, yes. However, since that area changes if you do anything like move to a keyframe, the indicator needs to be separate so it does not get overwritten/hidden by any actions the user may take.

I would also suggest a way to cancel an in progress render.

caryoscelus:
All good ideas, but there doesn't seem to be (m)any devs still interested in synfig.

wstaelens:
>           All good ideas, but there doesn't seem to be (m)any devs still interested in synfig.

All these pay wall, donate walls and different paying sites for tutorials etc are keeping me a bit away from it...

morevnaproject:
> All these pay wall, donate walls and different paying sites for tutorials etc are keeping me a bit away from it...

We do not have donate wall - on our download page it is clearly indicated that you can enter "zero" in the price field to get a binary build for free. Also, anonymous download link is provided. 

![screenshot_001](https://user-images.githubusercontent.com/332868/46422413-85c60300-c75e-11e8-97d6-84a013dd359a.png)

So, no wall - it is free for anyone who is able to read. ^__^

Having this suggestion to donate on the download page allows us to keep two developers contributing regularly to Synfig code. I invite you to check commits activity [here](https://github.com/synfig/synfig/graphs/commit-activity) and [here](https://github.com/synfig/synfig/commits/master) to see the impact they do.
All those regular contributions are possible thanks to donations we get from\ Download page.

Regarding paid tutorials - yes, we do have **one** paid training video course (translated in several languages), which also helps  to pay for efforts of our main developers. For those who do not want to get a paid course we have free tutorials and manual pages here - https://wiki.synfig.org/Category:Manual

Again, it is free for anyone who is able to read. ^__^

morevnaproject:
[...]

PR #699, opened 2019-01-04
626 383 render progressbar

BobSynfig:
Hi Konstantin,
here is a Render Progress Bar in Dock_Info :)
Only the last commit of course.
See https://github.com/synfig/synfig/issues/626 for more details
Concerns also https://github.com/synfig/synfig/issues/383
Cheers


morevnaproject:
Merged now! I especially like the idea to put progressbar into Info Panel. Possibly, we can consider moving some other widgets/buttons from canvas view toolbar to  Info Panel or some other panel...
Thank you!

PR #722, opened 2019-02-05
Improve outline smoothness

ThaFireDragonOfDeath:
**Change description**
In commit 90788ec4df1450785463c403556dfbda7d695aa2 the SAMPLES define was decreased from 50 to 10. This made the outline rough (see attachments for example). This patch reverts that change to make the outlines smooth again (the advanced outlines wasn't affected by this bug).

**Attachments**
Rendered outline in master (4aa255f9ad852e3de5a64046f243a8b9dc34d143):
outlinebug-original

Rendered outline in master with this commit:
outlinebug-fixed

Original .sif file:
[OutlineBug.zip](https://github.com/synfig/synfig/files/2832796/OutlineBug.zip)

### DIFF
--- a/synfig-core/src/modules/mod_geometry/outline.cpp
+++ b/synfig-core/src/modules/mod_geometry/outline.cpp
@@ -61,7 +61,7 @@ using namespace synfig;
 
 /* === M A C R O S ========================================================= */
 
-#define SAMPLES		10
+#define SAMPLES		50
 #define ROUND_END_FACTOR	(4)
 #define CUSP_THRESHOLD		(0.40)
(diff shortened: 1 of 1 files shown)


morevnaproject:
Can we keep samples =10 for draft rendering mode?

blackwarthog:
i think increasing of samples count to 50 will not affect to speed

blackwarthog:
as i remember now segments should be a splines, not lines, so we have bug in other place

morevnaproject:
> as i remember now segments should be a splines, not lines, so we have bug in other place

Okay, so I am going to merge this one and open a new issue about segments/splines.

PR #728, opened 2019-02-13
Fixed rename file extension when rendering based on 'target'

### LINKED ISSUE #627: Feature request: Rename file extension when rendering based on 'target'
1.2.2

**Issue description**:
As the title. Use case: I've saved as png. Now I want to render as gif. And it overwrites my png. When I change the target to gif my file extension in the textbox above should change to gif (or mpeg or whatever)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/63742324-feature-request-rename-file-extension-when-rendering-based-on-target?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ankit-kumar-dwivedi:
### **Topic:** Fixes #627 
Now every-time the target changes filename will be updated with new .ext if required.

### DIFF
--- a/synfig-studio/src/gui/render.cpp
+++ b/synfig-studio/src/gui/render.cpp
@@ -269,7 +269,7 @@ RenderSettings::set_target(synfig::String name)
 {
 	target_name=name;
 	//TODO: Replace this condition
-	tparam_button->set_sensitive(target_name.compare("ffmpeg") && target_name.compare("png-spritesheet")?false:true);
+	tparam_button->set_sensitive(!(target_name.compare("ffmpeg") && target_name.compare("png-spritesheet")));
 }
 
 void
--- a/synfig-studio/src/gui/render.cpp
+++ b/synfig-studio/src/gui/render.cpp
@@ -240,9 +240,16 @@ RenderSettings::set_entry_filename()
 void
 RenderSettings::on_comboboxtext_target_changed()
 {
+	string ext[] = {"null", ".bmp", ".png", ".dv", ".avi", ".gif", ".png", ".jpg", ".gif", ".mng", "null", "null", ".exr", ".png", ".png", ".ppm", ".yuv"};
 	int i = comboboxtext_target.get_active_row_number();
 	if (i < 0 || i >= (int)target_names.size()) return;
 	if (target_name == target_names[i]) return;
+	if (!(i == 0 || i == 10 || i == 11))
+	{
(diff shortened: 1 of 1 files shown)


morevnaproject:
Hi! Thank you for PR, this is a good start!
There is a following problem in current implementation: Synfig can have different set of target, depending on compilation options. I.e., user can have synfig compiled without "magick++" module. In this case, targets in dropdown list will have different order, which is not corresponding to your array -
```
string ext[] = {"null", ".bmp", ".png", ".dv", ".avi", ".gif", ".png", ".jpg", ".gif", ".mng", "null", "null", ".exr", ".png", ".png", ".ppm", ".yuv"};
```

ankit-kumar-dwivedi:
> user can have synfig compiled without "magick++" module. In this case, targets in dropdown list will have different order, which is not corresponding to your array -
> 
> ```
> string ext[] = {"null", ".bmp", ".png", ".dv", ".avi", ".gif", ".png", ".jpg", ".gif", ".mng", "null", "null", ".exr", ".png", ".png", ".ppm", ".yuv"};
> ```
@morevnaproject 
In that case I can simply replace the vector with an map as in that case order is not going to matter the extension being selected I'll get it in string from the target Dropbox.

ankit-kumar-dwivedi:
@morevnaproject please review the changes :smile_cat:

morevnaproject:
Hey, I did a quick test and it works as expected now! ^__^

Just want to suggest a few improvements to the code:

1. I think we can remove everything that maps to "null"  - ```{"Auto","null"},{"null","null"},{"null-tile","null"}```. 

If those entries will be removed, then ```ext.find(target_names[i])``` (line 250) will return ```ext.end()``` and that should be fine. 

Also, that will allow to simplify condition from ```if(itr != ext.end() && (itr->second)!="null")``` to ```if(itr != ext.end())``` (line 252).

2. I forgot one more possible target - "libav". It should map to ".avi" extension. Can you please add this as well? ^__^

ankit-kumar-dwivedi:
Done Changes !

morevnaproject:
Looks good to me! Thank you and congratulations with the first code contribution! ^__^

PR #749, opened 2019-03-06
Added Preview workarea rendering

### LINKED ISSUE #729: Add new workarea rendering mode: Preview
This continues issue #409.

Long story short: 

In 17812bc2eaf88be30a6b57587bc55da71265ae2f we have added downsampling for images. This greatly improves quality if your scene uses big images scaled to smaller size. Unfortunately, this operation demands additional CPU resources and you will notice general application slowdown when working with such scenes.

You can enable Draft rendering mode (it disables downasampling), but this solution doesn't fits for all cases, as it is hides many layers, which could be important for you. 

So, we need one more rendering mode, which displays all possible layers, but with some simplifications - i.e. no downsampling (maybe some other simplifications can be added).

I suggest to call this new rendering mode "Preview".

So, we will have 3 rendering modes:

- "Draft" (simplifications + some layers hidden).
- "Preview" (some simplifications, not influencing image look very much).
- And the last one, which renders without any simplifications - I suggest to call it "Final".

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70076080-add-new-workarea-rendering-mode-preview?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ankit-kumar-dwivedi:
Fixes #729 
I added the new rendering mode - "Cobra Preview" and made it the default renderer.
I verified it's working.

>So far we need only one Optimizer, which modifies how images are displayed: it disables supersampling and forces NearestNeighbour interpolation scaling method.

_The above line Implied to NEAREST INTERPOLATION so I used the OptimizerDraftTransformation_
For GUI I used Comboboxtext and replaced the Togglebutton
As I cannot directly place Combobox to Toolbar hence I placed the Combobox in a Gtk::Tool Item Container.
**The GUI looks as below:**

 
@morevnaproject Please review the changes made :smiley: 

PS: Want to drive your attention towards the depreciation error which says **throws is deprecated in C++ 11** while building synfig, I think in ETL

### DIFF
--- a/synfig-core/src/synfig/rendering/renderer.cpp
+++ b/synfig-core/src/synfig/rendering/renderer.cpp
@@ -1017,7 +1017,7 @@ Renderer::unregister_renderer(const String &name)
 const Renderer::Handle&
 Renderer::get_renderer(const String &name)
 {
-	static const char default_engine[] = "software-preview";
+	static const char default_engine[] = "software";
 	return get_renderers().count(name) > 0           ? get_renderers().find(name)->second
 		 : get_renderers().count(default_engine) > 0 ? get_renderers().find(default_engine)->second
 		 : blank;
--- a/synfig-core/src/synfig/rendering/renderer.cpp
+++ b/synfig-core/src/synfig/rendering/renderer.cpp
@@ -46,6 +46,7 @@
 
 #include "software/renderersw.h"
 #include "software/rendererdraftsw.h"
+#include "software/rendererpreviewsw.h"
 #include "software/rendererlowressw.h"
 #include "software/renderersafe.h"
 #ifdef WITH_OPENGL
@@ -97,6 +98,7 @@ Renderer::initialize_renderers()
 
 	// register renderers
 	register_renderer("software", new RendererSW());
+	register_renderer("software-preview", new RendererPreviewSW());
 	register_renderer("software-draft", new RendererDraftSW());
 	register_renderer("software-low2",  new RendererLowResSW(2));
 	register_renderer("software-low4",  new RendererLowResSW(4));
@@ -1015,7 +1017,7 @@ Renderer::unregister_renderer(const String &name)
 const Renderer::Handle&
 Renderer::get_renderer(const String &name)
 {
-	static const char default_engine[] = "software";
+	static const char default_engine[] = "software-preview";
 	return get_renderers().count(name) > 0           ? get_renderers().find(name)->second
 		 : get_renderers().count(default_engine) > 0 ? get_renderers().find(default_engine)->second
 		 : blank;
--- a/synfig-core/src/synfig/rendering/software/CMakeLists.txt
+++ b/synfig-core/src/synfig/rendering/software/CMakeLists.txt
@@ -3,6 +3,7 @@ target_sources(synfig
         "${CMAKE_CURRENT_LIST_DIR}/rendererdraftsw.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/rendererlowressw.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/renderersafe.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/rendererpreviewsw.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/renderersw.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/surfacesw.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/surfaceswpacked.cpp"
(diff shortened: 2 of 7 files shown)


morevnaproject:
@ankit-kumar-dwivedi Exciting! I am testing it now. ^__^

morevnaproject:
Tested and it works awesome!

Just a few concerns:

1. I am afraid, that this way to set default renderer also affects CLI rendering. When I tried to render file using CLI, it uses Preview renderer instead of Final. I suspect the same happen if we render via GUI using "File" -> "Render" command.

My proposal: revert 
```
static const char default_engine[] = "software-preview";
```
back to
```
static const char default_engine[] = "software-preview";
``` 
and modify code of synfig-studio (app.cpp?) to activate Preview renderer on start.


2. I suggest to change order of items in combo-box, so they will be like this:

- Draft
- Preview
- Final

In all other aspects this is awesome work and great contribution!

morevnaproject:
> PS: Want to drive your attention towards the depreciation error which says throws is deprecated in C++ 11 while building synfig, I think in ETL

Can you please submit an issue about that? Thank you! ^__^

ankit-kumar-dwivedi:
> ```
> static const char default_engine[] = "software-preview";
> ```
> 
> 
> back to
> 
> ```
> static const char default_engine[] = "software-preview";
> ```

@morevnaproject :confused: ??

ankit-kumar-dwivedi:
@morevnaproject have changed the order to draft , preview and final. 
Please confirm the change for app.cpp

morevnaproject:
> I guess you wish to say revert to
> ```
> static const char default_engine[] = "software";
> ```
> i.e how it was before right?

Right. Sorry, my mistake. ^__^

Or, you can consider alternative solution:

Keep default renderer as it specified in your code (preview is default):
```
static const char default_engine[] = "software-preview";
```
...and explicitly set renderer to "software" in synfig CLI. Also explicitly do the same in GUI when user does final render (via "File" -> "Render"). I think it can be passed as parameter when calling Renderer::get_renderer().

I do not know which solution is easier, just consider any of them. ^__^

ankit-kumar-dwivedi:
@morevnaproject I am following this

>     I am afraid, that this way to set default renderer also affects CLI rendering. When I tried to render file using CLI, it uses Preview renderer instead of Final. I suspect the same happen if we render via GUI using "File" -> "Render" command.
> 
> 
> My proposal: revert
> 
> ```
> static const char default_engine[] = "software-preview";
> ```
> 
> back to
> 
> ```
> static const char default_engine[] = "software";
> ```


I reverted back the default renderer to "software" but I am not able
> and modify code of synfig-studio (app.cpp?) to activate Preview renderer on start.

what I thought as a solution is to call 
	**synfigapp::Main::settings().set_value("pref.navigator_renderer",             "software-preview");**
but I feel this is wrong. 

on the other hand when I tried the alternate solution

> Or, you can consider alternative solution:
> 
> Keep default renderer as it specified in your code (preview is default):
> 
> ```
> static const char default_engine[] = "software-preview";
> ```
> 
> ...and explicitly set renderer to "software" in synfig CLI. Also explicitly do the same in GUI when user does final render (via "File" -> "Render"). I think it can be passed as parameter when calling Renderer::get_renderer().

there was no get_render() in render.cpp in gui

@morevnaproject can you tell me what am I doing wrong here! :'(

ankit-kumar-dwivedi:
@morevnaproject actually I was thinking that...
If replacing "software-draft" -> "software-preview" is causing all this problem. 

 > 1. I am afraid, that this way to set default renderer also affects CLI rendering. When I tried to render > file using CLI, it uses Preview renderer instead of Final. I suspect the same happen if we render via  > GUI using "File" -> "Render" command.

Then I doubt this behavior is not due to the changes I made. Have you tested in the same way when "software-draft" renderer was added?

ankit-kumar-dwivedi:
@morevnaproject I am getting the following on startup now :)
image
Sending the commit now.

ankit-kumar-dwivedi:
@morevnaproject please review the PR

morevnaproject:
Reviewing it now! ^__^

morevnaproject:
Works perfectly now! Thank you!

ankit-kumar-dwivedi:
@morevnaproject also my apologies for few absurd comments(though I deleted should've got mails)
Less sleep makes me dumb I think :p

morevnaproject:
No problem! Congratulations with a contribution that will make many users happy! ^__^

morevnaproject:
As continuation of this work we can think about other simplifications to introduce in this mode. Possibly we can figure out more simplification, which will not make significant impact on final image, but increase speed.

Issue #760, opened 2019-03-11
Disable Brush Tool


morevnaproject:
As result of discussion here - https://forums.synfig.org/t/remove-brush-tool-from-synfig-studio-stable-release/9572:

> Brush tool is basically unusable at its current status. That confuses many users and makes a bad impact on user experience.

So, I am proposing the following:

- Add an option to Preferences, which enables and disables experimental features. Call it “Enable experimental features”.
-  This option should be disabled by default.
-  Add a code to hide Brush Tool icon from Toolbox if “Enable experimental features” option disabled.

9inpachi:
@morevnaproject I will start looking into this. Let me take care of it.

morevnaproject:
@9inpachi Awesome! This issue is yours now. ^__^

9inpachi:
@morevnaproject I have setup the code for the "Enable experimental features" preference but can not fund the code where the tools are located.
If you can refer to some files, that would help a lot.
Thanks. :)

morevnaproject:
In Synfig tools are called "states". Please take a look here: 

https://github.com/synfig/synfig/blob/59badba9e0c9eeed52176a7ce1fce08a6692f2c6/synfig-studio/src/gui/app.cpp#L1727

P.S. I am looking forward for your email. as mentioned in https://github.com/synfig/synfig/issues/759#issuecomment-471425914 ^__^

9inpachi:
@morevnaproject 
Looking forward to your email. :)

morevnaproject:
@9inpachi I cannot find your message. Did you mailed to me via https://www.synfig.org/contact/ ? ^__^

9inpachi:
I did receive a confirmation email suggesting that my message has been sent.
I will try sending again.

morevnaproject:
Got your mail, thank you! ^__^

PR #764, opened 2019-03-12
Experimental features option. Disable brush tool.

9inpachi:
Closes: #760
The option for enabling/disabling experimental features has been added.
The default value for enabling experimental features is "false", as seen here.
https://github.com/synfig/synfig/blob/9b3fe68804dd3d5bfb5d6ac43a74fa72138b6c18/synfig-studio/src/gui/app.cpp#L306

This feature is currently being used to disable the "Brush Tool". It can be extended to isolate other experimental features as well.

Thanks for locating the code for "Brush Tool" @morevnaproject.

^_^

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -1724,7 +1724,7 @@ App::App(const synfig::String& basepath, int *argc, char ***argv):
 		/* other */
 		state_manager->add_state(&state_text);
 		if(!getenv("SYNFIG_DISABLE_SKETCH" )) state_manager->add_state(&state_sketch);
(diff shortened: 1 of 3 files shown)


morevnaproject:
Hi! Thank you for the PR! I have tested your changes and noticed that "Experimental features" is mentioned twice in UI layout. I suggest to make layout more compact, similar to how "Autosave" switch located -
screenshot_008

9inpachi:
The changes have been commited.
I included a section label so that more specific "Experimental Features" could be accommodated if there was a need. Like there could be multiple options under the Experimental Features section.
But I guess that is not needed right now.

morevnaproject:
Looks good now! Thank you! ^__^

PR #767, opened 2019-03-12
Filled rectangled layer merged to rectangle layer. Fixes #745

### LINKED ISSUE #745: Feature request: Rounded corners for Rectangle Layer
We need an option to draw Rectangle Layer with rounded corners - http://mathworld.wolfram.com/RoundedRectangle.html

roundedrectangle_1000

Note: Synfig already has a special layer called  "Filled Rectangle" (can be created by choosing "Layer" -> "New Layer" -> "Examples" -> "Filled Rectangle" from menu). This is not the same as Rectangle Layer. The Filled Rectangle layer has an option for rounded corners, but it doesn't draw them correctly in two aspects:
- the rounded parts has no antialiasing.
- the corners are not circles.

screenshot_001

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70890510-feature-request-rounded-corners-for-rectangle-layer?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

yshkp:
@morevnaproject Please review the changes.


morevnaproject:
Hi! Thank you for the PR! I tested it and found that "Expand amount" paramtere has no effect anymore. Can you please fix that? ^__^

yshkp:
@morevnaproject Yeah, it needs to be fixed first. What you think could be the reason for this?

morevnaproject:
That's because you do not use value of ```expand``` variable in your code.

screenshot_001

yshkp:
@morevnaproject I have fixed the `expand` parameter and I have written the expand variable I noticed when I checked my IDE where I did not save it so please don't mind it. ^_^
Now, you may test it.

morevnaproject:
It works now! ^__^

I have examined the code and noticed the following lines:
```
	std::vector<Point> list(4);

	list[0][0] = list[3][0] = p0[0] - expand;
	list[0][1] = list[1][1] = p0[1] - expand;
	list[2][0] = list[1][0] = p1[0] + expand;
	list[2][1] = list[3][1] = p1[1] + expand;

	set_stored_polygon(list);
```
I think they are not needed and should be removed.

And one more thing: can you please make "Keep Bevel Circular" parameter enabled by default?
Thank you! ^__^

yshkp:
@morevnaproject All your required changes are done. Now, it looks great. Please review.
Also, check the message I sent you via mail. ^_^

morevnaproject:
Great, thank you!

I just found one more thing we need to consider.

Since Filled Rectangle Layer is removed now, so all Synfig files that use it will not be valid anymore. That could make some users feel bad. ^__^
The solution is to modify file loading function, to treat all Filled Rectangle layers as Rectangle Layers.

morevnaproject:
I have created a simple test file - 
[filled-rectangle.sif.zip](https://github.com/synfig/synfig/files/2974053/filled-rectangle.sif.zip)

This is how it looks in version 1.3.11 -
screenshot_006

...and this is how it looks when opened in version without Filled Rectangle layer -
screenshot_007

yshkp:
> screenshot_007

@morevnaproject Source of filled rectangle was removed thats why these results came across. 

-  Is filled rectangle layer option still up there ?
And in loading function we just need to replace `filled rectangle` with `rectangle`. Right ?

morevnaproject:
> @morevnaproject Source of filled rectangle was removed that's why these results came across

Exactly. ^__^

> And in loading function we just need to replace "filled_rectangle" with "rectangle". Right ?

Yes. Replace "filled_rectangle" to "rectangle". You need to modify file to allow proper loading -
https://github.com/synfig/synfig/blob/master/synfig-core/src/synfig/loadcanvas.cpp

yshkp:
@morevnaproject Could you please tell how you created filled rectangle in the test file((version without filled rectangle) so that I can also test it?

morevnaproject:
@yshkp  No I have created filled rectangle in other version - in the one that has filled rectangle. THen I saved it and opened in a version without filled rectangle.

yshkp:
@morevnaproject Though the function doesn't call for filled rectangle directly and as far as I understood we need to check for if the filled rectangle was called and then to modify `parse_layer` function with replacing its attributes with rectangle layer. 
I hope I am in the right way. ^__^

morevnaproject:
@yshkp Yes, that should work! ^__^

yshkp:
@morevnaproject I have made required changes and now it's working fine. Please have a review.

morevnaproject:
@yshkp Great, testing it now! ^__^

morevnaproject:
Works perfectly now! Thank you for the awesome contribution! ^__^

PR #770, opened 2019-03-12
Added support for 'pacman' package manager in linux setup script

Rathod-Sahaab:
pacman is the package manager for many arch based distros
Headers are generally provided with binaries so *-devel isn't required


morevnaproject:
Looks good. Thank you! ^__^

PR #785, opened 2019-03-24
End_Time feature implemented

rockyhandsome1:
Hi @morevnaproject please review the changes.


morevnaproject:
Hi! I have tested the improvements and it works fine.

Still, looking at the code, I am afraid that getting End Time Widget is not the most elegant way to communicate CanvasView and FrameDial.

Instead, I suggest to create two public functions in FrameDial:
- ```set_end_time(time, fps)```
- ```get_end_time()```
So, you can call those functions from CanvasView and they can change appropriate values of End Time Widget.

Issue #771, opened 2019-03-14
Lock ratio button unexpectedly crops the canvas


Domarius:
**Synfig version & platform**:
SynfigStudio-1.3.11-testing-19.02.09-linux64-9583a.appimage

**Issue description**:
[As discussed here](https://forums.synfig.org/t/solved-render-at-different-size-than-canvas-without-clipping/9559/3), when rendering a project, clicking the "link" button (highlighted in the image below) causes the undesired behaviour of cropping the canvas to the new render size, eg. if the new size is smaller than the canvas size.

Whereas if you don't click this button, and enter a new size (eg. smaller than the canvas), the project will be scaled down to render at the new size.

![ratio lock](https://user-images.githubusercontent.com/7411541/54341829-9754f480-4686-11e9-8e91-dacaf5366f80.png)
**Steps to reproduce**:

Normal behaviour;
* Set canvas property to 1920x1080
* Insert some image/photo where cropping will be obvious (eg. a photo portrait), scale it to full size of the canvas.
* File, Render, type 1280x720 for the canvas width & height but do NOT click the link button
* Render (you can just render 1 image)
* Note that the full canvas is in the rendered image, it's just at a smaller resolution

Incorrect behaviour;
* File, Render
* Click the link button so that the 2 rings show as linked together
* Enter "1280" for the width, then click or tab away from the field to have the height automatically change to 720
* Render
* Note that the canvas has now been cropped to the smaller size, instead of scaled.

You can reset this incorrect behaviour by clicking the link button again so that the rings are now unlinked.

BobSynfig:
Hi,
Did you check the options in the tab "Other"?
There are options having consequences on this first tab.
Also please have a look at the wiki pages:
https://wiki.synfig.org/Canvas_Properties_Dialog
https://wiki.synfig.org/Image_Dimensions

A french user had a similar problem of understanding about these options and it has be explained and solved here https://forums.synfig.org/t/agrandir-image-pas-juste-le-calque/9579

Domarius:
Thanks for getting back to me.  I tried those options on the other tab, and it has no effect.  The pages you linked too, still don't explain this behaviour.  And the French user isn't talking about this specific problem.

Why would the "lock ratio" link button suddenly switch the output to cropped? This button should only keep the width & height in ratio with each other or not.  It shouldn't affect cropping vs scaling.   For that there should be a separate explicit option that says "Crop to new size" or "Scale to new size".

BobSynfig:
[...]

PR #797, opened 2019-03-28
Made image aspect checked by default

ankit-kumar-dwivedi:
Fixes: #771 
@morevnaproject please review this when you have time [^_^]

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -3968,7 +3968,7 @@ App::new_instance()
 	canvas->rend_desc().set_w(preferred_x_size);
 	canvas->rend_desc().set_h(preferred_y_size);
 	canvas->rend_desc().set_antialias(1);
-	canvas->rend_desc().set_flags(RendDesc::PX_ASPECT|RendDesc::IM_SPAN|RendDesc::IM_ASPECT);
+	canvas->rend_desc().set_flags(RendDesc::IM_SPAN|RendDesc::IM_ASPECT);
 	canvas->set_file_name(filename);
 	canvas->keyframe_list().add(synfig::Keyframe());
 
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -3968,7 +3968,7 @@ App::new_instance()
 	canvas->rend_desc().set_w(preferred_x_size);
 	canvas->rend_desc().set_h(preferred_y_size);
 	canvas->rend_desc().set_antialias(1);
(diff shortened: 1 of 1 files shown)


morevnaproject:
Thanks! Reviewing now. ^__^

synfig-studio/src/gui/renddesc.cpp
```diff
@@ -151,7 +154,7 @@ Widget_RendDesc::refresh()
 	entry_duration->set_value(rend_desc_.get_duration());
 
 	adjustment_fps->set_value(rend_desc_.get_frame_rate());
-
+	//zzzzzzzzzzzzzzzzzzzzz
```
morevnaproject: I suggest to remove this line. :satisfied:

morevnaproject:
It looks like there is some misunderstanding here. ^__^ As per my comment - https://github.com/synfig/synfig/issues/771#issuecomment-476298262 - we need "Image Aspect" checkbox enabled. This PR enables "Image Span" (which is already enabled by default).

synfig-studio/src/gui/renddesc.cpp
ankit-kumar-dwivedi: That's my way to mark things as important 😂

ankit-kumar-dwivedi:
> It looks like there is some misunderstanding here. ^__^ As per my comment - [#771 (comment)](https://github.com/synfig/synfig/issues/771#issuecomment-476298262) - we need "Image Aspect" checkbox enabled. This PR enables "Image Span" (which is already enabled by default).

Will fix this when I get home :-)

morevnaproject:
Also, I think that ```renddesc.cpp``` is not the best place for such modifications. I suggest to take a look at the code which responsible for creating new documents (in ```app.cpp```) - this is where it sets all parameters of renddesc for new files.

ankit-kumar-dwivedi:
> Also, I think that `renddesc.cpp` is not the best place for such modifications. I suggest to take a look at the code which responsible for creating new documents (in `app.cpp`) - this is where it sets all parameters of renddesc for new files.

Okay @morevnaproject I'll keep it on my mind! ^_^

ankit-kumar-dwivedi:
@morevnaproject done changes please have a look :smile:

morevnaproject:
Code now looks good! But, I have tested the changes and the resize didn't worked for me. From this comment - https://github.com/synfig/synfig/issues/771#issuecomment-473236624 - I figured out that we also need to disable "Pixel Aspect" checkbox (it is enabled by default).

ankit-kumar-dwivedi:
@morevnaproject disabled pixel aspect by default!

morevnaproject:
Looks god now. Thank you! ^__^

Issue #516, opened 2018-04-03
"Edit in External Tool" Options


libreartist:
**Synfig version & platform**: 1.31, Debian (Sid)

**Issue description**: I recently noticed in LibreOffice that it has right-click option on its objects (like vectors, images, sound files and video). If clicked, that particular object will be opened in external application (the default one for those kind of files in your system, e.g., gimp for raster images, inkscape for vector images, ardour for sound files, etc.), so that we can edit them and save them directly into the document, not externally. Thus, it allows a workflow where the parts can be updated while they being positioned in the document. Can it be done in Synfig, for its raster images, vector images and sound files -- if they be needing editing while the project has already crossed the half-way?

morevnaproject:
Sounds good!

morevnaproject:
Related to https://github.com/synfig/synfig/issues/513

morevnaproject:
Let's start by implementing this for Image Layers only.

What is expected:

1. New option in Preferences of Synfig Studio, which defines a path to application for opening image files ("krita" or "gimp" or similar).
2. User can right-click any image and choose "Edit in external application..." menu item. This should open a image file in application specified in Preferences.

ankit-kumar-dwivedi:
Thanks for referring this issue @morevnaproject have my exam on Monday so will be unresponsive for 2 days. 
And yes assign this one to me :grin:

morevnaproject:
Great! This issue is yours now. ^__^

ankit-kumar-dwivedi:
@morevnaproject working on it now :)

morevnaproject:
By default Synfig doesn't embeds images into its files. It just references them.
You can obtain image path from image layer - 

![screenshot_001](https://user-images.githubusercontent.com/332868/54048053-3f903680-420b-11e9-80bb-fd189a2f9a9e.png)

So, all you need to do is to launch binary (the one specified in preferences) and pass an image path as an argument to it. I.e.:

```
/usr/bin/krita /path/to/some/file.png
```

ankit-kumar-dwivedi:
@morevnaproject will send a PR tomorrow :)

ankit-kumar-dwivedi:
> So, all you need to do is to launch binary (the one specified in preferences) and pass an image path as an argument to it. I.e.:
> 
> ```
> /usr/bin/Krita /path/to/some/file.png
> ```

[...]

PR #805, opened 2019-04-06
Fixed part 2 of #516

ankit-kumar-dwivedi:
As this can be seen in fig now on a group containing a single image layer have an external edit tool option.
SS1
And when there are more than one image layer or any kind of layer the option is not shown.
SS2

@morevnaproject please review when you have time [^_^]

### DIFF
--- a/synfig-studio/src/gui/instance.cpp
+++ b/synfig-studio/src/gui/instance.cpp
@@ -1631,26 +1631,63 @@ void
 Instance::gather_uri(std::set<synfig::String> &x, const synfig::Layer::Handle &layer) const
 {
 	if (!layer || !layer->get_canvas()) return;
+	int count =0; //will be used to count layers in the group
+	
+	//check if the layer is group layer
+	ParamVocab vocab = layer->get_param_vocab();
+	synfig::Layer::Handle layerfinal,lay;
+	bool some_bool = false;
+	for(ParamVocab::const_iterator i = vocab.begin(); i != vocab.end(); ++i)
+	{
+		if(i->get_description() == "Group content")
+		{
+			etl::handle<synfig::Layer_PasteCanvas> p = etl::handle<synfig::Layer_PasteCanvas>::cast_dynamic(layer);
+			synfig::Canvas::Handle canvas = p->get_sub_canvas();
+			if(canvas)
+			{
+				for(IndependentContext i = canvas->get_independent_context(); *i; i++)
+				{
+					lay = (*i);
+					count++;
+				}
+				if(count==1)
+				{
+					some_bool = true;//yes only one layer 
+				}				
+			}
+		}
+	}
 
 	FileSystem::Handle file_system = layer->get_canvas()->get_file_system();
 	if (!file_system) return;
 
-	ParamVocab vocab = layer->get_param_vocab();
+	//if yes then the layer inside group should be processed not the group!
+	if(some_bool)
+	{
+		vocab = lay->get_param_vocab();
+		layerfinal = lay;
+	}
+	else
+	{
+		layerfinal = layer;
+	}
+	
+
 	for(ParamVocab::const_iterator i = vocab.begin(); i != vocab.end(); ++i)
 	{
 		if (i->get_hint() == "filename")
 		{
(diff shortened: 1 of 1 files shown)


morevnaproject:
Great! It is on review now. ^__^

ankit-kumar-dwivedi:
Okay @morevnaproject Please let me if there are improvements

synfig-studio/src/gui/instance.cpp
```diff
@@ -1631,26 +1631,63 @@ void
...
 	if (!layer || !layer->get_canvas()) return;
+	int count =0; //will be used to count layers in the group
+	
+	//check if the layer is group layer
+	ParamVocab vocab = layer->get_param_vocab();
+	synfig::Layer::Handle layerfinal,lay;
+	bool some_bool = false;
+	for(ParamVocab::const_iterator i = vocab.begin(); i != vocab.end(); ++i)
+	{
+		if(i->get_description() == "Group content")
```
morevnaproject: I suggest to use simpler method to check if layer is PasteCanvas.
You can blindly do ```dynamic_cast``` and then check if it returned a valid object. Like here:
https://github.com/synfig/synfig/blob/6eafe8e0af292b65c2dd7d39822b7531da045bdc/synfig-core/src/synfig/layers/layer_pastecanvas.cpp#L275-L276

synfig-studio/src/gui/instance.cpp
```diff
@@ -1631,26 +1631,63 @@ void
...
+		{
+			etl::handle<synfig::Layer_PasteCanvas> p = etl::handle<synfig::Layer_PasteCanvas>::cast_dynamic(layer);
+			synfig::Canvas::Handle canvas = p->get_sub_canvas();
+			if(canvas)
+			{
+				for(IndependentContext i = canvas->get_independent_context(); *i; i++)
+				{
+					lay = (*i);
```
morevnaproject: Refactoring: I suggest to use "child_layer" instead of "lay" as variable name.

synfig-studio/src/gui/instance.cpp
```diff
@@ -1631,26 +1631,63 @@ void
...
+					count++;
+				}
+				if(count==1)
```
morevnaproject: Here we can check if (count>1) and if true then break from the loop.

synfig-studio/src/gui/instance.cpp
```diff
@@ -1631,26 +1631,63 @@ void
...
+			}
+		}
+	}
 
 	FileSystem::Handle file_system = layer->get_canvas()->get_file_system();
 	if (!file_system) return;
 
-	ParamVocab vocab = layer->get_param_vocab();
+	//if yes then the layer inside group should be processed not the group!
+	if(some_bool)
```
morevnaproject: I suggest to check if (count == 1) instead. Then it is safe to not use some_bool at all.

synfig-studio/src/gui/instance.cpp
```diff
@@ -1631,26 +1631,63 @@ void
...
+	{
+		vocab = lay->get_param_vocab();
+		layerfinal = lay;
+	}
+	else
+	{
+		layerfinal = layer;
```
morevnaproject: I suggest to use "layer" instead of "layerfinal". Then, this code will be enough:

```
if (count == 1){
		vocab = lay->get_param_vocab();
		layer = lay;
	}

synfig-studio/src/gui/instance.cpp
ankit-kumar-dwivedi: Okay this looks way simpler thanks

synfig-studio/src/gui/instance.cpp
ankit-kumar-dwivedi: I guess it didn't worked due to being const

ankit-kumar-dwivedi:
@morevnaproject Done Changes!

synfig-studio/src/gui/instance.cpp
morevnaproject: Ah, I understand! ^__^

morevnaproject:
Looks good now! Thank you! ^__^

PR #808, opened 2019-04-09
Remove legacy PKG_CHECK_MODULES from m4/gnome.m4

onlyjob:
Remove broken, oudated, embedded copy of PKG_CHECK_MODULES from
 m4/gnome.m4. Removal causes synfigstudio to pick up the current version
 from the pkg-config package. The outdated version makes cross building
 fail.

Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=900433
From: Helmut Grohne <helmut@subdivi.de>

### DIFF
--- a/synfig-studio/m4/gnome.m4
+++ b/synfig-studio/m4/gnome.m4
@@ -64,63 +64,6 @@ else
 fi])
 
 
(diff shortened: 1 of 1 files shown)


morevnaproject:
Thank you for the fix!

Issue #773, opened 2019-03-14
Should ask for File when import image/sound layer is invoked


ankit-kumar-dwivedi:
**Synfig version & platform**:
Development Build

**Issue description**:
Linux Mint OS

How to recreate?
**layer->new layer->other->import image**
**layer->new layer->other->sound**

ankit-kumar-dwivedi:
**I am already working on a fix :)**

ankit-kumar-dwivedi:
@morevnaproject I am very strongly willing to remove the multiple code file present for the same work.
There is a dedicated file for only image import but also there is more general import function written which makes the "image import" redundant So I would like to tell you that I am willing to delete the dedicated code for only "sound" and only "image import" and use the function from general import for these two layers. Would love to hear your opinion on this (^_^)
Thank you !

morevnaproject:
Can you please be more specific which functions you are referencing (and in which files)? ^__^

ankit-kumar-dwivedi:
@morevnaproject Ah my bad I though 
https://github.com/synfig/synfig/blob/6b3c81c2fb73a7a96cfddb0e43de54fe7a61da2e/synfig-studio/src/gui/canvasview.cpp#L3381 
this to be only for images. Which created all this confusion sorry for that!

morevnaproject:
No problem! ^__^

ankit-kumar-dwivedi:
@morevnaproject do we really need these 2 options? when there is "import" which can import any file

morevnaproject:
Do you mean those two options? -

> layer->new layer->other->import image
> layer->new layer->other->sound

ankit-kumar-dwivedi:
Yes aren't these two options redundant @morvenaproject

morevnaproject:
Each option produces different type of layer - Image Layer and Sound Layer.

ankit-kumar-dwivedi:
> @morevnaproject do we really need these 2 options? when there is "import" which can import any file



> Yes aren't these two options redundant @morvenaproject
> […](#)

@morevnaproject I am sorry for my confusing comment  ¯\\_(⊙︿⊙)_/¯
I wanted to explain 2 points:

1.  The two options "Import image" and "sound" are meant to import image and sound files. 
![Screenshot from 2019-03-27 19-59-58](https://user-images.githubusercontent.com/37617951/55085447-b44ce700-50cc-11e9-8c5d-1a1fb6cdce55.png)
but when the same work can be done using the "File->import.."
[...]

PR #816, opened 2019-04-19
fix ask for file when layer->new layer->other->sound/image is invoked

ankit-kumar-dwivedi:
fixes: #773 

@morevnaproject 
Please review and suggest if changes are required

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -1652,40 +1652,43 @@ void
 CanvasView::add_layer(String x)
 {
 	Canvas::Handle canvas;
-
 	SelectionManager::LayerList layer_list(get_selection_manager()->get_selected_layers());
-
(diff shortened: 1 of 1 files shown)


morevnaproject:
Thank you for the fix! It prompts for file as expected now. But when I import image, the layer dimensions are always set to square (60x60 px) -

screenshot_001

So, it would be nice to retrieve image dimensions and set layer parameters accordingly.
Thank you! ^__^

ankit-kumar-dwivedi:
@morevnaproject was working on input dialog for vectorizer! will do the fix now [^_^]

ankit-kumar-dwivedi:
Done the required changes but while checking import audio I encounter #819

ankit-kumar-dwivedi:
@morevnaproject done changes please review

morevnaproject:
Looks good to me now! Thank you! ^__^

PR #818, opened 2019-04-23
core/geometry: Fix crash for degenerate looping primitives

bobbybee:
Note that we index two nodes back from the end, which is a buffer
underflow for degenerate primitives (lines, mostly), causing a crash on
select sif's (that were fine on older Synfig versions... Now we don't
crash, but hang rendering...)

Signed-off-by: Alyssa Rosenzweig <alyssa@rosenzweig.io>

### DIFF
--- a/synfig-core/src/modules/mod_geometry/outline.cpp
+++ b/synfig-core/src/modules/mod_geometry/outline.cpp
@@ -227,8 +227,11 @@ Outline::sync_vfunc()
 		Vector first_tangent = bline.front().get(blank).get_tangent2();
 		Vector last_tangent = first_point.get_tangent1();
 
-		// if we are looped and drawing sharp cusps, we'll need a value for the incoming tangent
-		if (loop && sharp_cusps && last_tangent.is_equal_to(Vector::zero())) {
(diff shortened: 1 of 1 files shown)


ice0:
Awesome! Thank you for cooperation!

morevnaproject:
I have a strong assumption that the code inside of "if" block works for the case when `bline.size() == 2`.

```
if (loop && sharp_cusps && last_tangent.is_equal_to(Vector::zero()) && bline.size() > 2) {
    ValueBase::List::const_iterator prev = iter; --prev;
    const BLinePoint &prev_point = prev->get(blank);
    etl::hermite<Vector> curve(
	    prev_point.get_vertex(),
	    first_point.get_vertex(),
	    prev_point.get_tangent2(),
	    first_point.get_tangent1() );
    const etl::derivative< etl::hermite<Vector> > deriv(curve);
    last_tangent = deriv(1.0 - CUSP_TANGENT_ADJUST);
}
```

So, I think the condition should be `bline.size() >= 2` instead of `bline.size() > 2`.

Please correct me if I am wrong. ^__^

morevnaproject:
Ping @blackwarthog

PR #840, opened 2019-06-09
Update autotools files

rodolforg:
Minor update to less warnings. Autoconf 2.60 is already 11 years old, so I think there is no problem here.
http://git.savannah.gnu.org/gitweb/?p=autoconf.git


ice0:
Looks good to me!
Thank you!

PR #866, opened 2019-06-30
pass arguments by reference for performance reasons

rodolforg:


morevnaproject:
@blackwarthog Can you please review this? ^__^

blackwarthog:
> @blackwarthog Can you please review this? ^__^

I'm afraid that we have a risk that RendDesc will destroyed while processing a function.
And i cannot predict behavior of function sigc::bind in synfig-studio/src/gui/asyncrenderer.cpp in async_render_tile()

https://github.com/synfig/synfig/blob/8620d102ed4ea69ecd1f3df6220948e4e12a1123/synfig-studio/src/gui/asyncrenderer.cpp#L138

Performance anvantage is very very small and also classes Target_* and AsyncRenderer is deprecated (but still in use) and should be refactored or removed in future (for performance reasons :)

So i think will be better to keep these params without references.

PR #867, opened 2019-06-30
fix some memory leaks on svg parser

rodolforg:
Some Converity memory leak reports fixed

### DIFF
--- a/synfig-core/src/modules/mod_svg/svg_parser.cpp
+++ b/synfig-core/src/modules/mod_svg/svg_parser.cpp
@@ -252,6 +252,7 @@ Svg_parser::parser_graphics(const xmlpp::Node* node,xmlpp::Element* root,String 
 		}
 		if(nodename.compare("g")==0){
 			parser_layer (node,root->add_child("layer"),parent_style,mtx);
+			free(mtx);
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -252,6 +252,7 @@ Svg_parser::parser_graphics(const xmlpp::Node* node,xmlpp::Element* root,String
 		}
 		if(nodename.compare("g")==0){
 			parser_layer (node,root->add_child("layer"),parent_style,mtx);
+			free(mtx);
```
blackwarthog: `mtx` can be NULL here

synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -302,6 +303,7 @@ Svg_parser::parser_graphics(const xmlpp::Node* node,xmlpp::Element* root,String
 				build_fill (child_fill,fill,NULL);
 			}
 			parser_effects(nodeElement,child_layer,parent_style,mtx);
+			free(mtx);
```
blackwarthog: `mtx` can be NULL here too

blackwarthog: Oops.. If arg is a null pointer, the function `free` does nothing.
ok

PR #868, opened 2019-06-30
Cov 308605

rodolforg:
Fix Coverity issue cov#308605

Maybe we should just use `std::find()` instead of `for(i=0,iter=studio::App::instance_list.begin();iter!=studio::App::instance_list.end() && ((*iter)!=x);iter++,i++)`

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_compselect.cpp
+++ b/synfig-studio/src/gui/widgets/widget_compselect.cpp
@@ -116,7 +116,7 @@ Widget_CompSelect::set_selected_instance(etl::loose_handle<studio::Instance> x)
 			set_active(i);
 		} else {
 			synfig::warning("Can't set selected instance! (already closed?)");
-			iter = studio::App::instance_list.begin();
+			return;
 		}
 	}
 	else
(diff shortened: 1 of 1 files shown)


morevnaproject:
@blackwarthog Please let me know your opinion on this.

blackwarthog:
On Wed, 03 Jul 2019 00:35:27 -0700
Konstantin Dmitriev <notifications@github.com> wrote:

> @blackwarthog Please let me know your opinion on this.

in widget_compselect.cpp:119, seems to be better to clear `x` ( `x.reset()` ) instead of return from function

https://github.com/synfig/synfig/blob/ede78dc50daebe91e2df296f7931188850ced31c/synfig-studio/src/gui/widgets/widget_compselect.cpp#L119

seems this code is dead, i cannot find any references to Widget_CompSelect in the code

i think it is a combobox for select one of sif-files currently opened in editor (Instance)
i don't know do we use it in future or not

blackwarthog:
> Maybe we should just use std::find() instead of
> for(i=0,iter=studio::App::instance_list.begin();iter!=studio::App::instance_list.end() && ((*iter)!=x);iter++,i++)

`find` does not return an index and we cannot to determine index by iterator - operator `-` is not applicable for iterators of linked lists

synfig-studio/src/gui/widgets/widget_compselect.cpp
```diff
@@ -111,12 +111,12 @@ Widget_CompSelect::set_selected_instance(etl::loose_handle<studio::Instance> x)
...
 			;
 
-		if (*iter==x) 
+		if (iter != studio::App::instance_list.end())
 		{
 			set_active(i);
 		} else {
 			synfig::warning("Can't set selected instance! (already closed?)");
-			iter = studio::App::instance_list.begin();
+			return;
```
blackwarthog: i think we need `x.reset()` instead of `return` here

PR #876, opened 2019-07-06
simplification of etl_config.h

rodolforg:
remove commented/unused `#define`s and get rid of #define for `extern "C"`


morevnaproject:
Merged now. Thank you! ^__^

PR #877, opened 2019-07-06
update POTFILES.in - new files added

rodolforg:
When running `make check`, it fails due to these missing files


morevnaproject:
Thank you! ^__^

PR #878, opened 2019-07-07
remove an unneeded mutex wrapper

rodolforg:
Its only "initialized" property is totally useless, so Mutex class can be used directly

### DIFF
--- a/synfig-core/src/synfig/main.cpp
+++ b/synfig-core/src/synfig/main.cpp
@@ -101,18 +101,7 @@ using namespace synfig;
 static etl::reference_counter synfig_ref_count_(0);
 Main *Main::instance = NULL;
 
-class GeneralIOMutexHolder {
-private:
-	Mutex mutex;
-	bool initialized;
-public:
-	GeneralIOMutexHolder(): initialized(true) { }
-	~GeneralIOMutexHolder() { initialized = false; }
-	void lock() { if (initialized) mutex.lock(); }
-	void unlock() { if (initialized) mutex.unlock(); }
-};
-
-GeneralIOMutexHolder general_io_mutex;
+Mutex general_io_mutex;
 
 /* === P R O C E D U R E S ================================================= */
 
--- a/synfig-core/src/synfig/main.cpp
+++ b/synfig-core/src/synfig/main.cpp
@@ -101,7 +101,7 @@ using namespace synfig;
 static etl::reference_counter synfig_ref_count_(0);
 Main *Main::instance = NULL;
 
-Mutex general_io_mutex;
+static Mutex general_io_mutex;
 
 /* === P R O C E D U R E S ================================================= */
 


rodolforg:
Just to be clear:
`initialized` is private without accessor methods. This property is set true in constructor, and just change (for no reason) to false in destructor. So it is useless.

morevnaproject:
This code is introduced in b297b1b5ab0ede919547dc05da2476d98003a0df by @blackwarthog 

I guess its purpose is to avoid calling `unlock()` on destroyed mutex object? @blackwarthog any comments?

rodolforg:
`general_io_mutex` is destructed only with program's ending, because it is a 'global' object.
I'll submit a patch to ensure that according to [this site](https://www.geeksforgeeks.org/static-objects-destroyed/).

blackwarthog:
The static keyword in global space does not change a destructors behavior.

In this case 'static' means that this variable will not export from this object file (cpp-file) to avoid collisions with other global variable with the same name in other files.

For global object you cannot predict an order of calling them destructors. It depends on compiler, compiler options and other factors. Standard says that order of calling constructors and destructors for global variables is undefined.

Wrapper used to avoid crashing while synfig closing. It's required for batch processing of video rendering. Any error code from synfig will stop whole rendering process.

Synfig has many static variables and we need to be sure that them destructors will not call mutex methods after mutex is destroyed. Destructors of static variables does not use multithreading so dummy-mutex is useful in this case.

But a 'static' keyword we actually need here, because we do not need to export this var for other .cpp files.

rodolforg:
I thought so, but that article says that those 'globals' set as static seems to be destructed after `main()` execution.
Anyway:
1. `general_io_mutex` is used only to avoid clash of `std::cerr`/`std::cout` outputs. (functions `synfig::error(const String &str)`, `synfig::warning(const String &str)`, and `synfig::info(const String &str)`).
2. I can't see why this wrapper does guarantee anything: if this `GeneralIOMutexHolder::unlock()` is called after object destruction, there is _no_ guarantee that its now gone `GeneralIOMutexHolder::initialized` contains a valid value (false) to avoid calling `Mutex::unlock()`.
3. By the way, there are some direct calls to `std::cerr`/`std::cout` in synfig-core and -studio source code.

blackwarthog:
> I thought so, but that article says that those 'globals' set as static seems to be destructed after `main()` execution.

it's right, any global variable will destructed after main() execution, and article says: static keyword will transform local variable to global (in some aspects) and this var as any global variable will destructed after main() execution



> Anyway:
> 1. `general_io_mutex` is used only to avoid clash of `std::cerr`/`std::cout` outputs. (functions
> `synfig::error(const String &str)`, `synfig::warning(const String &str)`, and `synfig::info(const String &str)`).

yes, it's the aim

> 2. I can't see why this wrapper does guarantee anything: if this `GeneralIOMutexHolder::unlock()`
> is called after object destruction, there is _no_ guarantee that its now gone
> `GeneralIOMutexHolder::initialized` contains a valid value (false) to avoid calling `Mutex::unlock()`.

it's not right, c++ does not allocate and does not deallocate memory for global variables, they are part of the program dump and memory for them reserved in the executable file at the compilation time. Memory of global vars will never used for other purposes.

> 3. By the way, there are some direct calls to `std::cerr`/`std::cout` in synfig-core and -studio source code.

It's a little bad practice and in some cases may cause errors. Originally synfig not used threads and we still have a lot of non-thread-safe code which causes suddenly falls down.

rodolforg:
> [...], and article says: static keyword will transform local variable to global (in some aspects) and this var as any global variable will destructed after main() execution

Yeah, I know it, but the article's point is the difference of destructor calling time between 'static' and not 'static' 'global' variables.

> it's not right, c++ does not allocate and does not deallocate memory for global variables, they are part of the program dump and memory for them reserved in the executable file at the compilation time. 
Memory of global vars will never used for other purposes.

When I coded to Sega Genesis, I saw this kind of stuff, but I was not sure if it is a standard thing. Thank you for clarifying it to me. I will close this PR then.

> > 3. By the way, there are some direct calls to `std::cerr`/`std::cout` in synfig-core and -studio source code.

> It's a little bad practice and in some cases may cause errors. Originally synfig not used threads and we still have a lot of non-thread-safe code which causes suddenly falls down.

So I'll fix it in another PR.

rodolforg:
But— what about std::cerr / cout / clog being thread-safe by default in C++11 ?

https://en.cppreference.com/w/cpp/io/cerr

So, we could just change that number of << calls from

```cpp
void
synfig::warning(const String &str)
{
	general_io_mutex.lock();
	cerr<<"\033[33m"<<"synfig("<<getpid()<<")"<<current_time().c_str()<<_("warning")<<": "<<str.c_str()<<"\033[0m"<<endl;
	general_io_mutex.unlock();
}
```
to

```cpp
void
synfig::warning(const String &str)
{
	std::ostringstream str;
	str << "\033[33m" << "synfig(" << getpid() << ")" << current_time().c_str() << _("warning") << ": " << str << "\033[0m" << endl;
	cerr << str.str();
}
```

blackwarthog:
> But— what about std::cerr / cout / clog being thread-safe by default in C++11 ?
> 
> https://en.cppreference.com/w/cpp/io/cerr

hm.... it's good
 
> So, we could just change that number of << calls from
> 
> ```cpp
> void
> synfig::warning(const String &str)
> {
> 	general_io_mutex.lock();
> 	cerr<<"\033[33m"<<"synfig("<<getpid()<<")"<<current_time().c_str()<<_("warning")<<": "<<str.c_str()<<"\033[0m"<<endl;
> 	general_io_mutex.unlock();
> }
> ```
> to
> 
> ```cpp
> void
> synfig::warning(const String &str)
> {
> 	std::ostringstream str;
> 	str << "\033[33m" << "synfig(" << getpid() << ")" << current_time().c_str() << _("warning") << ": " << str << "\033[0m" << endl;
> 	cerr << str.str();
> }
> ```


I'm not sure if it's a proper way. I've checked standard and it talls:

> §27.4.1.4 Concurrent access to a synchronized (27.5.3.4) standard iostream object’s formatted and unformatted input (27.7.2.1) and output (27.7.3.1) functions or a standard C stream by multiple threads shall not result
in a data race (1.10). [ Note: Users must still synchronize concurrent use of these objects and streams by
multiple threads if they wish to avoid interleaved characters. — end note ]

Authors of cplusplus.com understand it so:
> no data races are initiated, although no guarantees are given on the order in which characters from multiple threads are inserted ( http://www.cplusplus.com/reference/ostream/ostream/operator%3C%3C/ )

I think will be better to follow recomendations of standard and synchronize threads by mutex

PR #879, opened 2019-07-10
applies/forces etl:: namespace for ETL library

rodolforg:
The Synfig-core makes use of “etl::” namespace directly
almost everywhere.
so I think this kind of macro useless.

'Discussed' here:
https://forums.synfig.org/t/forcing-etl-namespace/9938


rodolforg:
Rebased to master

blackwarthog:
thanks

PR #880, opened 2019-07-10
remove redundant and unused function RendDesc::apply()

rodolforg:
One could simply use assignment operator;
the "apply" method implementation is just a wrapper
for the former.


morevnaproject:
Looks good, thank you! ^__^

PR #890, opened 2019-07-24
Fix some key mapping

rodolforg:


synfig-studio/src/gui/trees/layertree.cpp
```diff
@@ -105,8 +105,8 @@ using namespace studio;
 bool LayerTree::onKeyPress(GdkEventKey* event)
 {
 
-	switch (event->hardware_keycode) {
-		case 119: { // 'Del' key pressed
+	switch (event->keyval) {
+		case GDK_KEY_Delete: { // 'Del' key pressed
 			std::cout << "LayerTree::onKeyPress: Got delete!" << std::endl;
```
morevnaproject: I suggest to remove this line. ^__^

morevnaproject:
Hi! Thank you very much for this long-awaited fix! The DEL key now works perfectly!
Howm and End keys didn't worked for me for some reason, but I suspect this is a problem with my unusual keyboard. I will try with another one and get back to you. ^__^

synfig-studio/src/gui/trees/layertree.cpp
rodolforg: Really? :/ It took me a long time to track where/if delete was used due to non-enum key usage.
Same about method name not being "on_key_press_down()".
I read `git blame` and this code piece was written only in commits 27966f3919 and 65643eee77 .
Apparently, there is no particular reason to avoid `keyval` and `GDK_KEY_*` enums.
Should I really remove those 3 lines (the one following too).

rodolforg:
This is what I made:

1. When I [open a | create a new | change tab to a] Synfig file, I can press Home/End as for current releases.
2. If "canvas dock" is out of focus, those keys have its normal behavior for the dock on focus.
3. If on focus, it checks if none of its widgets makes use first, before forcing the frame seeking.

synfig-studio/src/gui/trees/layertree.cpp
morevnaproject: I meant only this line ^__^ -
```
std::cout << "LayerTree::onKeyPress: Got delete!" << std::endl;
```
(the one which produces debug output to console).

rodolforg: Oh. Done :)

rodolforg:
Any news about Home & End accel keys?
What do you do that doesn't work as expected?

morevnaproject:
I've tested them with other keyboard and Home/End keys worked perfectly! (the problem with other keyboard was that it had limited set of keys and I've used Home/End from numpad)
Thank you for the fixes!m ^__^

PR #892, opened 2019-07-24
Home and End keys of numpad can also change current time

rodolforg:


rodolforg:
Fix https://github.com/synfig/synfig/pull/890#issuecomment-514718843

morevnaproject:
Awesome! Works perfectly now, thank you! ^__^

PR #897, opened 2019-07-27
Fix more memory leaks in Studio

rodolforg:


morevnaproject:
Merged now. Thank you! ^__^

PR #901, opened 2019-07-30
Minor cleaning mainwindow

rodolforg:


morevnaproject:
Hi! I have tested this PR and found one problem. When I start Synfig and choosing "View" - "Show Menubar" checkbox it doesn't work first time. When I do same command again then it works. That happens only first time after Synfig starts  - all subsequent calls of "View" - "Show Menubar" work as expected.

rodolforg:
Ugh. Sorry.
But here in Gnome it vanishes the menu, how can you make it show again?

morevnaproject:
@rodolforg Yeah, that's a common problem. ^__^ You can use button in top-left corner of canvas to call menu:

screenshot_006

morevnaproject:
Merged now. Thank you! ^__^

PR #902, opened 2019-07-30
remove unused/unneeded localization function

rodolforg:
just using _() / gettext() already defined in this header file

### DIFF
--- a/synfig-studio/src/gui/localization.h
+++ b/synfig-studio/src/gui/localization.h
@@ -46,9 +46,6 @@
 
 /* === C L A S S E S & S T R U C T S ======================================= */
 
-inline const char* synfiggui_localize(const char *x)
-	{ return _(x); }
-
 /* === E N D =============================================================== */
 
 #endif


synfig-studio/src/gui/localization.h
```diff
@@ -46,9 +46,6 @@
 
 /* === C L A S S E S & S T R U C T S ======================================= */
 
-inline const char* synfiggui_localize(const char *x)
-	{ return _(x); }
-
 /* === E N D =============================================================== */
```
blackwarthog: this function required for compile-time check for valid localization header

we have three `localization.h` in synfig-core, in synfigapp and in gui.
they differs by _ definition:
```
    #define _(x) dgettext("synfig",x)       // core
    #define _(x) dgettext("synfigstudio",x) // app
    #define _(x) gettext(x)                 // gui
```

if somebody includes wrong header file, then localization breaks without any message
in macros for actions we used synfigapp_localize to force the compiler error in this case

also we have added synfigcore_localize and synfiggui_localize for possible use in future

blackwarthog:
really two of three language functions are not used yet, but we will keep these functions for possible compile-time checks in future

PR #903, opened 2019-07-31
fix envvar in build.conf.sample used by build scripts

rodolforg:


morevnaproject:
@rodolforg Thank you for the fix! ^__^

PR #906, opened 2019-08-04
Check for printf functions when building ETL using cmake

duskydawg:
Since vasprintf isn't available in all implementations of stdio.h, we should check before we define HAVE_VASPRINTF.

### DIFF
--- a/ETL/CMakeLists.txt
+++ b/ETL/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.0)
 
 project(ETL)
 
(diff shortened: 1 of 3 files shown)


blackwarthog:
Since we restricted c++ >= c++11 we does not need to use non-standart aprintf/vaprintf functions.
We can use vsnprintf instead.

duskydawg:
Amended to remove vasprintf and rebased to master.

blackwarthog:
done

blackwarthog:
i was worked at the same problem too and already did it.
sorry that i hadn't coordinate it with you.
thanks for your work

ETL/CMakeLists.txt
```diff
@@ -7,8 +7,8 @@ if (CMAKE_USE_PTHREADS_INIT)
     set(HAVE_LIBPTHREAD ON)
 endif()
 
-# TODO: check for it
-set(HAVE_VASPRINTF ON)
+include(CheckCXXSymbolExists)
+check_cxx_symbol_exists(vsnprintf stdio.h HAVE_VSNPRINTF)
```
blackwarthog: vsnprintf must exist in any c++11 compiler by standard, so i've removed any checks from CMakeLists.txt and configure.ac

PR #911, opened 2019-08-07
Fix 910 sketch buttons

rodolforg:


morevnaproject:
Merged now. Thank you! ^__^

PR #915, opened 2019-08-10
Fix #243: Save toolbox layout

### LINKED ISSUE #243: In the ToolBox tab, the color selection window does not keep his position saved.
I set the program interface to my liking and put the ToolBox tab at the top left and under this the options tool window. But when I start Synfig again, the window of color selection moves and is underhanded.

synfig_01

synfig_02
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38567649-in-the-toolbox-tab-the-color-selection-window-does-not-keep-his-position-saved?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

rodolforg:


synfig-studio/src/gui/docks/dock_toolbox.cpp
```diff
@@ -134,6 +134,23 @@ Dock_Toolbox::~Dock_Toolbox()
 		studio::App::dock_toolbox=NULL;
 }
 
+void Dock_Toolbox::write_layout_string(string& params) const
+{
+	char num_str[6];
+	snprintf(num_str, 5, "%d", tool_box_paned->get_position());
+	params += std::string(num_str);
```
morevnaproject: @rodolforg Sorry, I did not understanding - why use "+=" instead of "="? ^__^"

morevnaproject:
Also, there is one more issue: 
- Launch Synfig with modifications of this PR.
- Close Synfig.
- Launch version of Synfig **without** modifications of this PR (previous version).
- Result: Synfig starts with incorrect UI layout - almost all dock panels disappeared and only workarea window available.

screenshot_002


This happens because this PR changes how window layout is stored in settings file and Synfig fails to read layout.

Suggestion: Store Toolbox panel layout differently, in separate variable. Something like `pref.toolbox_paned_position`.

rodolforg:
I started this way, by using preferences. However, it wouldn't save the
handle position in custom workspace layouts.

Em sex, 16 de ago de 2019 07:33, Konstantin Dmitriev <
notifications@github.com> escreveu:

synfig-studio/src/gui/docks/dock_toolbox.cpp
rodolforg: I don't know. lol

morevnaproject:
> it wouldn't save the handle position in custom workspace layouts.

Hmm... I see, you right. So, we should think about other options to avoid breaking UI layout in older versions.

morevnaproject:
We can change name of settings file - use "settings-1.3.12" instead of "settings-1.3". How about that?

rodolforg:
Or two `docks=` lines?
the first one cleaner (without that change) and another like `docks_1_3_12=` ?

morevnaproject:
I like that idea!

> another like `docks_1_3_12=` ?

...just want to think about something other than `docks_1_3_12`. How about 'workspace.layout' instead of 'docks_1_3_12.layout'?

rodolforg:
maybe we should keep `docks` prefix and just change the suffix?
`docks.layout` would be left untouched by version > 1.3.12.
These newer and future versions would use `docks.last_layout`? or `docks.workspace_layout`…
I don't know.

morevnaproject:
Hi! I have tested latest changes. Now it doesn't damages configuration for previous versions of Synfig, but when starting a new version (the one with changes from this PR) it starts with messed up window layout - only workspace window and several dock panels floating, lie on [this image](https://user-images.githubusercontent.com/332868/63162096-ed6dff80-c04b-11e9-86f0-e0c665a57f9d.png).

I guess that's because `workspace.layout` is empty and should be set to some default value?

rodolforg:
> I guess that's because `workspace.layout` is empty and should be set to some default value?

I guess so. However, I can't figure out how to solve it.

morevnaproject:
@rodolforg Hi! Do you need any help with finishing this? ^__^

rodolforg:
Done. Thanks for the tip.

morevnaproject:
Merged now via 78cb621ddeb92a1f912a5b1eeafb30fe1c5602cc. Thank you! ^__^

PR #940, opened 2019-09-01
fix mod_ffmpeg with cmake build

eroen:
With cmake, the autotools definitions are not implicitly defined. This led to
the windows codepath being used everywhere, causing breakage on unix.

The necessary definitions have been manually added to the cmake build.


ice0:
Thank you!

PR #970, opened 2019-10-04
Removed unnecessary ETL/pen includes

ice0:


blackwarthog:
`apen` is still in use

apen without a base prototype (pen) looks strange

i think to keep pen template for now

ice0:
I do not remove ETL/pen, just clean some files.

blackwarthog:
oo... hm...

approved

PR #971, opened 2019-10-06
fix layertree crashing on closing studio

rodolforg:
That leak doesn't happen.
I think layertree should be responsible to delete them though.
However, synfig studio code is a bit tangled.


ice0:
Ok, thank you @rodolforg 
@blackwarthog  fixed this behavior by [this ](https://github.com/synfig/synfig/commit/e04f20a745b113bd5ac9dc122a31a59da55f1f85) commit.

Issue #451, opened 2017-11-02
Color wheel fills history too much


zozorg:
**Synfig version & platform**:
linux 1.3.4

**Issue description**:
When changing the color of an existing object with the color wheel, dragging the mouse creates a ton of history elements, presumably one for every color your mouse dragged over. This uselessly fills history and is painful to revert with ctrl+z.

morevnaproject:
Possibly we can have a checkbox, which tells if Synfig should dynamically update workarea during modifications in Colors window.

caryoscelus:
This can be better fixed by joining color change action sequence into one action. Not sure how easy that would be, though.

morevnaproject:
@rodolforg Can we solve that in the same way as you did in #982? (https://github.com/synfig/synfig/pull/982/commits/75a93af2e663ee29b92b54c05bf5fcecf5b3e4d0)

rodolforg:
It is not that easy in this case: there's is no signal for "user is 'dragging' in color widget".
And to make it worse, the widget is deprecated in Gtk :/

morevnaproject:
Oh, so replacing a color widget would be a higher priority.
I wonder, what Gtk offer intead of deprecated widget?

ebarranco:
A workaround, is calling the color wheel from the outline/fill color,  for some reason here does not generate steps in the history stack. [I show you here,](https://youtu.be/T-hXPxQ2fl0) but we definitely need a working HSV color selection solution.

rodolforg:
Well, I tested the new ColorChooser, but it has the same problem: I can't detect when user press mouse button (and drag it) over the specific color widget (the deprecated wheel or current plane).

So my proposal it to do our own widget - or wait for Gtk team to accept our possible patch.

Other possible solution: somehow we kidnap events from child widget of (deprecated) GtkColorSelection or (current) GtkColorChooserWidget to our parent.

This way we could process/detect left-mouse button dragging (beginning and ending) to make a Passiv action that group those several actions of color changing into a single one.
However I didn't find a way to steal those Gtk event priority to parent widget.

Keyikedalube:
What if **Apply** and **Cancel** buttons are added to the color dialog instead. That way when a user accepts changes from the dialog only one history is recorded, thus, fixing the inconvenient undo behavior.

rodolforg:
It's a solution, but you miss the visual feedback in image itself that helps you to properly choose a new color

Keyikedalube:
[...]

PR #982, opened 2019-10-22
Improves Graphs widget/panel (related to #267)

### LINKED ISSUE #267: Waypoint editable graph represention mode
Alternative timetrack dock/mode, which would present parameter values as a graph (as graphs dock currently does) *and* allow to edit waypoints, including their parameters (e.g. curve handles for curve interpolation)

rodolforg:
Improves Graph widget features.

Partially implements #267 requests.

Current status:
* [x] draw waypoints in curves
* [x] allow single waypoint selection
* [x] allow multiple waypoint selection
  * [x] via `shift` (adding to selection) and `control` (toggling selection state) keys
  * [x] via mouse dragging rectangle
  * [x] cancelable while mouse-dragging with `right-click`
* [x] allow waypoint value change by dragging orarrow keys
  * [x] `shift` key constraints dragging
  * [x] cancelable while dragging with `right-click` or `Esc`
* [x] allow multiple waypoint value change by dragging
* [x] allow to move (multiple) waypoint time by dragging or arrow keys
* [x] select all visible nodes with `Ctrl+A`
* [ ] select all nodes of single channel with `Ctrl+Shift+A`?
* [ ] ~~show control points for TCB~~
* [x] double click in waypoint shows edit dialog
* [ ] ~~create Bézier waypoint type~~
* [ ] ~~add control points for Bézier waypoints~~
* [x] zoom to pointer/mouse position
* [x] pan with `middle-button` click and drag
* [x] show tooltip for hovered point
* [x] `Del` removes selected waypoints
* [x] Double-click in a curve creates a new waypoint

Items ~~striked-through~~ will be implemented in another PR.

### DIFF
--- a/synfig-core/src/synfig/timepointcollect.cpp
+++ b/synfig-core/src/synfig/timepointcollect.cpp
@@ -139,6 +139,76 @@ synfig::waypoint_collect(set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 	return 0;
 }
 
+bool
+synfig::waypoint_search(Waypoint& waypoint, const UniqueID &uid, const etl::handle<Node> &node)
+{
+	// Check if we are a linkable value node
+	LinkableValueNode::Handle linkable_value_node;
+	linkable_value_node=linkable_value_node.cast_dynamic(node);
+	if(linkable_value_node)
+	{
+		const int link_count(linkable_value_node->link_count());
+		for(int i=0; i < link_count; i++) {
+			bool ret = waypoint_search(waypoint, uid, linkable_value_node->get_link(i).get());
+			if (ret)
+				return true;
+		}
+		return false;
+	}
+
+	// Check if we are a layer
+	Layer::Handle layer;
+	layer=layer.cast_dynamic(node);
+	if(layer)
+	{
+		const Layer::DynamicParamList& dyn_param_list(layer->dynamic_param_list());
+
+		for(Layer::DynamicParamList::const_iterator iter=dyn_param_list.begin(); iter!=dyn_param_list.end(); ++iter) {
+			bool ret = waypoint_search(waypoint, uid, iter->second);
+			if (ret)
+				return true;
+		}
+
+		ValueBase canvas_value(layer->get_param("canvas"));
+		if(canvas_value.get_type()==type_canvas)
+		{
+			bool ret = waypoint_search(waypoint, uid,
+									  Canvas::Handle(canvas_value.get(Canvas::Handle())));
+			if (ret)
+				return true;
+		}
+		return false;
+	}
+
+	// Check if we are a canvas
+	Canvas::Handle canvas;
+	canvas=canvas.cast_dynamic(node);
+	if(canvas)
+	{
+		for(Canvas::const_iterator iter = canvas->begin(); iter!=canvas->end(); ++iter) {
+			bool ret = waypoint_search(waypoint, uid, *iter);
+			if (ret)
+				return true;
+		}
+		return false;
+	}
+
+	// Check if we are an animated value node
+	ValueNode_Animated::Handle value_node_animated;
+	value_node_animated=value_node_animated.cast_dynamic(node);
+	if(value_node_animated)
+	{
+		ValueNode_AnimatedInterfaceConst::const_findresult result = value_node_animated->find_uid(uid);
+		if (!result.second)
+			return false;
+
+		waypoint = *result.first;
+		return true;
+	}
+
+	return false;
+}
+
 //! \writeme
 int
 synfig::activepoint_collect(set<Activepoint, std::less<UniqueID> >& /*activepoint_set*/,const Time& time, const etl::handle<Node>& node)
--- a/synfig-core/src/synfig/timepointcollect.h
+++ b/synfig-core/src/synfig/timepointcollect.h
@@ -44,6 +44,9 @@ namespace synfig {
 //! \writeme
 int waypoint_collect(std::set<Waypoint, std::less<UniqueID> >& waypoint_set,const Time& time, const etl::handle<Node>& node);
 
+//! Search for a specific waypoint (by its uid) in node.
+bool waypoint_search(Waypoint& waypoint, const UniqueID& uid, const etl::handle<Node>& node);
+
 //! \writeme
 int activepoint_collect(std::set<Activepoint, std::less<UniqueID> >& activepoint_set,const Time& time, const etl::handle<Node>& node);
 
--- a/synfig-core/src/synfig/valuenodes/valuenode_animatedinterface.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_animatedinterface.cpp
@@ -1111,8 +1111,6 @@ ValueNode_AnimatedInterfaceConst::waypoint_is_only_use_of_valuenode(Waypoint &wa
 void
 ValueNode_AnimatedInterfaceConst::erase(const UniqueID &x)
 {
-	// \todo printf?
-	printf("%s:%d erasing waypoint from %lx\n", __FILE__, __LINE__, uintptr_t(this));
 	WaypointList::iterator iter(find(x));
 	Waypoint waypoint(*iter);
 	assert(waypoint.get_value_node());
--- a/synfig-studio/src/gui/CMakeLists.txt
+++ b/synfig-studio/src/gui/CMakeLists.txt
@@ -50,6 +50,7 @@ target_sources(synfigstudio
         "${CMAKE_CURRENT_LIST_DIR}/splash.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/statemanager.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/valuelink.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/waypointrenderer.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/workarea.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/workspacehandler.cpp"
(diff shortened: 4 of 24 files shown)


rodolforg:
@morevnaproject Could you please check if is the building error my fault?
And test if curve points selection works how it should? (as well as @ebarranco )

ice0:
@rodolforg Hi, Rodolfo!
The build is failed because travis-ci is updated they OSX image. I will fix it.

morevnaproject:
@rodolforg This is absolutely epic! I have tested your latest changes and confirming that selection works perfectly! ^__^

screenshot_001

rodolforg:
@morevnaproject  now I started to drag :)
However, it has two issues I see:
* It requires to be in edit mode :(
* The undo history increases a lot

I need some help with Undo feature.
Does SynfigStudio use undo-group stuff like some frameworks?
How can I start one undo-group and edit it (add more actions to it while dragging)?

morevnaproject:
> - It requires to be in edit mode :(

Ah, I see - it requires to have Animation Mode on, right? I see you solved this by temporary enabling/disabling this mode. ^__^

morevnaproject:
> I see you solved this by temporary enabling/disabling this mode. ^__^

I've tested this. It works cool. ^__^

> - The undo history increases a lot
> 
> I need some help with Undo feature.
> Does SynfigStudio use undo-group stuff like some frameworks?

Unfortunately, there are no such thing as undo-group in Synfig.

The same issue happens when you edit color - it produces a lot of history entries. See #451.
I think it's okay to leave things as is for now and address this issue separately later.

rodolforg:
> I've tested this. It works cool. ^__^

So is it good enough?

I want to start the TCB handles only after selection and dragging are working as expected, in order to don't change both feature codes and make them a reference for the future timetrack widget rework.

> > Does SynfigStudio use undo-group stuff like some frameworks?

> Unfortunately, there are no such thing as undo-group in Synfig.

Oh... Ok. Maybe later then.
Thank you for testing.

morevnaproject:
> So is it good enough?

Actually, it would be really nice to allow dragging horizontally (this will change time of waypoint). Horizontal dragging should be snapped to frames.

> I want to start the TCB handles only after selection and dragging are working as expected, in order to don't change both feature codes and make them a reference for the future timetrack widget rework

:tada:

rodolforg:
Should I put some snap to other curve values?

morevnaproject:
> Should I put some snap to other curve values?

@rodolforg No, I think we do not need snap for vertical movement.

morevnaproject:
I found an issue with moving points horizontally - when the point we move overlaps with another point, then it changes value of both points.

Video - https://www.youtube.com/watch?v=udV1Wot1268

Is it possible to tweak horizontal snapping, so it will avoid overlapping with existing points?

rodolforg:
> Is it possible to tweak horizontal snapping, so it will avoid overlapping with existing points?

What do you mean? Avoid dragging movement pass beyond other existent curve points?
Or to skip frame of those existent waypoints?

Jose-Moreno:
@morevnaproject @rodolforg Hi, if I may, normally the value overlapping happens in software like after effects and blender where overlapping the "waypoint" /keyframe will replace the static waypoint value with the value from the active one.

However the specific issue seen in the video portrays the overlapping behavior is contrary to that of most programs and while certainly you can't have the a "waypoint" occupying the same temporal space as another (waypoint value should get replaced), in the case of this video Konstantin shows, I'd say the issue is that if the user does NOT release the waypoint immediately on top of other static waypoints, the value should be preserved and the moving waypoint should keep it's value while "floating" until the user releases it on any other part of the timeline.

_Example After Effects "Waypoint" Dragging_
_Note: X is Red, Y is Green_
after_effects_curves

Notice on the GIF how the active value "snaps" and is ready to replace the static waypoint, however if i keep dragging the waypoint it dos NOT replace the original static value when I keep moving past that.

In the case of the video at second `0:06` the waypoint that is being moved should indeed reform the curve but without modifying the value of the waypoint it just crossed over. That waypoint should just be preserved, otherwise this would become VERY frustrating for users that want to experiment with timing or if the users are moving a keyframe from one place to another and didn't use copy/paste especifically, for example.

morevnaproject:
> What do you mean? Avoid dragging movement pass beyond other existent curve points?
Or to skip frame of those existent waypoints?

I mean skip frame of those existent waypoints.

> normally the value overlapping happens in software like after effects and blender where overlapping the "waypoint" /keyframe will replace the static waypoint value with the value from the active one. ...

@Jose-Moreno The approach you described would be the best solution. I just was thinking about some minimal/easier way to implement. Thanks for  putting detailed notes about that! ^__^

rodolforg:
@Jose-Moreno I did as you showed me, thank you!

@morevnaproject I use your suggestion for converted/linked layer values. Its children work fine as Jose-Moreno suggested, but I couldn't do it that way for converted ones.

Could you please test with exported values and any other "unusual" case? I don't know Synfig that much…

morevnaproject:
@rodolforg Thank you for the fixes!

I have tested latest commit (28bba15d0183f9f4b) and found a weird behaviour - when I select multiple waypoints and slightly move mouse to left or right (really small movement, which doesn't make waypoints to go into other frame), then it makes other selected waypoints move up or down.

I have recorded a video to show how it looks like -
https://www.youtube.com/watch?v=eTQCspFVA_o&feature=youtu.be

rodolforg:
@morevnaproject  Are you sure this is the last commit? That debug message about erasing waypoint shouldn't appear anymore.
I'm investigating it anyway.

morevnaproject:
@rodolforg Let me double-check that...

morevnaproject:
> That debug message about erasing waypoint shouldn't appear anymore.

I figured that out. I was rebuilding last commit, but only did a build for synfig-studio (not for synfig-core). Now did a full rebuild and debug message has gone. 

The weird  behaviour I described in https://github.com/synfig/synfig/pull/982#issuecomment-554308960 still remains.

rodolforg:
@morevnaproject could you please try it now?

morevnaproject:
Tested 86f8b0449a8b78f9f3a31c86c205813b0b9082a1, still the same behaviour :(

rodolforg:
sorry, I pushed the wrong commit :P

morevnaproject:
=D

I've tested 2d318255ff and it behaves better. Still, it behaves weird when multiple points are selected and I move them up-down. See video - https://www.youtube.com/watch?v=wlCtZGGEwAQ

Notice how point from purple line moves vertically relatively to points of blue line.
This is especially visible at 00:54.

Here is the test file I have played with, just in case - 
[test.sifz.zip](https://github.com/synfig/synfig/files/3854423/test.sifz.zip)

rodolforg:
I believe now it is fixed.
I found another issue : dragging the same 3 points quickly in horizontal axis, some selected point would be unselected and and stay at wrong time. Fixed it also.

morevnaproject:
It works perfectly now for me! :tada: ^__^

morevnaproject:
Some info on how splines represented in TCB - https://www.cubic.org/docs/hermite.htm (see "The Kochanek-Bartels Splines (also called TCB-Splines)" section).

rodolforg:
@morevnaproject  could you tell me what do you think about this tooltip implementation?
Captura de tela de 2019-11-22 19-17-16
While dragging, the shown info is a bit different, but I can't take a screenshot =P

morevnaproject:
> could you tell me what do you think about this tooltip implementation?

I think it is awesome! ^__^

https://www.youtube.com/watch?v=2Bi--fpBhsM

rodolforg:
Ok. Only the "value" in tooltip that bothers me now because its values are not user friendly: no measurement unit shown neither converted to those used in GUI (degrees instead of radians, pixels instead of units, etc.)

But now I'll focus in TCB handles and possible bug fixes you point me.

rodolforg:
@morevnaproject should I show the handles of all TCB waypoints or just of those selected ones?

morevnaproject:
@rodolforg It makes sense to display handles only for selected waypoints.

rodolforg:
Hey @morevnaproject Do :)

1. Do you think I should finish this PR right before " first draft of TCB handles for Graph panel" (1cd7ce8) ?
This PR affects `synfig-studio/src/gui/widgets/widget_canvastimeslider.cpp`and `synfig-studio/src/gui/widgets/widget_timeslider.cpp` besides `dock_` and `widget_curves.cpp` obviously. So it might be good people review/try this huge PR to fix any possible regression.

2. Did you try and play with this last commit? ;)

morevnaproject:
1. I think this is a very good idea. For my taste it is already a very cool feature (even without TCB handles), and it works very good for me. If we will merge it, then it will be shipped with 1.4.0, which is also very cool. @blackwarthog @ice0 opinions?

2. Yes, I tested it. As I understood, this is a very early WIP of TCB handles, so I withheld any feedback. ^__^

ice0:
I think it would be nice to include this feature in 1.4

rodolforg:
>     2. Yes, I tested it. As I understood, this is a very early WIP of TCB handles, so I withheld any feedback. ^__^

It is indeed. However, I was analyzing how TCB interpolation works in Synfig when an animation mix TCB with other interpolators and then I found out that, in some cases, TCB has no influence in the curve.
What should I do regarding TCB handle in these cases?
And that's why I put it aside for a while. lol

rodolforg:
Done.

morevnaproject:
> What should I do regarding TCB handle in these cases?

If they has no influence, then it's okay to hide them, I think.

blackwarthog:
seems it works

morevnaproject:
Merged via a415edb6aee6717ddda4aeb77ae14a9afe163899

Thank you s much for your contributions! ^__^

PR #986, opened 2019-10-28
remove unused CompView class

rodolforg:
It isn't even listed in CMake files


ice0:
Hi, Rodolfo!
We currenly using autotools as our primary build system, and this class is listed inside `synfig-studio/src/gui/Makefile.am`
Btw, i did't find how it used.
@morevnaproject @blackwarthog any ideas?

morevnaproject:
Merged via 7f408dc4c10d09a1ea66fa7acbc1becfc38ba6c0

Issue #998, opened 2019-11-06
Set preview video zoom to "Fit" by default


Keyikedalube:
**Synfig version & platform**:
1.3.11
*Development version:
Revision: 20191027
Branch: master
Revision ID: d0285b*

**Issue description**:
When previewing video the actual preview window is always floating. I have to manually change zoom from *100%* to *Fit* every time I preview my animation.
Won't it be better to set the default zoom to *Fit* for better UX?

morevnaproject:
Closed via #999

PR #999, opened 2019-11-08
Fix issue #998 Changed default preview zoom to Fit

Keyikedalube:
Build and tested the new default value as Fit (instead of 100%) on my Debian system.
Now the preview window opens in Fit zoom by default.


morevnaproject:
Looks good to me. Thank you! ^__^

PR #1007, opened 2019-11-16
ENTER finish ("run") spline tool; ESC do cancel it - fixes #492

rodolforg:
partial fix #491

Besides the new key assignment, Escape now really cancels spline.
Before this commit, it just hid the spline drawing
(if you click 'run' toolbutton the spline appears).

PR #1008, opened 2019-11-17
finish (and loop) spline if first vertex is clicked

rodolforg:
As suggested in https://github.com/synfig/synfig/pull/1007#issuecomment-554748401

Sometimes it has issues if tangent handle is on the same place as first vertex: you can't finish if first vertex is not reachable.


morevnaproject:
Thank you! ^__^

PR #1016, opened 2019-11-24
remove unused ETL mutex and thread classes

rodolforg:
Smach class just fakes that it is using them
Resource uses it, but it can be replaced with standard C+11 std::mutex.

That ETL_LOCK_REFCOUNTS isn't used anywhere… Maybe these code pieces can be deleted too.


rodolforg:
Maybe we should get rid of all other custom mutexes too (synfig:: and Glib::).
Glib indeed deprecated in favor of standards std::mutex, std::recursive_mutex . RWLock style is not in C++11 though.

morevnaproject:
@blackwarthog Your opinion?

ice0:
> That ETL_LOCK_REFCOUNTS isn't used anywhere… Maybe these code pieces can be deleted too.
@blackwarthog said it doesn't needed and can be removed.

> Glib indeed deprecated in favor of standards std::mutex, std::recursive_mutex . RWLock style is not in C++11 though.
he agrees, but Glib::RWLock needs to stay for a while.

blackwarthog:
yes my is said

my is agrgrrree :)

rodolforg:
So I can change all of them?
I'll try to do it tonight

ice0:
@rodolforg sure.

rodolforg:
I still have to remove `Glib::Threads::mutex`es and `Glib::Threads::cond`s

rodolforg:
@blackwarthog , what do you think?

blackwarthog:
> @blackwarthog , what do you think?

I think we need to use atomic counter in Resource::Storage, or use mutex without ifdef.
And may be underscores in private field names are not required.

ice0:
I think using underscores for private members is a good idea. So I suggest to leave them.

blackwarthog:
> I think using underscores for private members is a good idea. So I suggest to leave them.

Fooo.. It's ugly. Code shall be beautyfull. Can you write a Symphony? Can you take a blank canvas and turn it into a masterpiece?

(not significant, may keep underscores)

rodolforg:
It was conflicting with std::mutex , so I had to change it.

What about IPC code slice for Windows? This thread never ends, I think.

blackwarthog:
> It was conflicting with std::mutex , so I had to change it.

i prefer to remove `using namespace std` or add namespace (::, std::, myclass::, or this->) explicitly


> What about IPC code slice for Windows? This thread never ends, I think.

You right. The thread terminates by the OS when the main process closed. Seems this bug here for years. We need to add an atomic finalization flag and calling to join.

ice0:
> This thread never ends, I think.

I suggest to finish with last two changes:

> I think we need to use atomic counter in Resource::Storage, or use mutex without ifdef.

> You right. The thread terminates by the OS when the main process closed. Seems this bug here for years. We need to add an atomic finalization flag and calling to join.

and merge this PR after testing.
It gets too big for review. So all new improvements must go to the new PR.

rodolforg:
If you prefer, I can do a push --force in order to reduce this PR changes to commit 999d4f7 :
remove unused variable and lockguards

rodolforg:
(and let these commits to another PR:


replace synfig::Mutex and synfig::RecMutex with std:: ones
0a75002

replace synfig::RWLock with Glib::Threads::RWLock
eb97502 
)

blackwarthog:
i suggest to restore mutex in resource::storage, it was inside of disabled ifdef by mistake

and merge this PR, with or without underscores it's have no matter

also create a new issue about infinite IPC thread

rodolforg:
@blackwarthog  so it should use atomic counter 5fafc03  AND use mutex?

blackwarthog:
> @blackwarthog  so it should use atomic counter 5fafc03  AND use mutex?

not, i had not seen last commits

i need to check how atomics work there, but i can do it after PR is merged
anyway it worked without any mutex before

PR can be merged

morevnaproject:
Merged via 102e916f077fd6f783d035dd4fb2500ec9ab831a. Thank you!

Issue #917, opened 2019-08-12
[Lottie exporter] 004-image-group-cutout.sif fails to export


morevnaproject:
**Synfig version & platform**:
Commit id: 394aeeafd73c646c

**Issue description**:
Test file - https://github.com/synfig/synfig-tests-regressions/blob/master/export/lottie/004-image-group-cutout.sif

Error:
```
synfig(14100) [22:26:55] info: Traceback (most recent call last):
  File "/home/konstantin/sources/synfig.lottie/_production/build/share/synfig/plugins/lottie-exporter/lottie-exporter.py", line 163, in <module>
    new_file_name = parse(FILE_NAME)
  File "/home/konstantin/sources/synfig.lottie/_production/build/share/synfig/plugins/lottie-exporter/lottie-exporter.py", line 68, in parse
    gen_layers(settings.lottie_format["layers"], canvas, canvas.get_num_layers() - 1)
  File "/home/konstantin/sources/synfig.lottie/_production/build/share/synfig/plugins/lottie-exporter/layers/driver.py", line 81, in gen_layers
    itr)
  File "/home/konstantin/sources/synfig.lottie/_production/build/share/synfig/plugins/lottie-exporter/layers/group.py", line 94, in gen_layer_group
    asset = add_precomp_asset(settings.lottie_format["assets"][-1], canvas, canvas.get_num_layers())
  File "/home/konstantin/sources/synfig.lottie/_production/build/share/synfig/plugins/lottie-exporter/sources/precomp.py", line 29, in add_precomp_asset
    layers.driver.gen_layers(lottie["layers"], canvas, layer_itr-1)
  File "/home/konstantin/sources/synfig.lottie/_production/build/share/synfig/plugins/lottie-exporter/layers/driver.py", line 81, in gen_layers
    itr)
  File "/home/konstantin/sources/synfig.lottie/_production/build/share/synfig/plugins/lottie-exporter/layers/group.py", line 94, in gen_layer_group
    asset = add_precomp_asset(settings.lottie_format["assets"][-1], canvas, canvas.get_num_layers())
  File "/home/konstantin/sources/synfig.lottie/_production/build/share/synfig/plugins/lottie-exporter/sources/precomp.py", line 29, in add_precomp_asset
    layers.driver.gen_layers(lottie["layers"], canvas, layer_itr-1)
[...]

PR #1018, opened 2019-11-25
[Lottie export] Add exponent convert method and support for stretch layer export

AnishGG:
I have added support for "exp" convert method. There are also some changes in the scale parameter of the group layer, which was not working as expected with convert methods.
@morevnaproject please have a look.
This resolves the error in #917, but the animation is mirrored for some groups in this issue. I will have a look into this.


AnishGG:
> but the animation is mirrored for some groups in this issue. I will have a look into this.

This is because of a `stretch` layer present in the animation. It is not yet supported.
@morevnaproject can we do this animation without the stretch layer?

I will try to provide support to `stretch` layer soon.
Thank you

morevnaproject:
Testing it now. Thank you!!

PR #1024, opened 2019-12-05
Prefer use of standard mutexes - part 1

rodolforg:


rodolforg:
These are the last mutexes simple to convert to C++11 standard ones. 

If approved, next PR will handle those with Condition Variables.

ice0:
Merged via https://github.com/synfig/synfig/commit/a3e556a6cc6533d08b2fe52bf48163b3276b6d6a. Thank you!

PR #1025, opened 2019-12-05
Suppress some warnings when rendering Synfig images

ice0:


ice0:
@morevnaproject I'm not sure, maybe we don’t need to increase the version of the layers. Given that the old logic remains unchanged.

morevnaproject:
I think it's good to increase version, just to be 100% safe.

morevnaproject:
Merged via 4dab30ea275ae194c0c8ed17cfafdac28ebe45ea.
Thank you!

PR #1032, opened 2019-12-08
Fix some typos

FirasH:
Fixed some typos in the script

### DIFF
--- a/1-setup-linux-native.sh
+++ b/1-setup-linux-native.sh
@@ -244,7 +244,7 @@ elif command -v pacman >/dev/null; then
             shared-mime-info \
             cmake make \
             python-lxml"
-    echo "Running pacman (you need root previleges to do that)..."
+    echo "Running pacman (you need root privileges to do that)..."
     echo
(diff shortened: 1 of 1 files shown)


1-setup-linux-native.sh
```diff
@@ -244,12 +244,12 @@ elif command -v pacman >/dev/null; then
             shared-mime-info \
             cmake make \
             python-lxml"
-    echo "Running pacman (you need root previllages to do that)..."
+    echo "Running pacman (you need root previleges to do that)..."
```
rodolforg: Shouldn't it be  "privileges"? ;)

FirasH: Yes you're right! My bad.

FirasH:
> I think it's better you to rebase for squash this last commit into the first one, and force push it.

I have enabled "Allow squash merging", is that enough to squash the 2 commits as the pull request is merged?

rodolforg:
I think it is :)

PR #1034, opened 2019-12-08
there is no reason for using float instead of Real

rodolforg:
In fact, Waypoint::Model uses Real and not float...


ice0:
Merged via https://github.com/synfig/synfig/commit/e8281c679184cf9e9e6a7554ccaac0a5bff761a2

Thank you!

PR #1075, opened 2020-01-06
remove trivial copy-constructors

rodolforg:
If it is defined, assignment copy method (operator =()) should be defined too.
Trivial copy constructors are defined by default. ;)

It avoids compiler warnings.


ice0:
Merged via https://github.com/synfig/synfig/commit/91540cd1d890b31fb9a504a86911694104ecd77c

Thank you!

PR #1076, opened 2020-01-06
define copy constructor if assignment operator is defined

rodolforg:
Reason: c++ Rule of 5 (by using copy-swap idiom)

### DIFF
--- a/synfig-core/src/synfig/value.cpp
+++ b/synfig-core/src/synfig/value.cpp
@@ -83,9 +83,46 @@ ValueBase::ValueBase(Type &x):
 	create(x);
 }
 
+ValueBase::ValueBase(const ValueBase& x)
+	: ValueBase(*x.type)
+{
+	if(data != x.data)
+	{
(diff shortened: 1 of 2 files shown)


rodolforg:
Oops, I messed up here

rodolforg:
And now I fixed :P

synfig-core/src/synfig/value.cpp
```diff
@@ -210,16 +218,15 @@ ValueBase::clear()
...
 	type=&type_nil;
 }
 
-
 Type&
 ValueBase::ident_type(const String &str)
 {
 	Type *type = Type::try_get_type_by_name(str);
-	return type == NULL ? type_nil : *type;
+	return !type ? type_nil : *type;
```
ice0: Maybe better to invert the condition now? Like:
```
return type ? *type : type_nil;
```

rodolforg: fixed and squashed

PR #1083, opened 2020-01-09
Avoid ambiguity in translation of Set (verb x noun)

rodolforg:
When setting layer "description" I found out that the "OK" button (labeled as "Set" in code) in my language is translated as if it mean (Layer) Set.

### DIFF
--- a/synfig-studio/src/gui/instance.cpp
+++ b/synfig-studio/src/gui/instance.cpp
@@ -1274,10 +1274,12 @@ Instance::process_action(synfig::String name, synfigapp::Action::ParamList param
 				// export and rename value dialog
 				if (entry.name == "ValueNodeRename") button2 = _("Rename");
 				// set layer description dialog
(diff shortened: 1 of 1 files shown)


morevnaproject:
Good catch! Maybe simply set it to "OK"? ^__^

rodolforg:
I'd do that change, but previous lines are another action verb (https://github.com/synfig/synfig/blob/c55b9a30412f8217b4f86cac1fb37e933c47efa9/synfig-studio/src/gui/instance.cpp#L1271)
So, for the sake of parallelism, I suggested that way.

If we used `pgettext()` it would allow to distinguish contexts :)

morevnaproject:
Oh, well...
 Well, for the sake of simplicity, let's go with "OK" label. ^__^

rodolforg:
Done

rodolforg:
"Add Layers to Set" dialog used "Export" button instead of "Add" or... "Ok"

PR #1087, opened 2020-01-10
replace button label with open icon for widget_filename

rodolforg:
The label contents ". . ." is nearly visible, so I thought it would be good to change it.


morevnaproject:
Looks cool!

screenshot_002

PR #1095, opened 2020-01-20
remove unused action restore-defaults-settings

rodolforg:
It's not an menu item and Dialog_Setup calls
App::restore_default_settings() method directly


ice0:
@morevnaproject can you take a look?

morevnaproject:
Looks good! Thank you!

PR #1096, opened 2020-01-20
remove strange #include in Dialog_Setup

rodolforg:
It was inserted by 366b6582272b8657c1d37909cc4c845de746cc82 . I don't know if it was supposed to be in another place...


morevnaproject:
I think those are the traces of gamma widget, which has been moved to Canvas Properties window.

rodolforg:
But why would it be _inside_ language list? =)

PR #1105, opened 2020-01-22
Fixed MLT build on MSYS via autobuild/build.sh script

ice0:


autobuild/build.sh
```diff
@@ -79,6 +78,7 @@ if [[ `uname -o` == "Msys" ]]; then
 	# copy MLT
 	MLT_REV=1   # Change this when something is changed inside of if block below
 	if [ ! -f "${PREFIX}/mlt-${VERSION_MLT}-${MLT_REV}.done" ]; then
+		${REPO_DIR}/autobuild/msys2/build_mlt.sh
```
morevnaproject: This command is called in first script:

https://github.com/synfig/synfig/blob/7fbaaab8c27a3c8290a11a7925292e8de502a159/1-setup-windows-msys2.sh#L48-L49

Should we run this conditionally then?

ice0: I think i had problems because i didn't rerun 1-setup... after last changes. As for now all builds fine, so i will remove it from here.

ice0: Fixed

ice0: Also add +x to 1-setup-windows-msys2.sh

Issue #1082, opened 2020-01-09
[Vectorization] Line thickness should depend on TL/BR


morevnaproject:
See this commit - https://github.com/synfig/synfig/commit/a8055a1d1af0c144ae2288b930aedd2da3305e84

Width division factor should depend from TL/BR points of image.

See video demonstrating the problem:
https://www.youtube.com/watch?v=h8R2qBGKQLY

So, the difference in  of TL/BR points from their initial positions should affect the width division factor.

PR #1108, opened 2020-01-22
make width depend on TL/BR

ankit-kumar-dwivedi:
@morevnaproject I used  `width = (width/2.5)*max(w_factor,h_factor)` I also tried sqrt(w_factor*h_factor) and results were similar considering almost same transformation along width and height. Let me know if this logic can be futher improved. 
@ice0 @blackwarthog suggestions?

Closes: #1082


morevnaproject:
Works perfectly now! Thank you! ^__^

PR #1134, opened 2020-01-30
keep undo history clean when deleting multiple curve waypoints

rodolforg:


rodolforg:
After user delete multiple waypoints simultaneously, `ctrl+Z` undoes one waypoint deletion at a time.

ice0:
Merged. Thank you!

Issue #790, opened 2019-03-25
Display “static intervals” between waypoints


morevnaproject:
If parameter value on a segment between two adjacent waypoints dosn't changes, then visualize that somehow.

Here's how that's done in Blender:
![screenshot_001](https://user-images.githubusercontent.com/332868/54899417-59679400-4f02-11e9-8203-abf3f8e6d2a4.png)

I suggest to display such intervals by connecting waypoints with a solid line:

![screenshot_002](https://user-images.githubusercontent.com/332868/54899679-312c6500-4f03-11e9-9e9b-e7ee232a5cd3.png)

yshkp:
@morevnaproject This is a good feature for waypoints and I think I can replace my deliverable of single click state with this one. 
PS- Also check comments on suggestions.

morevnaproject:
Yeah, I've just mentioned this in my review of your proposal draft. Sent it to you just now. ^__^

yshkp:
@morevnaproject Surely I will do what you have asked me by that time.

yshkp:
@morevnaproject As I have started working on the task that you gave me and on the early basis I came to an idea that: If it is allowed to select multiple rows which are rendered using cell renderer then automatically user can select multiple waypoints using CTRL key as the main issue was due to operating on a single row.
Please tell me if I am on the right track.

morevnaproject:
@yshkp Since this is a question about issue #762, can you please post it there? That way our discussion will be on topic in right place. ^__^

Issue #762, opened 2019-03-11
Allow to select multiple waypoints from different rows


morevnaproject:
Allow user to select multiple waypoints by holding CTRL.

Currently this is possible if waypoints belong to the same parameter.
But if waypoints belong to different parameters, then it is impossible to select them together.

yshkp:
@morevnaproject As it is related to my project which I proposed to you so you can assign it to me and further it will be continued during our project period.

morevnaproject:
@yshkp Assigned to you now. ^__^

yshkp:
@morevnaproject I will be submitting you a draft proposal tomorrow which will also mention this issue about the approach I thought. If you allow me I can also include this is in my calendar of coding period. 
I have also thought some improvements regarding waypoints handling which definitely improve it.

morevnaproject:
@yshkp Thanks for keeping me updated! Looking forward for your draft. ^__^
P.S. Please send it to me via email.

yshkp:
@morevnaproject Please check your mail. I sent you the link to my draft proposal. Please take a look and review. Any suggestions would be appreciated.
I hope it matches to your expectations. ^__^

morevnaproject:
Got your mail, thank you! ^__^

yshkp:
@morevnaproject As I have started working on the task that you gave me and on the early basis I came to an idea that: If it is allowed to select multiple rows which are rendered using cell renderer then automatically user can select multiple waypoints using CTRL key as the main issue was due to operating on a single row.
Please tell me if I am on the right track.

morevnaproject:
It's hard to tell without trying. I doubt this will work, but still you can try to enable multiple selection.

Again, the safest way is to re-write the widget completely. For reference on writing custom widgets you can check TimeSlider widget (gray area on top of TimeTrack Panel) - https://github.com/synfig/synfig/blob/master/synfig-studio/src/gui/widgets/widget_timeslider.cpp

Graps Panel is also a good example of custom widget related with time/parameters.

yshkp:
@morevnaproject Well, yes now I am going to implement it in re-writing the widget.
Just gone through some docs so I thought this could work so suggested you an idea. But it's good to have a custom widget so without wasting time now working on it. :)

morevnaproject:
Also, I suggest to create a copy of TImeTrack panel and do all work there. You can call it "TimeTrack-2". This will allow incremental development, without breaking existing functionality. ^__^

yshkp:
[...]

Issue #761, opened 2019-03-11
Drag-select multiple waypoints


morevnaproject:
Allow user to select multiple waypoints without tediously having to click them one by one. User should be able to click and drag a box around multiple ones to select them all. 

For this I suggest to add a "box-selection tool" to TimeTrack panel:

![screenshot_002](https://user-images.githubusercontent.com/332868/54120983-d6d3d480-442b-11e9-8481-47e1db9e97d0.png)

morevnaproject:
Implemented via #/1135.

Issue #529, opened 2018-04-23
Timetrack not refreshes when editing animation in imported file


morevnaproject:
**Synfig version & platform**:
All versions and platforms

**Issue description**:
When I am trying to edit animation of imported file, then Timetrack is not refreshed (no new waypoints displayed until you change frame).
See in this video at 2:06 - https://youtu.be/TVipENxNFZ0?t=2m6s

morevnaproject:
Tested with PR #1135.

Got a bit different behavior.

I will try to explain here what is expected and what is not working with this PR.

Sample file - 
[bug-parent-child-timetrack.zip](https://github.com/synfig/synfig/files/4288363/bug-parent-child-timetrack.zip)
The archive includes 2 files - "parent.sifz" and "child.sifz".

1. Open both files in Synfig.
2. Switch to "child.sifz" and enable animation mode.
3. Switch to "parent.sifz" 
4. Expand group called "[imported-file]" and click (select) "Circle" layer inside.
5. Look at Timetrack panel: BUG - Timetrack Panel should display keyframes of "child.sifz" (you can compare how it looks in old Timetrack and how in new:

![screenshot_002](https://user-images.githubusercontent.com/332868/75899986-d3243380-5e6e-11ea-8560-a32f744bd772.png)

![screenshot_003](https://user-images.githubusercontent.com/332868/75899993-d61f2400-5e6e-11ea-881a-d4bc7f645fb7.png)

6. Try to move Circle layer. Since on step 2 we enabled animation mode in "child.sifz", its Origin Parameter should get new waypoints when we move circle. And it actually does. But Timetrack is not updated.

If you have old timetrack active, then you need to move time cursor to trigger refresh - and waypoints will appear.

With new timetrack situation is better - you do not need to touch time cursor. Just hover mouse over timetrack and waypoints will appear.

Here is a video showing all those steps - https://youtu.be/6qc-1vntPp4

rodolforg:
Fixed the keyframe line drawing regression in that PR with https://github.com/synfig/synfig/pull/1135/commits/5e6877b04b5c19661ac058927a42f14f224ab118

Regarding the missed drawing update, I don't have this issue here, so I'll try it later.
https://youtu.be/ypsAn-XSn-k

morevnaproject:
> Fixed the keyframe line drawing regression in that PR with 5e6877b

Confirmed! Thank you so much! ^__^

> Regarding the missed drawing update, I don't have this issue here, so I'll try it later.
[...]

PR #1135, opened 2020-01-30
New timetrack

rodolforg:
I don't know all features of current TimeTrack, please let me know if something is missing:

- [x] Sync with Parameters dock (it has issues with two timetracks enabled. It works fine with only the new timetrack.

- [x] context/popup menu via right-click in waypoint (for single waypoint)

- [x] move selected waypoint (mouse drag and arrow keys)

- [x] shift + drag : copy selected waypoint
- [x] waypoint deletion via Delete key (**novelty**)

- [x] double-click opens Waypoint editor dialog

- [x] multiple selection - including across parameters (**novelty**) (#761 #762)

- [x] display static interval (**novelty**) (#790)

- [x] allow to scale waypoints (**novelty**) (#266 https://github.com/synfig/synfig/issues/266#issuecomment-480610390 )

- [x] parameters and Timeline panels show the same row selected. (**novelty**) (#246)

- [x] waypoint navigation  (**novelty**)
  key `n` and `b` to go to next or previous waypoint. By holding `shift`, user skips 5 waypoints.
(by design, it only works when only one waypoint is selected.)

- [ ] #529  fixed?

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -148,6 +148,7 @@
 #include "docks/dock_navigator.h"
 #include "docks/dock_soundwave.h"
 #include "docks/dock_timetrack.h"
+#include "docks/dock_timetrack2.h"
 #include "docks/dock_toolbox.h"
 
 #include "modules/module.h"
@@ -300,6 +301,7 @@ studio::Dock_LayerGroups   *dock_layer_groups;
 studio::Dock_Navigator     *dock_navigator;
 studio::Dock_SoundWave     *dock_soundwave;
 studio::Dock_Timetrack     *dock_timetrack;
+studio::Dock_Timetrack2    *dock_timetrack2;
 studio::Dock_Curves        *dock_curves;
 
 std::list< etl::handle< studio::Module > > module_list_;
@@ -1069,6 +1071,7 @@ DEFINE_ACTION("panel-curves",          _("Graphs"));
 DEFINE_ACTION("panel-groups",          _("Sets"));
 DEFINE_ACTION("panel-pal_edit",        _("Palette Editor"));
 DEFINE_ACTION("panel-soundwave",       _("Sound"));
+DEFINE_ACTION("panel-timetrack2",      _("Timetrack 2"));
 
 // actions in Help menu
 DEFINE_ACTION("help",           Gtk::Stock::HELP);
@@ -1239,6 +1242,7 @@ DEFINE_ACTION("keyframe-properties", _("Properties"));
 "		<menuitem action='panel-info' />"
 "		<menuitem action='panel-navigator' />"
 "		<menuitem action='panel-timetrack' />"
+"		<menuitem action='panel-timetrack2' />"
 "		<menuitem action='panel-curves' />"
 "		<menuitem action='panel-groups' />"
 "		<menuitem action='panel-pal_edit' />"
@@ -1618,6 +1622,10 @@ App::App(const synfig::String& basepath, int *argc, char ***argv):
 		studio_init_cb.task(_("Init Timetrack..."));
 		dock_timetrack = new studio::Dock_Timetrack();
 		dock_manager->register_dockable(*dock_timetrack);
+
+		studio_init_cb.task(_("Init Timetrack 2..."));
+		dock_timetrack2 = new studio::Dock_Timetrack2();
+		dock_manager->register_dockable(*dock_timetrack2);
 
 		studio_init_cb.task(_("Init Curve Editor..."));
 		dock_curves = new studio::Dock_Curves();
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -300,7 +300,7 @@ studio::Dock_Info          *dock_info;
 studio::Dock_LayerGroups   *dock_layer_groups;
 studio::Dock_Navigator     *dock_navigator;
 studio::Dock_SoundWave     *dock_soundwave;
-studio::Dock_Timetrack     *dock_timetrack_old;
+studio::Dock_Timetrack_Old     *dock_timetrack_old;
 studio::Dock_Timetrack2    *dock_timetrack;
 studio::Dock_Curves        *dock_curves;
 
@@ -1620,7 +1620,7 @@ App::App(const synfig::String& basepath, int *argc, char ***argv):
 		dock_manager->register_dockable(*dock_soundwave);
 
 		studio_init_cb.task(_("Init Timetrack (old)..."));
-		dock_timetrack_old = new studio::Dock_Timetrack();
+		dock_timetrack_old = new studio::Dock_Timetrack_Old();
 		dock_manager->register_dockable(*dock_timetrack_old);
 
 		studio_init_cb.task(_("Init Timetrack..."));
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -300,8 +300,8 @@ studio::Dock_Info          *dock_info;
 studio::Dock_LayerGroups   *dock_layer_groups;
 studio::Dock_Navigator     *dock_navigator;
 studio::Dock_SoundWave     *dock_soundwave;
-studio::Dock_Timetrack     *dock_timetrack;
-studio::Dock_Timetrack2    *dock_timetrack2;
+studio::Dock_Timetrack     *dock_timetrack_old;
+studio::Dock_Timetrack2    *dock_timetrack;
 studio::Dock_Curves        *dock_curves;
 
 std::list< etl::handle< studio::Module > > module_list_;
@@ -1066,12 +1066,12 @@ DEFINE_ACTION("panel-meta_data",       _("Canvas MetaData"));
 DEFINE_ACTION("panel-children",        _("Library"));
 DEFINE_ACTION("panel-info",            _("Info"));
 DEFINE_ACTION("panel-navigator",       _("Navigator"));
-DEFINE_ACTION("panel-timetrack",       _("Timetrack"));
+DEFINE_ACTION("panel-timetrack-old",   _("Timetrack (old)"));
 DEFINE_ACTION("panel-curves",          _("Graphs"));
 DEFINE_ACTION("panel-groups",          _("Sets"));
 DEFINE_ACTION("panel-pal_edit",        _("Palette Editor"));
 DEFINE_ACTION("panel-soundwave",       _("Sound"));
-DEFINE_ACTION("panel-timetrack2",      _("Timetrack 2"));
+DEFINE_ACTION("panel-timetrack",      _("Timetrack"));
 
 // actions in Help menu
 DEFINE_ACTION("help",           Gtk::Stock::HELP);
@@ -1241,8 +1241,8 @@ DEFINE_ACTION("keyframe-properties", _("Properties"));
 "		<menuitem action='panel-children' />"
 "		<menuitem action='panel-info' />"
 "		<menuitem action='panel-navigator' />"
+"		<menuitem action='panel-timetrack-old' />"
 "		<menuitem action='panel-timetrack' />"
-"		<menuitem action='panel-timetrack2' />"
 "		<menuitem action='panel-curves' />"
 "		<menuitem action='panel-groups' />"
 "		<menuitem action='panel-pal_edit' />"
@@ -1619,13 +1619,13 @@ App::App(const synfig::String& basepath, int *argc, char ***argv):
 		dock_soundwave = new studio::Dock_SoundWave();
 		dock_manager->register_dockable(*dock_soundwave);
 
-		studio_init_cb.task(_("Init Timetrack..."));
-		dock_timetrack = new studio::Dock_Timetrack();
-		dock_manager->register_dockable(*dock_timetrack);
+		studio_init_cb.task(_("Init Timetrack (old)..."));
+		dock_timetrack_old = new studio::Dock_Timetrack();
+		dock_manager->register_dockable(*dock_timetrack_old);
 
-		studio_init_cb.task(_("Init Timetrack 2..."));
-		dock_timetrack2 = new studio::Dock_Timetrack2();
-		dock_manager->register_dockable(*dock_timetrack2);
+		studio_init_cb.task(_("Init Timetrack..."));
+		dock_timetrack = new studio::Dock_Timetrack2();
+		dock_manager->register_dockable(*dock_timetrack);
 
 		studio_init_cb.task(_("Init Curve Editor..."));
 		dock_curves = new studio::Dock_Curves();
--- a/synfig-studio/src/gui/docks/dock_timetrack2.cpp
+++ b/synfig-studio/src/gui/docks/dock_timetrack2.cpp
@@ -35,6 +35,8 @@
 
 #include <synfig/general.h>
 
+#include <cstring>
+
 #endif
 
 using namespace studio;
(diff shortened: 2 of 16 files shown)


ice0:
Wow!

morevnaproject:
!!! O___O

morevnaproject:
At the moment I am finishing release of Synfig 1.3.12, and then I need to finalize our GSoC submission. Then I will review this. Thank you so much!!! ^__^

morevnaproject:
/remind me about this in 5 days

rodolforg:
> In the screenshot below (built from synfig master, not rodolforg version) the keyframes (start and end) don't appear fully making it hard for me to duplicate the keyframes if I wanted to. Hope I read this [commit message](https://github.com/synfig/synfig/pull/1135/commits/56078c3903148e5879bf02fe27cb951d582747f9 .

Hm… Sadly I just 'replicate' the current behavior of 2px displacement the current Timetrack has.

ice0:
As far as I remember, @morevnaproject also has some problems in this part.

morevnaproject:
Yes adding padding at the start and end will make life easier for many users ^__^ - #1145

morevnaproject:
I am building this PR now... ^__^

morevnaproject:
Tested and it is so awesome!

So far i found only one issue:
If i select any waypoints and hit Delete, then Timetrack is not updated. The animation in workarea updated - so I can see that waypoints are really deleted. To make changes reflected on Timetrack I need to click with on any place of Timetrack.

rodolforg:
> If i select any waypoints and hit Delete, then Timetrack is not updated. 

:o Are you sure? I know this phrase sucks but: "Here it works as expected"

morevnaproject:
Yes, here is a video ^__^ - https://www.youtube.com/watch?v=b6-2QW6PWds

Built with commit 188ac4a87edc40 merged into master at 00e3c96c42a7566.

morevnaproject:
I have built commit 188ac4 (without merging into master) and still can reproduce the same issue. ^__^"

rodolforg:
OMG, I'll have to check it out because here is my video https://youtu.be/6E1JXP8UuK4

morevnaproject:
Could it be because of difference in GTK versions? I am using Gtk 3.22.30.

BTW, I love semi-transparent waypoints - didn't discovered them at first! ^__^

rodolforg:
> Could it be because of difference in GTK versions? I am using Gtk 3.22.30.

Here I am with 3.24.11.

> BTW, I love semi-transparent waypoints - didn't discovered them at first! ^__^
They're like that for valuenodes, just like I did for Curves Dock.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -0,0 +1,825 @@
...
+			param_list.add("addcanvas", value_desc.get_value().get(synfig::Canvas::Handle()));
+		} else {
+			param_list.add("addvaluedesc", value_desc);
+		}
+
+		param_list.add("addtime", wi->time_point.get_time());
+	}
+	action->set_param_list(param_list);
+	canvas_interface->get_instance()->perform_action(action);
+}
```
rodolforg: @morevnaproject , could you try if a `queue_draw();` here solves your issue?

morevnaproject: This solves issue, but please see my next comment...

morevnaproject:
Hi! Apologies for dealy.
I have made more testing and discovered some interesting things:
1. Timetrack is also not updates when new waypoints are created. I do not know why I didn't discovered that initially - maybe I had other (old) Timetrack Panel active while animating. Or maybe because of next thing (2).
2. The issue with timetrack not updating is happens ONLY for new documents. If I open some existing document, then timetrack updates correctly (both for creating and deleting waypoints).

Please see this video for demonstration ^__^ - https://www.youtube.com/watch?v=_wwFc3nYwcc

morevnaproject:
P.S. All tests are done for 188ac4a

rodolforg:
> Hi! Apologies for dealy.
Me too! :)

>     2. The issue with timetrack not updating is happens ONLY for new documents. If I open some existing document, then timetrack updates correctly (both for creating and deleting waypoints).

Ok, I investigate it more and I found out when exactly it isn't updated: when the parameter is not a valuenode before you select layer.
* If you select a circle layer whose Origin is animated, and edit its animation it refreshs the timetrack.
* If you then starts to animate its radius, it doesn't visually refresh its waypoints in timetrack.
* Then you select none, select circle layer back and continue to animate its radius: everything works as expected.

The reason for it is that I set to only redraw the timetrack when a valuenode changes. I choose to do like that due to performance reasons (if it has a large parameter or waypoint list, maybe timetrack would be slowly re-rendered when user simply changes a not-animated parameter).

The problem is I only search for value nodes in the parameter list when it is set. I do not track if a regular value base changes to value node - because I don't know how to do it yet. If I can't find out when editor changes a regular parameter to an animated value node, ... well, I'll have to refresh everytime anything changes.

rodolforg:
@morevnaproject I think that issue you pointed is now fixed. I made a workaround and it seems to work fine.
The only thing left is… sync the Param panel height with TimeTrack Panel (last row issue).
I hope I solve it today.

rodolforg:
Not that "today", but it's done :)

I believe this PR is ready now and there's nothing left to do here. :)
Please tell if there is any bug @morevnaproject @ice0 @Jose-Moreno

synfig-studio/src/gui/widgets/widget_timetrack.cpp
rodolforg: I believe it's solved now.

rodolforg:
> I believe this PR is ready now and there's nothing left to do here. :)

Hm... Actually I forgot about this: https://github.com/synfig/synfig/issues/266#issuecomment-573763300 Should I do it?

And I found out issue #529 , and I found out my TimeTrack widget make Synfig crash. I'll fix it now

morevnaproject:
Hi! Thank you for the fixes! Previous issue is resolved.

I have tested more and found another issue:

1. Create new document.
2. Create circle, animate any of its parameters.
3. Hit Ctrl+S -> "Save as" dialog appears.
4. Close that dialog.
5. After that you will be unable to move waypoints.

Demonstration video -  https://youtu.be/NUJ2Hqpm0SM?t=95 (starting from 1:35)

> Hm... Actually I forgot about this: #266 (comment) Should I do it?

I think it would be nice to have panel with buttons, so user do not need to guess how to scale waypoints. Of course, it not needs "bounding box selection" button, because you implemented it in a very intuitive way. ^__^

rodolforg:
> And I found out issue #529 , and I found out my TimeTrack widget make Synfig crash. I'll fix it now
fixed

rodolforg:
>     5. After that you will be unable to move waypoints.
> Demonstration video - https://youtu.be/NUJ2Hqpm0SM?t=95 (starting from 1:35)

Fixed :)

> > Hm... Actually I forgot about this: #266 (comment) Should I do it?
> 
> I think it would be nice to have panel with buttons, so user do not need to guess how to scale waypoints.

My only concern is about cluttering the dock, reducing its size.

@morevnaproject  and could please check if #529 is addressed?

morevnaproject:
>>    Demonstration video - https://youtu.be/NUJ2Hqpm0SM?t=95 (starting from 1:35)
>
> Fixed :)

Confirmed! Thank you! ^__^

> My only concern is about cluttering the dock, reducing its size.

I think that's okay as long as toolbar is on the left (we have a plenty of horizontal space, while vertical space is usually limited). ^__^

morevnaproject:
Regarding #529 - it is not fixed. Well, actually... maybe it  is partially fixed.
I wrote detailed explanation in comment here - https://github.com/synfig/synfig/issues/529#issuecomment-594645256

It is possible this problem with updating timetrack might have deep roots and might be hard to fix, so, maybe it can be solved later. This is why I do not put explanations in this PR and instead placed them as comment into #529. ^__^

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <gui/localization.h>
+
+#include <synfig/general.h>
+
+#endif
+
+using namespace studio;
+
+Dock_Timetrack2::Dock_Timetrack2()
+	: Dock_CanvasSpecific("timetrack2", _("Timetrack"), Gtk::StockID("synfig-timetrack")),
```
morevnaproject: I suggest to change into this:
```
Dock_CanvasSpecific("timetrack", _("Timetrack"), Gtk::StockID("synfig-timetrack")),
```

...and change initialization of old timetrack (in `dock_timetrack.cpp`) to this:
```
Dock_CanvasSpecific("timetrack-old", _("Timetrack (old)"), Gtk::StockID("synfig-timetrack")),
```

synfig-studio/src/gui/app.cpp
```diff
@@ -1070,6 +1072,7 @@ DEFINE_ACTION("panel-curves",          _("Graphs"));
 DEFINE_ACTION("panel-groups",          _("Sets"));
 DEFINE_ACTION("panel-pal_edit",        _("Palette Editor"));
 DEFINE_ACTION("panel-soundwave",       _("Sound"));
+DEFINE_ACTION("panel-timetrack2",      _("Timetrack 2"));
```
morevnaproject: change to:
```
DEFINE_ACTION("panel-timetrack-old", _("Timetrack (old)"));
DEFINE_ACTION("panel-curves",          _("Graphs"));
DEFINE_ACTION("panel-groups",          _("Sets"));
DEFINE_ACTION("panel-pal_edit",        _("Palette Editor"));
DEFINE_ACTION("panel-soundwave",       _("Sound"));
DEFINE_ACTION("panel-timetrack",      _("Timetrack"));
```

synfig-studio/src/gui/app.cpp
```diff
@@ -1240,6 +1243,7 @@ DEFINE_ACTION("keyframe-properties", _("Properties"));
 "		<menuitem action='panel-info' />"
 "		<menuitem action='panel-navigator' />"
 "		<menuitem action='panel-timetrack' />"
+"		<menuitem action='panel-timetrack2' />"
```
morevnaproject: change to:
```
"		<menuitem action='panel-timetrack-old' />"
"		<menuitem action='panel-timetrack' />"
```

morevnaproject:
In general I feel this PR is ready for merge. 

I just suggest to change 
"timetrack" -> "timetrack-old"
and 
"timetrack2" -> "timetrack".
(see comments in my review).

After doing those replaces the users will get new timetrack instead of old, without changing anything in their configuration.

I think it is not needed to rename .cpp/.h files, as this will help to keep track of version history.

morevnaproject:
...and one more thing: I found that semi-transparent representation of waypoints makes users think that they are "inactive" (cannot be moved).
screenshot_005

Can we think about some different representation? Maybe remove transparency and have a double outline for them?

rodolforg:
> I think it is not needed to rename .cpp/.h files, as this will help to keep track of version history.

We can use "git mv" to rename files and its history log is preserved.

rodolforg:
Should I rename the classes too? eg. `Dock_Timetrack` to `Dock_TimetrackOld`

synfig-studio/src/gui/app.cpp
rodolforg: Done

synfig-studio/src/gui/app.cpp
rodolforg: Done

synfig-studio/src/gui/docks/dock_timetrack2.cpp
rodolforg: Done

rodolforg:
> Can we think about some different representation? Maybe remove transparency and have a double outline for them?

Or maybe a smaller waypoint?

rodolforg:
> I think that's okay as long as toolbar is on the left (we have a plenty of horizontal space, while vertical space is usually limited). ^__^

The button panel should be left or right timetrack widget? :)
Scale and Move buttons only?

morevnaproject:
> We can use "git mv" to rename files and its history log is preserved.

Well, sometimes I am filtering git history "by filename". I am not sure how this will be handled.
In any case, I am not against of renaming files, and I agree that it will be a cleaner solution. ^__^

> Should I rename the classes too? eg. Dock_Timetrack to Dock_TimetrackOld

That's good idea too! ^__^

>> Can we think about some different representation? Maybe remove transparency and have a double outline for them?
> Or maybe a smaller waypoint?

I still think about double outline, because this is more similar to representing "multiple waypoints" (for my taste ^__^). 

> The button panel should be left or right timetrack widget? :)

On the right, sorry for the mess. ^__^

> Scale and Move buttons only?

I can't think about any other buttons right now...

Well actually, we can have a buttons with each interpolation type. Clicking such button will apply this interpolation to selected waypoints. That would be cool. But I think it is better to leave for separate PR - I really would like to have this PR merged as soon as possible. ^__^

rodolforg:
> In any case, I am not against of renaming files, and I agree that it will be a cleaner solution. ^__^

Ok. I tested `git log`, and `--follow` is needed to follow file renaming. However, it doesn't work for `git blame` unless you specify previous commit :P So, I'm letting this as it is now.

> > Should I rename the classes too? eg. Dock_Timetrack to Dock_TimetrackOld
> 
> That's good idea too! ^__^

Done.

> > > Can we think about some different representation? Maybe remove transparency and have a double outline for them?
> > > Or maybe a smaller waypoint?
> 
> I still think about double outline, because this is more similar to representing "multiple waypoints" (for my taste ^__^).

Ok. I'll try it.

Hm... Actually, regular line width is 1.0; selected waypoints have width of 2.0.
So, linked-waypoints (not necessarily it means multiple waypoints) should use 2.0 and 4.0?

> > The button panel should be left or right timetrack widget? :)
> 
> On the right, sorry for the mess. ^__^
> 
> > Scale and Move buttons only?
> 
> I can't think about any other buttons right now...
> 
> Well actually, we can have a buttons with each interpolation type. Clicking such button will apply this interpolation to selected waypoints. That would be cool. But I think it is better to leave for separate PR - I really would like to have this PR merged as soon as possible. ^__^

Ok. Move, Duplicate and Scale.

ice0:
> Hm... Actually, regular line width is 1.0; selected waypoints have width of 2.0.
> So, linked-waypoints (not necessarily it means multiple waypoints) should use 2.0 and 4.0?

If linked waypoints can also be selected or not, then using width, in my opinion, is not a good idea. Better to change the color.

rodolforg:
> If linked waypoints can also be selected or not, then using width, in my opinion, is not a good idea. Better to change the color.

Sorry, my mistake: I mean converted waypoints. (anyway, it doesn't necessarily mean "multiple waypoints").

rodolforg:
@morevnaproject  sorry for the long time I spent on this.
Could you please test and evaluate it?
And I still don't know what to do with converted waypoint design…

morevnaproject:
Hi guys! Apologies for delay. Let me clarify what I was meaning with "double outline" ^__^ -
screenshot_001

morevnaproject:
Tested the PR. The toolbar on the right looks so cool! ^__^
screenshot_001

I suggest to merge this PR already and do all further improvements as separate PRs.

@rodolforg can you please rebase this branch on top of the latest master? ^__^

morevnaproject:
Merged now. Thank you so much!!! :tada:

morevnaproject:
I have added two issues for future improving of TimeTrack:
- https://github.com/synfig/synfig/issues/1266
- https://github.com/synfig/synfig/issues/1267

rodolforg:
@morevnaproject Could you please close the issues mentioned on this PR description?
If they're solved, of course.

PR #1159, opened 2020-02-13
Fix some typos and some non translatable strings

FirasH:
Includes typos and fix of non-translatable strings.


ice0:
Merged. Thank you so much!

PR #1166, opened 2020-02-20
Do not require to explicitly call FileSystemTemporary::discard_changes()

morevnaproject:
This is a better approach for 93f25328c15 and 28db040026 (partially reverts their changes).
It does not requires to explicitly call `FileSystemTemporary::discard_changes()`.

Instead, I just added a flag to FileSystemTemporary class (keep_files_when_destroyed),
which defines how to deal with temporary files after class instance is destroyed.


ice0:
Merged, thank you!

PR #1182, opened 2020-02-28
Improved dialog waypoint UX

### LINKED ISSUE #1178: Waypoint Editor should show TCB options only if it is a TCB point
Currently when you open a Waypoint editor dialog*, it always shows the TCB parameter values/controls even when it uses an interpolation method other than TCB.

This dialog should only show Tension, Continuity, Bias and Temporal Tension only if its input or output interpolation is TCB.

* You can view this dialog by double clicking a waypoint in timetrack or via its context menu.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/89150180-waypoint-editor-should-show-tcb-options-only-if-it-is-a-tcb-point?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1189: Conflicting mnemonic initials on Waypoint Editor
Building from source
Debian sid

**Issue description**:
I don't use mnemonic keys a lot but I noticed this dialog has the same mnemonic initials when I press Alt key
Screenshot from 2020-03-01 06-36-08

The only time I'd use mnemonic is for buttons such as OK and Cancel which usually are standard O and C for all types of desktop applications
I tried to cancel my edit on the above dialog and it refused to close. Looking closely I found out there are two Cs and Continuity got the focus.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/89395544-conflicting-mnemonic-initials?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Keyikedalube:
Here's how the improved waypoint dialog works

waypoint dialog

Fixes #1178 
Fixes #1189

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_waypoint.cpp
+++ b/synfig-studio/src/gui/widgets/widget_waypoint.cpp
@@ -266,9 +266,9 @@ Widget_Waypoint::set_waypoint(synfig::Waypoint &x)
 	if (
 		(Waypoint::Interpolation)before_options->get_value() == Interpolation::INTERPOLATION_TCB ||
 		(Waypoint::Interpolation)after_options->get_value() == Interpolation::INTERPOLATION_TCB)
-		show_tcb_params(true);
+		config_tcb_params(true);
 	else
-		show_tcb_params(false);
+		config_tcb_params(false);
 }
 
 const synfig::Waypoint &
@@ -290,9 +290,9 @@ Widget_Waypoint::get_waypoint()const
 }
 
 void
-Widget_Waypoint::show_tcb_params(bool show_labels)
+Widget_Waypoint::config_tcb_params(bool show_params)
 {
-	if (show_labels) {
+	if (show_params) {
 		// show labels
 		tcbFrame->get_label_widget()->show();
 		tensionLabel->show();
@@ -332,9 +332,9 @@ Widget_Waypoint::on_before_options_changed()
 	if (
 		(Waypoint::Interpolation)before_options->get_value() == Interpolation::INTERPOLATION_TCB ||
 		(Waypoint::Interpolation)after_options->get_value() == Interpolation::INTERPOLATION_TCB)
-		show_tcb_params(true);
+		config_tcb_params(true);
 	else
-		show_tcb_params(false);
+		config_tcb_params(false);
 }
 
 void
@@ -343,8 +343,8 @@ Widget_Waypoint::on_after_options_changed()
 	if (
 		(Waypoint::Interpolation)before_options->get_value() == Interpolation::INTERPOLATION_TCB ||
 		(Waypoint::Interpolation)after_options->get_value() == Interpolation::INTERPOLATION_TCB)
-		show_tcb_params(true);
+		config_tcb_params(true);
 	else
-		show_tcb_params(false);
+		config_tcb_params(false);
 }
 
--- a/synfig-studio/src/gui/widgets/widget_waypoint.cpp
+++ b/synfig-studio/src/gui/widgets/widget_waypoint.cpp
@@ -263,12 +263,7 @@ Widget_Waypoint::set_waypoint(synfig::Waypoint &x)
 	before_options->set_value((Waypoint::Interpolation)waypoint.get_before());
 	after_options->set_value((Waypoint::Interpolation)waypoint.get_after());
 	
-	if (
-		(Waypoint::Interpolation)before_options->get_value() == Interpolation::INTERPOLATION_TCB ||
-		(Waypoint::Interpolation)after_options->get_value() == Interpolation::INTERPOLATION_TCB)
-		config_tcb_params(true);
-	else
-		config_tcb_params(false);
+	update_tcb_params_visibility();
 }
 
 const synfig::Waypoint &
--- a/synfig-studio/src/gui/widgets/widget_waypoint.cpp
+++ b/synfig-studio/src/gui/widgets/widget_waypoint.cpp
@@ -173,7 +173,7 @@ Widget_Waypoint::Widget_Waypoint(etl::handle<synfig::Canvas> canvas):
 	interpolationTable->attach(*interpolationInLabel, 0, 1, 0, 1, Gtk::SHRINK | Gtk::FILL, Gtk::EXPAND | Gtk::FILL, 0, 0);
 	interpolationTable->attach(*before_options, 1, 2, 0, 1, Gtk::EXPAND | Gtk::FILL, Gtk::EXPAND | Gtk::FILL, 0, 0);
 
-	Gtk::Label *interpolationOutLabel = manage(new Gtk::Label(_("_Out Interpolation"), true));
+	Gtk::Label *interpolationOutLabel = manage(new Gtk::Label(_("O_ut Interpolation"), true));
 	interpolationOutLabel->set_alignment(0, 0.5);
 	interpolationOutLabel->set_mnemonic_widget(*after_options);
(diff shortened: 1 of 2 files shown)


synfig-studio/src/gui/widgets/widget_waypoint.cpp
```diff
@@ -294,5 +289,62 @@ Widget_Waypoint::get_waypoint()const
 	return waypoint;
 }
 
+void
+Widget_Waypoint::show_tcb_params(bool ok)
```
ice0: Maybe rename ok -> show_labels?

Keyikedalube: OK

Keyikedalube: The function does not only show labels but also show spin buttons and set adjustment values. Moreover, the function name says show_tcb_params so I think we need a better argument name than show_labels? If "ok" was too vague what about "enable or enabled"?

rodolforg:
Maybe show_params?

Keyikedalube:
Worth looking at too #1189

Keyikedalube:
@ice0 @rodolforg What about changing from

`show_tcb_params(bool show_labels)`

to

`config_tcb_params(bool show_params)`

That way both func name & arg name doesn't say *show* which is quite redundant. I'm all ears to other names :)

synfig-studio/src/gui/widgets/widget_waypoint.cpp
rodolforg: The problem with "ok" is it looks like to store success or failure

Keyikedalube: What about changing from

`show_tcb_params(bool show_labels)`

to

`config_tcb_params(bool show_params)`

That way both func name & arg name doesn't say *show* which is quite redundant. I'm all ears to other names too :)

ice0: Yeah, show_params is more clear

synfig-studio/src/gui/widgets/widget_waypoint.cpp
```diff
@@ -294,5 +289,62 @@ Widget_Waypoint::get_waypoint()const
...
+void
+Widget_Waypoint::on_after_options_changed()
+{
+	if (
+		(Waypoint::Interpolation)before_options->get_value() == Interpolation::INTERPOLATION_TCB ||
+		(Waypoint::Interpolation)after_options->get_value() == Interpolation::INTERPOLATION_TCB)
+		config_tcb_params(true);
+	else
+		config_tcb_params(false);
+}
```
rodolforg: Avoid code repetition: maybe later someone touches the code but don't do it in both places…
So, I suggest to create a method named `update_tcb_params_visibility()` or something like that.

These callbacks can be or not be removed.

synfig-studio/src/gui/widgets/widget_waypoint.cpp
```diff
@@ -294,5 +289,62 @@ Widget_Waypoint::get_waypoint()const
...
+		biasLabel->hide();
+		temporalTensionLabel->hide();
+		
+		// hide spin buttons
+		spin_tension->hide();
+		spin_continuity->hide();
+		spin_bias->hide();
+		spin_temporal_tension->hide();
+	}
+}
```
rodolforg: In order to prevent forgetting some widget to show/hide, prefer to use `Gtk::Widget::set_visible(bool visible = true)`.
This way, only the adjustments will be inside an `if` statement.

rodolforg: And maybe you should just show/hide the TCB frame.

synfig-studio/src/gui/widgets/widget_waypoint.cpp
Keyikedalube: I didn't notice I was implementing the same logic on those two signal functions.
Thank you!

synfig-studio/src/gui/widgets/widget_waypoint.cpp
Keyikedalube: Another new gtkmm xp gained!
Thanks again :+1:

synfig-studio/src/gui/widgets/widget_waypoint.cpp
```diff
@@ -269,13 +262,15 @@ Widget_Waypoint::set_waypoint(synfig::Waypoint &x)
...
 	before_options->set_value((Waypoint::Interpolation)waypoint.get_before());
 	after_options->set_value((Waypoint::Interpolation)waypoint.get_after());
-
-	adj_tension->set_value(waypoint.get_tension());
-	adj_continuity->set_value(waypoint.get_continuity());
-	adj_bias->set_value(waypoint.get_bias());
-	adj_temporal_tension->set_value(waypoint.get_temporal_tension());
-
+	
+	if (
```
ice0: If I understand correctly, this code can now be replaced by:
`Widget_Waypoint::update_tcb_params_visibility()`?

Keyikedalube: Thank you :+1:  Didn't see that. Committing the changes

Keyikedalube: Running studio make... Might take a while before pushing the new changes.

synfig-studio/src/gui/widgets/widget_waypoint.h
```diff
@@ -57,15 +58,18 @@ class Widget_Waypoint;
...
 	Widget_Time *time_widget;
 	mutable synfig::Waypoint waypoint;
 	synfig::Canvas::Handle canvas;
 
-	Widget_Enum *before_options,*after_options;
+	Widget_Enum *before_options, *after_options;
 
+	// TCB Parameter members
+	Gtk::Frame *tcbFrame;
+	Gtk::Label *tensionLabel, *continuityLabel, *biasLabel, *temporalTensionLabel;
```
rodolforg: I think this isn't necessary anymore :)

Keyikedalube: Strange...
Those declaration members are not even initialized on the implementation file.

Constructors are for initializing members. I suspect the compilation might fail if I remove those members. Let me try.

Keyikedalube: Yup, compilation failed.

And I moved those variables to the header file in the first commit :) Now I feel stupid I might even forget I made this PR two days from now XD

New line 182, 196, 203, 210, and 217
https://github.com/synfig/synfig/pull/1182/commits/220e76695b20fe63edfa42944b10e5a7ad33612b

Just out of curiosity, I removed those labels and redeclared in the constructor to see if compilation will fail. And it did. Because they are local scoped within the constructor.

rodolforg: Sorry. I selected line 
Gtk::Frame *tcbFrame;
By mistake.

Keyikedalube: OK, got what you meant. Move those labels back to the implementation file.
Committing the changes right away :)

ice0:
Merged. Thank you!

PR #1104, opened 2020-01-22
fix crash get invalid value_node

rodolforg:
fix #1102  - but the real problem reason is deeper:
something is really wrong with bone deletion. App crash on quit anyway.

PR #1188, opened 2020-02-29
Fix crash when removing a list item of a parameter

rodolforg:
If you select a vertex item of Spline parameter list and then right-click on it and choose `Remove item (Smart)`, app crashes.
Not anymore. Simple, but not the best, solution.

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_dynamiclist.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_dynamiclist.cpp
@@ -832,6 +832,11 @@ const synfig::Node::time_set	& ValueNode_DynamicList::ListEntry::get_times() con
 											end = timing_info.end();
 
 	//must remerge with all the other values because we don't know if we've changed...
+	if (!value_node) {
+		static const synfig::Node::time_set empty_set;
+		synfig::error("ValueNode Dynamic List invalid");
+		return empty_set;
+	}
 	times = value_node->get_times();
 
 	for(; j != end; ++j)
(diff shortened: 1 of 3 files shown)


rodolforg:
Depends on https://github.com/synfig/synfig/pull/1104.
Could you please review both PR?

ice0:
@blackwarthog found a deeper problem, so i will continue to review after he take decision to use hack or fix original problem.

rodolforg:
@ice0 This PR just avoids Synfig Studio from crashing. Crash yells a bad reputation for the software: work hour loss :(

The problem root seems to be bad (or none) reaction to item removal of parameter list: connected signals and queued redrawing before internal (studio) structure/data update.

ice0:
@rodolforg I agree. Just want to check if better solution is possible. Anyway we do not release, until this issue is closed.

blackwarthog:
After call method LayerParamTreeStore::rebuild() does the following:
- called LayerParamTreeStore
- items from tree removes one by one
- after removing of selected item section moves to the next item and calling signal_selection_changed()
- Dock_Curves listen this signal and initiates of redraw
- error raised when selection moved to last element in list, because curretly list have a one item less

also this behavior produse a lot of extra calls of signal_selection_changed and a lot of repaints of GUI.


GTK cannot handle this situation because TreeStore (child of TreeModel) does not know nothing about widgets and selection
TreeView widget and Selection does not know when TreeStore cleared
GTK have no signal on TreeModel completely clears, it replaced by one by one calling of item removing signals

We need to safely clear th TreeStore when its content was invalidated.
But GTK assumes that item will removed from model *before* removing from user logic.
So by GTK assumed workflow we need to keep model and linked with it user data in valid state.

We have ability to remove a single item, but it have some ancient problems. Developers before us was disabled this code, and just do full rebuild of the tree. Will be goot to fix it in future.

But anyway we need an ability to do full rebuild safely.

Solution can be in adding of our own signal_clearing into LayerParamTreeStore, and connect widget to it, to clear the selection before clearing of TreeStore.

ice0:
Merged. Thank you!

ice0:
Created a new issue to find the better solution later: https://github.com/synfig/synfig/issues/1287

Issue #989, opened 2019-10-29
[Lottie exporter] Add automatic testing of code


morevnaproject:
It would be nice to have a Travis CI job, which will automatically test new commits/PRs against a set of test files.

Test files are located here - https://gitlab.com/synfig/synfig-tests/tree/master/export/lottie

This could be just a simple call of lottie exporter python script. And it should detect export fails, like this one - #917. This way we can easily detect problems with exporter at early stages.

AnishGG:
I will have a look at this.

AnishGG:
For this issue, should the changes be made in this file? : https://github.com/synfig/synfig/blob/master/travis.sh
OR
should I make a different file whose execution will be called inside `travis.sh`

ice0:
@AnishGG if i understand correctly lottie-plugin doesn't need builded Synfig for work. 

So i suggest to do some sort of testing script on Python (with PyTest or something like it) which can be executed locally, and add it to .travis.yml as a new job later.

PR #1197, opened 2020-03-05
Automate Lottie exporter tests using Travis

AnishGG:
Resolves: #989 
- The job is currently allowed to fail because there are around 6 tests that are failing.

One doubt:
- Where should I place the script `lottie_test.sh`? Currently, it is inside the root directory, but I suppose this is not the right place for it.

### DIFF
--- a/lottie_test.sh
+++ b/lottie_test.sh
@@ -2,18 +2,27 @@
 # Don't abort on Error
 # set -e
 
-DIRECTORY=`pwd`
-program="$DIRECTORY/synfig-studio/plugins/lottie-exporter/lottie-exporter.py"
+WORKDIR=`dirname "$0"`
+pushd "${WORKDIR}" > /dev/null
+WORKDIR=`pwd`
+popd > /dev/null
+
+# Python program to run lottie
+LOTTIE="${WORKDIR}/synfig-studio/plugins/lottie-exporter/lottie-exporter.py"
 
 # Clone the tests directory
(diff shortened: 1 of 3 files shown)


lottie_test.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+# Don't abort on Error
+# set -e
+
+DIRECTORY=`pwd`
```
ice0: It is better to use the same approach as in the build script:

https://github.com/synfig/synfig/blob/4f6e09c576477df12dc2726efbcaee25f1914288/2-build-debug.sh#L28

lottie_test.sh
```diff
@@ -0,0 +1,36 @@
...
+        ((Error_count=Error_count+1))
+        ((pass=0))
+    }
+
+    # To display whether a file ran successfully or not
+    if [ $pass -eq 1 ]
+    then
+        echo "[OK] $filename";
+    else
+        echo "[FAIL] $filename";
```
ice0: I suggest colorizing the output (this will be better for UX)

AnishGG:
@ice0 have made the suggested changes. Could you review this ^__^

AnishGG:
@ice0 I have made the required changes. Please review ^_^

ice0:
Merged. Thank you!

morevnaproject:
@AnishGG Thanks a lot for this!

PR #1200, opened 2020-03-07
Extra time track space at start and end

### LINKED ISSUE #1145: Add extra time track space at start and end
Building from source
Debian sid

**Issue description**:
It's tedious to position my mouse at the starting keyframe (and ending keyframe too) to bring up the right-click context menu or drag it while animating.
Screenshot from 2020-02-03 20-15-16
It would be nice if more empty space is added for these two so that keyframes and waypoints icon appear fully visible and not half. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/88038680-add-extra-time-track-space-at-start-and-end?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Keyikedalube:
Used canvas manipulation for adding extra space at start position
And delta time manipulation for end position

Fixes #1145


morevnaproject:
Hi @Keyikedalube! Thank you for this improvement! I have tested PR and found a small issue - when you zoom timelline, then padding remains visible for gray areas of timeline even when you scroll. For green areas padding is not visible.

I think video explains the issue better - https://youtu.be/4_PF6HANTS8

synfig-studio/src/gui/timeplotdata.cpp
```diff
@@ -136,7 +138,7 @@ TimePlotData::recompute_time_bounds()
 void
 TimePlotData::recompute_geometry_data()
 {
-	k = widget.get_width()/(upper - lower);
+	k = (widget.get_width()-EXTRA_TIMETRACK_SPACE*2) / (upper-lower);
```
rodolforg: You're changing the behavior of a class method that is used by more than one class, ie. not only the one you're fixing. So be careful here: your fix can break other stuff.
TimePlotData is also used by Widget/Dock Sound and Graph, for example.

Therefore, instead of making such a fixed change by `#define` directive, use/create a class property and its accessor/mutator method.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -230,7 +233,7 @@ Widget_Timeslider::on_draw(const Cairo::RefPtr<Cairo::Context> &cr)
 	const etl::handle<TimeModel> & time_model = time_plot_data->time_model;
 
 	// Draw the time line...
-	double tpx = time_plot_data->get_pixel_t_coord(time_plot_data->time) + 0.5;
+	double tpx = time_plot_data->get_pixel_t_coord(time_plot_data->time) + EXTRA_TIMETRACK_SPACE;
```
rodolforg: Maybe `time_plot_data` should handle this extra space internally in its members `get_pixel_t_coord()` and `get_t_from_pixel_coord()`

rodolforg: and `get_double_pixel_t_coord()`

Keyikedalube:
@morevnaproject Thanks for pointing that out
I think it's better to paint the entire timeslider gray and not do any fancy extra whitespace indicator.

synfig-studio/src/gui/timeplotdata.cpp
Keyikedalube: Well, that was the intention. Change the root class and let subclass follow. And define directive was the easier method because it's readable. BTW I haven't checked sound and graph source code so yes you're correct my change is breaking that two classes.

Looking at the timeplotdata...

synfig-studio/src/gui/widgets/widget_timeslider.cpp
Keyikedalube: In my current implementation where I made the fix on `TimePlotData::recompute_geometry_data()`
managing **k** value in this function makes sense

If we had to let `get_pixel_t_coord()` take care of k then there is no point for recompute... function to exist because we'd have to
1. re-calculate k first (notice the get_width()-EXTRA_TIMETRACK_SPACE*2 before being divided by time difference)
2. and then return time coordinate

within the get_pixel_t_coord() function itself which kind of breaks the rule "let function take care of one thing only"

I feel ok with my current solution
The only drawback that new contributors implementing a time slider related widget take note of pushing their canvas drawing by 16px

But if you feel otherwise I understand
Maybe you can take over this PR and apply your creative fixes :)

Keyikedalube:
@morevnaproject merging this PR is up to rodolforg now
I've done my part but he has more expertise in this area
so you might want to assign this PR to him If he feels a better solution can be pushed

Keyikedalube:
New weekend update for graph and sound class I missed out last week
Hope I've taken care of all the time-related widgets visual alignment 
# Graph editor
Screenshot from 2020-03-14 16-35-55
# Sound wave
Screenshot from 2020-03-14 16-34-54

Keyikedalube:
# And bad news for graph editor
Canvas x coordinate has been translated to 16.0
But there is no input handling event for this class
so below is a screenshot demonstrating how mouse cursor has to be placed behind any waypoint to bring it in focus
1. typical - but waypoint is not catching the event
2. works - but that's a good example of bad UX 
Screenshot from 2020-03-14 16-57-41

I'm not going to work on this event handling hack, unfortunately

It's been fun poking and reading tons of synfig code
However, if there is a better way to implement the time track with proper start and end time position
Let me know and I'll close this PR

Issue #1191, opened 2020-03-01
Rename "Warp Layer" to "Perspective Layer"


morevnaproject:
We need to rename Warp Layer to Perspective Layer.

rodolforg:
It would be good to update wiki too

ARYA-BHARDWAJ:
may I work on this issue

ice0:
@ARYA-BHARDWAJ sure! It's your now! :)

ankit-kumar-dwivedi:
@morevnaproject please approve the related PR as well

PR #1206, opened 2020-03-08
Rename wrap

ARYA-BHARDWAJ:
This is regarding issue no  #1191
-warp layer is renamed to perspective layer 
-Identation fixed


ARYA-BHARDWAJ:
@ice0  I have fixed indentation this time and I hope this time all issues have been resolved

rodolforg:
It looks like that you `git mv` the `warp.h` to `perspective.h`, and that's good. But why didn't you do the same with `warp.c`? This way we lose all info that `git log` and `git blame` could bring us to help investigate issues. :)

PR #1209, opened 2020-03-08
fix i18n for Set Keyframe Description dialog

rodolforg:
* "Set" button -> "Ok", because it would be translated as "group" meaning otherwise

* Don't concatenate strings, because several languages don't follow genitive case ordering like Noun (owner) + Noun like English and German do.
Eg., in Portuguese, we use Noun + " de " + Noun(owner). So we need full text for i18n.


ice0:
Merged. Thank you!

PR #1222, opened 2020-03-10
warp layer is now renamed to perspective layer

ARYA-BHARDWAJ:
This is regarding issue no #1191 
- changes are made in folder synfig-core/src/modules/lyr_std
- wrap.h , wrap.cpp changed to perspective.h, perspective.cpp 
- Updated cmake.list

### DIFF
--- a/synfig-core/src/modules/lyr_std/perspective.cpp
+++ b/synfig-core/src/modules/lyr_std/perspective.cpp
@@ -67,7 +67,7 @@ using namespace lyr_std;
 /* === G L O B A L S ======================================================= */
 
 SYNFIG_LAYER_INIT(Perspective);
-SYNFIG_LAYER_SET_NAME(Perspective,"perspective");
+SYNFIG_LAYER_SET_NAME(Perspective,"warp");
 SYNFIG_LAYER_SET_LOCAL_NAME(Perspective,N_("Perspective"));
 SYNFIG_LAYER_SET_CATEGORY(Perspective,N_("Distortions"));
 SYNFIG_LAYER_SET_VERSION(Perspective,"0.2");
--- a/synfig-core/src/modules/lyr_std/perspective.cpp
+++ b/synfig-core/src/modules/lyr_std/perspective.cpp
@@ -1,6 +1,6 @@
 /* === S Y N F I G ========================================================= */
-/*!	\file warp.cpp
-**	\brief Implementation of the "Warp" layer
+/*!	\file perspective.cpp
+**	\brief Implementation of the "Perspective" layer
 **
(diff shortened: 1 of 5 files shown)


ARYA-BHARDWAJ:
@rodolforg please review

ice0:
@ARYA-BHARDWAJ please check, if you can open files with warp layer which is saved by old Synfig version.

ARYA-BHARDWAJ:
Yes  can open files with warp layer which is saved by old Synfig version but we have to add perspective layer.

ARYA-BHARDWAJ:
No, we can't open files saved by old synfig version. Can you suggest me how to solve it ??

synfig-core/src/modules/lyr_std/perspective.cpp
```diff
@@ -66,12 +66,12 @@ using namespace lyr_std;
...
 /* === G L O B A L S ======================================================= */
 
-SYNFIG_LAYER_INIT(Warp);
-SYNFIG_LAYER_SET_NAME(Warp,"warp");
-SYNFIG_LAYER_SET_LOCAL_NAME(Warp,N_("Warp"));
-SYNFIG_LAYER_SET_CATEGORY(Warp,N_("Distortions"));
-SYNFIG_LAYER_SET_VERSION(Warp,"0.2");
-SYNFIG_LAYER_SET_CVS_ID(Warp,"$Id$");
+SYNFIG_LAYER_INIT(Perspective);
+SYNFIG_LAYER_SET_NAME(Perspective,"perspective");
```
ice0: To support old versions, i think you need to set name here back to "warp".
It is not visible in GUI, but used for file saving/reading.

So try to change it back:
```
SYNFIG_LAYER_SET_NAME(Perspective,"warp");
```
and check if this will be enough,

ARYA-BHARDWAJ: Yeah it worked thanks.

ice0:
Awesome! We need to wait until @morevnaproject also reviews it. But he is busy, and it may take some time. Therefore, I suggest you choose a new issue to work on it if you want.

morevnaproject:
Hi @ARYA-BHARDWAJ! Apologies for such a long delay.

The PR looks good. I just noticed you have added one extra empty file -  `log/trace.log`. Can you please remove it from PR? ^__^

morevnaproject:
> I think those changes should be squashed before merge.

I am afraid that git will loose track of file renames after squash. I do not see anything wrong in merging that without squash. ^__^

It would be nice to rebase, though for a clean merge commit.

Issue #1150, opened 2020-02-08
CI: Run "autobuild/transifex-push-template.sh" every time when "translations" branch updated


morevnaproject:
We need to run "autobuild/transifex-push-template.sh" every time when "translations" branch updated.

AnishGG:
Does this require to add a new job to the "translations" branch which will be just a simple call to `autobuild/transfex-push-template.sh`?

ice0:
@AnishGG Yes, I think that additional job will be better because it gives a clearer understanding of what will happen when it is launched.

AnishGG:
Then I can take this issue!

morevnaproject:
Hi everyone! I am afraid that the job at its current implementation will fail. Did anyone checked if it works? ^__^

AnishGG:
When I opened a PR to `synfig:translations` branch, the job was failing. But @ice0 suggested that it will fail because of an error in .po file.
Here is the job from that PR: https://travis-ci.com/github/synfig/synfig/jobs/296861674

ice0:
@morevnaproject sure, i tested it. Job works fine.

ice0:
Ok. We need to add one more step and install transifex binary. The whole process is described here:
https://docs.transifex.com/transifex-github-integrations/github-tx-client

@morevnaproject can you do a test project or something where we can push translations for testing purposes?

Also I suggest to add one more job to do translations validation.

morevnaproject:
> Ok. We need to add one more step and install transifex binary. The whole process is described here:
>
> https://docs.transifex.com/client/installing-the-client
> https://docs.transifex.com/transifex-github-integrations/github-tx-client

Oh, I found that Transifex now capable to work with GitHub directly, without CI -
https://docs.transifex.com/transifex-github-integrations/github-tx-ui

This seems to be easier than configuring tx-client. ^__^"

> Also I suggest to add one more job to do translations validation (#1250)

That's great idea! @AnishGG can you do that? ^__^

AnishGG:
> > Also I suggest to add one more job to do translations validation (#1250)
> 
> That's great idea! @AnishGG can you do that? ^__^

Yup, I will make a P.R. regarding this soon.

PR #1223, opened 2020-03-11
New job for running "autobuild/transifex-push-template.sh"

AnishGG:
Issue: #1150 
@ice0 
I have pushed a new job and removed all other jobs for the sake of fast testing. But the job is failing.
Is this expected behaviour?


ice0:
@AnishGG Yes, there is an error in the .po file. So the script should fail.

AnishGG:
Alright!
I will add all other jobs and ask for a review then.

.travis.yml
```diff
@@ -90,6 +90,12 @@ matrix:
     script: ./travis.sh
     after_script: ccache -s # show ccache stats
 
+  - os: linux
+    name: "Transifex push template Ubuntu 16.04 Xenial"
+    sudo: required
+    dist: xenial
+    script: bash autobuild/transifex-push-translations.sh
```
ice0: Also i suggest to change shell inside `transifex-push-translations.sh` from `/bin/sh` to `/bin/bash`, as `pushd/popd` cannot be executed by `sh`

AnishGG:
> Please do not forget to add condition:
> job must run only on `translations` branch.
> 
> https://docs.travis-ci.com/user/conditional-builds-stages-jobs/

Oh, I see.
According to my knowledge, if we push a commit to a specific branch(say 'X'), it should trigger the branch X's `.travis.yml`, right?
Please correct me if I am wrong.

ice0:
> According to my knowledge, if we push a commit to a specific branch(say 'X'), it should trigger the branch X's `.travis.yml`, right?
> Please correct me if I am wrong.

Yes.

AnishGG:
> > According to my knowledge, if we push a commit to a specific branch(say 'X'), it should trigger the branch X's `.travis.yml`, right?
> > Please correct me if I am wrong.
> 
> Yes, it triggers `.travis.yml` from that branch.

Then we don't need an `if` condition, right? Because we will be editing only that branch's `.travis.yml`.

ice0:
> Then we don't need an `if` condition, right? Because we will be editing only that branch's `.travis.yml`.

The main idea was to update `translations` branch from `master` branch every time when we need to update translations. If we do different versions of `.travis.yml` when we need to resolve merge conflicts every time when we do updates to this branch.

AnishGG:
> The main idea was to update `translations` branch from `master` branch every time when we need to update translations. If we do different versions of `.travis.yml` when we need to resolve merge conflicts every time when we do updates to this branch.

Oh!
Got it now. Then I will introduce the `if` condition!
One more doubt: should I make the P.R. to `master` branch or `translations` branch then?

ice0:
> Oh!
> Got it now. Then I will introduce the `if` condition!
> One more doubt: should I make the P.R. to `master` branch or `translations` branch then?

To `master` branch :)

AnishGG:
> > Oh!
> > Got it now. Then I will introduce the `if` condition!
> > One more doubt: should I make the P.R. to `master` branch or `translations` branch then?
> 
> To `master` branch :)

Alright!
I might have to open a new PR then. Will close this if required

ice0:
You can change branch by editing this PR.

AnishGG:
Okay, will try that!

Issue #1131, opened 2020-01-30
[CMake] Change the layout of the builded files inside the cmake-build directory


ice0:
Check the directory structure which is created by 2-build-debug.sh or 2-build-release.sh
Modify the CMake scripts to make the same structure.

Keyikedalube:
Last time I tried 2-build-cmake.sh on Fedora the built files were stored in a folder named something *portable(?)*
After switching to Debian 2-build-cmake.sh is not usable [#928](https://github.com/synfig/synfig/issues/928#issuecomment-575589239) So don't correctly remember how builded files are stored for cmake.

But here's how I understand this issue. You want to replace _debug/ and _production/ directory depending on arg passed to build-cmake and not create a separate build directory?

ice0:
Here current CMake status. (https://github.com/synfig/synfig/issues/928)

And commands you need to execute:
```
git clone https://github.com/synfig/synfig
cd synfig
mkdir build-cmake && cd build-cmake
cmake ..  -DCMAKE_INSTALL_PREFIX=~/synfig-cmake
make -j4
make install
export PATH=~/synfig-cmake/bin/:$PATH
export LD_LIBRARY_PATH=~/synfig-cmake/lib/:$LD_LIBRARY_PATH
export SYNFIG_ROOT=~/synfig-cmake/
make build_images -j4
make install
# fix incorrect install path for icons
mkdir ~/synfig-cmake/share/pixmaps/synfigstudio
mv ~/synfig-cmake/share/pixmaps/*.png ~/synfig-cmake/share/pixmaps/synfigstudio
```

As you can see, you need to do `make install` before `make build_images`.
If you try to run `synfig` (console binary needed for build_images target), it doesn't start, because it cannot find `libsynfig.so` in it's build directory. And to do that, we run `make install` command and it creates `bin` and `lib` directories, and binary can be executed at this time.
The goal of this issue is to do the same thing, but not at `install` step, but right at build step. This will allow to user, just do `cmake ..` to do the whole build.

ice0:
Directory structure, in CMake terminology:

```
${PROJECT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/bin
${PROJECT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/etc
${PROJECT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/lib
...
and so on
```

Keyikedalube:
[...]

PR #1224, opened 2020-03-11
[CMake] Updated CMake script to change layout of build files.

AYESDIE:
Fixes on #1131. 

Changes:
- Updated `CMakeLists.txt`. Inside the build folder, it makes `bin` and `lib` (~~ETL~~, ~~include~~). Execution still require [these commands](https://github.com/synfig/synfig/issues/1131#issuecomment-586873899).
- Updated `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -241,3 +241,9 @@ synfig-studio/doc
 /ETL/nbproject/
 /synfig-core/nbproject/
 /synfig-studio/nbproject/
+
+# ------------------------------------------------------------------------
+# CLion projects
+# ------------------------------------------------------------------------
+/.idea
+cmake-build-*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 2.8)
 
 project(synfig-studio)
 
+set(SYNFIG_BUILD_ROOT ${CMAKE_BINARY_DIR}/output)
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${SYNFIG_BUILD_ROOT}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${SYNFIG_BUILD_ROOT}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${SYNFIG_BUILD_ROOT}/bin)
+
 add_subdirectory(ETL)
 include_directories(ETL)
 include_directories(${PROJECT_BINARY_DIR}/ETL) # to find the generated etl_profile.h
--- a/synfig-core/src/modules/CMakeLists.txt
+++ b/synfig-core/src/modules/CMakeLists.txt
@@ -43,6 +43,9 @@ set(MODS_ENABLED
 #    mptr_mplayer # - "This code has vulnerabilities"
 )
 
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${SYNFIG_BUILD_ROOT}/lib/synfig/modules)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${SYNFIG_BUILD_ROOT}/lib/synfig/modules)
+
 ## Magick++ support
 option(WITH_MAGICPP "Enable Magic++ module (output format support)" OFF)
 option(WITHOUT_MAGICPP "Disable Magic++ module (output format support)" OFF)


ice0:
Hi, @AYESDIE!

This can be done with 3 lines of code.
Please see here: https://cmake.org/cmake/help/v3.0/manual/cmake-variables.7.html

```
CMAKE_RUNTIME_OUTPUT_DIRECTORY
CMAKE_LIBRARY_OUTPUT_DIRECTORY
CMAKE_ARCHIVE_OUTPUT_DIRECTORY
```

also `synfig-core/src/modules/CMakeLists.txt` modules must goes to `lib/synfig/modules/`

AYESDIE:
@ice0 Ah, rather than going for a per-target basis, updating those directory in the root `CMakeLists.txt` seems better. I'll work on the changes and push it soon. 

Should I make changes on how `ETL` and `include` is built?

ice0:
> updating those directory in the root CMakeLists.txt seems better

Yes. Also this will be much easier to support in the future.

> Should I make changes on how ETL and include is built?

I did not think about them. Therefore, I suggest just skipping them for now and back to them later.

AYESDIE:
> Yes. Also this will be much easier to support in the future.
Agreed :)

I force pushed the new changes. I removed the `ETL` and `include` build part as of now.

CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 2.8)
 
 project(synfig-studio)
 
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/modules)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/modules)
```
ice0: I suggest to set `CMAKE_LIBRARY_OUTPUT_DIRECTORY ` to `${CMAKE_BINARY_DIR}/lib/`
and inside `synfig-core/src/modules/CMakeLists.txt` set to `${CMAKE_BINARY_DIR}/lib/modules/`

synfig-core/src/synfig/CMakeLists.txt
```diff
@@ -158,3 +158,9 @@ install(
     FILE "${PROJECT_NAME}-config.cmake"
     DESTINATION "lib/cmake/${PROJECT_NAME}"
 )
+
```
ice0: This code can be removed, if you change logic a little bit.

synfig-studio/src/synfigapp/CMakeLists.txt
```diff
@@ -29,3 +29,9 @@ install(
     TARGETS synfigapp
     DESTINATION lib
 )
+
```
ice0: This code can be removed too :)

.gitignore
```diff
@@ -241,3 +241,8 @@ synfig-studio/doc
 /ETL/nbproject/
 /synfig-core/nbproject/
 /synfig-studio/nbproject/
+
+# ------------------------------------------------------------------------
+# CLion projects
+# ------------------------------------------------------------------------
+/.idea
```
ice0: If you use CLion, please add cmake-build-* too.

ice0:
@AYESDIE, also i want to ask, which OS you are using?

AYESDIE:
I am on Ubuntu 19.10.

ice0:
> I am on Ubuntu 19.10.

Awesome!

Also forgot to mention, please change the path to ${CMAKE_BINARY_DIR}/**output**/->lib/bin/etc

CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 2.8)
 
 project(synfig-studio)
 
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
```
ice0: Please change path to `${CMAKE_BINARY_DIR}/output/lib/`

Almost forgot about it.

CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 2.8)
...
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
```
ice0: output/bin

synfig-core/src/modules/CMakeLists.txt
```diff
@@ -43,6 +43,9 @@ set(MODS_ENABLED
 #    mptr_mplayer # - "This code has vulnerabilities"
 )
 
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/modules)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/modules)
```
ice0: here too

AYESDIE:
I was building on my end to make sure things were correct places :))

synfig-core/src/synfig/CMakeLists.txt
```diff
@@ -145,7 +145,7 @@ install(
 install(
     TARGETS synfig
     EXPORT synfig
-    LIBRARY DESTINATION lib
+    LIBRARY DESTINATION ${CMAKE_BINARY_DIR}/output/lib
```
ice0: here need to be relative path

synfig-studio/src/synfigapp/CMakeLists.txt
```diff
@@ -27,5 +27,5 @@ include(vectorizer/CMakeLists.txt)
 
 install(
     TARGETS synfigapp
-    DESTINATION lib
+    DESTINATION ${CMAKE_BINARY_DIR}/output/lib
```
ice0: here too, change it back to lib

ice0:
Ok, also please check `make install` target, and you see why i add new comments :)

synfig-core/src/synfig/CMakeLists.txt
AYESDIE: But `2-build-debug.sh` builds it inside a `build` folder. 
```
  - _debug
    - build
      - lib
        - libsynfig.so  
```

Making structure like:
```
  - cmake-build-debug
    - output
      - lib
        - libsynfig.so
```

synfig-studio/src/synfigapp/CMakeLists.txt
AYESDIE: Same as above.

synfig-core/src/synfig/CMakeLists.txt
AYESDIE: Ah, okay, nevermind. It works the way you told too.

AYESDIE:
> Ok, also please check make install target, and you see why i add new comments :)

It worked the way it was before too, since I kept on checking it with `make install` on each commit, so I thought it was correct. But I realize I overcomplicated things abit.

ice0:
Awesome work! This is exactly what I wanted!

Therefore, I propose to continue tomorrow.

P.S. I will merge your changes after all the checks have been completed.

AYESDIE:
Thank you for guiding :) 

The next thing will be to refactor the `build_images` target, right?  #1131 to be exact?

ice0:
Yes. But i want to add one more simple issue for tomorrow.
We need to add /etc/modules.cfg and /share/synfig/*.glade files to do synfigstudio runnable.

In this way, new students will be able to use IDEs such as CLion or NetBeans to build Synfig Studio and check they changes.

AYESDIE:
Ah, sounds good to me. I'll take a look at it tomorrow :)

CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 2.8)
 
 project(synfig-studio)
 
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/output/lib)
```
ice0: As we planning to use that path further i suggest to extract that path `${CMAKE_BINARY_DIR}/output/` into a separate variable `SYNFIG_BUILD_ROOT `

synfig-core/src/modules/CMakeLists.txt
```diff
@@ -43,6 +43,9 @@ set(MODS_ENABLED
 #    mptr_mplayer # - "This code has vulnerabilities"
 )
 
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/output/lib/modules)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/output/lib/modules)
```
ice0: Just checked. This path must be `lib/synfig/modules` not `lib/modules`.

CMakeLists.txt
AYESDIE: That makes sense. Would make things easier to maintain in the future.

ice0:
Merged! Thank you!

Issue #1225, opened 2020-03-11
[CMake] modules.cfg and *.glade files


ice0:
We need to add /etc/modules.cfg and /share/synfig/*.glade files to do synfigstudio runnable.

ice0:
If you try to run `./synfig` from `output/bin/ ` directory - all works fine.
But if you try to run `./synfigstudio` - it exits with error 
```
synfig(766) [10:51:23] error: FileError: Failed to open file 
“share/synfig/ui/canvas_options.glade”: No such file or directory
```
We need to copy *.glade files from `synfig-studio/ui` directory to `$SYNFIG_BUILD_ROOT/share/synfig/ui/`

ice0:
After what it still cannot find the files (we need to change a path searching logic), but as a temporary solution we can set environment variable:
```
SYNFIG_ROOT=/path/ti/output/dir ./synfigstudio
```
and `./synfigstudio`  should start with broken images.

ice0:
Also we need to write `modules.cfg` file generated by `synfig-core/src/modules/CMakeLists.txt` to `$SYNFIG_BUILD_ROOT/etc`

AYESDIE:
I worked out on it. Should I open a PR for it?

ice0:
Assigned to you. Just do PR when you will be ready.

PR #1228, opened 2020-03-12
[CMake] Update CMake script for /share and /etc.

AYESDIE:
Fixes: #1225

### DIFF
--- a/synfig-core/src/modules/CMakeLists.txt
+++ b/synfig-core/src/modules/CMakeLists.txt
@@ -73,6 +73,12 @@ endforeach(MOD)
 
 file(WRITE synfig_modules.cfg ${SYNFIG_MODULES_CONTENT})
 
+configure_file(
+    synfig_modules.cfg
+    ${SYNFIG_BUILD_ROOT}/etc/synfig_modules.cfg
+    COPYONLY
+)
+
 install(
     FILES synfig_modules.cfg
     DESTINATION etc
--- a/synfig-studio/ui/CMakeLists.txt
+++ b/synfig-studio/ui/CMakeLists.txt
@@ -5,6 +5,12 @@ set(UIFILES
 )
 
 foreach (UIFILE IN ITEMS ${UIFILES})
+    configure_file(
+        ${CMAKE_CURRENT_SOURCE_DIR}/${UIFILE}
+        ${SYNFIG_BUILD_ROOT}/share/synfig/ui/${UIFILE}
+        COPYONLY
+    )
+
     install(
         FILES ${CMAKE_CURRENT_SOURCE_DIR}/${UIFILE}
         DESTINATION share/synfig/ui


ice0:
@AYESDIE you need to copy them on build stage, not on install stage. 

The main idea is to allow user just open project in CLion (for example) and just build it and run. 
As for now it requires to do `make install` and choose another binary (not produced by `make` command)

AYESDIE:
The building process is still hectic. But I think once this is done, we will be able to run `synfig` for `build_images` target from our `cmake-build` directly which will make things easier :)

AYESDIE:
Oh, wait, I messed it abit. I'll push update soon.

ice0:
2 additions:
1. Try to not require CMake higher than 3.0. Synfig is distributed across multiple platforms, so it’s better not to require a higher version of CMake than you really need. (Not all of them have required version)
2. https://cmake.org/cmake/help/v3.0/command/file.html

> WRITE will write a message into a file called ‘filename’. It overwrites the file if it already exists, and creates the file if it does not exist. (If the file is a build input, use configure_file to update the file only when its content changes.)

So better to use `configure_file` for that.

AYESDIE:
@ice0 This has been bugging me a bit. Would we not require those *.glade and modules.cfg file if we decide to `make install`? If we don't go for `install` and just copy those, they won't be installed to their appropriate location when we do `make install`. 

This seems to be a case specific thing where we want to run it on something like CLion.

ice0:
They will be installed if you leave `install` command as is.

For example you just need to do something like:
```
foreach (UIFILE IN ITEMS ${UIFILES})
   # copying file to SYNFIG_BUILD_ROOT too
   configure_file(...)

    install(
        FILES ${CMAKE_CURRENT_SOURCE_DIR}/${UIFILE}
        DESTINATION share/synfig/ui
    )
endforeach()
```

AYESDIE:
Thanks, and I think this should do it. 

I checked the `~/synfig-cmake` and `cmake-build/output` directory and it works.

synfig-core/src/modules/CMakeLists.txt
```diff
@@ -73,6 +73,11 @@ endforeach(MOD)
 
 file(WRITE synfig_modules.cfg ${SYNFIG_MODULES_CONTENT})
 
+configure_file(
```
ice0: Please add COPYONLY, as it is faster and does not require parsing the file.

ice0:
In addition, after these changes, can you check if you can just build and run the synfigstudio target in CLion.

AYESDIE:
Opening it requires changing configuration from CLion itself. On specifying `SYNFIG_ROOT` the application opened.

ice0:
Awesome!

Can you take https://github.com/synfig/synfig/issues/1230?

AYESDIE:
Sure :)

ice0:
Merged. Thank you!

PR #1241, opened 2020-03-13
Bump version for settings file

morevnaproject:
It's obvious that we'd better to change name for settings file.


ice0:
Merged. Thank you!

PR #1244, opened 2020-03-14
OSX build: Remove hardcoded references to Python version

morevnaproject:


3-package-osx-dmg.sh
```diff
@@ -199,8 +199,8 @@ mkapp()
 	mkdir -p "${APPCONTENTS}/Frameworks/Python.framework/Versions/${PYTHON_VERSION}/lib/python${PYTHON_VERSION}/"
 	rsync -av --exclude "__pycache__" "${MACPORTS}${PKG_PREFIX}/Frameworks/Python.framework/Versions/${PYTHON_VERSION}/lib/python${PYTHON_VERSION}/" "${APPCONTENTS}/Frameworks/Python.framework/Versions/${PYTHON_VERSION}/lib/python${PYTHON_VERSION}/"
 	pushd "${APPCONTENTS}/Frameworks/Python.framework/Versions/${PYTHON_VERSION}/lib/python${PYTHON_VERSION}/" > /dev/null
-	mkdir -p ../../../../../../lib/python3.6/site-packages
-	ln -sf ../../../../../../lib/python3.6/site-packages site-packages
+	mkdir -p ../../../../../../lib/python${PYTHON_VERSION}/site-packages
```
ice0: Why not to use ${APPCONTENTS} here? I think it will be the same:

`${APPCONTENTS}/lib/python${PYTHON_VERSION}`

morevnaproject: In next line (`ln -sf ...`) we need to specify relative path (to make sure symlink is correct if AppDir is launched from any location. So, I have used same relative path for `mkdir ...` statement. 

In this case, it is easier to see that paths in both lines are the same. If we will need to fix path, then we will simply fix it to the **same** value in both places. This is simpler and less chances for mistake here. ^__^

ice0: Got it.

ice0:
Merged. Thank you!

PR #1245, opened 2020-03-14
Fix issue #1216

### LINKED ISSUE #1216: Waypoint types list is not consistent across Synfig Studio
Master: 4f6e09

**Issue description**:
In Synfig Studio there are various places where the user can choose the type of waypoint to use.

I have noticed that those lists are not always consistent.

For example in the timeline we have:
order_timeline

While editing a specific waypoint the order is different:
order_waypoint

Probably there are other places where this order is not consistent. It would be nice to have it consistent across all Synfig Studio.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/90008123-waypoint-types-list-is-not-consistent-across-synfig-studio?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

FirasH:
Fixes #1216 
Consistent order of Waypoint types: Clamped, TCB, Constant, Ease, Linear


FirasH:
#1216

PR #1246, opened 2020-03-15
Fix issue #1231

### LINKED ISSUE #1231: Setting environment dependencies on openSUSE leads to conflicting commands
Master: c3054c4
openSUSE Leap 15.1

**Issue description**:
If you want to setup a freshly installed openSUSE machine using the provided script:
"1-setup-linux-native.sh"

The command:
"if command -v apt-get >/dev/null; then"

which includes apt-get is detected by the OS and thus the part needed to set openSUSE using zypper is not chosen:
"elif which zypper >/dev/null; then"

Currently I have simply removed all the code before the code for zypper and it works correctly. Probably the order should be from the most specific package manager (e.g zypper or yum) to the most generic like apt-get.

FirasH:
Fixes #1231 
Change the order of the Package Managers for Linux system to avoid ambiguities with apt-get

### DIFF
--- a/1-setup-linux-native.sh
+++ b/1-setup-linux-native.sh
@@ -25,94 +25,7 @@
 set -e
 
 echo "Checking dependencies..."
-if command -v apt-get >/dev/null; then
-        if [ ! -f /etc/altlinux-release ]; then
-            #
-            #  Ubuntu/Debian
-            #
-            PKG_LIST=" \
-                build-essential \
-                autoconf automake autopoint \
-                shared-mime-info \
-                libltdl3-dev \
-                libtool \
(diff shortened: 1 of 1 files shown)


FirasH:
I have tested  it on openSUSE and it works perfectly.
Could someone please verify if it works on Ubuntu/Debian?

FirasH:
Fixes #1231

FirasH:
Fixes #1231

ice0:
Checked on Ubuntu - works fine. @morevnaproject please review.

morevnaproject:
Looks good to me! ^__^

PR #1254, opened 2020-03-17
[Lottie Exporter] Add support for Cos convert and Integer convert method

saurabh-ramola:
Added cos convert method for Lottie Exporter plugin.
@AnishGG , @ice0 , please review :smiley: .

### DIFF
--- a/synfig-studio/plugins/lottie-exporter/common/Param.py
+++ b/synfig-studio/plugins/lottie-exporter/common/Param.py
@@ -487,6 +487,37 @@ class Param:
                 self.expression = ret
                 return ret, self.expression_controllers
 
+            elif self.param[0].tag == "cos":
+                self.subparams["cos"].extract_subparams()
+                angle, eff_1 = self.subparams["cos"].subparams["angle"].recur_animate("region_angle")
+                amp, eff_2 = self.subparams["cos"].subparams["amp"].recur_animate("real")
+                self.expression_controllers.extend(eff_1)
+                self.expression_controllers.extend(eff_2)
+                
+                if self.dimension == 2:
+                    ret = "mul(Math.cos(degreesToRadians({angle})), [{amp}, {amp}])"
+                else:
+                    ret = "mul(Math.cos(degreesToRadians({angle})),{amp})"
+                ret = ret.format(angle=angle,amp=amp)
+
+                self.expression = ret
+                return ret, self.expression_controllers
+
+            elif self.param[0].tag == "fromint":
+                self.subparams["fromint"].extract_subparams()
+                if self.dimension == 2:
+                    link, eff_1 = self.subparams["fromint"].subparams["link"].recur_animate("scalar_multiply")
+                    self.expression_controllers.extend(eff_1)
+                    ret = "[mul(Math.round({link}), {PIX_PER_UNIT}), mul(Math.round({link}), {PIX_PER_UNIT})]"
+                    ret = ret.format(link=link, PIX_PER_UNIT=settings.PIX_PER_UNIT)
+                else:
+                    link, eff_1 = self.subparams["fromint"].subparams["link"].recur_animate("scalar_multiply")
+                    self.expression_controllers.extend(eff_1)
+                    ret = "mul(Math.round({link}), {PIX_PER_UNIT})"
+                    ret = ret.format(link=link, PIX_PER_UNIT=settings.PIX_PER_UNIT)
+                self.expression = ret
+                return ret, self.expression_controllers
+
         else:
             self.single_animate(anim_type)
             # Insert the animation into the effect
@@ -799,6 +830,25 @@ class Param:
                 else:
                     ret = math.sin(angle)*amp
 
+            elif self.param[0].tag == "cos":
+                angle = self.subparams["cos"].subparams["angle"].__get_value(frame)
+                amp = self.subparams["cos"].subparams["amp"].__get_value(frame)
+                angle = math.radians(angle)
+                if isinstance(amp, list):
+                    ret = [0, 0]
+                    ret[0] = math.cos(angle) * amp[0]
+                    ret[1] = math.cos(angle) * amp[1]
+                else:
+                    ret = math.cos(angle)*amp
+
+            elif self.param[0].tag == "fromint":
+                link = self.subparams["fromint"].subparams["link"].__get_value(frame)
+                if isinstance(link, list):
+                    ret = [0, 0]
+                    ret[0] = round(link[0])*settings.PIX_PER_UNIT
+                    ret[1] = round(link[1])*settings.PIX_PER_UNIT
+                else:
+                    ret = round(link)*settings.PIX_PER_UNIT
 
         else:
             ret = self.get_single_value(frame)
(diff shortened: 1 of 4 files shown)


synfig-studio/plugins/lottie-exporter/settings.py
```diff
@@ -69,7 +69,7 @@
 GROUP_LAYER = {"group", "switch"}
 SKELETON_LAYER = {"skeleton"}
 UNKNOWN_LAYER = "unknown_layer"
-CONVERT_METHODS = {"add", "average", "composite", "exp", "linear", "radial_composite", "scale", "subtract", "switch", "weighted_average", "bone_link", "bone", "bone_root","sine"}
+CONVERT_METHODS = {"add", "average", "composite", "exp", "linear", "radial_composite", "scale", "subtract", "switch", "weighted_average", "bone_link", "bone", "bone_root","sine","cos"}
```
AnishGG: Can you arrange these alphabetically?

synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -799,6 +815,19 @@ def __get_value(self, frame):
...
                     ret = math.sin(angle)*amp
 
+            elif self.param[0].tag == "cos":
+                angle = self.subparams["cos"].subparams["angle"].__get_value(frame)
+                amp = self.subparams["cos"].subparams["amp"].__get_value(frame)
+                angle = math.radians(angle)
+                
+                if isinstance(amp, list):
+                    ret = [0, 0]
+
```
AnishGG: No need of line break here.

synfig-studio/plugins/lottie-exporter/settings.py
saurabh-ramola: Done :smiley: .

synfig-studio/plugins/lottie-exporter/common/Param.py
saurabh-ramola: Done :smiley: .

AnishGG:
Looks good to me. ^__^
Keep up the good work!

saurabh-ramola:
Added Integer convert method for Lottie Exporter plugin.
@AnishGG , please review :smiley: .

saurabh-ramola:
Added support for Integer convert method.

synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -799,6 +828,25 @@ def __get_value(self, frame):
...
+                    ret = math.cos(angle)*amp
+
+            elif self.param[0].tag == "fromint":
+                link = self.subparams["fromint"].subparams["link"].__get_value(frame)
+                if isinstance(link, list):
+                    ret = [0, 0]
+                    ret[0] = int(link[0]/60)*60.0
+                    ret[1] = int(link[1]/60)*60.0
+                else:
+                    ret = int(link/60)*60.0
```
AnishGG: If you add the following line here:
`print(frame, ret)`
You will be able to see that the actual values produced by lottie differ from the values produced here.
I am attaching the `.sif` used by me
[integer_circle.zip](https://github.com/synfig/synfig/files/4348879/integer_circle.zip)

Also, instead of 60 here, I guess you need to use [settings.PIX_PER_UNIT](https://github.com/synfig/synfig/blob/77164e8371269afb0490287f200f6c35fe067e88/synfig-studio/plugins/lottie-exporter/common/misc.py#L33)

AnishGG: You are getting a wrong value, mostly because `fromint` uses `round()` instead of `int()`.
Am I right here @ice0 ?

saurabh-ramola: Changed int with round and it produces the correct values. So I guess this is what needed to be done :smiley: .
Thank you for the help ^_^

saurabh-ramola:
@AnishGG, I have also added support for Integer convert method in simple_circle. Please have a look. ^_^

synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -487,6 +487,37 @@ def recur_animate(self, anim_type):
...
+
+                self.expression = ret
+                return ret, self.expression_controllers
+
+            elif self.param[0].tag == "fromint":
+                self.subparams["fromint"].extract_subparams()
+                if self.dimension == 2:
+                    link, eff_1 = self.subparams["fromint"].subparams["link"].recur_animate("points")
+                    self.expression_controllers.extend(eff_1)
+                    ret = "[{link},{link}]"
```
AnishGG: Here simple circle implementation is giving different output than expected.
I am attaching the synfig file on which it fails:
[integer_circle.zip](https://github.com/synfig/synfig/files/4350165/integer_circle.zip)

AnishGG: I suspect this is happening because you are using `animation type` as `real` and `points`, whereas we need something discrete.

AnishGG: BTW, why have you used different `anim_type` in the `if-else` conditions?

synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -487,6 +487,37 @@ def recur_animate(self, anim_type):
...
 
+            elif self.param[0].tag == "cos":
+                self.subparams["cos"].extract_subparams()
+                angle, eff_1 = self.subparams["cos"].subparams["angle"].recur_animate("region_angle")
+                amp, eff_2 = self.subparams["cos"].subparams["amp"].recur_animate("real")
+                self.expression_controllers.extend(eff_1)
+                self.expression_controllers.extend(eff_2)
+                
+                if self.dimension == 2:
+                    ret = "mul(Math.cos(degreesToRadians({angle})), [{amp}, {amp}])"
```
AnishGG: For simple circle, we need to give diameter  to the lottie file, but here we are giving radius as input.
This needs to be changed "only for support of simple circle".
Somehow it should be given {amp}*2 only for simple circle

saurabh-ramola: This might fail under some circumstances like when there is a sine function being called recursively inside another sine function.

AnishGG: Yup, you are right. Using {amp}*2 will fail when called in recursively. Hence I purpose that you modify the original animation(I mean modify the xml itselves, just like [here](https://github.com/synfig/synfig/blob/77164e8371269afb0490287f200f6c35fe067e88/synfig-studio/plugins/lottie-exporter/synfig/animation.py#L17)) and multiply the circle_radius by 2 at the root level

saurabh-ramola: Resolved now. :D

synfig-studio/plugins/lottie-exporter/common/Param.py
saurabh-ramola: Thanks for the clarification. This is resolved now!

synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -846,6 +896,19 @@ def add_convert_link(self, val):
         self.param.append(root)
         self.SUBPARAMS_EXTRACTED = 0
 
+    def scale_convert_link(self, val):
```
AnishGG: Nice work here!

saurabh-ramola:
@ice0 Does this need any changes ? ^_^

saurabh-ramola:
> Yes. Please rebase :)
@ice0 Done :smiley:

ice0:
Merged. Thank you!

PR #1259, opened 2020-03-19
Set LayerEncapsulate created group children_lock parameter to static …

### LINKED ISSUE #780: "Lock Selection" parameter of Group Layer should be static by default
See https://wiki.synfig.org/Static_Parameters

When Group is created, it has "Blend Method" and "Z Range" parameters marked as static by default -
screenshot_008

It is reasonable to have "Lock Selection" parameter static by default as well.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/71538320-lock-selection-parameter-of-group-layer-should-be-static-by-default?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

FirasH:
…by default.

Fixes #780

### DIFF
--- a/synfig-studio/src/synfigapp/actions/layerencapsulate.cpp
+++ b/synfig-studio/src/synfigapp/actions/layerencapsulate.cpp
@@ -150,7 +150,6 @@ Action::LayerEncapsulate::lowest_depth()const
 void
 Action::LayerEncapsulate::prepare()
 {
-
 	if(!first_time())
 		return;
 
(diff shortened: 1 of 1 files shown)


rodolforg:
Looks good to me.
Nice catch: `set_param()` replaces the whole `ValueBase`.
An alternative would be use `get_param()`, change the value itself and then call `set_param()`.

ice0:
Merged. Thank you!

PR #1270, opened 2020-03-23
DashItem Offset and Length variables are not consistent when editing them

### LINKED ISSUE #1265: DashItem Offset and Length variables are not consistent when editing them
**Issue description**:

0. Create a new document
1. Create an Advanced Outline: `Toolbox > Spline Tool > enable "Advanced Layer"`
2. Select it and in the `Parameters` Window open the sub-menu `Dash Item List > Dash Item 001`

Here you can see the list of parameters. If you select either `Offset` or `Length` the value is displayed without unit (by default 0.100000):
view

but if you try to edit it, it will display 6.00000px:
edit

I assume those two values should be consistent, because as it is now, you have to "guess" the value which corresponds to px and viceversa.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/90417922-dashitem-offset-and-length-variables-are-not-consistent-when-editing-them?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

FirasH:
Fixes #1265

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_composite.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_composite.cpp
@@ -763,12 +763,10 @@ ValueNode_Composite::get_children_vocab_vfunc()const
 		ret.push_back(ParamDesc(ValueBase(),"offset")
 			.set_local_name(_("Offset"))
 			.set_description(_("The offset length of the Dash Item over the Spline"))
-			.set_is_distance()
 		);
 		ret.push_back(ParamDesc(ValueBase(),"length")
 			.set_local_name(_("Length"))
 			.set_description(_("The length of the Dash Item"))
-			.set_is_distance()
 		);
 		ret.push_back(ParamDesc(ValueBase(),"side_before")
 			.set_local_name(_("Side Type Before"))


ice0:
@morevnaproject please review.

rodolforg:
Or maybe we should make value editor and viewer widgets consistent, instead of removing an parameter description info?

ice0:
> Or maybe we should make value editor and viewer widgets consistent, instead of removing an parameter description info?

I think this will be better solution.

FirasH:
@rodolforg for sure it's a more complete and deep solution.
Do you already have an idea of how to do that practically in the code?

rodolforg:
@FirasH first we need to know what value presentation should be used: relative (0.1, 10%) or absolute (pixel, for example)

`Widget_ValueBase` (in widget_value.cpp), `ValueBase_Entry` and `CellRenderer_ValueBase` (bothe in cellrenderer_value.cpp) are responsible to display a ValueBase value for viewing or editing.

Take a look first on how Gtk::TreeView works for edition first if you don't know Gtk approach:
https://developer.gnome.org/gtkmm-tutorial/stable/sec-treeview-examples.html.en#sec-editable-cells-example
https://en.wikibooks.org/wiki/GTK%2B_By_Example/Tree_View/Editable_Cells

morevnaproject:
> @FirasH first we need to know what value presentation should be used: relative (0.1, 10%) or absolute (pixel, for example)

That depends on the situation - for various parameters the answer is different. In some cases we need relative value. And in some cases the parameter should give value in pixels.

In this case (for DashItem Offset and Length) we need a relative parameter. So, I think this PR is ready to merge. ^__^

rodolforg:
@morevnaproject  Do you think we should provide another hint (`.set_relative()`) or just not using hint `.set_distance()` is always good?

morevnaproject:
@rodolforg  I think it's okay to go as it is - not use `.set_distance()` ^__^

PR #1273, opened 2020-03-23
[MSYS2] Enable building of images (Autotools)

morevnaproject:
This enables building image with Autotools on Windows.

I've been forced to disable Magick++, because it failed to build for me.
At the same time, ImageMagick is mandatory requirement to build images with Autotools (it is used to convert some icon files).


ice0:
Merged. Thank you!

Issue #1218, opened 2020-03-10
[Lottie exporter] Use "File" -> "Export for web" menu comand to call "Lottie Exporter"


morevnaproject:
Currently Lottie Exporter is implemented as a Synfig Plugin.
This puts some limitation at its functionality.

For example:
- It is not possible to set output filename
- It is not possible to set additional parameters (for example #1217)
- After calling plugin Synfig file is closed and re-opened every time. That takes time and also clears undo history.

Considering all above, I suggest to re-organize Lottie Exporter from plugin to native menu command.

I mean modify Synfig C++ code in the following way:
1.  Add "File" -> "Export for web" menu command.
2. This command will open a dialog similar to "Save As..." dialog.
3. When user choosing location and filename, he clicks "Export/Save" and C++ code is saving a temporary copy of current document and executes a python script which does actual conversion. At this point we can pass additional parameters to python script, which come from options of dialog at step 2.

morevnaproject:
@AnishGG Please take a look. ^__^

AnishGG:
Sure!

AnishGG:
@morevnaproject 
For this, I will have to create a new dialog(just like https://github.com/synfig/synfig/blob/master/synfig-studio/src/gui/dialogs/canvasoptions.cpp), right?
P.S. I am new to gtkmm, will need a little bit help here.

AnishGG:
I was trying to go through the code and implement this. I will need some clarifications. :P
Here is the branch I am committing the code to https://github.com/AnishGG/synfig/tree/export_for_web_ui
For now, I have just copied the `canvas_options` for most of the parts.
I am trying to call the `lottie exporter` plugin from within the `dialog_exportforweb.cpp`, but I am failing to do so here: https://github.com/AnishGG/synfig/blob/e0ea6945624b8bfca9cee2b7e49ff6420010d167/synfig-studio/src/gui/dialogs/dialog_exportforweb.cpp#L214

@ice0 @morevnaproject any suggestions from your side please. ^__^

ice0:
@AnishGG this is because you need to get an instance of the active `CanvasView`, and not just run the method from the `Instance` class. So in your example, it will be something like:
```
canvas_view_->get_instance()->run_plugin(this->plugin_path);
```

PR #1289, opened 2020-03-26
Export plugins

mbasaglia:
Adds more metadata to the plugins to support a proper export system.

I've also updated the lottie exporter to use this system.

This is said plugin description as an example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<plugin type="exporter">
   <_name>Export to Lottie format</_name>
   <exec>lottie-exporter.py</exec>
   <extension>.json</extension>
   <description>Lottie Animation (*.json)</description>
</plugin> 
```

if the `type` attribute is present and its value is "exporter", the plugin will be added to a list of exporters. The `<extension>` and `<description>` elements are used by the new "Export" dialog to set up the file filters.

Such plugins no longer show in the Plugins menu, they are only accessible through File > Export...

### DIFF
--- a/synfig-studio/plugins/lottie-exporter/common/Param.py
+++ b/synfig-studio/plugins/lottie-exporter/common/Param.py
@@ -433,6 +433,7 @@ class Param:
                 return ret, self.expression_controllers
 
             elif self.param[0].tag == "bone_link":
+                import pdb; pdb.set_trace(); pass
                 self.subparams["bone_link"].extract_subparams()
                 guid = self.subparams["bone_link"].subparams["bone"][0].attrib["guid"]
                 bone = self.get_bone_from_canvas(guid)
@@ -800,6 +801,7 @@ class Param:
                     ret = link_on * switch + link_off * (1 - switch)
 
             elif self.param[0].tag == "bone_link":
+                import pdb; pdb.set_trace(); pass
                 guid = self.subparams["bone_link"].subparams["bone"][0].attrib["guid"]
                 bone = self.get_bone_from_canvas(guid)
                 ret_origin, ret_angle, lls, rls = bone.__get_value(frame)
@@ -986,6 +988,7 @@ class Param:
                 self.subparams["scale"].subparams["scalar"].update_frame_window(window)
 
             elif node.tag == "bone_link":
+                import pdb; pdb.set_trace(); pass
                 self.subparams["bone_link"].extract_subparams()
                 guid = self.subparams["bone_link"].subparams["bone"][0].attrib["guid"]
                 bone = self.get_bone_from_canvas(guid)
--- a/synfig-studio/plugins/lottie-exporter/common/Param.py
+++ b/synfig-studio/plugins/lottie-exporter/common/Param.py
@@ -433,7 +433,6 @@ class Param:
                 return ret, self.expression_controllers
 
             elif self.param[0].tag == "bone_link":
-                import pdb; pdb.set_trace(); pass
                 self.subparams["bone_link"].extract_subparams()
                 guid = self.subparams["bone_link"].subparams["bone"][0].attrib["guid"]
                 bone = self.get_bone_from_canvas(guid)
@@ -801,7 +800,6 @@ class Param:
                     ret = link_on * switch + link_off * (1 - switch)
 
             elif self.param[0].tag == "bone_link":
-                import pdb; pdb.set_trace(); pass
                 guid = self.subparams["bone_link"].subparams["bone"][0].attrib["guid"]
                 bone = self.get_bone_from_canvas(guid)
                 ret_origin, ret_angle, lls, rls = bone.__get_value(frame)
@@ -988,7 +986,6 @@ class Param:
                 self.subparams["scale"].subparams["scalar"].update_frame_window(window)
 
             elif node.tag == "bone_link":
-                import pdb; pdb.set_trace(); pass
                 self.subparams["bone_link"].extract_subparams()
                 guid = self.subparams["bone_link"].subparams["bone"][0].attrib["guid"]
                 bone = self.get_bone_from_canvas(guid)
--- a/synfig-studio/src/gui/instance.cpp
+++ b/synfig-studio/src/gui/instance.cpp
@@ -223,25 +223,128 @@ Instance::set_redo_status(bool x)
 	signal_undo_redo_status_changed()();
 }
 
+bool
+studio::Instance::run_plugin_with_arguments(std::string plugin_path, const std::vector<std::string>& args)
+{
+	OneMoment one_moment;
+
+	bool result = false;
+	String output;
+	String command = "";
+
+	// Path to python binary can be overridden
+	// with SYNFIG_PYTHON_BINARY env variable:
+	const char* custom_python_binary=getenv("SYNFIG_PYTHON_BINARY");
+	if(custom_python_binary) {
+		command=custom_python_binary;
+		if (!App::check_python_version(command)) {
+			output=_("Error: You need to have Python 3 installed.");
+			command="";
+		}
+	} else {
+	// Set path to python binary depending on the os type.
+	// For Windows case Python binary is expected
+	// at INSTALL_PREFIX/python/python.exe
+		std::list< String > binary_choices;
+		binary_choices.push_back("python");
+		binary_choices.push_back("python3");
+		std::list< String >::iterator iter;
+		for(iter=binary_choices.begin();iter!=binary_choices.end();iter++)
+		{
+			String python_path;
+#ifdef _WIN32
+			python_path = "\"" + App::get_base_path()+ETL_DIRECTORY_SEPARATOR+"python"+ETL_DIRECTORY_SEPARATOR+*iter+".exe" + "\"";
+#else
+			python_path = *iter;
+#endif
+			if (App::check_python_version(python_path))
+			{
+				command = python_path;
+				break;
+			}
+
+		}
+	}
+
+	if (command.empty())
+	{
+		output=_("Error: No Python 3 binary found.\n\nHint: You can set SYNFIG_PYTHON_BINARY environment variable pointing at your custom python installation.");
+	} else {
+		synfig::info("Python 3 binary found: "+command);
+
+
+		// Construct the full command:
+		command = command+" \""+plugin_path+"\"";
+        for ( const auto& arg : args )
+        {
+            command += " \"" + arg + "\"";
+        }
+        command += " 2>&1";
+#ifdef _WIN32
+		// This covers the dumb cmd.exe behavior.
+		// See: http://eli.thegreenplace.net/2011/01/28/on-spaces-in-the-paths-of-programs-and-files-on-windows/
+		command = "\"" + command + "\"";
+#endif
+
+		FILE* pipe = popen(command.c_str(), "r");
+		if (!pipe) {
+			output = "ERROR: pipe failed!";
+		} else {
+			char buffer[128];
+			while(!feof(pipe)) {
+				if(fgets(buffer, 128, pipe) != NULL)
+						output += buffer;
+			}
+
+			if (output != "" ){
+				synfig::info(output);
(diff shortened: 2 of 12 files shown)


AnishGG:
@mbasaglia seems like great work! Did you test this locally? The travis build is failing with some errors.
Also as @morevnaproject proposed, we need a behaviour like this: https://github.com/synfig/synfig/issues/1218 where a file's undo history should not be erased. ^_^

AnishGG:
I tried building your PR locally and it worked. Don't know why is it failing on travis.
@ice0 can you have a look here. ^__^

synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -433,6 +433,7 @@ def recur_animate(self, anim_type):
                 return ret, self.expression_controllers
 
             elif self.param[0].tag == "bone_link":
+                import pdb; pdb.set_trace(); pass
```
AnishGG: why do we need debugger here?

synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -753,6 +754,7 @@ def __get_value(self, frame):
                     ret = link_on * switch + link_off * (1 - switch)
 
             elif self.param[0].tag == "bone_link":
+                import pdb; pdb.set_trace(); pass
```
AnishGG: same here

synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -893,6 +895,7 @@ def update_frame_window(self, window):
                 self.subparams["scale"].subparams["scalar"].update_frame_window(window)
 
             elif node.tag == "bone_link":
+                import pdb; pdb.set_trace(); pass
```
AnishGG: here as well

mbasaglia:
has sorry about pdb, I was using that when I was making my telegram exporter and forgot about it -_-

mbasaglia:
> Also as @morevnaproject proposed, we need a behaviour like this: #1218 where a file's undo history should not be erased. ^_^

It should already work on this branch, I've modified run_plugin with an argument that disables reloading the canvas.

I've also extracted the code that actually runs the plugin into a different function, which doesn't save a temp file, so it might be used in the future for things import plugins (that wouldn't require a canvas)

mbasaglia:
By the way, I was also considering adding support for a single plugin directory to define multiple plugins.

In the case of export, it might be helpful to define multiple exporters based on a single codebase.

Something like if the root element of the plugin xml file is `<plugins>`, it can define multiple `<plugin>` elements inside it.

mbasaglia:
Looks like the build errors on travis are about `Glib::RefPtr::get`, probably different Glib versions than what I have

AnishGG:
@ice0 could you review this please. ^__^

AnishGG:
Also @mbasaglia, as mentioned in #1218, could you change the menu item name from `export` to `export for web`.

mbasaglia:
> Also @mbasaglia, as mentioned in #1218, could you change the menu item name from `export` to `export for web`.

This is more general than just Lottie though, it can be used for any exporter. Main reason I made this is that I want to make my own exporters.

I already have a Telegram sticker exporter, and I might also add a SVG exporter.

mbasaglia:
I've added an extra commit just to make the XML parsing a bit simpler / more readable. I've also changed the `type` from an element into an attribute since that seems more appropriate.

synfig-studio/src/gui/instance.cpp
```diff
@@ -223,25 +223,128 @@ Instance::set_redo_status(bool x)
...
+	String output;
+	String command = "";
+
+	// Path to python binary can be overridden
+	// with SYNFIG_PYTHON_BINARY env variable:
+	const char* custom_python_binary=getenv("SYNFIG_PYTHON_BINARY");
+	if(custom_python_binary) {
+		command=custom_python_binary;
+		if (!App::check_python_version(command)) {
+			output=_("Error: You need to have Python 3 installed.");
```
rodolforg: This message is overriding by line 271 below. Besides, it doesn't make clear the searched python binary was from the env var.

synfig-studio/src/gui/instance.cpp
```diff
@@ -223,25 +223,128 @@ Instance::set_redo_status(bool x)
...
 }
 
+bool
+studio::Instance::run_plugin_with_arguments(std::string plugin_path, const std::vector<std::string>& args)
+{
+	OneMoment one_moment;
+
+	bool result = false;
+	String output;
+	String command = "";
```
rodolforg: No need of this empty assignment :)

synfig-studio/src/gui/instance.cpp
```diff
@@ -223,25 +223,128 @@ Instance::set_redo_status(bool x)
...
+	// at INSTALL_PREFIX/python/python.exe
+		std::list< String > binary_choices;
+		binary_choices.push_back("python");
+		binary_choices.push_back("python3");
+		std::list< String >::iterator iter;
+		for(iter=binary_choices.begin();iter!=binary_choices.end();iter++)
+		{
+			String python_path;
+#ifdef _WIN32
+			python_path = "\"" + App::get_base_path()+ETL_DIRECTORY_SEPARATOR+"python"+ETL_DIRECTORY_SEPARATOR+*iter+".exe" + "\"";
```
rodolforg: I know there's a lot of places that uses `ETL_DIRECTORY_SEPARATOR`, but I do believe it is better for readability to just use "/" instead: both Win, Mac & Linux accepts it as separator.

synfig-studio/src/gui/instance.cpp
```diff
@@ -223,25 +223,128 @@ Instance::set_redo_status(bool x)
...
+        command += " 2>&1";
+#ifdef _WIN32
+		// This covers the dumb cmd.exe behavior.
+		// See: http://eli.thegreenplace.net/2011/01/28/on-spaces-in-the-paths-of-programs-and-files-on-windows/
+		command = "\"" + command + "\"";
+#endif
+
+		FILE* pipe = popen(command.c_str(), "r");
+		if (!pipe) {
+			output = "ERROR: pipe failed!";
```
rodolforg: Hmm should it be translatable?

synfig-studio/src/gui/instance.cpp
```diff
@@ -223,25 +223,128 @@ Instance::set_redo_status(bool x)
...
+		synfig::info("Python 3 binary found: "+command);
+
+
+		// Construct the full command:
+		command = command+" \""+plugin_path+"\"";
+        for ( const auto& arg : args )
+        {
+            command += " \"" + arg + "\"";
+        }
+        command += " 2>&1";
```
rodolforg: Identation?

synfig-studio/src/gui/instance.cpp
```diff
@@ -223,25 +223,128 @@ Instance::set_redo_status(bool x)
...
+#endif
+
+		FILE* pipe = popen(command.c_str(), "r");
+		if (!pipe) {
+			output = "ERROR: pipe failed!";
+		} else {
+			char buffer[128];
+			while(!feof(pipe)) {
+				if(fgets(buffer, 128, pipe) != NULL)
+						output += buffer;
```
rodolforg: Identation

synfig-studio/src/gui/instance.cpp
```diff
@@ -223,25 +223,128 @@ Instance::set_redo_status(bool x)
 	signal_undo_redo_status_changed()();
 }
 
+bool
+studio::Instance::run_plugin_with_arguments(std::string plugin_path, const std::vector<std::string>& args)
```
rodolforg: I'd split up this function in two:
* One detects a valid python binary filepath - and maybe build the full command line
* Other runs the script (open pipe and everything)

synfig-studio/src/gui/instance.cpp
```diff
@@ -223,25 +223,128 @@ Instance::set_redo_status(bool x)
...
 
-	String message = strprintf(_("Do you really want to run plugin for file \"%s\"?" ),
-				this->get_canvas()->get_name().c_str());
+    if ( modify_canvas )
+    {
+        String message = strprintf(_("Do you really want to run plugin for file \"%s\"?" ),
+                    this->get_canvas()->get_name().c_str());
 
-	String details = strprintf(_("This operation cannot be undone and all undo history will be cleared."));
+        String details = strprintf(_("This operation cannot be undone and all undo history will be cleared."));
```
rodolforg: You don't need to use `strprintf` if there isn't formatting placeholders.

synfig-studio/src/gui/instance.cpp
```diff
@@ -296,93 +400,11 @@ studio::Instance::run_plugin(std::string plugin_path)
...
-			App::dialog_message_1b(
-					"Error",
-					output,
-					"details",
-					_("Close"));
-
-			one_moment.show();
-
-		} else {
+		if (result && modify_canvas){
```
rodolforg: identation?

synfig-studio/src/gui/pluginmanager.cpp
```diff
@@ -137,6 +137,8 @@ PluginManager::load_plugin( const std::string &path )
 	PluginManager::plugin p;
 	std::string plugindir = dirname(path);
 	p.id=plugindir;
+
+    std::list<plugin>* target_list = &list_;
```
rodolforg: identation

synfig-studio/src/gui/instance.cpp
mbasaglia: I wanted to keep this merge request small so I haven't refactored much of the plugin code.

But I think it would also be a good idea to move most of the code in run_plugin into the plugin manager (or maybe even have plugin struct become a proper class and have it there).

`run_plugin_with_arguments` is really just extracting part of the code that used to be in `run_plugin`.

synfig-studio/src/gui/instance.cpp
mbasaglia: It's code that was already there, this kind of changes are beyond the scope of this merge request IMO.

rodolforg: Ok.

morevnaproject:
@mbasaglia Thank you for the great addition! I like how it work. ^__^

Just one request:
Currently the Export dialog allows to select file type (just one option at the moment - "Lottie JSON"). But when export is done it creates two files - "json" and "html". It would be nice to have two filetype options - "Lottie JSON" and "Lottie HTML". 

But that request can be a separate issue, I think this  PR is ready to merge. ^__^

mbasaglia:
> Just one request:
> Currently the Export dialog allows to select file type (just one option at the moment - "Lottie JSON"). But when export is done it creates two files - "json" and "html". It would be nice to have two filetype options - "Lottie JSON" and "Lottie HTML".

Yeah I'd like to add the ability to do something like that, maybe having some way of defining multiple plugins in the same directory.

It could be implemented by having multiple exporters in a single plugin but I think allowing multiple plugins in the same directory (sharing the same codebase) is more future-proof.

As I was also considering adding support for "import" plugins, so you could have a single installed plugin, that defines various functionality like both import and export in some format or the like.

The way inkscape handles this is by searching for files with a specific extension in the plugin directory (instead of just plugin.xml like synfig does). An alternative could be to add support different root element in plugin.xml that defines multiple `<plugin>` elements inside it.

I can work on this but I think the decision on how this is handled should be made by someone with more experience in this project.

morevnaproject:
@mbasaglia Can you please rebase your branch on top of the latest master, so I can make a clean merge? ^__^

mbasaglia:
@morevnaproject pushed rebased commits

morevnaproject:
@mbasaglia I have created new issue to continue discussion - https://github.com/synfig/synfig/issues/1309

rodolforg:
@ice0 @morevnaproject  maybe backport this and #1313 to v1.4.2?

PR #1293, opened 2020-03-27
fix memory leaks for blur

rodolforg:
Coverity Report ID: 308890, 308875, 308868, 308866, 308837, 308830, 308823,  308805


ice0:
Merged. Thank you!

Issue #1309, opened 2020-03-31
[Lottie] Allow to choose export format


morevnaproject:
As of #1289 export to Lottie is called via "File" -> "Export" format.

Currently the Export dialog allows to select file type (just one option at the moment - "Lottie JSON"). But when export is done it creates two files - "json" and "html". It would be nice to have two filetype options - "Lottie JSON" and "Lottie HTML".

![screenshot_003](https://user-images.githubusercontent.com/332868/78042869-d5f64380-739c-11ea-8d73-905534d93f89.png)

mbasaglia:
Link to my previous comment: https://github.com/synfig/synfig/pull/1289#issuecomment-606635662

morevnaproject:
In response to https://github.com/synfig/synfig/pull/1289#issuecomment-606635662:

> It could be implemented by having multiple exporters in a single plugin but I think allowing multiple plugins in the same directory (sharing the same codebase) is more future-proof.

I think import/export plugins should be separated from other plugins, listed in "Plugins" menu. I feel they are completely different "beast". ^__^

mbasaglia:
> I think import/export plugins should be separated from other plugins, listed in "Plugins" menu. I feel they are completely different "beast". ^__^

Yeah after my export implementation they are stored in a separate list in the plugin manager.

I was thinking about reworking the way plugins are handled, having a proper Plugin class instead of a struct with direct member access.
Having accessor functions to plugin data could also allow translated strings to work better as they'd return the value based on the current locale (currently the plugin name is localized based when the plugin is loaded).

This would open up the possibility of having different kinds of plugin, maybe even using different interpreters instead of just python?

AnishGG:
> But when export is done it creates two files - "json" and "html".

Actually there is one more file `.log`, that is also created. It stores the logs, if some kind of failure occurs.



> This would open up the possibility of having different kinds of plugin, maybe even using different interpreters instead of just python?

This would be amazing!

mbasaglia:
The need for logs could also be removed, maybe synfig can handle stdout/stderr from plugins a bit differently than what it currently does

mbasaglia:
Here's my proposal for improvements to the Plugin system:

## Import/Export plugin description

```xml
<plugin>
    <!-- translatable, not strictly needed for exporters
        but if in the future there is going to be
        something like a dialog to enable/disable plugins
        or whatever it might be worth keeping -->
    <name>Plugin name</name>

[...]

Issue #1252, opened 2020-03-17
Items in Plugins menu ignore language set in preference


morevnaproject:
**Synfig version & platform**:
1.3.13, Linux

**Issue description**:
When I set application language to English (via "Edit" -> "Preferences"), the language of items in "Plugins" menu remains in system locale (Russian in my case).

![screenshot_005](https://user-images.githubusercontent.com/332868/76856067-14f8a500-6885-11ea-9410-4cdce20a9de5.png)

How to reproduce:
1. Ensure that your system locale is any other than English.
2. Start Synfig
3. Select "Edit" -> "Preferences" from menu
4. Choose any language, different from your system locale 
![screenshot_006](https://user-images.githubusercontent.com/332868/76856140-38235480-6885-11ea-88e1-5f9492fe19b9.png)
5. Click OK to apply new settings.
6. Restart Synfig.
7. Click "Plugins" menu and observe items in your system language.

man-of-code:
Hey, @morevnaproject Can I work on this??

ice0:
@man-of-code Sure, assigning you to this.

man-of-code:
@morevnaproject @ice0 Thank you!!
Any idea where to start ??

FirasH:
I think this has been fixed in 51ba106

rodolforg:
I confirm it is solved on master e64c7d6784fa10ab8b1753c0c03ab120269fa1c0

PR #1313, opened 2020-04-01
Rework the plugin system

mbasaglia:
Implements most of the things mentioned here: https://github.com/synfig/synfig/issues/1309#issuecomment-606829467

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -3301,13 +3301,26 @@ App::dialog_import_file(std::string &filename, const std::string& preference)
 
 	Gtk::FileChooserDialog *dialog = new Gtk::FileChooserDialog(*App::main_window, _("Please select a file"), Gtk::FILE_CHOOSER_ACTION_OPEN);
 
+	bool found = false;
 	for ( const ImportExport& exp : App::plugin_manager.importers() )
 	{
+		found = true;
 		Glib::RefPtr<Gtk::FileFilter> filter = Gtk::FileFilter::create();
 		filter->set_name(exp.description.get());
 		for ( const std::string& extension : exp.extensions )
 			filter->add_pattern("*" + extension);
 		dialog->add_filter(filter);
+	}
+
+	if ( !found )
+	{
+		dialog_message_1b(
+			"Error",
+			_("There are no file importers available"),
+			"details",
+			_("Close")
+		);
+		return {};
 	}
 
 	dialog->set_current_folder(prev_path);
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -2988,27 +2988,39 @@ App::dialog_open_file_with_history_button(const std::string &title, std::string 
 
 	// File filters
 	// Synfig Documents
-	Glib::RefPtr<Gtk::FileFilter> filter_supported = Gtk::FileFilter::create();
-	filter_supported->set_name(_("Synfig files (*.sif, *.sifz, *.sfg)"));
-	filter_supported->add_mime_type("application/x-sif");
-	filter_supported->add_pattern("*.sif");
-	filter_supported->add_pattern("*.sifz");
-	filter_supported->add_pattern("*.sfg");
-	dialog->add_filter(filter_supported);
+	Glib::RefPtr<Gtk::FileFilter> filter_builtin = Gtk::FileFilter::create();
+	filter_builtin->set_name(_("Synfig files (*.sif, *.sifz, *.sfg)"));
+	filter_builtin->add_mime_type("application/x-sif");
+	filter_builtin->add_pattern("*.sif");
+	filter_builtin->add_pattern("*.sifz");
+	filter_builtin->add_pattern("*.sfg");
+	dialog->add_filter(filter_builtin);
 
 	// Any files
 	Glib::RefPtr<Gtk::FileFilter> filter_any = Gtk::FileFilter::create();
 	filter_any->set_name(_("Any files"));
 	filter_any->add_pattern("*");
 	dialog->add_filter(filter_any);
-	dialog->set_filter(filter_any);
+
+	// Supported files
+	Glib::RefPtr<Gtk::FileFilter> filter_supported = Gtk::FileFilter::create();
+	filter_supported->set_name(_("All supported files"));
+	filter_supported->add_pattern("*.sif");
+	filter_supported->add_pattern("*.sifz");
+	filter_supported->add_pattern("*.sfg");
+	dialog->add_filter(filter_supported);
+	dialog->set_filter(filter_supported);
+
 
 	for ( const ImportExport& exp : plugin_manager.importers() )
 	{
 		Glib::RefPtr<Gtk::FileFilter> filter = Gtk::FileFilter::create();
 		filter->set_name(exp.description.get());
 		for ( const std::string& extension : exp.extensions )
+		{
 			filter->add_pattern("*" + extension);
+			filter_supported->add_pattern("*" + extension);
+		}
 		dialog->add_filter(filter);
 	}
 
@@ -3030,7 +3042,7 @@ App::dialog_open_file_with_history_button(const std::string &title, std::string 
 		auto filter = dialog->get_filter();
 
 		plugin_importer = "";
-		if ( filter == filter_any )
+		if ( filter == filter_any || filter == filter_supported )
 		{
 			if ( !dialog_select_importer(filename, plugin_importer) )
 				return false;
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -2906,6 +2906,64 @@ on_open_dialog_with_history_selection_changed(Gtk::FileChooserDialog *dialog, Gt
 	history_button->set_sensitive(!dialog->get_filename().empty());
 }
 
+/*
+
+Finds which importer to use for the given filename
+
+Returns false if the user has canceled the import.
+
+plugin is set to the script identifier for the importer,
+or an empty string if the file should be opened as a normal sif file
+
+*/
+bool
+App::dialog_select_importer(const std::string& filename, std::string& plugin)
+{
+	synfig::String ext = filename_extension(filename);
+
+	Gtk::Dialog dialog(_("Select importer"), true);
+	dialog.add_button(_("Cancel"), Gtk::RESPONSE_REJECT)->set_image_from_icon_name("gtk-cancel", Gtk::ICON_SIZE_BUTTON);
+	dialog.add_button(_("OK"), Gtk::RESPONSE_ACCEPT)->set_image_from_icon_name("gtk-ok", Gtk::ICON_SIZE_BUTTON);
+
+	Gtk::Label label(_("Please select the importer to use for this file"));
+	dialog.get_content_area()->pack_start(label);
+
+	Gtk::ComboBoxText combo_importers;
+	dialog.get_content_area()->pack_end(combo_importers);
+
+	int count = 0;
+	plugin = "";
+	for ( const auto& importer : App::plugin_manager.importers() )
+	{
+		if ( importer.has_extension(ext) )
+		{
+			plugin = importer.id;
+			combo_importers.append(importer.id, importer.description.get());
+			if ( count == 0 )
+				combo_importers.set_active_id(importer.id);
+			count++;
+		}
+	}
+
+	if ( count == 1 )
+		return true;
+
+	if ( count == 0 )
+		return true;
+
+	dialog.show_all();
+
+	if ( dialog.run() != Gtk::RESPONSE_ACCEPT )
+	{
+		plugin = "";
+		return false;
+	}
+
+	plugin = combo_importers.get_active_id();
+	return true;
+
+}
+
 bool
 App::dialog_open_file_with_history_button(const std::string &title, std::string &filename, bool &show_history, std::string preference, std::string& plugin_importer)
 {
@@ -2972,17 +3030,10 @@ App::dialog_open_file_with_history_button(const std::string &title, std::string 
 		auto filter = dialog->get_filter();
 
 		plugin_importer = "";
-		synfig::String ext = filename_extension(filename);
 		if ( filter == filter_any )
 		{
-			for ( const auto& importer : App::plugin_manager.importers() )
-			{
-				if ( importer.has_extension(ext) )
-				{
-					plugin_importer = importer.id;
-					break;
-				}
-			}
+			if ( !dialog_select_importer(filename, plugin_importer) )
+				return false;
 		}
 		else
 		{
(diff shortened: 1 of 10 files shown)


ice0:
It's ok, what travis build is errored, i will fix it ASAP.

ice0:
Ok, Travis build fixed. After rebase all should be fine.

mbasaglia:
I've added support for importers defined in plugins.

To test this you can use my plugin: http://mattia.basaglia.gitlab.io/tgs/downloads.html
It supports a couple different formats, probably the easiest to test is SVG.

morevnaproject:
@mbasaglia Thank you for an awesome addition! Please check my comments here - https://github.com/synfig/synfig/issues/1309#issuecomment-607334235 ^__^

morevnaproject:
I have tested this PR, here are my thoughts:

1. We now have two "Import" menu commands:
screenshot_001

I was unable to test properly "Import file..." command, because importer gave me an error when I made an attempt to import Lottie json file. But that's different issue.

I guess that "Import file..." opens imported file in a new tab (not adds to currently opened document). So, I think the meaning of it is more like "File" - "Open" command.

I think that we should merge "Import file..." and "Open" commands for better UX.

2. Export command works perfectly, I have tested with html and json export. ^__^

mbasaglia:
yeah it does work more like open, I'll move the code to that dialog. 

to test import the easiest way is to create a simple drawing in synfig, export with "[TGS] SVG" and import the same file.

morevnaproject:
Thank you! I want to say that I am absolutely excited about this feature and which possibilities it  opens for further development.

For example, a very common case when newbie user tries to import some image using "File" - "Open" command. Synfig give them error, because images intended to be added via "File" - "import" command. But newbie users are not aware about that and they often reach out to support asking what's wrong.

With new plugins system we can create a very simple plugin, which allows to open images via "File" - "Open" command. The plugin will do simple thing - it will create a new document with imported image. ^__^

mbasaglia:
I've pushed some changed that do that and also fixed some issues.

mbasaglia:
Yea

mbasaglia:
CI fail seems unrelated to my changes

```
$ ./1-setup-osx-brew.sh

Checking adwaita-icon-theme...

==> Downloading https://homebrew.bintray.com/bottles/gnome-icon-theme-3.22.0.yos

Error: Failed to download resource "gnome-icon-theme"

Download failed: https://homebrew.bintray.com/bottles/gnome-icon-theme-3.22.0.yosemite.bottle.tar.gz

[...]

==> Downloading https://download.gnome.org/sources/adwaita-icon-theme/3.22/adwai

Error: Failed to download resource "gnome-icon-theme"

Download failed: https://download.gnome.org/sources/adwaita-icon-theme/3.22/adwaita-icon-theme-3.22.0.tar.xz

The command "./1-setup-osx-brew.sh" failed and exited with 1 during .
```

AnishGG:
I have restarted the job. Sometimes the download fails because of network errors at the server.

morevnaproject:
> I've pushed some changes that do that and also fixed some issues.

Thank you! I will be able to test it on Tuesday. ^__^

mbasaglia:
By the way, as a side effect of this PR, #1252 should be much easier to fix as all that is required is to recreate / update the menu items when language is changed.
(Previously plugin names were loaded at startup and not using the correct language).

morevnaproject:
Hi! I did one more round of testing and found the following:

The Lottie plugin (which is only capable for export currently) for some reason is displayed in "File" - "Open" dialog. The plugin doesn't has any importers declared, so I guess it shouldn't be there. ^__^ 

(I do not have any other export/import plugins installed, besides Lottie, which is shipped with this version of Synfig).

screenshot_001

morevnaproject:
Also, I suggest to make "Open history" button inactive when any filetype handled by plugin is selected.
screenshot_003

mbasaglia:
> Also, I suggest to make "Open history" button inactive when any filetype handled by plugin is selected.
> screenshot_003

I'm not too familiar with Gtk(mm) and I wasn't able to find a signal for when the filter changes (doesn't help that the documentation isn't great) do you know how that's done?

https://developer.gnome.org/gtkmm/stable/classGtk_1_1FileChooserDialog.html

rodolforg:
As far as I can tell, you can't access the filter combobox (and its signals) in a normal way, only what filter is currently active.

So I think "Open history" should behave differently according to the selected filter – if it's possible.

Regarding the Gtkmm docs, I use it, but often I read the Gtk docs themselves, as they're more ordered and cleaner than `mm` counterpart.

mbasaglia:
> As far as I can tell, you can't access the filter combobox (and its signals) in a normal way, only what filter is currently active.
> 
> So I think "Open history" should behave differently according to the selected filter – if it's possible.

Currently for plugins "open history" just acts like open. 

So I guess it isn't possible to disable the open history button, even if it's possible to write the code that enables/disables based on the current filter, it can't be run when the filter changes if we can't hook to the combo signal.

rodolforg:
Couldn't the "Open history" action callback check what is the selected filter instead of  (sadly) disabling the button?

mbasaglia:
I'm not sure I follow, what should happen if someone clicks it when a plugin importer is selected?

mbasaglia:
This is the relevant code btw, clicking on "Open history" just sets an output variable when clicked from the dialog

https://github.com/synfig/synfig/blob/742010d34f33049693e3496dd76d7d7460cd8053/synfig-studio/src/gui/app.cpp#L2970

And this is where it's used, looks like it only does something for ".sfg" files? So clicking on .sif, .sifz doesn't do anything special either (from what I gather)

https://github.com/synfig/synfig/blob/742010d34f33049693e3496dd76d7d7460cd8053/synfig-studio/src/gui/app.cpp#L4196-L4238

morevnaproject:
> And this is where it's used, looks like it only does something for ".sfg" files?

Yes, the "Open history" works for ".sfg" format only.

I am not insist on reworking behavior of "Open history" button right now, in this exact PR. Just brought that up that maybe this could be fixed in the future if not now. ^__^

I am testing the PR again now. Thanks to everyone! ^__^

morevnaproject:
I found the crash is happening now with current code.
To reproduce it I have installed TGS plugin (as it is providing importers, we need them to trigger the crash).

Also, you will need 2 Lottie JSON (or any other importable) files to reproduce the crash.
I have used those ones - 
[lottie-2.zip](https://github.com/synfig/synfig/files/4464135/lottie-2.zip)

1. In Synfig choose "File" - "Open".
2. Navigate to directory where you have saved your Lottie JSON files.
3. From filetype dropdown choose "[TGS} Lottie JSON" (otherwise you will not see json files in open dialog)
screenshot_006
4. Choose first file "ball-1.json" and click open.
5. File will be opened.
6. After that choose "File" - "Open" again from menu
7. Notice that filetype dropdown has empty value now. I guess this is the reason of the bug.
screenshot_007
8. Select "ball-1.json" and click Open button.
9. Synfig crashes.

morevnaproject:
Video demonstration for issue -  https://youtu.be/7Po9pR-fg5M

morevnaproject:
Issue 2:

1. Launch Synfig, choose "File" - "Open".
2. From filetype dropdown choose "Any files"
screenshot_010
3. Choose file "ball-1.json" and click Open.
4. Error appears (because Synfig tries to read file as SIF).
screenshot_008

morevnaproject:
Finally, suggestion:
It would be nice to have "All supported formats" option in file type drowpdown box and have it selected by default in "Open FIle" dialog. Example how it looks in Krita:
screenshot_011

Then user will immediately see all files that can be opened by Synfig and that's a big advantage. ^__^

mbasaglia:
One thing to consider regarding selecting the plugin with "All supported formats" is what to do if two different plugins handle the same extension.

The easy solution would be to select the first one, but this choice might not be intuitive.

Maybe show a second dialog with a combo box if there are multiple matches?

mbasaglia:
I've set the dialog to use "Any files" by default, this also fixes the blank entry.
I've also added support to load plugins from "Any files" (just selects the first one for now).

One thing about the "open history" button, maybe it can be removed and we can add another entry in the filter combo? It might be more intuitive since it's an operation specific to a single file type.

morevnaproject:
> Maybe show a second dialog with a combo box if there are multiple matches?

I like this idea!

mbasaglia:
I've implemented the dialog to disambiguate plugins.

With this feature I've also added support for open recent (so imported files are listed there, and if there's ambiguity the dialog is shown)

morevnaproject:
Awesome!
screenshot_003

morevnaproject:
> I've set the dialog to use "Any files" by default, this also fixes the blank entry.

Can we have "All supported types" instead "Any files" by default?
I mean that with "Any files" we have many files visible, which cannot be opened. This can be confusing. ^__^"

Also, I found one more issue:
1. Launch Synfig and choose "File" - "Open".
2. Select any SVG file (i.e. tiger. Click Open.
3. File is loaded.
4. Make some edits.
5. Choose "File" - "Save". File marked as saved. Note that the name of saved file is "tiger.svg".
screenshot_005
6. Close file.
7. Choose "File" - "Open". And open same file again.
8. File opened without changes. That means no changes were saved at step 5.

I think the solution would be to ask to save into Synfig format at step 5 - display "Save As..." dialog (just like GIMP does).

mbasaglia:
I've added the extra file filter and made synfig handle files opened with a plugin in a similar way to how it handles new files. 

So for these files it will still show tiger.svg after opening, but when saving it recognizes it doesn't have a proper file name set and will prompt the file dialog (after which the name on the tab will be updated with the name of the saved file)

morevnaproject:
> So for these files it will still show tiger.svg after opening, but when saving it recognizes it doesn't have a proper file name set

Awesome, this is exactly what's needed. Testing now.
Thank you!

morevnaproject:
Testing complete. I think it is ready to merge now.

I haven't looked at the code, though. @ice0 or @rodolforg can you please take a look before merge? ^__^

synfig-studio/src/gui/app.cpp
```diff
@@ -1210,15 +1210,11 @@ DEFINE_ACTION("keyframe-properties", _("Properties"));
 "	<menu action='menu-plugins'>"
 ;
 
-	list<PluginManager::plugin> plugin_list = studio::App::plugin_manager.get_list();
-	for(list<PluginManager::plugin>::const_iterator p=plugin_list.begin();p!=plugin_list.end();++p) {
-
+	for ( const auto& p : studio::App::plugin_manager.plugins() ) {
```
rodolforg, line 1213: Just a cosmetic change for readability: keep "plugin" name instead of "p"

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1517,14 +1517,13 @@ CanvasView::init_menus()
 		sigc::mem_fun0(canvas_properties,&CanvasProperties::present)
 	);
 
-	std::list<PluginManager::plugin> plugin_list = App::plugin_manager.get_list();
     auto instance = get_instance().get();
-	for(std::list<PluginManager::plugin>::const_iterator p = plugin_list.begin(); p != plugin_list.end(); ++p)
+	for ( const auto& p : App::plugin_manager.plugins() )
```
rodolforg, line 1521: again, pls better naming :) "plugin" even for this short code slice

synfig-studio/src/gui/instance.cpp
```diff
@@ -359,10 +260,13 @@ studio::Instance::run_plugin(std::string plugin_path, bool modify_canvas, std::v
 	// Generate temporary file name
 	String filename_original = get_canvas()->get_file_name();
 	String filename_processed;
+	String filename_prefix;
+	if ( !is_absolute_path(filename_original) )
+		filename_prefix = temporary_filesystem->get_temporary_directory() + ETL_DIRECTORY_SEPARATOR;
```
rodolforg, line 265: Someone can check if this temporary directory works fine on Windows?
If it points to a regular (=not admin) user has write access on the returned folder?
It could not be inside the folder Synfig is installed, but a system-defined one.
@mbasaglia @ice0

mbasaglia: on my system it doesn't even use /tmp, it uses ~/.sinfig/tmp

rodolforg: My worry is on MS Windows :(

I had a bad experience on using it installed on a Windows PC  by IT crew. I couldn't save the file :O

mbasaglia: I think the filesystem gets constructed using `App::get_temporary_directory()` which is defined like that:
```c++
App::get_temporary_directory()
{
	return synfigapp::Main::get_user_app_directory() + ETL_DIRECTORY_SEPARATOR + "tmp";
}
```

mbasaglia: by the way, `temporary_filesystem->get_temporary_directory()` was already used in this function

rodolforg: I don't think it's a reason to not merge this commit.
It just sounds a bell on my rings :)

morevnaproject: No worries, this dir is in `$HOME/Synfig` on Windows. It is proven to work as it is used for running plugins and autosave functionality. ^__^

morevnaproject:
@mbasaglia Can you please rebase on top of current master? ^__^

synfig-studio/src/gui/instance.cpp
rodolforg: Well, I did had problems with 1.3.11
https://github.com/synfig/synfig/issues/1052

Actually, that's why I started to contributed to Synfig... tried to cleanup etl... in order to understand code. But I'm not a regular Windows user and building is too slow for Windows. So, I started to do other things... lol
Someday I gave up and opened the issue

mbasaglia:
> @mbasaglia Can you please rebase on top of current master? ^__^

After rebase invoking plugins causes a segmentation fault on `stream_in->rdbuf();` in `run_plugin`

mbasaglia:
It happens on master as well

rodolforg:
Did you check what was the merge conflict on rebase?
This could help to find the commit that is probably guilty

mbasaglia:
Conflict was just some argument being changed from `string` to `std::string`, as I mentioned I get the same issue on a `master` checkout

rodolforg:
Oh
Not plugin related? :/
(And I meant the guilty commit in master branch)

mbasaglia:
it affects run_plugin, but it seems regardless of my changes

mbasaglia:
hmm, it seems to only affect some files

rodolforg:
Can you check if until commit 6e6eee54232375bfd30dca43b4a4a745d7d21d43 it still worked?

If so, try to `git bisect` from this commit on ;)

mbasaglia:
Found a fix

mbasaglia:
um, I re-based again. Any chance of this getting merged?

ice0:
Sorry. Was busy last days.

Merged. Thank you!

mbasaglia:
Cool, I'll be documenting the new stuff in the XML file on the wiki

PR #1317, opened 2020-04-01
remove useless Importer::get_frame overloaded method

rodolforg:
This version with rectangle parameters aren't used anywhere


ice0:
Merged. Thank you!

PR #1328, opened 2020-04-07
avoid duplicated dynamic_cast (other cases)

rodolforg:
Improves efficiency, readability maybe, and vanishes some Coverity reports
I've made some minor cleanup too


ice0:
Merged. Thank you!

PR #1334, opened 2020-04-11
Warning when opening file of older version. Closes #1307.

### LINKED ISSUE #1307: Warning when opening file of older version
This issue is a followup to those two issues -
https://github.com/synfig/synfig/issues/1284
https://github.com/synfig/synfig/issues/314#issuecomment-603703084

We now got outline layer, which is breaks "forward compatibility".

I.e. if user opens old file in new version and saves it, then the resulting file cannot be correctly opened in old version again (he will get different placement of widthpoints and incorrect placement  of Spline Vertex converter).

I think it would be nice to add warning dialog when opening files with canvas version of 1.0 or lower. It should warn user that he is opening file of older version and saving this file might make it impossible to open correctly in old version again.

@rodolforg @ice0 what do you think? ^__^

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/90865875-warning-when-opening-file-of-older-version?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

morevnaproject:
Closes #1307.

I have used existing functionality which displays warnings on file loading.

The warning looks a bit ugly, but can be improved later.

screenshot_014

### DIFF
--- a/synfig-core/src/synfig/loadcanvas.cpp
+++ b/synfig-core/src/synfig/loadcanvas.cpp
@@ -3232,18 +3232,13 @@ CanvasParser::parse_canvas(xmlpp::Element *element,Canvas::Handle parent,bool in
 
 	Gamma gamma = canvas->rend_desc().get_gamma();
 	String version = canvas->get_version();
-	if ( version == "0.1"
-	  || version == "0.2"
-	  || version == "0.3"
-	  || version == "0.4"
-	  || version == "0.5"
-	  || version == "0.6"
-	  || version == "0.7"
-	  || version == "0.8"
-	  || version == "0.9"
-	  || version == "1.0" )
+	if ( version == "1.0" || (version[0] == '0' && version[1] == '.') )
 	{
 		gamma.set(2.2);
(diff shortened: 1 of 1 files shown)


rodolforg:
What is ugly? That wide empty space?
Or the message contents?
Or this solution itself?

morevnaproject:
> That wide empty space?

Yes, that wide empty space.

Also, I had to put manual linebreaks into message, to make sure it is clearly visible. ^__^

rodolforg:
For me, it is ok.
Maybe mention the problematic layers?

Do you want someone try to fix the empty space?

morevnaproject:
> Maybe mention the problematic layers?

The list of problematic layers can vary depending on how old previous version of Synfig. ^__^

> Do you want someone try to fix the empty space?

Yes, but this is not top priority task. I have added an issue about that - #1376

PR #1338, opened 2020-04-12
show workarea task messages as tooltip too

rodolforg:
sometimes they don't fit in due to screen size vs. message length


ice0:
Merged. Thank you!

Issue #1284, opened 2020-03-25
Incorrect position of width points on outline when opened old files in 1.3.13


morevnaproject:
**Synfig version & platform**:
Synfig 1.3.12, 1.3.13

**Issue description**:

[outline-test-2.sif.zip](https://github.com/synfig/synfig/files/4380047/outline-test-2.sif.zip)

In Synfig 1.0.2 / 1.2.2 / 1.3.11:
![screenshot_008](https://user-images.githubusercontent.com/332868/77522396-c798bb00-6eb6-11ea-98b2-af15d601015a.png)

In Synfig 1.3.12, 1.3.13:
![screenshot_009](https://user-images.githubusercontent.com/332868/77522442-d7b09a80-6eb6-11ea-85ec-a12af250afbc.png)

Detailed description of this issue: https://github.com/synfig/synfig/issues/314#issuecomment-603703084

How to solve:
Modify file loading function. It should check if file is old (check canvas version) and if it is, then for all looped outlines change order of vertices (take last vertex and make it first).

morevnaproject:
Good news:
- Synfig 1.3.11 uses canvas version 1.0
- Synfig 1.3.12 uses canvas version 1.1

https://github.com/synfig/synfig/blob/1db7f4d4952355dd844a6e0fb4924ecda2f3df3c/synfig-core/src/synfig/canvas.h#L117-L131

morevnaproject:
@rodolforg Thank you for the fix! ^__^

morevnaproject:
I have noticed this condition in code:
```
if (gran_parent && gran_parent->get_name() == "layer" && gran_parent->get_attribute_value("type") == "advanced_outline") {
```
which makes the fix applied to Advanced outlines only.

I think we should apply this fix to ALL blines - including Regular Outlines, Regions, even if they are not belong to layer (exported).
This will fix issue demonstrated in https://github.com/synfig/synfig/issues/314#issuecomment-601336943

We need to apply fix to all blines, because any bline can report position, in case if it is referenced in Spline Vertex convert (like in case of https://github.com/synfig/synfig/issues/314#issuecomment-601336943 - the advanced outline is attached to Regular Outline, which reports wrong poisions and so the artwork gets distorted)

morevnaproject:
P.S. Of course, I was talking about all _looped_ blines. For unlooped we
shouldn’t change anything. ^__^
-- 
http://morevnaproject.org/

morevnaproject:
[...]

PR #1347, opened 2020-04-13
fix bline point order when saving for older canvas (< synfig 1.4)

rodolforg:
this is due to changes discussed here https://github.com/synfig/synfig/issues/1284

They will still have trouble in 1.3.x versions :/

### DIFF
--- a/synfig-core/src/synfig/savecanvas.cpp
+++ b/synfig-core/src/synfig/savecanvas.cpp
@@ -579,10 +579,16 @@ xmlpp::Element* encode_dynamic_list(xmlpp::Element* root,ValueNode_DynamicList::
 	ValueNode_WPList::ConstHandle wplist_value_node(ValueNode_WPList::ConstHandle::cast_dynamic(value_node));
 	ValueNode_DIList::ConstHandle dilist_value_node(ValueNode_DIList::ConstHandle::cast_dynamic(value_node));
 
+	bool must_rotate_point_list = false;
+
 	if(bline_value_node)
 	{
 		if(bline_value_node->get_loop())
+		{
 			root->set_attribute("loop","true");
+			if (save_canvas_version < RELEASE_VERSION_1_4_0) // or get_file_version()?
+				must_rotate_point_list = true;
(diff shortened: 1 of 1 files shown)


synfig-core/src/synfig/savecanvas.cpp
```diff
@@ -579,10 +579,16 @@ xmlpp::Element* encode_dynamic_list(xmlpp::Element* root,ValueNode_DynamicList::
...
 	ValueNode_DIList::ConstHandle dilist_value_node(ValueNode_DIList::ConstHandle::cast_dynamic(value_node));
 
+	bool must_rotate_point_list = false;
+
 	if(bline_value_node)
 	{
 		if(bline_value_node->get_loop())
+		{
 			root->set_attribute("loop","true");
+			if (save_canvas_version < RELEASE_VERSION_1_4_0) // or get_file_version()?
```
rodolforg, line 589: @blackwarthog could you point me the difference between `save_canvas_version` and `get_file_version()` ?
Can they store different values or are they always the same?

morevnaproject:
Great addition! 

I think it is okay if users still have trouble with 1.3.x, since this is a development version (i.e. "unoffilcial/unstable" release). ^__^

morevnaproject:
It works perfectly, thank you so much! ^__^

synfig-core/src/synfig/savecanvas.cpp
morevnaproject: `get_file_version()` returns `save_canvas_version`, so I think they are always the same.

morevnaproject:
Oh, I just remembered that we also need to lower canvas version when saving file in old version.

Otherwise following situation is possible:
- user opens (creates) file in Synfig 1.4
- adds looped advanced outlines with widthpoints
- user saves file in older format
- user opens closes file and opens it again in Synfig 1.4 -> widthpoint positions are wrong.

I have added PR #1365 about that. ^__^

rodolforg:
Nice catch :)

PR #1353, opened 2020-04-15
Show interpolation type directly in waypoint's context menu.

morevnaproject:
Previously it was hidden in sub-menu:

screenshot_007

...now it is available directly after right-click:

screenshot_006

(no need to do complex mouse movements ^__^).


ice0:
Merged. Thank you!

PR #1365, opened 2020-04-18
fix bline point order when saving for older canvas (part 2)

morevnaproject:
Complementary fix for #1347

It fixes following situation:
- user opens (creates) file in Synfig 1.4
- adds looped advanced outlines with widthpoints
- user saves file in older format
- user opens closes file and opens it again in Synfig 1.4 -> widthpoint positions are wrong.


ice0:
Merged. Thank you!

Issue #1346, opened 2020-04-13
Sound panel should list available audio layers


rodolforg:
The Sound panel lets user choose an audio from filesystem by browsing instead of show a combobox with the audio layers the scene has.

Once user selects a sound layer, it should automatically obey the delay parameter value of this layer and set it in sound panel.

A possible problem is if user uses multiple sound layers with same name. The combobox contents would have "identical" items (delay parameters may be different).

It should keep tracking scene for sound layers additions and removals and renames.

morevnaproject:
Great addition! I would love to have that functionality in Synfig! ^__^

rodolforg:
Any suggestion for the repeated layer names?

Issue #650, opened 2018-10-18
Combobox initial dropdown state


wstaelens:
**Synfig version & platform**:
SF 1.2.2
Windows 10 x64 Pro

**Issue description**:

![synfig white combo](https://user-images.githubusercontent.com/7447285/47138669-5664d980-d2ba-11e8-982e-b37a4029e13f.png)

first time i see a lot of white when opening. the items should be starting on top...

ice0:
This is a known GTK+ combobox problem (discussed here https://bugs.launchpad.net/ubuntu/+source/gtk+2.0/+bug/388633), so i am not sure it can be fixed at this time.

wstaelens:
Time to move to .NET Standard/.NET Core ? ;-) ;-)
Hope to see it soon fixed, as this is really a basic control.... Not expecting to still see such bugs in GTK+...

ice0:
As a workaround, we can move the default value to the bottom of the list. @morevnaproject, what do you think?

wstaelens:
I think it would be better to bomb that gitlab bug ticket so that it gets some attention and hopefully a fix.

morevnaproject:
Interesting information from http://libregraphicsworld.org/blog/entry/week-recap-23-december-2018 -

> Kevin Stoffler contributed a patch that works around the design decision of GTK+ developers regarding drop-down lists. Instead of always showing the currently selected item of a drop-down list under the cursor (at the cost of showing a lot of blank space above), GIMP now renders the whole list in a much nicer way (while losing the item selection). The patch was applied to both unit selection and blend mode selection. Smaller UX changes involve things like keeping aspect by default when scaling with the Unified Transform tool.

morevnaproject:
https://twitter.com/GIMP_Official/status/1075541003468267520

morevnaproject:
Related commit - https://gitlab.gnome.org/GNOME/gimp/commit/a9a979b2d0a9a3ef25d63c90a1401994cca2a818

BobSynfig:
Test done ok

Add this line in _/synfig-studio/src/gui/widgets/widget_enum.cpp_ (line 65)
>this->set_wrap_width(1); //#650

As a result, depending the size and the position of the window, the default value may not be selected by default.

![650](https://user-images.githubusercontent.com/45494804/50453303-ee78e200-093f-11e9-8959-5286753e804d.png)

morevnaproject:
The change is applied now. Thanks to everyone for participation!

PR #1370, opened 2020-04-20
Soundwave dock now loads sound layers from canvas

rodolforg:
Fix https://github.com/synfig/synfig/issues/1346

### DIFF
--- a/synfig-studio/src/gui/docks/dock_soundwave.cpp
+++ b/synfig-studio/src/gui/docks/dock_soundwave.cpp
@@ -201,18 +201,9 @@ private:
 			etl::handle<synfig::Layer_Sound> layer_sound = etl::handle<synfig::Layer_Sound>::cast_dynamic(layer);
 			if (!layer_sound)
 				return;
-			std::string guid = layer_sound->get_guid().get_string();
-			Gtk::TreeModel::iterator found_iter;
-			file_combo.get_model()->foreach_iter([guid, &found_iter](const Gtk::TreeModel::iterator& iter) -> bool {
-				Glib::ustring row_guid;
-				iter->get_value(1, row_guid);
-				if (row_guid == guid) {
-					found_iter = iter;
-					return true;
-				}
-				return false;
-			});
+			Gtk::TreeModel::iterator found_iter = find_layer_iter(layer_sound);
 			if (found_iter) {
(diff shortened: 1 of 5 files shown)


rodolforg:
It now has issues with Canvas Time model beginning being shifted from 0f. I'll fix that later.

Anyway, tell me what you think about how we should name items in sound layer combobox…

Currently it uses filename listed in sound layer parameter.

rodolforg:
I see one problem here: popping the combobox is not good :(

rodolforg:
@morevnaproject could you test it?

morevnaproject:
Testing now! Thank you! ^__^

morevnaproject:
The feature works great!

For name in combobox I suggestto display it in the following way:
* [**LAYER_NAME**] - **/path/to/file.wav**

Also I suggest to shorten path to file if it is too long.
I.e. for "/very/very/very/very/very/long/path/to/file.wav" display it like "...ry/long/path/to/file.wav".

rodolforg:
Is this way better?

morevnaproject:
I tested it and for some reason it doesn't shows filepath -
screenshot_004

rodolforg:
It only shows if it is different of layer description ("name").
Should it be always visible?

morevnaproject:
> Should it be always visible?

Yes, I think it's better to make it always visible. ^__^

Also, I just  realized that it would be nice if combobox choices will appear above of combobox itself. Just like we did for Blend Method's parameter combobox - https://github.com/synfig/synfig/issues/650#issuecomment-450001367

rodolforg:
I realize that `synfig::CanvasFileNaming::make_short_filename()` doesn't ellipsize long strings, it creates a relative path (which usually shortens the name).

How many chars would have a long filename?
In order to implement your example:

> I.e. for "/very/very/very/very/very/long/path/to/file.wav" display it like "...ry/long/path/to/file.wav".

morevnaproject:
> How many chars would have a long filename?

I guess, 30 is enough. ^__^

rodolforg:
Done :)

morevnaproject:
Thank you, it works perfectly now! Can you rebase, please? ^__^

rodolforg:
Done :)

morevnaproject:
Thank you for an awesome feature! ^__^

PR #1382, opened 2020-04-26
Moved state machine (smach) from ETL to synfig-studio/src/gui

ice0:


rodolforg:
So you ignored the test file?

We could ignore the intermediate file "smach" and replace with "_smach.h", don't you think?

ice0:
> So you ignored the test file?

Yes. There is no tests for synfig-studio for now, so i just disable it for a time. Will be back to it later.

> We could ignore the intermediate file "smach" and replace with "_smach.h", don't you think?

I agree. But smach need other improvements, so this is just an intermediate commit that should solve 2 problems:
1) Move smach from ETL to synfig-studio and make sure nothing is broken.
2) Do not loose the git history for this file.

Issue #1398, opened 2020-05-05
Interest in snap build


ppd:
Is there interest in providing a snap build? I have a quick proof-of-concept in my fork's snap branch: https://github.com/ppd/synfig/tree/snap

I don't really use synfig enough for me to consider maintaining the package myself. But if the project would like to carry something like this upstream, we could turn this into something mergeable.

It's not a hugely complex package anyway. A few CMake fixes were necessary, but all in all it's fairly straightforward.

pixelr0:
+1 for a snap package

ice0:
Sure! That will be awesome!

ppd:
What does make sense to ship in the snap? ffmpeg for mod_ffmpeg makes sense. What else?

ice0:
> A couple of things:
> 
> 1. What does make sense to ship in the snap? ffmpeg for mod_ffmpeg, libdv-bin for mod_dv, imagemagick for mod_imagemagick. What else?

Synfig have a lot of dependencies: sigc++, cairomm, glib, etc. 
You can check the full list in out flatpak build: https://github.com/flathub/org.synfig.SynfigStudio/blob/1.4-beta/org.synfig.SynfigStudio.yaml


> 2. Someone should register a user "synfig" in the snap store and also register the snap name "synfigstudio"

Done. Waiting for synfigstudio snap name to be approved. This name is reserved at the moment.

> 3. The appdata & desktop files are not being built by CMake atm. This is in progress though, isn't it?

Yes. I added issue for that, so it will be done during GSoC 2020.

ppd:
> Synfig have a lot of dependencies

Thank you. I was just curious about implicit dependencies that are called at runtime. The snap itself builds and runs already.

ice0:
This is what is called at runtime:

which
https://github.com/synfig/synfig/blob/a4bbdd72e57be5b449874eab8a6b47425af8d4fb/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp#L190

ffmpeg 
https://github.com/synfig/synfig/blob/a4bbdd72e57be5b449874eab8a6b47425af8d4fb/synfig-core/src/modules/mod_ffmpeg/mptr_ffmpeg.cpp#L155

convert
https://github.com/synfig/synfig/blob/9d2e9afb0770b91d311ecc3f128e32736ba8b22d/synfig-core/src/modules/mod_imagemagick/mptr_imagemagick.cpp#L157


encodedv
[...]

PR #1399, opened 2020-05-06
Cmake fixes

ppd:
A few CMake related fixes that came up while creating a snap package (#1398)

All in all, basic stuff. To quote the commits:

> Add include & library dirs for each pkg_config module
> This allows building & linking with libraries in non-standard paths.


ice0:
Merged. Thank you!

PR #1400, opened 2020-05-06
Add support for building snap packages

ppd:
As per #1398, this is the draft PR for adding support for building snap packages of synfigstudio.

~~Includes the separately proposed CMake fixes.~~ Much is left to discuss and agree upon, obviously.

To build & run:

```shell
# install snapcraft
snap install --classic snapcraft
# run snapcraft to build
snapcraft
# install resulting package
snap install --dangerous synfigstudio[...]_amd64.snap
# run it from your desktop menu/dash or from the cli:
snap run synfigstudio # or just synfigstudio
```


snap/gui/synfigstudio.desktop
```diff
@@ -0,0 +1,63 @@
+[Desktop Entry]
```
ppd, line 1: This file is just straight from the deb but needs to be built by CMake eventually

ppd:
Seems like the localization is not being built with CMake. Ist that in progress? There's a GSOC project for something CMake related, it seems.

ice0:
> Seems like the localization is not being built with CMake. Is that in progress?

Yes it will be done during the GSoC 2020.

Currently we are using autotools build for flatpak. You can check it here: https://github.com/flathub/org.synfig.SynfigStudio/blob/1.4-beta/org.synfig.SynfigStudio.yaml

snap/snapcraft.yaml
```diff
@@ -0,0 +1,117 @@
...
+    after: [synfig-core]
+    plugin: autotools
+    source: synfig-studio
+    configflags: *cfgflags
+    override-build: *build
+    parse-info: [usr/share/appdata/org.synfig.SynfigStudio.appdata.xml]
+    override-pull: |
+      snapcraftctl pull
+      # consider grade to be stable if commit is tagged
+      git describe --exact-match HEAD && grade="stable" || grade="devel"
```
ppd, line 93: Do we want this? Grade does not mean much. It's mostly a safeguard against pushing "devel" grade releases to stable channels.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,117 @@
...
+      sed -i -E "s|^Icon=(.*)|Icon=\${SNAP}/usr/share/icons/hicolor/scalable/apps/\1.svg|g" \
+        usr/share/applications/org.synfig.SynfigStudio.desktop
+
+  cleanup:
+    after: [synfig-studio]
+    plugin: nil
+    build-snaps: [core18, gnome-3-34-1804]
+    override-prime: |
+      set -eux
+      for lib in "libharfbuzz" "libglimm" "libgiomm" "libQt" "libflite" "qt5"; do
```
ppd, line 108: All this gets pulled in by the stage-packages. Libharfbuzz, libglimm and libgiomm are shipped in the gnome-3-34-1804 snap and must not be included. The rest is for space-saving; I don't think any of it is necessary for synfigstudio or its dependencies.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,117 @@
...
+      intltoolize --force
+      snapcraftctl build
+
+  synfig-studio:
+    after: [synfig-core]
+    plugin: autotools
+    source: synfig-studio
+    configflags: *cfgflags
+    override-build: *build
+    parse-info: [usr/share/appdata/org.synfig.SynfigStudio.appdata.xml]
```
ppd, line 89: We pull description + version from the appdata file. Maybe we want to set the version according to git instead? E.g. via git describe or by appending the commit hash.

ppd:
@ice0 let me hear what you think; I started reviews on all parts I consider a matter of opinion.

snap/snapcraft.yaml
ice0: I like the idea to have a safeguard.

ice0:
ping @morevnaproject

snap/snapcraft.yaml
ppd: Let's keep it then

snap/snapcraft.yaml
ppd: Atm, each revision in the snap store would be versioned "1.5.0". It's a bit confusing for the user. New revision, but the version stays the same. We could do something like "1.5.0~48b6b5d".

snap/snapcraft.yaml
ppd: :exploding_head: a typo, of course

snap/snapcraft.yaml
morevnaproject: I think for nightly builds we need a version with commit hash and date of commit. Like this - 1.5.0-20200514_48b6b5d.
In this case the builds can be sorted.

For releases (development or stable) it is enough to have just version number - i.e. 1.5.0.

All stable/development releases have a corresponding tag on them (i.e. v1.2.0, v1.2.1, v1.3.1, etc). Maybe this can be used to distinguish nightly builds from releases (if needed).

ppd: Okay. So we do it like this:

```
version = read from appdata file
if commit is not tagged:
  version = version + date + hash
```

ppd: The parse-info stuff clashes with setting the version manually, so I have pulled the appdata summary and description into snapcraft.yaml and only get the version from there.

ppd:
I resolved all discussions and improved the cleanup part. Any further comments or questions?

ice0:
Merged. Thank you!

PR #1404, opened 2020-05-10
hide GUID int constructor

rodolforg:
to avoid anybody to mess with its int parameter that could only be zero.

It fixes compiler warnings about unused parameter


ice0:
Merged. Thank you!

PR #1418, opened 2020-05-17
Cleanup module.cpp a bit

rodolforg:
and add module path specific to particular user in Windows and OSX too


ice0:
@rodolforg any ideas about this alerts?

ice0:
Merged. Thank you!

PR #1423, opened 2020-05-17
Fixed smach (state machine) compilation on MSVC

ice0:
This part compiles differently in gcc/clang and MSVC compilers. Since the behavior of the compiler is not specified for virtual methods (C++11 standard says: § 14.7.1.10. "It is unspecified whether or not an implementation implicitly instantiates a virtual member function of a class template if the virtual member function would not otherwise be instantiated").

Unlike GCC/Clang, MSVC expects the template class to be fully defined at the moment, so this code does not compile. Therefore, we move the `enter_state` method to derived classes.

Minimal working example:
```
template <class T>
class A {
  virtual void* Run() {
    return new T();
  }
};

class C;

class B : A<C> {
  
};
```

MSVC fails with the error:
```
<source>(4): error C2027: use of undefined type 'C'
<source>(8): note: see declaration of 'C'
<source>(3): note: while compiling class template member function 'void *A<C>::Run(void)'
<source>(10): note: see reference to class template instantiation 'A<C>' being compiled
Compiler returned: 2
```

This code can be tested online at https://godbolt.org

### DIFF
--- a/synfig-studio/src/gui/_smach.h
+++ b/synfig-studio/src/gui/_smach.h
@@ -217,10 +217,49 @@ public:
 
 	protected:
 
-		virtual void* enter_state(context_type* machine_context)const
+		/*
+		This part compiles differently in gcc/clang and MSVC compilers.
+		Since the behavior of the compiler is not specified for virtual methods
+		(C++11 standard says: § 14.7.1.10. "It is unspecified whether or not an
+		implementation implicitly instantiates a virtual member function of a
+		class template if the virtual member function would not otherwise be instantiated").
+
+		Unlike GCC/Clang, MSVC expects the template class to be fully defined
+		at the moment, so this code does not compile. Therefore, we move
+		the `enter_state` method to derived classes.
+
+		Minimal working example:
+		```
+		template <class T>
+		class A {
+		virtual void* Run() {
+			return new T();
+		}
+		};
+
+		class C;
+
+		class B : A<C> {
+		
+		};
+		```
+
+		MSVC fails with the error:
+		```
+		<source>(4): error C2027: use of undefined type 'C'
+		<source>(8): note: see declaration of 'C'
+		<source>(3): note: while compiling class template member function 'void *A<C>::Run(void)'
+		<source>(10): note: see reference to class template instantiation 'A<C>' being compiled
+		Compiler returned: 2
+		```
+
+		This code can be tested online at https://godbolt.org		
+		*/
+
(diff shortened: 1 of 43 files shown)


synfig-studio/src/gui/_smach.h
```diff
@@ -217,10 +217,49 @@ class smach
...
+		<source>(8): note: see declaration of 'C'
+		<source>(3): note: while compiling class template member function 'void *A<C>::Run(void)'
+		<source>(10): note: see reference to class template instantiation 'A<C>' being compiled
+		Compiler returned: 2
+		```
+
+		This code can be tested online at https://godbolt.org		
+		*/
+
+		/*virtual void* enter_state(context_type* machine_context)const
```
rodolforg, line 259: I think would be better to let:
`virtual void* enter_state(context_type* machine_context)const = 0;`

if MSVC allows it.

The state would be obliged to implement it

synfig-studio/src/gui/_smach.h
```diff
@@ -217,10 +217,49 @@ class smach
...
+		MSVC fails with the error:
+		```
+		<source>(4): error C2027: use of undefined type 'C'
+		<source>(8): note: see declaration of 'C'
+		<source>(3): note: while compiling class template member function 'void *A<C>::Run(void)'
+		<source>(10): note: see reference to class template instantiation 'A<C>' being compiled
+		Compiler returned: 2
+		```
+
+		This code can be tested online at https://godbolt.org
```
rodolforg, line 256: Cool!

synfig-studio/src/gui/_smach.h
ice0: I will check this.

ice0:
> How was it compiling before the move to synfig-studio?

This is MSVC related problem. Currently we use MinGW for Windows builds.

rodolforg:
Oh, I thought it was compiling since your last commits ;)

ice0:
> Oh, I thought it was compiling since your last commits ;)

Not yet, but I am working on it :)

synfig-studio/src/gui/_smach.h
ice0: Just checked:
It is already virtual in `state_base`, and it requires you to implement it anyway. So maybe leave it as it is?

rodolforg: Oh. Yupe :)

PR #1424, opened 2020-05-18
don't try to delete/move/copy zero waypoints

rodolforg:
If zero waypoints is selected in Timetrack and user presses Delete key,
an error message popped up as synfigapp::Action is not ready (not enough parameters).


ice0:
Merged. Thank you!

PR #1425, opened 2020-05-18
activate widget_filename when user chooses file via button/file dialog

rodolforg:
otherwise the parameter value change is only propagated when widget looses focus or user explicitly activates (by pressing Enter/Return).

If it's a Sound layer, clicking on Play button right after choose a file (and the edit-widget not yet loosing focus), it would play the previous audio.


rodolforg:
@morevnaproject for UX, it should be cherry-picked to 1.3.x, I guess

ice0:
Merged. Thank you!

morevnaproject:
> @morevnaproject for UX, it should be cherry-picked to 1.3.x, I guess

Done! Thank you! ^__^

PR #1426, opened 2020-05-18
Cleaned clock functions. Fixed compilation on MSVC.

ice0:


ETL/ETL/_clock_system.h
```diff
@@ -29,25 +29,11 @@
...
-#else
-# ifdef __GNUG__
-#  include <time.h>
-#  define __sys_clock	::clock
-#  define __sys_time	::time
-# else
-typedef int clock_t;
-typedef int time_t;
-extern clock_t _clock();
-extern time_t _time(time_t *);
```
rodolforg, line 45: I can't find those two functions above anywhere in synfig code, but here. I used this command line.
`$ git grep \\b_time\s*\(`
So I think this fallback was actually never reached and it's fine to delete.

ETL/ETL/_clock_system.h
```diff
@@ -29,25 +29,11 @@
...
-#  define CLOCKS_PER_SEC 1000
-#  define __sys_clock	_clock
-#  define __sys_time	_time
-# endif
-#endif
+#ifndef __sys_clock
+#include <ctime>
+#define __sys_clock	::clock
```
rodolforg, line 34: Could we eliminate `__sys_clock`?
Anyway, macro names starting with `__` are reserved and should not be note used. But it's used here and as headr file guards (__ETL__CLOCK_SYSTEM_H , __SYNFIG_xxxxx_H and so on).

ETL/ETL/_clock_system.h
ice0: I think i was remove it :)
Am I miss something?

ETL/ETL/_clock_system.h
ice0: Sure. I can’t do it now because I have a bunch of modified files and it will be difficult for me. But you can help here if you want. I also want to come back here later, because I think that `ETL/time` can be transferred to `synfig-core`.

ice0:
> > In programming environments where clock_t is a 32-bit integer type and CLOCKS_PER_SEC is one million, clock() will start failing in less than 36 minutes of processor time for signed clock_t, or 72 minutes for unsigned clock_t.
> 
> So are 32-bit systems a trouble? I don't know how this clock() is used.

I think all logic is build around 
```#  define CLOCKS_PER_SEC 1000

ETL/ETL/_clock_system.h
rodolforg: I just meant I agree :)

ETL/ETL/_clock_system.h
rodolforg: Ok. :) Not for this PR anyway

Issue #1393, opened 2020-04-30
File corruption: Copy-paste layers with exported values from one file to another


morevnaproject:
There is a very common situation when user can get his file corrupted by simply copy-pasting layers from one file to another. This happens if layer contains exported values (note that bones are exported values too!).

Steps to reproduce this situation:
1. Launch Synfig and create new file.
2. Create a simple Circle Layer.
3. Select Circle Layer, right-click its Radius parameter and select "Export".
4. Enter name for exported value, i.e. "test". Click Export.
5. Save file as "file1.sifz"
6. Hit Ctl+C to copy circle layer to clipboard.
6. Create a new file in other tab.
7. Hit Ctrl+V to paste circle from clipboard. Circle will appear on workspace. Notice, that layer is copied, but it is still references exported value from "file1.sifz"
8. Save file as "file2.sifz"
9. Quit Synfig.
10. Now remove (or rename) "file1.sifz".
11. Launch Synfig and try to open "file2.sifz".
12. Error:
![screenshot_001](https://user-images.githubusercontent.com/332868/80699264-df8eda80-8b05-11ea-934b-7c4702fdffe9.png)

@rodolforg @ice0 @FirasH  I think this is a top-priority issue to solve for next stable release (1.4.1 or 1.6.0).

What should be done here:

1. Change how Synfig deal with pasting layers when layer has exported values.
2. Change how Synfig handles situations with broken links to exported values in external documents.

morevnaproject:
Broken files reported by users regularly.
Examples: #786, #1387

morevnaproject:
> 1. Change how Synfig deal with pasting layers when layer has exported values.
> 2. Change how Synfig handles situations with broken links to exported values in external documents.

I think (1) should be a scope of this issue. (2) should be submitted a separate issue.

morevnaproject:
> 1. Change how Synfig deal with pasting layers when layer has exported values.

Here is my proposal for solving this - https://synfig-docs-dev.readthedocs.io/en/latest/projects/copy-paste-exported-values.html

Gilraiser:
Hello, I would like to show you an error that has to do with the topic. Here I present a video and two example files of how bones made are corrupted when copied to another file.

Original file:
[example-01.sif.zip](https://github.com/synfig/synfig/files/4663967/example-01.sif.zip)

Copied file:
[example-02.sifz.zip](https://github.com/synfig/synfig/files/4663965/example-02.sifz.zip)

Video with the demonstration:
[...]

PR #1438, opened 2020-05-23
Trying to fix skeleton copy

rodolforg:


rodolforg:
fixed LGTM reports

rodolforg:
It seems to work with: 
[bone-dup-test.zip](https://github.com/synfig/synfig/files/4671201/bone-dup-test.zip)
But now I noticed that it doesn't work with Gilraiser example https://github.com/synfig/synfig/issues/1393#issuecomment-632256581

rodolforg:
Now it just duplicates the skeleton itself, don't try to link anything to its bones.
What do you think, @morevnaproject @ice0 ?

morevnaproject:
@rodolforg Thank you so much for taking on this issue! I am reviewing it now. ^__^

morevnaproject:
> Now it just duplicates the skeleton itself, don't try to link anything to its bones.
> What do you think, @morevnaproject @ice0 ?

I see. If we duplicate a group (which contains character parts and its skeleton), then we will get a copy of this character, but his parts are not linked to skeleton.

Of course, I would prefer to have a copy which preserves links.

But, of course the behavior introduced by this PR is **much** better than situation without this PR (I mean duplicated bones and corrupted files).

morevnaproject:
I think if we merge this PR at its current state to stable version of Synfig (1.3.x), then we will avoid **a lot** of bug reports after release.

rodolforg:
@morevnaproject I removed draft status then.

> Of course, I would prefer to have a copy which preserves links.

The problem is that we don't know if the linked layers are or not cloned too.
Unless we make an exception in layer duplicate action to investigate this everytime a skeleton layer is to be duplicated.

PR #1439, opened 2020-05-23
quick workaround to crash in new timetrack

rodolforg:
- it happened when switching tab from a doc to another
it just avoids the crash, the cause still has to be investigated


ice0:
Merged. Thank you!

PR #1444, opened 2020-05-24
Animation Speed of Noise Distort Layer set to static by default

### LINKED ISSUE #1443: "Animation Speed" parameter of Noise Distort layer should be static by default
This issue is analog of #780. 

Solution is similar (see #1259).

FirasH:
Fixes #1443


morevnaproject:
Thank you! ^__^

PR #1445, opened 2020-05-24
Restore Defaults is applied only on OK pressed

### LINKED ISSUE #1415: Shouldn't Preferences > Restore Defaults give the user a chance to confirm?
Building from source
Debian sid

**Issue description**:
Screenshot from 2020-05-16 09-30-36

When I hit the Restore Defaults button, synfig restores everything back to factory settings AND saves it without the user either confirming the changes with OK button or Canceling the changes.

My expected outcome for Restore Defaults is
1. Change to factory default values
2. Wait for the user to either
    - Confirm the change (OK)
                    or
    - Revert the change (Cancel)

FirasH:
Fixes #1415

### DIFF
--- a/synfig-studio/src/gui/instance.cpp
+++ b/synfig-studio/src/gui/instance.cpp
@@ -758,7 +758,7 @@ Instance::safe_revert()
 	{
 		if (!App::dialog_message_2b(
 			_("Revert to saved"),
-			_("You will lose any changes you have made since your last save."
+			_("You will lose any changes you have made since your last save. "
 				"Are you sure?"),
 			Gtk::MESSAGE_QUESTION,
 			_("Cancel"),
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -2289,48 +2289,6 @@ void App::edit_custom_workspace_list()
 }
 
 void
-App::restore_default_settings()
-{
-	std::ostringstream temp_time_format;
-	temp_time_format << Time::FORMAT_FRAMES;
-	synfigapp::Main::settings().set_value("pref.time_format",                    temp_time_format.str());
-
-	synfigapp::Main::settings().set_value("pref.distance_system",                "pt");
-	synfigapp::Main::settings().set_value("pref.file_history.size",              "25");
-	synfigapp::Main::settings().set_value("pref.autosave_backup",                "1");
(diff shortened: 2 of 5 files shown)


FirasH:
Includes also a missing space in the Revert to saved confirm dialog.

FirasH:
There is a functional issue with the current implementation.

Some values have their signal connected to a function which does change the settings of Synfig.
This means that if for example autosave is disabled, when `Cancel` is pressed it won't be reverted to current settings:
https://github.com/synfig/synfig/blob/6d3f75ffc8a3450587c8105a738e1b6db79f7956/synfig-studio/src/gui/dialogs/dialog_setup.cpp#L829-L834

Probably the best way to deal with it is to remove those parts that in practice don't give any advantage to the user.

Keyikedalube:
I think these portions of code are safe to be removed when you do your final code cleanup:
- https://github.com/FirasH/synfig/blob/75fd11d512a23b38910efec2c2dbd35382e9d986/synfig-studio/src/gui/app.cpp#L2271-L2310
- https://github.com/FirasH/synfig/blob/75fd11d512a23b38910efec2c2dbd35382e9d986/synfig-studio/src/gui/app.h#L345

synfig-studio/src/gui/instance.cpp
```diff
@@ -974,7 +974,7 @@ Instance::safe_revert()
 	{
 		if (!App::dialog_message_2b(
 			_("Revert to saved"),
-			_("You will lose any changes you have made since your last save."
+			_("You will lose any changes you have made since your last save. "
```
ice0, line 977: Is extra space needed?

rodolforg: It's because "are you share" ;)
Maybe a newline `\n` would be better.

FirasH: Without the space here is how they look:
no_space

ice0: New line as suggested by @rodolforg would be better, but that is ok too.

FirasH:
@ice0 this request is ready. Do you need me to force-push it?

ice0:
@FirasH yes, can you rebase it?

FirasH:
> @FirasH yes, can you rebase it?

Done :+1:

ice0:
Merged. Thank you!

PR #1446, opened 2020-05-24
[Lottie Exporter] Add support for radial gradient layer

saurabh-ramola:
@AnishGG , Radial Gradient layer is now complete. Can you please test it and let me know if it is correct?

PR #1456, opened 2020-05-30
[Lottie exporter] Refactor lottie modules

AnishGG:
According to suggestion by @ice0 (https://github.com/synfig/synfig/pull/1446#discussion_r431296941), I have refactored some modules in lottie for better understanding.


ice0:
Merged. Thank you!

PR #1461, opened 2020-06-01
Check if target output has same name as existing files

### LINKED ISSUE #1360: Ask for confirmation before overwriting exported target
Master: ccf39ec

**Issue description**:
When exporting an animation to any arbitrary target, Synfig Studio never informs the user if the action is going to overwrite an existing file.

I suggest to add one check (for example with a popup dialog), before proceeding with the rendering of the target.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91175546-ask-for-confirmation-before-overwriting-exported-target?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

FirasH:
Fixes #1360

### DIFF
--- a/synfig-studio/src/gui/render.cpp
+++ b/synfig-studio/src/gui/render.cpp
@@ -449,7 +449,7 @@ RenderSettings::check_target_destination()
 	}
 	else
 	{
-		message = strprintf(_("\"%d\" files with the same name already exist. "
+		message = strprintf(_("%d files with the same name already exist. "
 							"Do you want to replace them?"),
 							n_frames_overwrite);
 	
--- a/synfig-studio/src/gui/render.cpp
+++ b/synfig-studio/src/gui/render.cpp
@@ -384,9 +384,6 @@ RenderSettings::check_target_destination()
 		return false;
 	}
 	
-	struct stat s;
-	int stat_return = -1; //Default to error
-	
 	//Retrieve current render settings
 	RendDesc rend_desc(widget_rend_desc.get_rend_desc());
 	
@@ -428,7 +425,8 @@ RenderSettings::check_target_destination()
 	}
 	
 	//Check name of target or first image of the sequence
-	stat_return = stat(full_name.c_str(), &s);
+	struct stat s;
+	int stat_return = stat(full_name.c_str(), &s);
 	
 	//Error on the selected path	
 	if(stat_return == -1 && errno != ENOENT)
--- a/synfig-studio/src/gui/render.cpp
+++ b/synfig-studio/src/gui/render.cpp
@@ -340,10 +340,7 @@ bool
 RenderSettings::check_target_destination()
 {
 	String filename=entry_filename.get_text();
-	String full_name(filename); //include eventually sequence_separator + time for image sequences
 	calculated_target_name=target_name;
-	
-	int n_frames_overwrite = 0;
 
 	if(filename.empty())
 	{
@@ -384,6 +381,11 @@ RenderSettings::check_target_destination()
 		return false;
 	}
 	
+	String full_name(filename); //include possible sequence_separator + time for image sequences
+	String extension(filename_extension(filename));
+	bool ext_multi_file = false; //output target is an image sequence
+	int n_frames_overwrite = 0;
+	
 	//Retrieve current render settings
 	RendDesc rend_desc(widget_rend_desc.get_rend_desc());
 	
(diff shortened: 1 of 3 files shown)


FirasH:
In case of an image sequence only the first output file name is checked, not the whole sequence.

Example scenario:
- existing files with numbering from 0010 to 0050
- export image sequence from 0009 to 0080
this will overwrite without warning the user (as in all cases without this patch)

If the structure of the code make sense, this last check should imho be implemented to avoid an error and overwrite a portion of a previously rendered sequence.

synfig-studio/src/gui/render.cpp
```diff
@@ -348,11 +349,89 @@ RenderSettings::on_render_pressed()
...
+	ext_multi_it = ext_multi.find(calculated_target_name);
+	
+	//calculated_target_name is a candidate with known output target (not Auto)
+	if(ext_multi_it != ext_multi.end())
+	{
+		//Image sequence: filename + sequence_separator + time
+		if(!toggle_single_frame.get_active() &&
+				(rend_desc.get_frame_end() - rend_desc.get_frame_start()) > 0)
+			full_name = filename_sans_extension(filename) +
+				App::sequence_separator +
```
FirasH, line 375: Maybe better to use `tparam.sequence_separator`?

FirasH:
> I don't like long functions (class methods) with multiple purposes.
> Maybe you should separate the image-sequence part?

I can move the part of the function used to check the file name etc. to another method, maybe something like `RenderSettings::check_target_destination()`?

> Actually, I think target codec itself should give us the hint if it outputs image sequence (maybe with the given target parameters). What do you think?

mmm, if I get what you mean we could use `tparam.video_codec` instead of checking for the matching `calculated_target_name`?

FirasH:
The code works, I'll try to squash it as much as possible.

Some notes:
- as @rodolforg said it is easier to use `Glib::file_test()`, need to replace that
- check if using `tparam.video_codec` simplifies drastically the implementation without the need to use that long `if-else` statement (or some other viable solution could do as well)

Any advice is welcome :grin:.

FirasH:
The long `if-else` is necessary because if the target is not explicitly selected (left on `Auto`) Synfig has to figure out the format to use.

In some cases, the same extension could have a sequence or not, like `.png` for an image sequence of `png` or a single file for `png-spritesheet`.

synfig-studio/src/gui/render.cpp
```diff
@@ -339,39 +374,116 @@ RenderSettings::on_render_pressed()
...
+	//If format is selected in comboboxtext_target
+	std::map<std::string,std::string> ext_multi = {{"bmp",".bmp"},{"cairo_png",".png"},
+					{"imagemagick",".png"}, {"jpeg",".jpg"},{"mng",".mng"},
+					{"openexr",".exr"},{"png",".png"},{"ppm",".ppm"}};
+	
+	std::map<std::string,std::string>::iterator ext_multi_it;
+	ext_multi_it = ext_multi.find(calculated_target_name);
+	
+	//calculated_target_name is a candidate with known output target (not Auto)
+	if(ext_multi_it != ext_multi.end())
```
rodolforg, line 400: Do you see both if-else branches share a lot of same code?
Can you "optimize" it? Avoiding code repetition is good.

rodolforg:
> mmm, if I get what you mean we could use `tparam.video_codec` instead of checking for the matching `calculated_target_name`?

I think Target class _should_ provide a way to tell what is supported by target rendering:

- export as movie, image sequence or static image (single frame or an atlas (like spritesheet) or none
- export sound or not
- alpha as background or not
- what params are supported - currently they're are hardcoded in synfigstudio gui, instead of provided by target implementation

FirasH:
> I think Target class _should_ provide a way to tell what is supported by target rendering:
> 
>     * export as movie, image sequence or static image (single frame or an atlas (like spritesheet) or none
> 
>     * export sound or not
> 
>     * alpha as background or not
> 
>     * what params are supported - currently they're are hardcoded in synfigstudio gui, instead of provided by target implementation
> 
>     * maybe something more?

That is for sure a more complete solution.
If we want to do that the parameters of the target should be assigned **before** checking if other files have the same name and then use that info instead of searching for it.

Currently the target is defined in `submit_next_render_pass()`, this means that we can get info from target only **after** rendering has started (and their parameters assigned).

Correct me if I'm wrong :+1:

FirasH:
A sneak peek of current implementation:
target overwrite sneak peek

synfig-studio/src/gui/render.cpp
```diff
@@ -339,39 +371,104 @@ RenderSettings::on_render_pressed()
...
-	render_passes.clear();
-	if (toggle_extract_alpha.get_active())
+	
+	String full_name(filename); //include possible sequence_separator + time for image sequences
+	String extension(filename_extension(filename));
+	bool ext_multi_file = false; //output target is an image sequence
+	int n_frames_overwrite = 0;
+	
+	//Retrieve current render settings
+	RendDesc rend_desc(widget_rend_desc.get_rend_desc());
```
rodolforg, line 390: It could be a const reference instead of copy.

synfig-studio/src/gui/render.cpp
```diff
@@ -339,39 +371,104 @@ RenderSettings::on_render_pressed()
...
+					{".jpg"},{".exr"},{".ppm"}};
+	
+		ext_multi_file = (find(ext_multi_auto.begin(), ext_multi_auto.end(),
+				filename_extension(filename)) != ext_multi_auto.end());
+	}
 
-	} else {
-		render_passes.push_back(make_pair(TARGET_ALPHA_MODE_KEEP, filename));
+	//Image sequence: filename + sequence_separator + time
+	if(!toggle_single_frame.get_active() && ext_multi_file &&
```
rodolforg, line 418: I wonder if it would be better to place this `if` statement before investigate if it is a multiple-files target.
`if(!toggle_single_frame.get_active() && (rend_desc.get_frame_end() - rend_desc.get_frame_start()) > 0)`

rodolforg:
> A sneak peek of current implementation:

Why the quotes around the number of overwritten files?

FirasH:
> > A sneak peek of current implementation:
> 
> Why the quotes around the number of overwritten files?

There isn't any particular reason, just a left over from a copy-paste.
I'll remove them.

synfig-studio/src/gui/render.cpp
FirasH: I think I can simplify it a bit and move the condition to avoid useless checks.

synfig-studio/src/gui/render.cpp
FirasH: not really sure how to do that

FirasH:
> So much better to read :)
> 
> I could not test it though; did you test with spritesheet target?

`png-spritesheet` wasn't correctly detected, not it's fixed.

Tested with all formats and seems working correctly.

synfig-studio/src/gui/render.cpp
```diff
@@ -339,39 +371,100 @@ RenderSettings::on_render_pressed()
...
+		ext_multi_it = ext_multi.find(calculated_target_name);
+	
+		//calculated_target_name is a candidate with known output target (not Auto)
+		if(ext_multi_it != ext_multi.end())
+		{
+			extension = ext_multi_it->second;
+			ext_multi_file = true;
+		}
+		//otherwise Auto is selected
+		else if(calculated_target_name != "png-spritesheet")
```
FirasH, line 410: Better to move this check before to avoid useless nesting of conditions

synfig-studio/src/gui/render.cpp
```diff
@@ -339,39 +371,100 @@ RenderSettings::on_render_pressed()
...
+	{
+		message = strprintf(_("A file named \"%s\" already exists. "
+							"Do you want to replace it?"),
+							basename(filename).c_str());
 	
-	progress_logger->clear();
-	submit_next_render_pass();
+		details = strprintf(_("The file already exists in \"%s\". "
+							"Replacing it will overwrite its contents."),
+							basename(dirname(filename)).c_str());
```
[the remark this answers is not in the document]
FirasH: Used that because it's the same thing used for `.sif` files:
https://github.com/synfig/synfig/blob/b6995dc5864f43239b4e5ac9b119222328af6af1/synfig-studio/src/gui/instance.cpp#L545-L552

FirasH:
> Last note: Not required but a must: If user chooses to select another filename, set focus to `entry_filename`

Actually the whole dialog is out of focus if `Use Another Name...` is selected... Not sure why, I'll take a look.

rodolforg:
> > Last note: Not required but a must: If user chooses to select another filename, set focus to `entry_filename`
> 
> Actually the whole dialog is out of focus if `Use Another Name...` is selected... Not sure why, I'll take a look.

You can force with `Widget::grab_focus()`.

synfig-studio/src/gui/render.cpp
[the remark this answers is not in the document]
rodolforg: Ouch.

FirasH:
> > > Last note: Not required but a must: If user chooses to select another filename, set focus to `entry_filename`
> > 
> > 
> > Actually the whole dialog is out of focus if `Use Another Name...` is selected... Not sure why, I'll take a look.
> 
> You can force with `Widget::grab_focus()`, I suppose.

I can set the window to active using `present()`. Happens also when selecting the destination with `Choose...`.

`entry_filename.grab_focus()` does select the whole string of the destination, although I'm not sure how useful would that be.

synfig-studio/src/gui/render.cpp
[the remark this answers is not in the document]
FirasH: Maybe is better to use absolute path?

rodolforg:
> `entry_filename.grab_focus()` does select the whole string of the destination, although I'm not sure how useful would that be.

I think it's to enforce the feedback what user chose.

synfig-studio/src/gui/render.cpp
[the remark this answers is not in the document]
rodolforg: If so, as it can be long, it would be better to place in a separated line.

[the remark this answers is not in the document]
FirasH: I fixed a couple of things but still there's this minor issue (IMHO that's why it's all on a single line):
new_line_path
The text is centered. We could place it to the left somehow, but still in case of an RTL language it wouldn't be good looking either...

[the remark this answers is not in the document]
FirasH: btw if it's very long, it breaks the line on the slash without increasing the size of the dialog.

FirasH:
@rodolforg I think everything is ready, let me know what you think

rodolforg:
I think it is ;)
Thank you!

ice0:
Merged. Thank you!

Issue #1227, opened 2020-03-12
[CMake] Generate localization files from *.po files


ice0:
.po files located inside `synfig-core/po` and `synfig-studio/po` folders.
We need to generate .mo files and put it to `$SYNFIG_BUILD_ROOT/share/locale/` folder.

Good explanation of localization process can be found [here](https://www.labri.fr/perso/fleury/posts/programming/a-quick-gettext-tutorial.html)

P.S. CMake has built-in functions to work with [localizations](https://cmake.org/cmake/help/v3.0/module/FindGettext.html).

PR #1462, opened 2020-06-03
Generate l10n files from .po files.

AYESDIE:
Fixes: #1227

Samples:
System Default (English):
image

French:
image

Russian:
image


CMakeLists.txt
```diff
@@ -14,4 +14,6 @@ include_directories(${PROJECT_BINARY_DIR}/ETL) # to find the generated etl_profi
 add_subdirectory(synfig-core)
 
 include_directories(synfig-core/src)
-add_subdirectory(synfig-studio)
\ No newline at end of file
+add_subdirectory(synfig-studio)
+
+include(cmake/Gettext.cmake)
```
ice0, line 19: Please add a new line at the end of the file.

cmake/Gettext.cmake
```diff
@@ -0,0 +1,83 @@
+set(LANGUAGES
```
ice0, line 1: You do not need to fill the language list manually. Each `po` folder has a `LINGUAS` file. Read the list of languages using the [file](https://cmake.org/cmake/help/v3.0/command/file.html) function.

synfig-core/po/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+file(READ LINGUAS LANGUAGES)
```
ice0, line 1: Why not to use `file(STRINGS ...` here?
So you don't need that part:
```
string(REPLACE "\n" ";" LANGUAGES "${LANGUAGES}")
list(APPEND LANGUAGES ${LANGUAGES})
```

synfig-core/po/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+
+set(PO_DIR ${SYNFIG_BUILD_ROOT}/share/locale)
+string(REPLACE "\n" ";" LANGUAGES "${LANGUAGES}")
+list(APPEND LANGUAGES ${LANGUAGES})
+
+foreach(LANGUAGE IN ITEMS ${LANGUAGES})
+    list(APPEND LANGUAGE_LIST "${CMAKE_CURRENT_SOURCE_DIR}/${LANGUAGE}.po")
```
ice0, line 8: Better to rename `LANGUAGE_LIST ` to `PO_FILES`

synfig-core/po/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+    foreach(LANGUAGE ${LANGUAGES})
+        add_custom_command(
+                TARGET pofiles_1 POST_BUILD
+                COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_BINARY_DIR}/${LANGUAGE}.gmo
+                ${PO_DIR}/${LANGUAGE}/LC_MESSAGES/synfig.mo
+        )
+
+        install(
```
ice0, line 25: `gettext_process_po_files` can add INSTALL target by self, so you don't need to do that.

ice0:
Also where is a problem with this code:
```
        add_custom_command(
                TARGET pofiles_1 POST_BUILD
                ...
        )
```
This code will overwrite all files every time, even if only one file is changed. I don’t have a good solution right now, but I think it can be implemented better. Probably no. Maybe not.

synfig-core/po/CMakeLists.txt
AYESDIE: I was thinking about using that INSTALL, however, it will not install the files as the bash script does. It won't rename those files to `synfig.mo` and `synfigstudio.mo` but instead, it will use the base names, so something like this.
image
In fact, to properly use this, we also need to pass the `lang` parameter, but then, using that method, it will result in a type of code which I sent before, like the one that had 108 pofiles targets.

synfig-core/po/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+file(STRINGS LINGUAS LANGUAGES)
```
ice0, line 1: I think we can put all of this strings inside `if(GETTEXT_FOUND)` condition.
After that you can move all this code to the function.

AYESDIE: I think making a `/cmake` folder and moving this code to a function in a file over there could be a good option too.

ice0: I support that.

AYESDIE:
I was thinking it would be better to make `pofiles` build only those `.mo` which aren't generated yet, which is something similar to what we are trying to achieve on `build_images` target. 

I wanted to ask if we should also consider changes in the file, like, build those files which are changed (this can also apply for `build_images` target)?

ice0:
> I wanted to ask if we should also consider changes in the file, like, rebuild those files which are changed (this can also apply for `build_images` target)?

Yes, we need to rebuild changed files.

synfig-core/po/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+include(${CMAKE_SOURCE_DIR}/cmake/gettext.cmake)
+file(STRINGS LINGUAS PO_LANGUAGES)
+
+gettext_process_po_files(
```
ice0, line 4: I suggest adding a little description of why we do not use the CMake `FindGettext` module for this.

cmake/gettext.cmake
```diff
@@ -0,0 +1,47 @@
+find_program(GETTEXT_MSGFMT_EXECUTABLE msgfmt)
```
ice0, line 1: Also please check what happens if mgsfmt is not found.

synfig-core/po/CMakeLists.txt
AYESDIE: I put the description in `/cmake/gettext.cmake`.

cmake/Modules/SynfigGettext.cmake
```diff
@@ -0,0 +1,70 @@
+#[=============================================================[
+
+Despite CMake providing functions to process .po files,
```
ice0, line 3: Here is my addtion to comment:

Although CMake provides functions for processing .po files, it cannot be used because GETTEXT_PROCESS_PO_FILES installs files with an invalid name. For example: for an en.gmo file, we should get the ${LOCALE_DIR}/${language}/LC_MESSAGES/${project_name}.mo, not ${LOCALE_DIR}/${language}/LC_MESSAGES/en.mo, as it does, so we need to get the possibility to pass ${project_name} to that function and use it for file naming.
We also added an issue to CMake:
https://gitlab.kitware.com/cmake/cmake/-/issues/20792

CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.1)
```
ice0, line 1: Which function requires version 3.1?

AYESDIE: We are using version 3.1 as a minimum in `synfig-studio/CMakeLists.txt` and `synfig-core/CMakeLists.txt`.

ice0: Yeah, they need to be fixed too. Added issue here: https://github.com/synfig/synfig/issues/1509

ice0:
Merged. Thank you!

Issue #1442, opened 2020-05-24
Synfig 1.3.14 crashes frequently on Windows


morevnaproject:
**Synfig version & platform**:
1.3.14

**Issue description**:

Originally reported here - https://forums.synfig.org/t/synfig-1-3-12-crashes-frequently/10647/27

Synfig crashes on attempt to edit spline.

Videos:

-  https://youtu.be/dTcu7AVWAHM
-  https://youtu.be/ePltkOkQQi8
-  https://youtu.be/hFjJyA0RVGU

File - [retesz_2M_2.sifz.zip](https://github.com/synfig/synfig/files/4672973/retesz_2M_2.sifz.zip)

morevnaproject:
I was unable to reproduce that on Linux. Can someone help here? @ice0

morevnaproject:
@FadhilKwan

FadhilKwan:
@morevnaproject yes ._.?

I tried to play with this file for quite a while now, and still couldn't find out why it crashes at certain times (still feels random to me).

However, i found something else that might be somewhat related...
You don't need to click any vertex points at all to find out if the crash will occur or not.

You can just click any line (or maybe even any layer), then hover your mouse to the parameter panel and try to scroll it. That also works (in terms of crashing Synfig).

That should be relatively the same thing, since when you click any point, the parameter panel will automatically scroll and select that exact point.

With this discovery on hand, i then decided to close the "Parameters" panel and tried to replicate what's in the video again. This time, no crashes occured.

But, the crash can still happen. The only way to do that is to scroll the "Timetrack" panel down,
(the "Graph" panel may can also do that, but i couldn't confirm that yet since you can't scroll far down in "Graph" panel.)

I even tried this with one of my old artwork (which was created in v1.2.2) once and this issue also happened with that file as well.

Still, I couldn't find what triggers the crash, though. This thing comes up sometimes with no clear sign of whether it'll happen or not.

rodolforg:
It doesn't seem to crash if ` --console` if passed as argument.

@FadhilKwan could you confirm it? Edit the `.bat` appending this option on the end of line of synfig executable call

FadhilKwan:
> It doesn't seem to crash if --console is passed as argument.

yeah, It kinda occurs..... much less now. (....?)

Most of the time it won't crash, but once you close the file and load it up again on the same Synfig window, scrolling the "Timeline" (or maybe clicking a point) then crashes it again almost all the time.

[...]

PR #1475, opened 2020-06-07
try to catch all exceptions thrown in GTK event callback/handlers

rodolforg:
This huge commit was made based on the discovery of the error reported
by Svarov-RZM in Github issue comment https://github.com/synfig/synfig/issues/1442#issuecomment-633446649

After some investigation (reported in https://github.com/synfig/synfig/issues/1442#issuecomment-633718069 ), the problem is some GTK callback events call some synfigapp actions that may throw exceptions to report
errors that would never be caught except by the main App class and main()
standard function. So, in an user perspective: Synfig Studio crashes.

For this reason, I surrounded all callback contents (60) with a try-catch
statement.

This commit avoids the crash, prints the error (or tries to) and make the
callback return the failure (if the callback can do it), propagating the error without a sudden close of the software.

I also renamed one to follow the convention used by the rest of the code:
onKeyPressed() -> on_key_pressed_event().

Besides, I ensure every callback to be private or protected:
there's no reason to they be public.

Finally, I checked if the callbacks methods were being called by other methods that were not signal calls. Only Workarea::on_drawing_area_event()
is used that way, but inside the same class in the methods: on_hruler_event() and on_vruler_event(). So I think it's safe, because
I made them look to the return value and handle it.

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -136,6 +136,8 @@
 
 #include <gui/localization.h>
 
+#include <gui/exception_guard.h>
+
 #endif
 
 /* === U S I N G =========================================================== */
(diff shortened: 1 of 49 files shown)


rodolforg:
fix LGTM reports and improve Windows exception report

ice0:
Merged. Thank you!

morevnaproject:
I have copied those commits into 1.3.x and set "testing" branch on them. Let's wait when it build binary packages and ask people to test.

Is this a right plan to go? ^_^

morevnaproject:
Build completed:
- https://coolbug.org/downloads/morevna/builds/SynfigStudio-1.3.15-testing-2020.06.09-win64-9a6e0.exe\
- https://coolbug.org/downloads/morevna/builds/SynfigStudio-1.3.15-testing-2020.06.09-win64-9a6e0.zip

PR #1484, opened 2020-06-10
Resource management of CMake build.

AYESDIE:
Fixes the issue in CMake build where app resource path is incorrect.


ice0:
Merged. Thank you!

PR #1485, opened 2020-06-10
Skeleton tool, GSoC 2020

Aa20475:
Implementing the Skeleton tool as a part of Google Summer of Code 2020.

Many more commits to come! :wink:


Aa20475:
> Again, indentation with spaces :(

:broken_heart:  Will fix it in the next commit! :+1:

morevnaproject:
Hi @Aa20475! Did you managed to fix indentation? Please push your new commits here. ^__^

morevnaproject:
@Aa20475 I have tested your code provided here. Here's what I found:

1. **Creating bone by dragging a mouse isn't working, when I create first bone.**

2. **The tool immediately switches back to previous tool after creating first bone.**

Steps to reproduce:

- Make sure you have no skeleton layer selected.
- Select Skeleton Tool
- Click on empty space of workarea.
- New skeleton layer and bone is created, but the tool immediately switches to previous tool.

Video - https://youtu.be/yPX3HGh3dC8

3. **Segfault when adding a second bone.**

Steps to reproduce:

- Make sure you have no skeleton layer selected.
- Select Skeleton Tool
- Click on empty space of workarea.
- New skeleton layer and bone is created,  the tool immediately switches to previous tool.
- Select Skeleton Tool again.
- Click on any place of workarea.
- --> Segfault

Video -  https://youtu.be/InIz8_kOW30

Aa20475:
@morevnaproject  They are fixed in the latest commit.

morevnaproject:
@Aa20475 Please let me know, do latest commits fix any of issues from [my previous comment](https://github.com/synfig/synfig/pull/1485#issuecomment-647293371)?

Aa20475:
Yes! I think all of them are resolved. 😅

morevnaproject:
I have tested latest changes and now it creates a new skeleton layer every time (even when I already have existing skeleton layer selected) - https://youtu.be/huwHZI1qKYU

It shouldn't create new skeleton layer when there is a skeleton layer selected already.

Aa20475:
@morevnaproject Yeah! I didn't actually set up an initial Active Bone. If the active bone is not set, then I thought of adding a new layer. If we are to add a bone with no parent again in the same layer, doesn't that mean there will be multiple roots now? If you want me to set an active bone  initially, I can add that. :sweat_smile:

morevnaproject:
Well, as it is said [in proposal](https://synfig-docs-dev.readthedocs.io/en/latest/projects/skeleton.html):

> Newly created bone is set as a child to the “Active bone”. **The newly created bone will then be set to “Active”.**

So, when I create a new bone it is supposed to be set as active.

> If we are to add a bone with no parent again in the same layer, doesn't that mean there will be multiple roots now?

Yes, if we have skeleton layer selected and no active bone, then I expect that the tool will create a new bone without a parent (on the existing selected layer).

Aa20475:
Okay! Will make the changes!

morevnaproject:
> The main issue I’m facing rn is the Origin fix. I believe we need to add Global origin to bone. Or somehow keep track of everything.

First of all, you need to fix the issue I mentioned in https://github.com/synfig/synfig/pull/1485#issuecomment-648635931

Without this fix I cannot test problem with origin - currently when I try create a bone it is created on new layer and without a parent.

Aa20475:
I also re-based the branch. :sweat_smile:  Also, the bone select is inconsistent too due to the origin thing.

morevnaproject:
I tested latest commit and it fails to build with the following error - 

```
/bin/sh ../../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H   -I../.. -I/home/konstantin/sources/synfig/synfig-core/src   -I/usr/include -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -pthread -I/usr/include/giomm-2.4 -I/usr/lib/x86_64-linux-gnu/giomm-2.4/include -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -I/usr/include/libxml++-2.6 -I/usr/lib/x86_64-linux-gnu/libxml++-2.6/include -I/usr/include/libxml2 -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/pango-1.0 -I/usr/include/harfbuzz -I/usr/include/pango-1.0 -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -D_REENTRANT -I/usr/include/mlt++ -I/usr/include/mlt -I/home/konstantin/sources/synfig/ETL -I/home/konstantin/sources/synfig/_production/build/include/ETL -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -DSYNFIG_NO_DEPRECATED -DLOCALEDIR=\"/home/konstantin/sources/synfig/_production/build/share/locale\" -DLIBDIR="\"/home/konstantin/sources/synfig/_production/build/lib\"" -DSYSCONFDIR="\"/home/konstantin/sources/synfig/_production/build/etc\"" -fdiagnostics-color=always -O2 -DNDEBUG -W -Wall -MT valuenodes/libsynfig_la-valuenode_staticlist.lo -MD -MP -MF valuenodes/.deps/libsynfig_la-valuenode_staticlist.Tpo -c -o valuenodes/libsynfig_la-valuenode_staticlist.lo `test -f 'valuenodes/valuenode_staticlist.cpp' || echo '/home/konstantin/sources/synfig/synfig-core/src/synfig/'`valuenodes/valuenode_staticlist.cpp
libtool: compile:  g++ -DHAVE_CONFIG_H -I../.. -I/home/konstantin/sources/synfig/synfig-core/src -I/usr/include -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -pthread -I/usr/include/giomm-2.4 -I/usr/lib/x86_64-linux-gnu/giomm-2.4/include -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -I/usr/include/libxml++-2.6 -I/usr/lib/x86_64-linux-gnu/libxml++-2.6/include -I/usr/include/libxml2 -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/pango-1.0 -I/usr/include/harfbuzz -I/usr/include/pango-1.0 -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -D_REENTRANT -I/usr/include/mlt++ -I/usr/include/mlt -I/home/konstantin/sources/synfig/ETL -I/home/konstantin/sources/synfig/_production/build/include/ETL -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -DSYNFIG_NO_DEPRECATED -DLOCALEDIR=\"/home/konstantin/sources/synfig/_production/build/share/locale\" -DLIBDIR=\"/home/konstantin/sources/synfig/_production/build/lib\" -DSYSCONFDIR=\"/home/konstantin/sources/synfig/_production/build/etc\" -fdiagnostics-color=always -O2 -DNDEBUG -W -Wall -MT valuenodes/libsynfig_la-valuenode_staticlist.lo -MD -MP -MF valuenodes/.deps/libsynfig_la-valuenode_staticlist.Tpo -c /home/konstantin/sources/synfig/synfig-core/src/synfig/valuenodes/valuenode_staticlist.cpp  -fPIC -DPIC -o valuenodes/.libs/libsynfig_la-valuenode_staticlist.o
/home/konstantin/sources/synfig/synfig-core/src/synfig/valuenodes/valuenode_staticlist.cpp: In member function ‘virtual synfig::ValueNode::Handle synfig::ValueNode_StaticList::clone(synfig::Canvas::LooseHandle, const synfig::GUID&) const’:
/home/konstantin/sources/synfig/synfig-core/src/synfig/valuenodes/valuenode_staticlist.cpp:434:21: error: ‘fix_bones_referenced_by’ is not a member of ‘synfig::ValueNode_Bone’
     ValueNode_Bone::fix_bones_referenced_by(*iter, *ret_iter, false, clone_map);
                     ^~~~~~~~~~~~~~~~~~~~~~~
Makefile:4148: recipe for target 'valuenodes/libsynfig_la-valuenode_staticlist.lo' failed
make[3]: *** [valuenodes/libsynfig_la-valuenode_staticlist.lo] Error 1
```

Aa20475:
I'll do it asap.

Aa20475:
@morevnaproject  Please check the latest commit and let me know if you have any issue. I had a rebase issue there. :sweat_smile:

morevnaproject:
Testing now ^__^

morevnaproject:
It fails again:

```
mv -f .deps/libsynfig_la-type.Tpo .deps/libsynfig_la-type.Plo
/bin/sh ../../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H   -I../.. -I/home/konstantin/sources/synfig/synfig-core/src   -I/usr/include -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -pthread -I/usr/include/giomm-2.4 -I/usr/lib/x86_64-linux-gnu/giomm-2.4/include -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -I/usr/include/libxml++-2.6 -I/usr/lib/x86_64-linux-gnu/libxml++-2.6/include -I/usr/include/libxml2 -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/pango-1.0 -I/usr/include/harfbuzz -I/usr/include/pango-1.0 -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -D_REENTRANT -I/usr/include/mlt++ -I/usr/include/mlt -I/home/konstantin/sources/synfig/ETL -I/home/konstantin/sources/synfig/_production/build/include/ETL -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -DSYNFIG_NO_DEPRECATED -DLOCALEDIR=\"/home/konstantin/sources/synfig/_production/build/share/locale\" -DLIBDIR="\"/home/konstantin/sources/synfig/_production/build/lib\"" -DSYSCONFDIR="\"/home/konstantin/sources/synfig/_production/build/etc\"" -fdiagnostics-color=always -O2 -DNDEBUG -W -Wall -MT libsynfig_la-value.lo -MD -MP -MF .deps/libsynfig_la-value.Tpo -c -o libsynfig_la-value.lo `test -f 'value.cpp' || echo '/home/konstantin/sources/synfig/synfig-core/src/synfig/'`value.cpp
In file included from /home/konstantin/sources/synfig/synfig-core/src/synfig/valuenodes/valuenode_bone.h:32:0,
                 from /home/konstantin/sources/synfig/synfig-core/src/synfig/base_types.cpp:41:
/home/konstantin/sources/synfig/synfig-core/src/synfig/bone.h:82:8: error: redeclaration of ‘synfig::Angle synfig::Bone::angle_’
  Angle angle_;
        ^~~~~~
/home/konstantin/sources/synfig/synfig-core/src/synfig/bone.h:78:8: note: previous declaration ‘synfig::Angle synfig::Bone::angle_’
  Angle angle_;
        ^~~~~~
```

morevnaproject:
I think you shouldn't introduce global origin and global angles now. Those values can be retrieved in other way.
Let's solve [this one problem](https://github.com/synfig/synfig/pull/1485#issuecomment-648635931) first. Do not try to solve several problems at once.

morevnaproject:
Building now...

morevnaproject:
Compiled fine, now it creates bones as expected. Yes, the origin issue remains, but we can solve this tomorrow.

I recommend you to take a look at "Link To Bone" action - it takes a group and links it to selected bone (see lesson 7 of our video training course). This action takes absolute transformation of Group (TG) and transformation of Bone (TB)  as input values. The interesting part is that it calculates offset for transformation (OT) in such way that `TB + OT = TG`.
So, Group remains in place after it is linked to Bone. This might give you some hints on how to correctly calculate origin for new bones.

Aa20475:
>This action takes absolute transformation of Group (TG) and transformation of Bone (TB) as input values. The interesting part is that it calculates offset for transformation (OT) in such way that TB + OT = TG.

I was actually thinking of writing such code xD Thanks! Will definitely check it out!  

>Compiled fine, now it creates bones as expected.

Also, I realized, with whatever choices we are giving to the user, the condition of multiple bone never arises ✌️

Aa20475:
https://github.com/synfig/synfig/blob/13dab0681156014e7c0c84db0e6750c284decfc8/synfig-studio/src/synfigapp/actions/valuedescbonelink.cpp#L216-L223

@morevnaproject  Is this the code you meant?

morevnaproject:
@Aa20475 Yes! ^__^

morevnaproject:
> I tried back tracking the back_transform function functionality. It seems like it is just translating the transform according to the value given to it which is same as changing the coordinate system. In our tool, if we have to do that, we need to iterate across all the bones again right? I’m confused.

Here, in the code you referenced above there is a `get_bone_transformation` function:
screenshot_004

I suggest to check code what this function does.

morevnaproject:
@Aa20475  So please examine the code of `get_bone_transformation` function and write here explanation of what this function does. If you do not understand some parts of this code, then let me know. Looking forward for your reply.

morevnaproject:
> Hey! What’s a setup point?

> https://github.com/synfig/synfig/blob/master/synfig-core/src/synfig/bone.h#L145-L150

> This comment. I guess this `animated_matrix_`  will do the trick to get the global coordinates.

I guess it means "initial setup position" for bone. This is something like "rest pose" concept in  Blender (do not know if you familiar with it?). I.e. each bone internally has "initial placement" and the actual position, which is calculated relatively to "initial position".

But currently in Synfig all bones has same "initial position" - it is the same as non-transformed bone with the following parameters:

screenshot_001

morevnaproject:
> This comment. I guess this animated_matrix_ will do the trick to get the global coordinates.

@Aa20475 You can check that easily.  Just add a code to output values from `animated_matrix_` when bone is created. Then compile and run the code. Create a skeleton and add bones (using regular process, not the skeleton tool) - then you can see the output in console and see if this is what you want or not.

Please do this test and push commit to this branch, so I can see result too.

Also, I am still waiting for explanation of `get_bone_transformation` function. Please post it here.
(It shouldn't be too detailed, just a brief explanation, please ^__^).

Aa20475:
`get_bone_transformation` is basically returning the `animated_matrix_` of the `bone_` Valuenode of `ValueNode_BoneLink` . From what I understand, it is checking for constraints on the bone link and resetting some values in that matrix. Basically, we are adding all the selected items into the `value_desc_list` of `ValueNode_BoneLink` and saving the intial bone link values in `base_value_` in that particular line.

morevnaproject:
That's right. So, now let's see what `animated_matrix_` gives to us. Please make a code changes requested in https://github.com/synfig/synfig/pull/1485#issuecomment-650132058 and examine how `animated_matrix_` works. Also, push your changes here.

morevnaproject:
Great. Now can you print matrix in slightly different way? Just print translation (offset) and rotation defined by this matrix.

Aa20475:
Yes! In my testing I realised, we can do `bone.get_animated_matrix.transform()` in our tool to get actual coordinates! 

Will do the changes you asked and will be on testing that! :D

morevnaproject:
So, now can you answer - is this `animated_matrix_` gives you what you need? ^__^

Aa20475:
Yes! I'll try to figure it out from here and ask you if I face any problem!

synfig-core/src/synfig/layers/layer_skeleton.cpp
```diff
@@ -167,14 +167,14 @@ Layer_Skeleton::sync_vfunc()
  		const Bone &bone = i->get(Bone());
  		Matrix matrix = bone.get_animated_matrix();
  		Vector v;
- 		cout<<i-list.begin()<<endl;
+ 		cout<<bone.get_name()<<endl;
+ 		cout<<"Angle:"<<acos(matrix.axis(0)[0])*180/(2*acos(0.0))<<endl;
```
morevnaproject, line 171: This code doesn't give correct value. For angle of bone = `-45` it shows `45`. And when angle of bone = `45` it displays `45` too.

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -0,0 +1,767 @@
...
+
+	refresh_tool_options();
+	App::dialog_tool_options->present();
+
+	// Turn off layer clicking
+	get_work_area()->set_allow_layer_clicks(false);
+
+	//ducks
+    get_work_area()->set_type_mask(get_work_area()->get_type_mask()-Duck::TYPE_TANGENT-Duck::TYPE_WIDTH);
+    get_canvas_view()->toggle_duck_mask(Duck::TYPE_NONE);
```
morevnaproject, line 397: Please fix indentation here (no spaces).

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -0,0 +1,767 @@
...
+
+Smach::event_result
+StateBone_Context::event_mouse_release_handler(const Smach::event& x)
+{
+	const EventMouse& event(*reinterpret_cast<const EventMouse*>(&x));
+
+	Point releaseOrigin(get_work_area()->snap_point_to_grid(event.pos));
+	synfig::info("Mouse release at "+get_point_string(releaseOrigin));
+
+    Layer::Handle layer = get_canvas_interface()->get_selection_manager()->get_selected_layer();
```
morevnaproject, line 574: Again, please fix indentation in this line and lines below.

Aa20475:
The `animated_matrix_` has first two rows representing the angle and the last row representing the offset translation. 
The angle rows are of form [[cos(a) sin(a) sin(a)][-sin(a) cos(a) cos(a)]] so I had to take an atan.

rodolforg:
@Aa20475 , As you often have issues with indentation, I suggest to use a git pre-commit hook to try to avoid it.

Here is one (I didn't test it though)

https://github.com/jumanjihouse/pre-commit-hooks/blob/master/pre_commit_hooks/forbid-space-in-indent

This long one is for the inverse case: forbid tabbed indentation (but adaptable to space):
https://github.com/martinjoiner/portable-code-pre-commit-hook/blob/master/pre-commit

morevnaproject:
@Aa20475 The last commit reverts all indentation to spaces again...

morevnaproject:
I guess your IDE is configured to use spaces for indentation.
Can you please check that?

Aa20475:
Yes! I actually changed that. I'll recheck. :)

Aa20475:
Can you point out the code where the parameter dock changed signals are processed? :eyes:  :confused:

morevnaproject:
> I’m working on either-
>  1. Set the origin to the click origin and reset the animated_matrix
>  2. Set the animated_matrix offset to click origin and leave the origin as is.
>
> Just letting you know. Please let me know what you think.

Thank you for keeping me updated!

I think you shouldn't do any modification of `animated_matrix`.

You probably need the following:

1. Get coordinates of point (P) where user clicked (coordinates should be in canvas space).
2. Get transformation coordinates (T) of parent bone (i.e. current active bones) - via its `animated_matrix`.
3. Origin of new bone (O) can be found from this formula: P = O*T.
4. So, you need to find reverse transformation of T (let's call it T1).
5. So, O = P*T1

morevnaproject:
> Can you point out the code where the parameter dock changed signals are processed? eyes confused

Can you please clarify for what purpose you need that?

Aa20475:
Yess! I actually thought of doing that. But I  thought of taking the easier way :sweat_smile:  of modifying the `animated_matrix`. :sweat_smile: 

> Can you please clarify for what purpose you need that?

I actually thought of checking out what's happening behind the scenes when we change the origin to get a more clear idea ._.

Aa20475:
> Get transformation coordinates (T) of parent bone (i.e. current active bones) - via its animated_matrix.

Okay! Now I got my mistake. I took the child bone matrix. :grimacing:  Sorry my bad.

morevnaproject:
> I actually thought of checking out what's happening behind the scenes when we change the origin to get a more clear idea

You can get this information directly from bone.

morevnaproject:
@Aa20475 Please try to avoid commits, where you fix a few things at once. It is recommended to have one commit, which  does one purpose. This way it is easier to examine changes in the code.

I am testing your latest commit now. ^__^

Aa20475:
Regarding the next stage of the project i.e, Adding "Make child to Active Bone" to the popup menu. I've inspected the code for `CanvasView::popup_param_menu` and realised that I'll have to make a new `VALUEDESC` action `valuedescchangeparent` and write an `is_candidate` function accordingly.

But I'm not sure on how to provide the active Bone value desc to that action as it's value is actually stored in the stateContext only. I'm currently working on workarounds for this. Any suggestions are welcomed. :slightly_smiling_face:

morevnaproject:
The tool works much better now!
Here are issues what I found:

1. Origin is set correctly if scale of parent bone is default. But if bone is stretched, then origin of new bone is shifted from the point of click.

2. Active bone is not always indicated correctly. Steps to reproduce: 
- Create first bone using Skeleton Tool.
- Modify any point of created bone.
- Create second bone -> it is not highlighted as active.

See video illustrating both problems here - https://youtu.be/9g_fELcy_QQ

morevnaproject:
> Adding "Make child to Active Bone" to the popup menu

After testing of Skeleton Tool for some time I think that we should make this work a bit different.
Instead of "Make child to Active Bone" command I suggest to make "Make Parent to Active Bone".

I came to this from this logic:

**Question:** When user wants this command?
**Answer:** He wants to use this command when he accidentally created some bone from wrong parent.

I.e. the user created a new bone, then he realized "Oh, man, this bone should have other parent!"
Since this bone is already active (every new bone automatically become active bone), he just needs to right-click any other bone and choose "Make Parent to Active Bone" command.
Then he can continue creating the skeleton. ^__^

Aa20475:
>Since this bone is already active (every new bone automatically become active bone), he just needs to right-click any other bone and choose "Make Parent to Active Bone" command.

Woah! That's a very interesting way to see it! Cool! Will do that! @morevnaproject

Aa20475:
I know the latest commit is a bit crude coz I passed Active bone as a valueNode instead of a valuedesc. I'm working on it. I wanted to see the action on the menu :smiley: , so just got it working! Will perfect it! :)

morevnaproject:
@Aa20475 Testing your changes now.

P.S. I suggest to look on how "Link to Bone" action works ^__^ - https://www.youtube.com/watch?v=oPnTrFdgOSE

morevnaproject:
> Added active bone to workarea for highlighting && Perfect origin placement

I confirm: the origin issue is fixed now! Congratulations! ^__^

> 2. Active bone is not always indicated correctly. Steps to reproduce: ...

This is also fixed. ^__^

Here are other issues I found during testing:

1. I am unable to set active bone at some circumstances. Here I am attaching file ([set-active-bone-bug.sifz.zip](https://github.com/synfig/synfig/files/4857180/set-active-bone-bug.sifz.zip)), where I cannot set this bone as active -
screenshot_001
When I click this bone it creates new bone instead.
Video - https://youtu.be/2x51K-oBFRw

2. I have noticed that skeleton tool do not uses actions. Instead, it adds bones directly to Skeleton layer. As result, it is not possible to undo actions. Please change behavior of Skeleton Tool, so it will use actions for creating bones.
Video illustration of the problem - https://youtu.be/dI-g1uOjN_A

Aa20475:
> I confirm: the origin issue is fixed now! Congratulations! ^__^

That's a relief! Yay!

>Here are other issues I found during testing:

Okay! Will start working on them!

morevnaproject:
> I’ve been working on the undo action. The issue I’m facing is that after performing valuedescchildbone action I’m changing the parameters accordingly. So, after an undo and a redo, the bone parameters are getting set to default. For this, I’m planning on writing a new action valuedescchangebone and override the event_undo_handler and specify two undos and similarly the redos.
>
> What’s your thoughts on this?
>
> I personally think modifying valuedescchildbone to accommodate changes in ChildBone. But I’m not sure on how to do it. > Please give some hint on this.


I think it is good idea to modify `valuedesccreatechildbone.cpp`. ^__^

> For this, I’m planning on writing a new action valuedescchangebone and override the event_undo_handler and specify two undos and similarly the redos.

This way will work too, but it can be done simpler:
1. Instead of writing `valuedescchangebone` you can use `Action::ValueDescSet` (`valuedescset.cpp`).
2. It is possible to group several actions together into "composite" action. See `synfigapp::Action::PassiveGrouper` in code of Circle Tool:

https://github.com/synfig/synfig/blob/5862c7611232cb180d8eb2934b17b2f306404f10/synfig-studio/src/gui/states/state_circle.cpp#L848-L853

Still, I think the idea to modify `valuedesccreatechildbone.cpp` is better, but I wanted to mention everything above just to give you some ideas about features of Synfig's action system. ^__^

Aa20475:
>See synfigapp::Action::PassiveGrouper 

I was actually about to propose this xD Cool! 

> I think the idea to modify valuedesccreatechildbone.cpp is better

But then I'll have to modify the `item` parameter in `ValueNodeStaticListInsert` accordingly! Will do that! Thanks!  :D

morevnaproject:
> But then I'll have to modify the `item` parameter in `ValueNodeStaticListInsert` accordingly!

Hmmm... Do you mean `Action::ValueNodeStaticListInsertSmart`?  (it is used in CreateChildBone)
Maybe you can replace `Action::ValueNodeStaticListInsertSmart` to `Action::ValueNodeStaticListInsert`?

The `Action::ValueNodeStaticListInsert` has `item` parameter, so you probaly can prepare bone with particular parameters and pass it as item. ^__^

Aa20475:
>The Action::ValueNodeStaticListInsert has item parameter, so you probaly can prepare bone with particular parameters and pass it as item. ^__^

Yep! Thinking of that! Was trying to figure out the difference between `ValueNodeStaticListInsert` and `ValueNodeStaticListInsertSmart` . Will do! :D

Aa20475:
I have modified `valuedesccreatechildbone` to accomodate some of the child bone parameters. The action works fine without disturbing the older context menu option. 

To modify the active bone, I wrote event handlers with a simple two stack undo-redo system. It is still not perfect but I can't figure why the handlers are not being called. Someone please help! :eyes:

synfig-studio/src/synfigapp/actions/valuedesccreatechildbone.cpp
```diff
@@ -63,7 +64,11 @@ ACTION_SET_CVS_ID(Action::ValueDescCreateChildBone,"$Id$");
 /* === M E T H O D S ======================================================= */
 
 Action::ValueDescCreateChildBone::ValueDescCreateChildBone():
-	time(0)
+	time(0),
+	origin(ValueBase(Point(1.1,0))),
+	scalelx(ValueBase(1.0)),
+	angle(ValueBase(Angle::deg(0.0))),
+	tool(false)
```
morevnaproject, line 71: I guess this line is not needed? ^__^

morevnaproject:
> It is still not perfect but I can't figure why the handlers are not being called.

I need some time to dig into this code. In the meantime, I suggest to switch on fixing other issues.
I.e. this one ^__^ -

> 1. I am unable to set active bone at some circumstances. Here I am attaching file ([set-active-bone-bug.sifz.zip](https://github.com/synfig/synfig/files/4857180/set-active-bone-bug.sifz.zip)), where I cannot set this bone as active -
>
> screenshot_001
>
> When I click this bone it creates new bone instead.
> Video - https://youtu.be/2x51K-oBFRw

rodolforg:
It's really annoying and difficult to review the code with all that non-related changes embed to this PR (like indentation fixes in unrelated files). You should propose these changes in separate PRs.

The git commit history here is very polluted too. Sometimes when you make a mistake in a work-in-progress PR, you should edit previous commit(s) and force a push to the shared PR. There is no need to a separated commit to fix an indentation mistake you did here in this PR. Same for simple mistake you did here 2d967fee63d4af16a7c3ef57eb3857c692c38f59.

Finally, I believe there should not be a commit that yells building errors. It compromises any future investigation (eg. via git bisect) for introduced bugs. Like this one e5e86346a1348f26977d5d887dbfd9f84a29aafc . After you discovered it doesn't compile, you could simply amend the commit and force-push it.

Force-push is really bad in "publicly shared" branches (like master and stable branches). However, for a in-progress PR it is fine.

Please understand: it's okay to make mistakes, but you should fix them in a better way :) If you could rewrite this git history and separate unrelated code changes, you would improve this PR quality in the source code point of view. And make it easier to review/follow.

If you need any help on this history rewriting, I can guide you through it.

Aa20475:
Okayy.. I'm really sorry.. I'll be on it.. 😬😅😨

morevnaproject:
> It's really annoying and difficult to review the code with all that non-related changes embed to this PR (like indentation fixes in unrelated files).

@rodolforg I agree  about that! Thank you for bringing this up! ^__^

But regarding other points:

> you should edit previous commit(s) and force a push to the shared PR

> there should not be a commit that yells building errors

...in my opinion student should avoid force-pushing, even to WIP branch.

This is because having full commit history helps to track student's progress during all coding period. I can watch commit history and see when particular work was done and how student came to particular decisions in code.

Also, I encourage students to push non-working code for following reason:
Students often afraid to push non-working code and if they stuck with solving some thing (which happens often), then  there are no commits from them in a long period. As result, their work becomes non-transparent, it isn't clear if student works or not.
I prefer to have regular (even non-working) commits from student to clearly see his progress.

Also, if student pushes non-working commit there are more chances he can get help with his problem. ^__^

morevnaproject:
So, @Aa20475 here is my request:

Please manually change back (revert) all lines, which do indentation fixes only. And then commit those changes (on top of existing commits).

Aa20475:
As you say @morevnaproject  :wink: :sweat_smile:

rodolforg:
Oh. I hope I didn't sound harsh. Sorry.

Konstantin, I agree with student (and any contributor, like me :) ) pushing non working code. It's all about learning and helping each other.

But I think once the problem is solved, the history should be fixed, for the reason I told. If you want to preserve the learning progress, what about to create a temporary fixing commit right after the trouble one?

And having a commit just for a newly introduced indentation is just... non-sense :)

morevnaproject:
> But I think once the problem is solved, the history should be fixed, for the reason I told.

I agree!
I think we can squash all history into single commit during merge. ^__^

Of course I do not like the strategy of squashing commits for big features - I prefer to keep history of small clean commits, if possible (like you do in your PRs ^__^).
But in this case, as we have a learning student here (and his work is part of GSoC), I feel it is important to keep commit history. For example, it allows to analyze his progress when we take decision whether to pass or fail this student. ^__^

> If you want to preserve the learning progress, what about to create a temporary fixing commit right after the trouble one?

Do you mean rewriting history and inserting a new commit, which is fixing trouble commit?

> And having a commit just for a newly introduced indentation is just... non-sense :)

I agree with that. ^__^

rodolforg:
I mean rewriting to leave the fixing commit near the trouble one, in order to squash them later.

---

I didn't mean to squash all to a single commit on the end of GSoC. I think it's bad and pointless to let the history with commits that don't let code to be builded. Or with those ones created to fix indentations or typos that were just introduced in the PR. :)

morevnaproject:
> I mean rewriting to leave the fixing commit near the trouble one, in order to squash them later.

> I didn't mean to squash all to a single commit on the end of GSoC. I think it's bad and pointless to let the history with commits that don't let code to be builded. Or with those ones created to fix indentations or typos that were just introduced in the PR. :)

I understand now. ^__^

I just see that student is currently struggling with code, so I was afraid that adding one more task related with managing commits will be too much. ^__^ I mean that for person without much experience in managing commits that can be a headache and also student can get his work lost by mistake. Of course, that's all about learning and with proper guidance it is easy to avoid mistakes.

So, I just wanted to let student to concentrate on one thing (code) at the moment. ^__^

And to avoid "junky" commits I planned to squash this PR at the end. I think its main changes will be concentrated in new `state_bone.cpp/.h` file + a few minor changes in other files. So I thought it will be okay in this case.

morevnaproject:
@rodolforg If you feel that it is a good moment to re-write commit history in this PR, then I am perfectly fine about that. ^__^

morevnaproject:
> To modify the active bone, I wrote event handlers with a simple two stack undo-redo system. It is still not perfect but I can't figure why the handlers are not being called. Someone please help! eyes

I have looked at the code.
You have used code for `event_undo_handler` from `state_tool`. 
I tested `state_tool` and found that event handler doesn't work there at all. ^__^

Some other handlers (like `event_mouse_button_down_handler`) are working.

> To modify the active bone, I wrote event handlers with a simple two stack undo-redo system.

I think stack will not work as solution here.

The problem is that in Synfig you can undo actions in random order. I.e. you can undo actions that you did at the very beginning of document creation - just use checkboxes in History Panel - https://wiki.synfig.org/History_Panel

Of course, in many situations this can give weird  results and even lead to crash. But this still a powerful feature if used carefully.

So, I think there should be some other logic of handling active bone during undo.

For example:

- when user makes undo you checking if the bone you removing is your current active bone
- if yes, then (before removing) retrieve its parent bone and make it active.

The logic is simple: if undo operation cancels creation of some bone, then you can always "guess" that active bone is its parent.

morevnaproject:
After thinking a bit more on undo problem:
I suggest to make a separate action for SetActiveBone user operation. This way you will not need undo event handler.

Aa20475:
Okay! That will do I guess! Thankss! :D

Aa20475:
>The problem is that in Synfig you can undo actions in random order. I.e. you can undo actions that you did at the very beginning of document creation - just use checkboxes in History Panel - https://wiki.synfig.org/History_Panel

Woah! Didn't know that! :open_mouth: 

>The logic is simple: if undo operation cancels creation of some bone, then you can always "guess" that active bone is its parent.

Alrightey!

morevnaproject:
> The logic is simple: if undo operation cancels creation of some bone, then you can always "guess" that active bone is its parent.

Consider, that you will not need to do any "guess" if you implement SetActiveBone operation as action.

morevnaproject:
@rodolforg @iceo FYI:
I have pushed current code into separate branch of main Synfig repository. Also, I will submit remaining tasks as separate issues (will be marked with "[Skeleton Tool]" prefix in name).

Aa20475:
Hey! Why not write a virtual `undo` function for `valuedesccreatechildbone`? In this we can modify the active bone highlighting accordingly? I think this will cause issues with the previous skeleton making process but we can add a bool to say if we want to change the active bone?

morevnaproject:
@Aa20475 Hmmm... This can be a good solution! You can try it. ^__^

Aa20475:
Hey!

I cannot acccess the `canvas_view` or anything related to rendering  in the action right?

morevnaproject:
From action you can access `canvas_interface`:

screenshot_004

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -217,7 +229,12 @@ StateBone_Context::load_settings()
 		if(settings.get_value("bone.bone_width",value) && value!="")
 			set_bone_width(Distance(atof(value.c_str()),App::distance_system));
 		else
-			set_bone_width(Distance(1,App::distance_system)); // default width
+			set_bone_width(Distance(6.6,App::distance_system)); // default width
```
rodolforg, line 232: as 6.6 is used more than 2 times, I suggest to make it a constant value that you refer to it when it concerns.

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -245,6 +263,8 @@ void
 StateBone_Context::reset()
 {
 	active_bone = -1;
+	set_id(_("NewSkeleton"));
+	set_bone_width(Distance(6.6,App::distance_system)); // default width
```
rodolforg, line 267: 6.6 :)

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -217,7 +229,12 @@ StateBone_Context::load_settings()
 		if(settings.get_value("bone.bone_width",value) && value!="")
 			set_bone_width(Distance(atof(value.c_str()),App::distance_system));
 		else
-			set_bone_width(Distance(1,App::distance_system)); // default width
+			set_bone_width(Distance(6.6,App::distance_system)); // default width
+
+		if(settings.get_value("bone.tip_width",value) && value!="")
+			set_tip_width(Distance(atof(value.c_str()),App::distance_system));
+		else
+			set_tip_width(Distance(6.6,App::distance_system)); // default width
```
rodolforg, line 237: 6.6 :)

synfig-studio/src/gui/states/state_bone.cpp
rodolforg: Btw, is it right to use `App::distance_system`? If user changed the unit system, wouldn't it change the default width?
Is 6.6 what? `Distance::SYSTEM_UNITS` ? `Distance::SYSTEM_PIXELS`? `Distance::SYSTEM_INCHES`?

`App::distance_system` is the system unit user chose.

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -217,7 +229,12 @@ StateBone_Context::load_settings()
 		if(settings.get_value("bone.bone_width",value) && value!="")
 			set_bone_width(Distance(atof(value.c_str()),App::distance_system));
 		else
-			set_bone_width(Distance(1,App::distance_system)); // default width
+			set_bone_width(Distance(6.6,App::distance_system)); // default width
+
+		if(settings.get_value("bone.tip_width",value) && value!="")
```
rodolforg, line 234: prefer `!value.empty()` rather `value!=""`.
Although it normally has the same behavior, if for any reason synfig or glib changes to `std::wstring`, it wouldn't compile. :)
Yes, I know it not probable, but… :)

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -345,12 +373,18 @@ StateBone_Context::StateBone_Context(CanvasView *canvas_view) :
 	);
 	// 2, default bone width
 	options_table.attach(bone_width_label,
-						 0, 1, 6, 7, Gtk::EXPAND|Gtk::FILL, Gtk::FILL, 0, 0
+						 0, 1, 2, 3, Gtk::EXPAND|Gtk::FILL, Gtk::FILL, 0, 0
```
rodolforg, line 376: Now I noticed that you're using deprecated Gtk::Table. Could you switch to Gtk::Grid?

One example of changing was done here 52eb2c84

synfig-studio/src/gui/workarea.cpp
```diff
@@ -2179,6 +2179,16 @@ WorkArea::set_selected_value_node(etl::loose_handle<synfig::ValueNode> x)
 	}
 }
 
+void
+WorkArea::set_active_bone_value_node(etl::loose_handle<synfig::ValueNode> x)
```
rodolforg, line 2183: Wouldn't be good to check if the valuenode refers to a Bone?

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
```diff
@@ -0,0 +1,160 @@
+/* === S Y N F I G ========================================================= */
+/*!	\file ValueDescMakeParentToActive.cpp
+**	\brief Template File
+**
+**	$Id$
+**
+**	\legal
+**	......... ... 2013 Ivan Mahonin
```
rodolforg, line 8: ? :)

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
```diff
@@ -0,0 +1,160 @@
+/* === S Y N F I G ========================================================= */
+/*!	\file ValueDescMakeParentToActive.cpp
```
rodolforg, line 2: I don't quite like this name:
1. The description makes clear it is related to Bone, mas the name doesn't
2. By the vocab/parameter list, it doesn't set the active bone as parent, but any bone you pass

So maybe it should be named "ValueDescSetParentBone" ?
I'm unsure about "ValueDesc" slice, but I don't recall how actions are named by now.

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
```diff
@@ -0,0 +1,160 @@
...
+bool
+Action::ValueDescMakeParentToActive::is_ready()const
+{
+	if (!value_desc)
+		return false;
+	return Action::CanvasSpecific::is_ready();
+}
+
+void
+Action::ValueDescMakeParentToActive::prepare()
```
rodolforg, line 147: Does it work?

Action::Super  is supposed to be a superset of regular actions. `prepare()` is used to build the action chain.
I don't see you doing nothing here. What am I missing?

Aa20475: Actually this action is still work under progress. I'm currently working on adding support to deformation layer.

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
Aa20475: This adds any bone we pass as the parent to active bone👀 so I guess the description is fine?

synfig-studio/src/gui/workarea.cpp
Aa20475: Yes! I'm working on modifying them. Currently, I'm stuck with Undo system and adding deformation layer 😅 I'll get onto this asap

synfig-studio/src/gui/states/state_bone.cpp
Aa20475: It's system_pixels. Will make the change.

synfig-studio/src/gui/states/state_bone.cpp
Aa20475: Okay! I actually took inspiration from other states. 😅 Will update the code!

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
rodolforg: It's not :) You have to pass which is the "active one". And it should be that way, IMO.

Another issue is the missing word "Bone" in class and action name.

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
rodolforg: Oh okay. But why should it support multiple actions here?
Isn't it a single action?

Aa20475: Yes it is.. I'll make the necessary changes

Aa20475:
In the latest commit, there are many repetitions in code of Skeleton and skeleton deform in `state_bone` and `valuedesccreatechildbone`. I'll work on clearing them! :v:

morevnaproject:
Hi! Building latest commit (8a2f44b) fails with an error for me:
```
/bin/sh ../../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I/home/konstantin/sources/synfig.single/synfig-studio/src/synfigapp -I../..  -I/home/konstantin/sources/synfig.single/synfig-studio/src -I/home/konstantin/sources/synfig.single/synfig-studio/libltdl  -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -pthread -D_REENTRANT -I/home/konstantin/sources/synfig.single/synfig-core/src -I/home/konstantin/sources/synfig.single/ETL -I/home/konstantin/sources/synfig.single/_production/build/include/ETL -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/OpenEXR -I/usr/include/giomm-2.4 -I/usr/lib/x86_64-linux-gnu/giomm-2.4/include -I/usr/include/libxml++-2.6 -I/usr/lib/x86_64-linux-gnu/libxml++-2.6/include -I/usr/include/libxml2 -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/pango-1.0 -I/usr/include/harfbuzz -I/usr/include/pango-1.0 -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/mlt++ -I/usr/include/mlt -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -DLOCALEDIR=\"/home/konstantin/sources/synfig.single/_production/build/share/locale\" -fdiagnostics-color=always -O2 -DNDEBUG -MT actions/libsynfigapp_la-valuenodestaticlistinsertsmart.lo -MD -MP -MF actions/.deps/libsynfigapp_la-valuenodestaticlistinsertsmart.Tpo -c -o actions/libsynfigapp_la-valuenodestaticlistinsertsmart.lo `test -f 'actions/valuenodestaticlistinsertsmart.cpp' || echo '/home/konstantin/sources/synfig.single/synfig-studio/src/synfigapp/'`actions/valuenodestaticlistinsertsmart.cpp
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I/home/konstantin/sources/synfig.single/synfig-studio/src/synfigapp -I../.. -I/home/konstantin/sources/synfig.single/synfig-studio/src -I/home/konstantin/sources/synfig.single/synfig-studio/libltdl -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -fopenmp -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -pthread -D_REENTRANT -I/home/konstantin/sources/synfig.single/synfig-core/src -I/home/konstantin/sources/synfig.single/ETL -I/home/konstantin/sources/synfig.single/_production/build/include/ETL -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/x86_64-linux-gnu//ImageMagick-6 -I/usr/include/ImageMagick-6 -I/usr/include/OpenEXR -I/usr/include/giomm-2.4 -I/usr/lib/x86_64-linux-gnu/giomm-2.4/include -I/usr/include/libxml++-2.6 -I/usr/lib/x86_64-linux-gnu/libxml++-2.6/include -I/usr/include/libxml2 -I/usr/include/glibmm-2.4 -I/usr/lib/x86_64-linux-gnu/glibmm-2.4/include -I/usr/include/pango-1.0 -I/usr/include/harfbuzz -I/usr/include/pango-1.0 -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/mlt++ -I/usr/include/mlt -I/usr/include/sigc++-2.0 -I/usr/lib/x86_64-linux-gnu/sigc++-2.0/include -DLOCALEDIR=\"/home/konstantin/sources/synfig.single/_production/build/share/locale\" -fdiagnostics-color=always -O2 -DNDEBUG -MT actions/libsynfigapp_la-valuenodestaticlistinsertsmart.lo -MD -MP -MF actions/.deps/libsynfigapp_la-valuenodestaticlistinsertsmart.Tpo -c /home/konstantin/sources/synfig.single/synfig-studio/src/synfigapp/actions/valuenodestaticlistinsertsmart.cpp  -fPIC -DPIC -o actions/.libs/libsynfigapp_la-valuenodestaticlistinsertsmart.o
/home/konstantin/sources/synfig.single/synfig-studio/src/synfigapp/actions/valuenodestaticlistinsertsmart.cpp: In static member function ‘static bool synfigapp::Action::ValueNodeStaticListInsertSmart::is_candidate(const synfigapp::Action::ParamList&)’:
/home/konstantin/sources/synfig.single/synfig-studio/src/synfigapp/actions/valuenodestaticlistinsertsmart.cpp:97:6: error: no match for ‘operator!’ (operand type is ‘const synfigapp::Action::ParamList’)
  if (!(get_param_vocab(),x))
      ^~~~~~~~~~~~~~~~~~~~~~
```

I am on gcc 7.3.0, Linux Mint. ^__^

morevnaproject:
Wait... sorry, that was a change I accidentally made in mys source tree. Fixed now. ^__^

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
morevnaproject: I agree with @rodolforg - it is better to rename this action to "ValueDescSetParentBone".
It would be even better to name it "ValueDescBoneSetParent", as we already have "ValueDescBoneLink" (that way the `valuedescbonesetparent.cpp` will be located together with `valuedescbonelink.cpp` in file list.

Following the same logic, it would be nice to rename "valuedesccreatechildbone.cpp/.h"  to "valuedescbonecreatechild.cpp/.h". But this should be done in separate PR. It would be nice to open a separate issue about that. @Aa20475 can you do that? ^__^

And of course, it's better to rename "active_bone" in param_vocab to "parent". ^__^

morevnaproject:
@Aa20475 Can you please address issues mentioned by @rodolforg? ^__^

Aa20475:
Yes sure..

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
Aa20475: Yes will do that! ^___^

Aa20475: >And of course, it's better to rename "active_bone" in param_vocab to "parent". ^__^

Btw, isn't active bone the child here? :eyes:

Aa20475:
Hey! Sorry for the multiple fix commit. Will not repeat it again!

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
Aa20475: Please verify the changes.

Aa20475:
The commit https://github.com/synfig/synfig/pull/1485/commits/3f9e930fffd0ca05fddca6779db47dd148cd920e doesn't work well. Working on fixing it.

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
morevnaproject: > Btw, isn't active bone the child here? :eyes:

You right, my mistake. ^__^

morevnaproject: It looks good now.

synfig-studio/src/gui/states/state_bone.cpp
morevnaproject: This issue is still present in current code.

synfig-studio/src/gui/states/state_bone.cpp
morevnaproject: This issue is still present in current code.

synfig-studio/src/gui/states/state_bone.cpp
morevnaproject: This issue is still present in current code.

synfig-studio/src/gui/states/state_bone.cpp
morevnaproject: This issue is still present in current code.

synfig-studio/src/gui/workarea.cpp
morevnaproject: This issue is still present in current code.

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
morevnaproject: This issue is still present in current code.

Aa20475: What should I exactly change here? :sweat_smile:

synfig-studio/src/gui/states/state_bone.cpp
Aa20475: I cleaned this.

synfig-studio/src/gui/states/state_bone.cpp
Aa20475: I fixed this tooo

synfig-studio/src/gui/states/state_bone.cpp
Aa20475: I decided to have it's `distance_system` fixed to `SYSTEM_UNITS`.

synfig-studio/src/gui/workarea.cpp
Aa20475: Resolved in latest commit.

synfig-studio/src/gui/states/state_bone.cpp
Aa20475: I made the changes here. I will do the tweaking needed to make it look neat.

synfig-studio/src/synfigapp/actions/valuedescmakeparenttoactive.cpp
morevnaproject: You should write your name instead on "Ivan Mahonin", as you are author of this code. ^__^

Aa20475: Ohhh okayy!will do it 😃😃 (I'm exciteddd)

Aa20475: Ohhh okayy!will do it 😃😃

morevnaproject: Looking forward to it! ^__^

Aa20475:
I saw all your suggestions in the issues. I'll fix them asap 😊

morevnaproject:
@Aa20475 I think this PR is ready for final review and merging, starting from commit 10d4510cfaa546800ac1274d2070f6fca25a3e2d.

But we cannot do review/merge process if there will be ongoing work in this branch,

So, can you please  create new branch (i.e. skeleton_tool2) and put all commits after 10d4510cfaa546800ac1274d2070f6fca25a3e2d there.

After that, please reset head of "skeleton_tool" branch to 10d4510cfaa546800ac1274d2070f6fca25a3e2d.

So, we will merge basic functionality of Bone Tool and all future improvements will be made as separate PRs.

Aa20475:
Okay. I'm glad to hear that. I'll do that. Also, if we are going to go into the next Coding period, I assume we'll be working on bone linking?

Aa20475:
Hey! I don't understand why the appveyor build is failing :confused:

morevnaproject:
> I assume we'll be working on bone linking?

No, first all other issues should be solved from this milestone - https://github.com/synfig/synfig/milestone/10
(I have ordered them by priority - most important are at the top). ^__^

morevnaproject:
@Aa20475 This PR is on review now, so please do not push to this branch anymore (unless requested by reviewers).
All further changes should go into "skeleton_tool2" branch.

The only case when we expect new commits here is only when some change requested from reviewers.

@ice0 @rodolforg Does anyone wants to review code of this PR? ^__^

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -0,0 +1,1121 @@
...
+	// Refresh the work area
+	get_work_area()->queue_draw();
+
+	App::dock_toolbox->refresh();
+}
+
+Smach::event_result
+StateBone_Context::event_stop_handler(const Smach::event &x)
+{
+	synfig::info("Skeleton Tool stop handler called");
```
morevnaproject, line 516: I think this can be removed.

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -0,0 +1,1121 @@
...
+	reset();
+	return Smach::RESULT_ACCEPT;
+}
+
+Smach::event_result
+StateBone_Context::event_refresh_handler(const Smach::event& x)
+{
+	synfig::info("Skeleton Tool refresh handler called");
```
morevnaproject, line 524: Please remove that too.

synfig-studio/src/gui/states/state_bone.h
```diff
@@ -0,0 +1,56 @@
+/* === S Y N F I G ========================================================= */
+/*!	\file state_bone.h
+**	\brief Template Header
+**
+**	$Id$
+**
+**	\legal
+**  Copyright (c) 2020 Aditya Abhiram J
```
morevnaproject, line 8: Indentation with spaces.

morevnaproject:
Thank you! Testing now...

PR #1487, opened 2020-06-11
minor improvements in Info panel

rodolforg:
* Replace deprecated Gtk::Table
* Use bold attribute to highlight the labels X, Y, R, G, B
* Make expandable only the vertical separator (position/color)
* Add tooltip for color components to clarify what the value between
  brackets means

### DIFF
--- a/synfig-studio/src/gui/docks/dock_info.cpp
+++ b/synfig-studio/src/gui/docks/dock_info.cpp
@@ -106,42 +106,82 @@ studio::Dock_Info::Dock_Info()
 {
 	set_use_scrolled(false);
 
-	Gtk::Table *table = manage(new Gtk::Table);
+	Gtk::Grid *table = manage(new Gtk::Grid);
+	table->set_column_spacing(3);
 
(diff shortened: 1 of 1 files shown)


rodolforg:
**Before**:
Captura de tela de 2020-06-11 04-11-06
**After**:
Captura de tela de 2020-06-11 04-03-54

FirasH:
Just a thought... is it possible to move the percentace inside the progress bar?

rodolforg:
Directly via GTK, no. It has no settings, AFAIK. The only available settings I see are the text string and a CSS node (that I suppose doesn't allow to move into the progress bar).

For its default theme (Adwaita), GTK renders a thin progress bar. Synfig forces it to a minimum 20px height in `app.cpp`:
`	data += "progress, trough 					{ min-height: 20px; }\n";`

But it would be possible, yes, via Gtk::Overlay, but it wouldn't make sense if progress bar become its standard, thin height.

ice0:
Merged. Thank you!

Issue #1249, opened 2020-03-16
[CMake] Set definitions and compile flags based on CMAKE_BUILD_TYPE


ice0:
Currently if you set CMAKE_BUILD_TYPE you must get different binaries with different splash screens, different About dialog, etc.

So we need to set some defines based on build type.
For example:
```
(for debug builds)
_DEBUG 
SHOW_EXTRA_INFO # Show extra info in the about dialog

compile_options: -fno-inline-functions -fno-inline -g -O0 -Wall
```

Here is a good article explaining some moments.
https://blog.kitware.com/cmake-and-the-default-build-type/

* Also need to set default build type to Debug.
* Need to add CMake message with Build type

ice0:
For Debug build:
```
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=undefined  -fsanitize=address  -fno-omit-frame-pointer")
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined  -fsanitize=address -fno-omit-frame-pointer")
```

AYESDIE:
I had a couple of question.

Where is the bash script for *release* build located?
Do we need to work on all the build types `Debug`, `Release`, `RelWithDebInfo` and `MinSizeRel`, or just the two `Debug` and `Release`.

ice0:
> Where is the bash script for release build located?
Here is the script: https://github.com/synfig/synfig/blob/master/2-build-production.sh

Also you may need to check autotools script, it have logic for chooses different splash screens. Maybe need to set some flag for that.

> Do we need to work on all the build types Debug, Release, RelWithDebInfo and MinSizeRel, or just the two Debug and Release.

Currently we need a possibility to change flags for Debug and Release only,

AYESDIE:
I found the autotools logic for all those, but are there other things that need to be looked at other than splash screen, about section and compiler flags?

ice0:
I think that will be enough. In any case, we can add other options later.

PR #1500, opened 2020-06-16
Different CMake builds based on CMAKE_BUILD_TYPE.

AYESDIE:
Fixes: #1249

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,9 +9,28 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${SYNFIG_BUILD_ROOT}/lib)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${SYNFIG_BUILD_ROOT}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${SYNFIG_BUILD_ROOT}/bin)
 
+if (NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Debug)
+endif()
+
+set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fno-inline-functions -fno-inline -g -O0 -Wall")
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}  -fno-inline-functions -fno-inline -g -O0 -Wall")
+
+# This can be enabled once the address sanitization related errors are resolved.
+#set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fsanitize=undefined  -fsanitize=address -fno-omit-frame-pointer")
+#set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}  -fsanitize=undefined  -fsanitize=address  -fno-omit-frame-pointer")
+
+message("-- CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
+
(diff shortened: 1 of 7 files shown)


synfig-studio/src/gui/dialogs/about.cpp
```diff
@@ -244,7 +244,11 @@ About::About()
 	extra_info += etl::strprintf(_("Built on %s" /* at %s */ "\n\n"), __DATE__ /* , __TIME__ */ );
 
 	extra_info += etl::strprintf(_("Built with:\n"));
-	extra_info += etl::strprintf(_("ETL %s\n"), ETL_VERSION);
+
+	#ifdef ETL_VERSION
```
ice0, line 248: I think it is better to define ETL_VERSION as autotools did.

synfig-studio/images/CMakeLists.txt
```diff
@@ -269,7 +268,40 @@ foreach (IMAGE IN ITEMS ${IMAGES})
...
+# Splash Screen
+if(${CMAKE_BUILD_TYPE} MATCHES Debug)
+    # Get git details for splash_screen_development.sif.in
+    configure_file(
+        ${CMAKE_CURRENT_SOURCE_DIR}/splash_screen_development.sif.in
+        ${CMAKE_CURRENT_BINARY_DIR}/splash_screen_development.sif
+        @ONLY
+    )
+
+    add_custom_command(
```
ice0, line 280: Why not put the image path ($ {CMAKE_CURRENT_BINARY_DIR} /splash_screen_development.sif) in the variable and move the `add_custom_command` out from the condition?

synfig-studio/src/gui/CMakeLists.txt
```diff
@@ -19,6 +19,21 @@ add_definitions(-DSYNFIG_DATADIR="${SYNFIG_DATA_PREFIX}")
 add_definitions(-DLOCALEDIR="${DATA_PREFIX}/share/locale")
 add_definitions(-DCMAKE_BUILD)
 
+if (CMAKE_BUILD_TYPE MATCHES Debug)
+    add_definitions("-DSHOW_EXTRA_INFO")
+
+    execute_process(
```
ice0, line 25: Since this code is not portable and cannot be run under Windows (without MSYS), it is better to manually set `ETL_VERSION="1.5.0"` and move this code to the main CMakeLists.txt

CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.1)
```
ice0, line 1: Which function requires version 3.1?

AYESDIE: We are using version 3.1 as a minimum in `synfig-studio/CMakeLists.txt` and `synfig-core/CMakeLists.txt`.

ice0: Yeah, they need to be fixed too. Added issue here: https://github.com/synfig/synfig/issues/1509

CMakeLists.txt
```diff
@@ -8,9 +9,41 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${SYNFIG_BUILD_ROOT}/lib)
...
+message("-- CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
+
+include(SynfigFindGit)
+
 add_subdirectory(ETL)
 include_directories(ETL)
 include_directories(${PROJECT_BINARY_DIR}/ETL) # to find the generated etl_profile.h
+
+
+execute_process(
```
ice0, line 32: Please remove this part as it is no longer needed.

ice0:
Merged. Thank you!

PR #1501, opened 2020-06-18
Copy plugins, brushes and sounds to the appropriate place in CMake build.

AYESDIE:
Up until now, even though the build didn't require them, files in `plugins`, `brushes` and `sounds` folders weren't being copied to CMake build folder. This should fix that issue.


synfig-studio/brushes/CD_concept/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+    Big_BG.myb
+    Big_BG_prev.png
+    Eraser.myb
+    Eraser_prev.png
+    readme.txt
+)
+
+set(CURRENT_OUTPUT_PATH "share/synfig/brushes/CD_concept")
+
+foreach (BRUSH_FILE IN ITEMS ${BRUSH_CD_CONCEPT_FILES})
```
ice0, line 15: This code looks like a good candidate to make a function from it and use it everywhere to copy/install files.

ice0: Also in this case why not just use a file(COPY...) to copy the entire directory?

synfig-studio/plugins/lottie-exporter/effects/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+set(INSTALL_DESTINATION "share/synfig/plugins/lottie-exporter/effects")
+set(PLUGIN_LOTTIE_EXPORTER_EFFECTS_FILES
+    allmask.py
```
ice0, line 3: Maybe group lottie plugin files to something like:
```
set(PLUGIN_LOTTIE_EXPORTER
    effects/allmask.py
    ...
    helpers/bezier.py
```
and so on?

AYESDIE: Won't the way it's done be more consistent? Even other projects follow something similar where each directory has it's own *CMakeLists.txt* file.

ice0: The main idea is to make the code cleaner and more understandable. But when I look at this code, I can’t understand what it is trying to do. Is there any logic inside subfolders? Does it change some files? Also this code will be difficult to maintain.

One of the reasons to creating makefiles in each directory is just a comfortable way to run the `make` command without going to the parent directory. But here we are not compiling anything, just copying the directory to some place.

And about other projects. Here is a piece of Inkscape code, for example::
https://gitlab.com/inkscape/inkscape/-/blob/master/src/trace/CMakeLists.txt

synfig-studio/plugins/view-unhide-all-layers/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+
+file(COPY ${PLUGIN_VIEW_UNHIDE_ALL_LAYERS_FILES}
+    DESTINATION ${SYNFIG_BUILD_ROOT}/${INSTALL_DESTINATION}
+)
+
+install(FILES ${PLUGIN_LOTTIE_EXPORTER_SYNFIG_FILES}
+    DESTINATION ${INSTALL_DESTINATION}
+)
+
+find_program(INTLTOOL_MERGE intltool-merge)
```
ice0, line 14: Better to make a function for intltool and use it everywhere. 
Also better to add a warning instead of error if intltool not found, so code can be compiled, just without translations.

ice0:
Merged. Thank you!

PR #1503, opened 2020-06-20
Revert "fix sync scrolling of params and timetrack"

morevnaproject:
Reverts commit aeb01c14cafdb409f5a67a865a5ad8b5fa06b0b7.

This applies fix for issue #1442 to master branch.

Actually, we need to check if #1442 takes place on master (for this we have to build packages and test). It is possible the issue is not takes place on master because of new timetrack.

Also, commit aeb01c14cafdb seems to be fixing several things at once. Maybe we should split it into parts?

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -3674,22 +3674,6 @@ CanvasView::set_ext_widget(const String& x, Gtk::Widget* y, bool own)
 		keyframe_tree=dynamic_cast<KeyframeTree*>(y);
 }
 
-AdjustmentGroup::Handle
-CanvasView::get_adjustment_group(const synfig::String& x)
-{
-	AdjustmentGroupBook::const_iterator i = adjustment_group_book_.find(x);
-	return i == adjustment_group_book_.end() ? AdjustmentGroup::Handle() : i->second;
-}
-
-void
-CanvasView::set_adjustment_group(const synfig::String& x, AdjustmentGroup::Handle y)
-{
-	if (y)
-		adjustment_group_book_[x] = y;
-	else
-		adjustment_group_book_.erase(x);
-}
-
 Gtk::UIManager::ui_merge_id
 CanvasView::get_popup_id()
 {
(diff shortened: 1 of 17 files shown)


morevnaproject:
> Actually, we need to check if #1442 takes place on master (for this we have to build packages and test)

@ice0 @Svarov-RZM @FadhilKwan Can you please check this build? - 
[SynfigStudio-1.5.0-testing-2020.06.20-win64-60bf6.zip (108.2 MB)](https://archive.morevnaproject.org/s/mX9Y9Z47CYJGwRK/download) 

Does it crashes the same way as in issue #1442 (when selecting spline)?

Svarov-RZM:
> Does it crashes the same way as in issue #1442 (when selecting spline)?

Yeah, version 1.5.0 also crashes the same way.

rodolforg:
@morevnaproject  so this build is from the master branch without the reverting commit, right?

ice0:
@rodolforg I think yes. I tested on latest master with latest MSYS (it has newer GTK) and I cannot reproduce the problem.  But I think better to test HEAD commit, because this PR #1475 is also needed.

morevnaproject:
> @morevnaproject so this build is from the master branch without the reverting commit, right?

Right.

> Actually, we need to check if #1442 takes place on master

I have posted the build to verify this statement.

@Svarov-RZM Thank you for testing, now we know that the issue persist in master branch too. ^__^

Important thing to note: @ice0 and others reported that they weren't able to reproduce this issue when rebuilding from source (debug and production). My guess is that this crash related with particular version of GTK. 

We use gtk-3.22.12 when building our packages for windows and Linux.

The latest version from 3.22.x branch is 3.22.30.

We can upgrade version of GTK and check. 
But I think we should upgrade after 1.4.0 release only.

rodolforg:
So, should I fix the error of this PR in Dock_Timetrack2?

morevnaproject:
@rodolforg I think there is no need to do that yet. First, we'll try to update GTK version ^__^ -

> We can upgrade version of GTK and check.

ice0:
@morevnaproject 
Does this issue still persist or can I close this PR?

morevnaproject:
The issue has gone since we updated gtk/glib in morevna-builds - https://github.com/morevnaproject/morevna-builds/pull/90

PR #1508, opened 2020-06-22
The long road to use Icon theme

rodolforg:
starting to move to Gtk::IconTheme instead of deprecated IconFactory and StockID

### DIFF
--- a/synfig-studio/images/Makefile.am
+++ b/synfig-studio/images/Makefile.am
@@ -418,7 +418,10 @@ appicondir="${datadir}/icons/hicolor"
 imagedir="${synfig_datadir}/images"
 image_DATA = $(IMAGES)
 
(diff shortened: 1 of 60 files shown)


rodolforg:
I did not modify the CMake files :(

rodolforg:
Just to clarify: I paused my work on this PR until the new stable version be released.
Reason: this PR requires a lot of changes in action handling to get rid of deprecated StockID (as far as I can tell: by creating something similar binding action with presentation info like label text, icon and tooltips).

A lot of work that should be tested right away it's approved by revisers and merged to master, by more users to check if anything is wrong or missing :(

ice0:
Merged via #2632.

P.S. `Widget_InterpolationEnum` is not merged, because I'm not sure if it's needed.

PR #1514, opened 2020-06-24
Edit tooltips for style consistency

FirasH:
Fixed some inconsistent tooltips (caps and form)

### DIFF
--- a/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
+++ b/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
@@ -705,7 +705,7 @@ Layer_Freetype::get_param_vocab(void)const
 
 	ret.push_back(ParamDesc("use_kerning")
 		.set_local_name(_("Kerning"))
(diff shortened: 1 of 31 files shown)


synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
```diff
@@ -705,12 +705,12 @@ Layer_Freetype::get_param_vocab(void)const
 
 	ret.push_back(ParamDesc("use_kerning")
 		.set_local_name(_("Kerning"))
-		.set_description(_("Enables/Disables font kerning (If the font supports it)"))
+		.set_description(_("When checked. enables font kerning (If the font supports it)"))
```
ice0, line 708: Looks like there should be a comma here?

FirasH: Right, nice catch!

ice0:
Merged. Thank you!

PR #1515, opened 2020-06-26
[Lottie Exporter] Added support for Power Convert method

saurabh-ramola:
Added support for Power convert method for Lottie format.
Testing it before I make it ready to be reviewed.


AnishGG:
@saurabh-ramola I tested this PR, and did not get the expected results.
I am attaching an example where the results differ:
Original screenshot: 
Screenshot from 2020-06-28 13-55-22
After lottie export: 
Screenshot from 2020-06-28 13-56-06

As you can see, the rectangle corners should touch the outline layer, but that is not the case.

AnishGG:
Here is the test file: 
[power.zip](https://github.com/synfig/synfig/files/4841910/power.zip)

ice0:
Merged. Thank you!

PR #1516, opened 2020-06-27
[Lottie Exporter] Added support for Vector X and Vector Y convert method

saurabh-ramola:
Added support for Vector X and Vector Y convert method in Lottie exporter plugin.

### DIFF
--- a/synfig-studio/plugins/lottie-exporter/common/Param.py
+++ b/synfig-studio/plugins/lottie-exporter/common/Param.py
@@ -541,6 +541,28 @@ class Param:
                 self.expression = ret
                 return ret, self.expression_controllers
 
+            elif self.param[0].tag == "vectorx":
+                self.subparams["vectorx"].extract_subparams()
+                vector, eff_1 = self.subparams["vectorx"].subparams["vector"].recur_animate("vector")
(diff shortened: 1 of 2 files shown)


synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -541,6 +541,33 @@ def recur_animate(self, anim_type):
...
+                ret = ret.format(x=vector)
+
+                self.expression = ret
+                return ret, self.expression_controllers
+
+            elif self.param[0].tag == "vectory":
+                self.subparams["vectory"].extract_subparams()
+                vector, eff_1 = self.subparams["vectory"].subparams["vector"].recur_animate("vector")
+                self.expression_controllers.extend(eff_1)
+                if self.dimension == 2:
```
AnishGG, line 561: I think we can write this statement as:
`ret = "[{y}[1], {y}[1]]" if self.dimension == 2 else "{y}[1]"`
This is more compact and removes the warnings on lgtm

saurabh-ramola: Changed the statement ^__^

ice0:
Merged. Thank you!

Issue #1402, opened 2020-05-08
CMake must build appdata & desktop files


ice0:
https://github.com/synfig/synfig/pull/1400/files/989c4ea0cfda65a85d040e42875753eb0f820a41#r420665001

AYESDIE:
Closed by #1519.

Issue #1412, opened 2020-05-16
Possible taskbar missing window icon issue


Keyikedalube:
*Previous issue was closed by me because GNOME doesn't have a concept for taskbar like Windows, Cinnamon, KDE, etc Instead it has a dash and a top panel*
*However, I don't have a Windows system with me so referenced this new issue for further investigation.*

The same happens on Windows 10 when I add to the lower toolbar as favourite:
![added_to_toolbar](https://user-images.githubusercontent.com/6705690/77150153-87d76b00-6a93-11ea-8cb1-5a0afe845f1d.png)

when it's not added as favourite, the icon is visible:
![not_added_to_toolbar](https://user-images.githubusercontent.com/6705690/77150316-ec92c580-6a93-11ea-871f-fd12a77f56a4.png)

(The screenshot is made with 1.3.13 but the same happened with 1.3.12)

_Originally posted by @FirasH in https://github.com/synfig/synfig/issues/1063#issuecomment-601601663_

FirasH:
Still the same issue with 1.3.14

Keyikedalube:
Hello @FirasH 
Would you try this test project https://github.com/Keyikedalube/window-icon-test/tree/master and let me know the outcome

I used Gtk::Window `set_icon` method to synfig icon. It's a window icon so I'm guessing the icon will be displayed on MS Windows taskbar.
Try adding it as favorite on your taskbar and let me know if the icon is still displayed
If it does I might be able to solve this issue :)

Below is the step by step procedure to clone, build, and run the test project:
```
git clone https://github.com/Keyikedalube/window-icon-test.git
cd window-icon-test

mkdir build-dir
cd build-dir

cmake ..
make

./window-icon-test
```

FirasH:
I'll try it with mingw-w64.

FirasH:
I get an error when trying to build:

```
MINGW64 ~/window-icon-test/build-dir
$ make
[ 20%] Creating gresources XML file (.gresource.xml)
CMake Error: Error processing file: /home/Postazione\ 2/window-icon-test/GCR_CMake/macros/BuildTargetScript.cmake
make[2]: *** [CMakeFiles/window-icon-test.dir/build.make:89: .gresource.xml] Errore 1
make[1]: *** [CMakeFiles/Makefile2:123: CMakeFiles/window-icon-test.dir/all] Errore 2
make: *** [Makefile:104: all] Errore 2
```

Keyikedalube:
[...]

PR #1519, opened 2020-07-02
CPack: .deb package

AYESDIE:
There are still changes that are to be made to this.

But in the meantime, to test things out, the following commands can be used.
```
mkdir build && cd build
cmake ../
make 
make build_images
make build_pofiles
make plugins_xml
make package
```
Once this is done, there will be a .deb file and a .tar.gz file in the build folder.
image

Once .deb package is installed from either terminal or from UI, you can run Synfig Studio by running this on the command line.
```
synfigstudio
```

To-do:
- [x] Set up version properly.
- [x] Set up dependencies and the dependency version properly.
- [x] Set up an icon.
- [x] Set up a better description rather than *README.md*.
image


AYESDIE:
I added changes for #1402 here, so now it shows on menu. 
image

P.S: Working on fixing that icon issue.

ice0:
The problem with the icon is possible due to an incorrect icon name. May also be related to this https://github.com/synfig/synfig/issues/1412

Snap fixes:
https://github.com/synfig/synfig/blob/97940e960d81b7fd02c070bfd1d1416e5ab96b39/snap/snapcraft.yaml#L120

Flatpak:
https://github.com/flathub/org.synfig.SynfigStudio/blob/1409de8e34f107340aec0dfa15317a7bb3da3759/org.synfig.SynfigStudio.yaml#L220

synfig-studio/images/CMakeLists.txt
```diff
@@ -272,6 +272,10 @@ endforeach()
 # Synfig Splash Screen
 if(${CMAKE_BUILD_TYPE} MATCHES Debug)
     set(SYNFIG_SPLASH_SCREEN ${CMAKE_CURRENT_BINARY_DIR}/splash_screen_development.sif)
+    file(
+        COPY ${CMAKE_CURRENT_SOURCE_DIR}/logo.sif
+        DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
+    )
```
Keyikedalube, line 278: Stumbled on this [StackOverflow answer](https://stackoverflow.com/a/34800667/13138783)
https://github.com/synfig/synfig/pull/1525#issuecomment-653833203

Better and intelligent CMake code to copy:
```cmake
configure_file(logo.sif logo.sif COPYONLY)
```

**input** is automatically mapped to `CMAKE_CURRENT_SOURCE_DIR`
and **output** to `CMAKE_CURRENT_BINARY_DIR`
so no need to explicitly enter the path variables :)

AYESDIE: `configure_file` is used to work with `*.in` files to substitute variable values referenced in the file content. Yes, copying is an alternate use of `configure_file`, however, `file(COPY ...)` preserves timestamps of input file as well. 

[CMake configure_file](https://cmake.org/cmake/help/v3.0/command/configure_file.html)
> Copies a file <input> to file <output> and substitutes variable values referenced in the file content.

[CMake file](https://cmake.org/cmake/help/v3.0/command/file.html#copy)
> Copying preserves input file timestamps, and optimizes out a file if it exists at the destination with the same timestamp.

You can check one use of `configure_file` [here](https://github.com/synfig/synfig/blob/master/synfig-studio/images/CMakeLists.txt#L276).

AYESDIE:
Changes:
- Added a basic CPackConfig.cmake script and PACKAGING.md to generate .deb packages.

- Fixes #1402.

- Updated `SYNFIG_INTLTOOL_MERGE` to account for .desktop.in files.

- Made changes to move icon to the specified location in cmake builds.

- Added dependencies on `synfig_bin` and `synfigstudio` to make building process easier.

So now, building can be done by:
```
mkdir build && cd build
cmake ../
make package -j8
```
This will generate a *.deb* package in the build folder.
So now, no need to build targets like `build_images` or `appdata_xml` on our own.

synfig-studio/images/CMakeLists.txt
Keyikedalube: https://github.com/synfig/synfig/pull/1525#issuecomment-653851114

synfig-studio/CMakeLists.txt
```diff
@@ -15,3 +16,35 @@ add_subdirectory(plugins)
...
+
+SYNFIG_INTLTOOL_MERGE(
+    DESKTOP
+    TARGET_NAME desktop_file
+    INPUT_FILE ${CMAKE_CURRENT_SOURCE_DIR}/org.synfig.SynfigStudio.desktop.in
+    OUTPUT_FILE "org.synfig.SynfigStudio.desktop"
+    INSTALL_DESTINATION "share/applications"
+)
+
+add_custom_command(
```
ice0, line 38: Better to do this part, like Inkscape did:
https://gitlab.com/inkscape/inkscape/-/blob/master/org.inkscape.Inkscape.desktop.template

cmake/SynfigIntltool.cmake
```diff
@@ -1,26 +1,38 @@
 find_program(INTLTOOL_MERGE intltool-merge)
 
 function(SYNFIG_INTLTOOL_MERGE)
```
ice0, line 3: Perhaps it is better to rename it to something with `STUDIO` in the name, so that it indicates that this function can only be used for a studio project?

synfig-studio/CMakeLists.txt
AYESDIE: They used `configure_file` in a way we used it to configure `splash_screen_development.sif.in` where it replaces the `${INKSCAPE_ICONPATH}` with the value of that variable. Making changes to this would also require changes in autotools script, and I'm not too used with scripting in that.

ice0: Agree.

ice0:
Merged. Thank you!

PR #1527, opened 2020-07-05
fix #1523 WidthPoint worngly computed on non-homogenous Advanced Outline

### LINKED ISSUE #1523: WidthPoints are incorrectly interpolated on non-homogenous Advanced Outline
Master branch (4aab7317323adb9aeb7a7e78e1e30a9ca00991e2)
(I have tested with patch from #1518, but the issue also happens without this patch)

**Issue description**:

When Advanced Outline has Homogenous parameter disabled, its thickness changes not corresponding to widthpoints.
This is especially visible when you use flat cusps:

screenshot_001

Here is a video to show the behavior in detail - https://youtu.be/N8nGyQNppIY

rodolforg:


morevnaproject:
It works perfectly now! Thank you! ^__^

PR #1528, opened 2020-07-05
simplify access to default brush resource path

rodolforg:


rodolforg:
I believe that was the last GUI resource left behind XD

ice0:
Merged. Thank you!

PR #1529, opened 2020-07-05
don't re-add the first brush path to brush path list !

rodolforg:


rodolforg:
'Depends' on #1528 

I'll rebase if needed when #1528 is merged

ice0:
Merged. Thank you!

Issue #1517, opened 2020-07-01
Synfig cannot open files with arabic filenames (Windows)


morevnaproject:
**Synfig version & platform**:
1.3.x and master

**Issue description**:
Synfig can open files with Russian filenames (on Russian locale), but fails to open if filename has Arabic characters.

morevnaproject:
I checked - it works fine in GIMP.
Investigating this file - https://github.com/GNOME/gimp/blob/mainline/app/main.c ...

morevnaproject:
This is an interesting piece of code - https://github.com/GNOME/gimp/blob/mainline/app/main.c#L524-L528

Found this - `g_win32_get_command_line ()` - https://developer.gnome.org/glib/stable/glib-Windows-Compatibility-Functions.html#g-win32-get-command-line

PR #1531, opened 2020-07-06
Synfig couldn't open files with arabic filenames (Windows) - #1517

rodolforg:


rodolforg:
I couldn't test it though!
I just followed what @morevnaproject pointed in #1517

morevnaproject:
Hi!
Thank you for the patch! 
I am digging it as well. Just wondering, why not use simply this? ^__^ -

```
#ifdef _WIN32
  argv = g_win32_get_command_line();
#else
  argv = g_strdupv (argv);
#endif
```

rodolforg:
:)

I did that way to release memory on exit.
As code can be interrupted by exceptions and returned immediatly, that was one way to assure the memory release.

morevnaproject:
I see now! Testing it...
Thank you!

morevnaproject:
I have tested PR and for some reason this solution didn't worked - the filename wasn't converted properly.

I have used `argv = g_win32_get_command_line();` with following `g_strfreev(argv);` and it worked.

Still that wasn't enough to fix the issue, as the files were still failed to open (although the filename was decoded properly). So, I made other fixes in this PR - https://github.com/synfig/synfig/pull/1532

rodolforg:
Ouch.
I pushed a new commit to show what would fix this problem, if you want to use it in your PR

morevnaproject:
Will test it! ^__^

morevnaproject:
It works now! ^__^

morevnaproject:
I have added changes to #1532. Thank you!

PR #1532, opened 2020-07-06
Fix opening files with Arabic names

### LINKED ISSUE #1517: Synfig cannot open files with arabic filenames (Windows)

morevnaproject:
Closes #1517.

Previously we used `Glib::locale_to_utf8()` function, and it worked well for simple encodings (i.e. Russian). But for languages which use wide characters (UTF-16) (i.e. Arabic) this didn't worked. So now we use `g_win32_get_command_line()` function instead (https://developer.gnome.org/glib/stable/glib-Windows-Compatibility-Functions.html#g-win32-get-command-line).

But that was not enough. After adding `g_win32_get_command_line()` function the filename was converted properly, but file opening procedure still failing 

After debugging I found that we have to do similar replacement for places where we access filesystem. 
The solution is to use Glib's `g_fopen()` function instead of `fopen()` - https://people.gnome.org/~ryanl/glib-docs/glib-File-Utilities.html#g-fopen.

It is possible we need to use `g_fopen()` (and [similar Glib's wrappers](https://people.gnome.org/~ryanl/glib-docs/glib-File-Utilities.html)) in other places...

### DIFF
--- a/synfig-core/src/modules/mod_bmp/trgt_bmp.cpp
+++ b/synfig-core/src/modules/mod_bmp/trgt_bmp.cpp
@@ -30,6 +30,7 @@
 #	include <config.h>
 #endif
 
+#include <glib/gstdio.h>
 #include "trgt_bmp.h"
 #include <synfig/general.h>
 #include <synfig/localization.h>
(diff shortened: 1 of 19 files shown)


morevnaproject:
It is basically working now. 

Still need to look for places where `Glib::locale_to_utf8()` function used and check where it should be replaced...

ice0:
@rodolforg agree.
But I think we can move all these things to the FileSystem class later, so switching dependencies will be a lot easier.

morevnaproject:
Please don’t merge it yet - I forgot to push one more commit. ^__^

morevnaproject:
Should be ready to merge now. ^__^

ice0:
Merged. Thank you!

morevnaproject:
Woops, I forgot to rebase. Sorry! ^__^"

PR #1533, opened 2020-07-06
Set defaults for Homogenous param of Advanced Outline: on and static

morevnaproject:


rodolforg:
I believe it's more intuitive.

ice0:
Merged. Thank you!

PR #1534, opened 2020-07-06
[CMake] Fixed MacOS build

ice0:
Also now you can use XCode for developing.

Xcode

Generating Xcode project:
```
cmake ../synfig -GXcode
```

if you get error 
```
CMake Error:
Xcode 1.5 not supported.
```
then you need to switch tools:
```
sudo xcode-select --switch /Applications/Xcode.app/
```


morevnaproject:
Blast! ^__^ :tada:

PR #1544, opened 2020-07-08
fix initial value for filename entry of Render Settings dialog

rodolforg:


morevnaproject:
Good catch!

ice0:
Merged. Thank you!

PR #1564, opened 2020-07-18
fix crash when user deletes the sound layer being shown in Sound panel

rodolforg:


ice0:
Merged. Thank you!

rodolforg:
@morevnaproject I suggest this commit to 1.3.x too

PR #1568, opened 2020-07-20
Fix rendering to non-latin filenames via ffmpeg on Windows. Closes #1567.

### LINKED ISSUE #1567: [Windows] Output filename with non-latin characters get corrupted when rendering to ffmpeg target
master branch, 1.3.x branch

**Issue description**:
Output filename with non-latin characters get corrupted when rendering to ffmpeg target on WIndows

morevnaproject:


morevnaproject:
Hmmm... possibly, it would be easier to use `Glib::filename_from_utf8()`? - https://developer.gnome.org/glibmm/stable/group__CharsetConv.html#ga2bd94064ad97b43324a7854b62f0835b

morevnaproject:
I tried to use `Glib::filename_from_utf8()` together with `popen()` and this not works - got  corrupted filenames again.

screenshot_001

So I am leaving the fix as it is for now.

ice0:
> I tried to use Glib::filename_from_utf8() together with popen() and this not works - got corrupted filenames again.

Yes, because it need to use `_wpopen` on windows to work (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/popen-wpopen?view=vs-2019)

ice0:
Yes, because you still need to use `_wpopen` on windows to work (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/popen-wpopen?view=vs-2019)

morevnaproject:
The problem is that `Glib::filename_from_utf8()` returns `std::string`, while for `_wpopen()` we need `std::wstring` (which is converted to `wchar_t`).

ice0:
Merged. Thank you!

PR #1571, opened 2020-07-20
fix uncaught exceptions in destructors

rodolforg:
a destructor shouldn't throw exception anyway:
https://rules.sonarsource.com/cpp/RSPEC-3654

fix Coverity issues CID 308620, 308960, 308965, 308978, 308612, 344421, 344423, 308639


ice0:
Merged. Thank you!

PR #1573, opened 2020-07-21
Dialog windows now return true only if button with id=1 (accept) is clicked

### LINKED ISSUE #1572: Existing files are overwitten if warning dialog is closed
**Synfig version**:
master f4a59e9

**Issue description**:
1. Create a new animation
2. Export to an arbitrary target (e.g. .png or .mp4)
3. Export again to the a target with the same name in order to display the warning dialog about overwriting an existing file
4. Close the dialog (press `Esc`)

Synfig will export the target overwriting the existing files.

It should work like `Use Another Name...`

FirasH:
Fix #1572

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -3614,7 +3614,10 @@ App::dialog_message_2b(const std::string &message,
 	dialog.add_button(button1, 0);
 	dialog.add_button(button2, 1);
 
-	return	dialog.run();
+	if(dialog.run() == 1)
+		return true;
+
+	return false;
 }
 
 


synfig-studio/src/gui/app.cpp
```diff
@@ -3614,7 +3614,10 @@ App::dialog_message_2b(const std::string &message,
 	dialog.add_button(button1, 0);
 	dialog.add_button(button2, 1);
 
-	return	dialog.run();
+	if(dialog.run() == 1)
```
rodolforg, line 3617: Are you sure 1 should be the correct/OK button?

Anyway Gtk::RESPONSE_NONE (= -1) is expected when user closes dialog, I guess

FirasH: In practice it works... but I don't remember where I read it in the documentation.
I think the easiest way is to "catch" the event where a positive response is given (only one case) and should be 1, because there are many cases like the one you reported or Gtk::RESPONSE_DELETE_EVENT.

Anyway I'll check it later and reply here.

FirasH: @rodolforg I checked and it's correct.
The returned id value is defined with the button here as 1:
https://github.com/synfig/synfig/blob/6e3297813bc4f825a9019c837be420857f65cd21/synfig-studio/src/gui/app.cpp#L3615

rodolforg: Nice! Approved then ;)

ice0:
Merged. Thank you!

ice0:
@morevnaproject maybe backport to 1.3.x

Issue #721, opened 2019-02-05
User-friendly rendering dialogue


morevnaproject:
I wish Synfig could have simpler rendering dialogue. Current one is a mess.

Actually, instead of "File"-"Render..." command I would like to have 2 commands:
- "File" -> "Export Animation..." (exports video file - avi, mp4, gif).
- "File" -> "Export Frame" (exports current frame as image).

Both those commands should a dialog similar to "Save file", with a minimum set of necessary options (subject for discussion).

Irelokeh:
I Admire this Idea too and I am wondering if it would supplement to put render setting and configurations in a separate, dedicated dialog preferably in preferences menu  so that it can be configured once and not need to be presented every time the user wants to make a render. 

I will make mockups later today to visualize this.

ThobyV:
I Admire this Idea too and I am wondering if it would supplement to put render setting and configurations in a separate, dedicated dialog preferably in preferences menu  so that it can be configured once and not need to be presented every time the user wants to make a render.

morevnaproject:
@ThobyV In some aspect, this is already like that. Please check "Canvas" -> "Properties" menu command. It will present almost the same dialog, very similar to render settings.

ThobyV:
Finally. I am  back on my computer and ready. I have been away for a while I am sorry for that.

ThobyV:
@morevnaproject  I have checked `render settings`. I will make a wireframe soon to show how I feel the render dialog should control Information and present actions.

morevnaproject:
Looking forward to it! ^__^

b3u:
@morevnaproject I quite like the current setup actually. It's nice and orderly. Is there a particular thing you think is confusing?

morevnaproject:
Hi @b3u ! 
I have outlined my vision about rendering dialog(s) in the first comment  - https://github.com/synfig/synfig/issues/721#issue-406833090 ^__^

FirasH:
I just want to add that currently some settings are not clear for the users.

For example, when I want to render a frame or a sequence, I always have to check the wiki page:
https://wiki.synfig.org/Render_dialog
because there are no explanatory tooltips (Quality/Anti-Aliasing or the "Other" tab that I have no idea what can those options do)

Keyikedalube:
> Both those commands should a dialog similar to "Save file"

That means settings section will have to become global ie., implemented in the Preferences dialog
[...]

Issue #777, opened 2019-03-17
Simplify canvas settings window


morevnaproject:
I am proposing to create a simplified Canvas settings window, which will allow to easily set dimensions of working document. 

At this moment Canvas Properties window looks like this (many options and 3 tabs!):
![screenshot_018](https://user-images.githubusercontent.com/332868/54493291-beb6f600-4900-11e9-8910-c4d7d7c512f4.png)

We need something like this (taken from Flash):

![flash_s](https://user-images.githubusercontent.com/332868/54493285-9c24dd00-4900-11e9-877e-553432f66117.png)

Useful reading on this topic: http://www.segtsy.com/2013/12/a-significant-synfig-setback/

rockyhandsome1:
Do you mean that the entire properties window to be replaced by the documents settings window. If yes then the user will be left with just few options..

morevnaproject:
@rockyhandsome1 My suggestion to create a simple window and add "Advanced..." button there. Clicking it will display the current window with all complicated options.

morevnaproject:
So, I assume we need to cover following cases:

1. **"Canvas" -> "Resize"** - Resizing canvas together with its content.
By default width linked with height to keep aspect ratio. User can unlink those values and in this case canvas expands appropriately.
This is current default behavior of "Canvas->Properties" window with Locks and Links set as follows:
![screenshot_007](https://user-images.githubusercontent.com/332868/55499956-66f2eb80-5671-11e9-92d8-3c1989455085.png)
2. **"Canvas" -> "Set Dimensions"** - Expanding/shrinking canvas (without changing its content size). 
By default width and height do not depend on each other. But user can link them to keep aspect ratio,
This is an equivalent of following settings for Locks and Links in "Canvas->Properties" window:
![screenshot_008](https://user-images.githubusercontent.com/332868/55500205-07491000-5672-11e9-8644-b273aa728498.png)
3. **"File" -> "Render"** - Set size for rendering output. Resizes image together with its content.
By default width linked with height to keep aspect ratio. User can unlink those values and in this case content of canvas is stretched as shown in https://github.com/synfig/synfig/issues/771#issuecomment-478359716
This is an equivalent of following settings for Locks and Links in "Canvas->Properties" window:
![screenshot_009](https://user-images.githubusercontent.com/332868/55500292-38294500-5672-11e9-871d-1f8697b44a33.png)

bajajvinamr:
Is this issue open at the moment and if yes, i would like to work upon it

[...]

PR #1577, opened 2020-07-22
Implemented canvasproperties dialog glade file

### LINKED ISSUE #1552: Implement canvasproperties glade file
Building from master branch
Debian sid

**Issue description**:
To ease UI designers, it's wise to separate the handwritten UI code from canvasproperties.cpp to its own glade (UI) file
And implement only the logic inside the .cpp file

That way it'll be much easier to focus only on the design aspects without having to refer to GTKMM API docs all the time

Keyikedalube:
- Slight improvement to UI alignment

- Use filesystem instead of gresource due to CMake feature issue

- Duplicated `Widget_RendDesc` class to `CanvasProperties` class
    Issue #721 wants a new render layout with minimal options so
    Widget_RendDesc is safe to be deprecated

Fixes #1552


Keyikedalube:
# How to build & test
```bash
mkdir build
cd build
cmake  .. -DCMAKE_INSTALL_PREFIX=$PWD/synfig-root
make all -j4 && make install
export SYNFIG_ROOT=$PWD/synfig-root && env LD_LIBRARY_PATH=$SYNFIG_ROOT/lib/ PATH=$PATH:$SYNFIG_ROOT/bin synfigstudio
```

Keyikedalube:
# Current pending issue
1. Launch synfigstudio build from this PR
2. Modify any values on the dialog content (let's say Image width)
3. Press OK
4. Then undo changes Ctrl+Z
5. Reopen Canvas Properties dialog
6. You'll notice changes are not updated (ie., width should return to previous default value)

Peek 2020-07-22 20-10

The remedy is to close the dialog and reopen it again for your undo to effect the dialog contents

I think it's because of signal events. The dialog doesn't know when the canvas editor is being undo-ed. And I don't know the proper way to connect the undo/redo event so that way when changes take effect on canvas window the dialog should respond too (or specifically synfig::RendDesc)

synfig-studio/ui/synfig-catalog.xml
```diff
@@ -2,14 +2,18 @@
 <glade-catalog name="synfigstudio" depends="gtk+" version="0.2" domain="synfigstudio">
 
   <glade-widget-classes>
+    <glade-widget-class name="gtkmm__CustomObject_widget_gammapattern" title="Gamma Pattern" />
     <glade-widget-class name="gtkmm__CustomObject_widget_vector" generic-name="vector" title="Vector" parent="GtkEntry" toplevel="false" default-width="100" />
     <glade-widget-class name="gtkmm__CustomObject_widget_time" since="0.2" generic-name="time" title="Time" parent="GtkEntry" toplevel="false" default-width="100" />
+    <glade-widget-class name="gtkmm__CustomObject_widget_link" title="Link" />
```
rodolforg, line 8: you should set the "parent" attribute as GtkToggleButton. It would make it appear better on glade ;)

synfig-studio/ui/canvas_properties.glade
```diff
@@ -0,0 +1,1367 @@
...
+    <property name="can_focus">False</property>
+    <property name="stock">synfig-utils_chain_link_off</property>
+    <property name="icon_size">3</property>
+  </object>
+  <object class="GtkBox" id="inner_box">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="orientation">vertical</property>
+    <child>
+      <object class="GtkAlignment">
```
rodolforg, line 88: According to docs: 
```
 gtk_alignment_new has been deprecated since version 3.14 and should not be used in newly-written code.
Use GtkWidget alignment and margin properties
```

synfig-studio/src/gui/widgets/CMakeLists.txt
```diff
@@ -11,7 +11,8 @@ target_sources(synfigstudio
         "${CMAKE_CURRENT_LIST_DIR}/widget_enum.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/widget_sublayer.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/widget_filename.cpp"
-        "${CMAKE_CURRENT_LIST_DIR}/widget_gradient.cpp"
+		"${CMAKE_CURRENT_LIST_DIR}/widget_gammapattern.cpp"
```
rodolforg, line 14: indentation issue

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -228,8 +229,10 @@ MainWindow::init_menus()
 
 void MainWindow::register_custom_widget_types()
 {
+	Widget_GammaPattern::register_type();
```
rodolforg, line 232: Maybe we should start to use some ordering logic here: alphabetically or by "age" (creation?).

What do you think?

synfig-studio/src/gui/CMakeLists.txt
```diff
@@ -28,7 +28,12 @@ endif()
 ## Targets
 ##
 
-add_executable(synfigstudio main.cpp)
+add_executable(synfigstudio main.cpp ${CMAKE_CURRENT_BINARY_DIR}/../../ui/resource.c)
+add_dependencies(synfigstudio build_resource)
+set_source_files_properties(
+	${CMAKE_CURRENT_BINARY_DIR}/../../ui/resource.c
```
rodolforg, line 34: I'm not a CMake guy at all, so please ignore if it is a stupid question: is it the better way to point it to `resource.c`?
And should it be referred twice?

synfig-studio/ui/synfig-catalog.xml
```diff
@@ -2,14 +2,18 @@
 <glade-catalog name="synfigstudio" depends="gtk+" version="0.2" domain="synfigstudio">
 
   <glade-widget-classes>
+    <glade-widget-class name="gtkmm__CustomObject_widget_gammapattern" title="Gamma Pattern" />
```
rodolforg, line 5: please add the "since" attribute :)

synfig-studio/ui/synfig-catalog.xml
```diff
@@ -2,14 +2,18 @@
 <glade-catalog name="synfigstudio" depends="gtk+" version="0.2" domain="synfigstudio">
```
rodolforg, line 2: maybe bump the version?

synfig-studio/ui/synfig-catalog.xml
```diff
@@ -2,14 +2,18 @@
...
   <glade-widget-classes>
+    <glade-widget-class name="gtkmm__CustomObject_widget_gammapattern" title="Gamma Pattern" />
     <glade-widget-class name="gtkmm__CustomObject_widget_vector" generic-name="vector" title="Vector" parent="GtkEntry" toplevel="false" default-width="100" />
     <glade-widget-class name="gtkmm__CustomObject_widget_time" since="0.2" generic-name="time" title="Time" parent="GtkEntry" toplevel="false" default-width="100" />
+    <glade-widget-class name="gtkmm__CustomObject_widget_link" title="Link" />
   </glade-widget-classes>
 
 
   <glade-widget-group name="synfigwidgets" title="Synfig Widgets" >
+    <glade-widget-class-ref name="gtkmm__CustomObject_widget_gammapattern" />
```
rodolforg, line 13: same ordering question I posed earlier

synfig-studio/src/gui/widgets/widget_link.h
```diff
@@ -46,12 +46,23 @@ class Widget_Link: public Gtk::ToggleButton
 	Gtk::Image *icon_off_;
 	Gtk::Image *icon_on_;
 
+	void init();
```
rodolforg, line 49: why two `init()`?
What about the constructors that would call `init()` actually call `init("", "")` instead? :)

synfig-studio/ui/synfig-catalog.xml
Keyikedalube: I implemented Widget_Link as a custom widget but if you open glade file you'll see that I inserted the standard GtkToggleButton. Also, in canvas_properties.cpp file I manually implemented button image and tooltips.

But if you want I can delete mine and instead call the custom widget :) But reimplementing Glib::property for Widget_Link's sake is too minimal don't you think?

synfig-studio/src/gui/widgets/CMakeLists.txt
Keyikedalube: Thanks! Working on fixing that now

synfig-studio/src/gui/mainwindow.cpp
Keyikedalube: I prefer alphabetically :)

Keyikedalube: Making the changes

synfig-studio/src/gui/CMakeLists.txt
Keyikedalube: Yes, resource.c is a generated file
At line 32 synfigstudio depends on build_resource target `add_dependencies(synfigstudio build_resource)`
So before the first arg is build second arg is built first but only after the user executes make command and not cmake

Now CMake doesn't know where resource.c is coming from because it isn't generated yet
That's why we are explicitly telling CMake to not fail when generating build files by turning on resource.c GENERATED file property as true

Keyikedalube: There is an alternative though

We can instead remove that resource.c path from line 31 and also eliminate line 33-35 and instead insert resource.c as a header file inside canvas_properties.h file. The problem would be that IDE will always redline that include file because it doesn't exist in src dir (we're doing out-of-source build). Plus we have to move the generated resource.c file from src/synfig-studio/ui build dir to src/synfig-studio/src/gui/dialogs build dir

Using the above method will make CMake script look clean and readable because TBH the current one is ugly now but effective

rodolforg: I really don't know CMake :( I trust you, I just thought it was strange :)

Maybe @ice0 could opine here

synfig-studio/src/gui/dialogs/canvasproperties.h
```diff
@@ -43,39 +55,137 @@
...
+		~UpdateLock(){locked--;}
+	};
+
 	//Gtk::TreeView* meta_data_tree_view;
 	//void on_button_meta_data_add();
 	//void on_button_meta_data_delete();
 
 public:
-	CanvasProperties(Gtk::Window& parent,etl::handle<synfigapp::CanvasInterface> canvas_interface);
+	CanvasProperties(Gtk::Window &, etl::handle<synfigapp::CanvasInterface>);
```
rodolforg, line 147: no problem at all, but why bother to remove the parameter names here?

Keyikedalube: My habit. I prefer the header file as a design template, so in practice, I only need to know what types the arguments being passed are.

Let's say I use parameter names here too. But when I implement the constructor in .cpp file the IDE doesn't autocomplete those names for me. So I'm typing the same thing twice both for declaration and definition file.

Honestly, it's just my taste. But if you prefer to keep this file consistent with other files, I could bring back those parameter name again. Conforming to project standard is better than having multiple code writing styles from different contributors

synfig-studio/ui/synfig-catalog.xml
Keyikedalube: bumping to 0.3

Also, do I have to bump this from glade file?
```xml
...
<requires lib="synfigstudio" version="0.2"/>
...
```

synfig-studio/ui/synfig-catalog.xml
Keyikedalube: done :)

synfig-studio/src/gui/widgets/widget_link.h
Keyikedalube:
```c++
...
Widget_Link::Widget_Link():
	Glib::ObjectBase("widget_link")
{
	init();
}

Widget_Link::Widget_Link(
		const string &tlt_inactive,
		const string &tlt_active):
	Glib::ObjectBase("widget_link")
{
	init(tlt_inactive, tlt_active);
}
...
```
In default constructor, there are no string parameters. How would it call `init("", "")`..?

synfig-studio/ui/synfig-catalog.xml
Keyikedalube: Adding to _widget_gammapattern and _widget_link...

Also, I notice _widget_vector is missing the since attribute. Intentional?

synfig-studio/ui/synfig-catalog.xml
rodolforg: Do you mean this line?
```xml
 <object class="GtkToggleButton" id="toggle_wh_r"> 
```

Well, you could use 
```
 <object class="gtkmm__CustomObject_widget_link" id="toggle_wh_r"> 
```
This class is now available by you catalog, so it appears in glade widget list. And, if you set the parent attribute in catalog, it would appear as a toggle button.

As Widget_Link can (is?) used elsewhere, programmer should always care about changing the image and tooltip (repeating code…).

You don't have to reimplement Glib::property, just create it. A bit ugly though, but…

Keyikedalube: Yup

Alright

synfig-studio/src/gui/dialogs/canvasproperties.h
rodolforg: I just think it's a good "autodoc"/tip for those who try to use the method :)

synfig-studio/src/gui/widgets/widget_link.h
rodolforg: As far as I can see, `init()` and `init(const string&, const string&)` are equal, except the former doesn't set the tooltips.
So it calls that way, without any tooltips. Am I wrong?

synfig-studio/ui/synfig-catalog.xml
rodolforg: yes, it should. Sorry

synfig-studio/ui/synfig-catalog.xml
rodolforg: It's because it was defined since the beginning :)

rodolforg: Ideally, I'd implemented all existent widgets in catalog at once. But I had other plans/priorities to do… :/

synfig-studio/src/gui/dialogs/canvasproperties.h
Keyikedalube: Alright, bringing back those names :)

synfig-studio/ui/synfig-catalog.xml
Keyikedalube: bumping it...

synfig-studio/ui/synfig-catalog.xml
Keyikedalube: OK, resolved :)

synfig-studio/src/gui/widgets/widget_link.h
Keyikedalube: Yes and it's need for wrap_new

synfig-studio/ui/canvas_properties.glade
Keyikedalube: Strange, I use the verify option in Glade. Didn't report anything about deprecated GtkAlignment object

rodolforg: Strange indeed

synfig-studio/ui/synfig-catalog.xml
Keyikedalube: Check this commit 0e6f409c8e6e6052dcdb452e13d2c09dcb39d69d
Hope I did the Glib::Property right like you asked

synfig-studio/ui/canvas_properties.glade
Keyikedalube: Removed deprecated GtkAlignment from glade
commit 0e6f409c8e6e6052dcdb452e13d2c09dcb39d69d

synfig-studio/ui/synfig-catalog.xml
Keyikedalube: > you should set the "parent" attribute as GtkToggleButton. It would make it appear better on glade ;)

Didn't work

Still got the same output like last time
Screenshot from 2020-07-23 14-40-56

Keyikedalube:
It'd be nice if someone could help me with updating the autotools build script...

synfig-studio/ui/synfig-catalog.xml
rodolforg: Oh I think you didn't XD
You'd replace the tooltips member properties with Glibmm instrospectable ones.

The `property_mine_` from my example would be actually something like `property_tooltip_link_` and `property_tooltip_unlink_` .
In constructor initialization list:
```c++
property_tooltip_link_(*this, "tooltip_link", "Link");
property_tooltip_unlink_(*this, "tooltip_unlink", "Unlink");
```
(tooltip_active and tooltip_inactive as I think it is right now)

then you get (and set) the string via the proxy `property_tooltip_link()`:
```c++
set_tooltip(property_tooltip_link())
```

Keyikedalube: I don't really understand how we are gonna send the custom tooltip_text to wrap_new which takes in the default constructor without arguments because now `tooltip_active` and `tooltip_inactive` strings are empty.

whereas for this previous code that we already had:
https://github.com/synfig/synfig/blob/8b93aa1c2baf37ca687010c617db988632df79fb/synfig-studio/src/gui/renddesc.cpp#L597-L598
We know those strings are custom initialized for each widget

Now in this PR we just have:
https://github.com/Keyikedalube/synfig/blob/d1ebfd83a56d567a20758fdd17cdcfae6185ad69/synfig-studio/src/gui/dialogs/canvasproperties.cpp#L306-L307

That get_widget calls the default constructor without string arguments:
https://github.com/Keyikedalube/synfig/blob/d1ebfd83a56d567a20758fdd17cdcfae6185ad69/synfig-studio/src/gui/widgets/widget_link.cpp#L158-L178

The strings are not custom initialized.

A total noob question but I haven't gotten into practice with deriving custom widgets. This is new.

However, if your ultimate design plan is to set the tooltip manually from Widget_Link implementation class
I think simply setting property tooltip_text inside the glade file would suffice because that works too

BTW, can you show me a sample project similar to this so I can familiarize how this is done?

Keyikedalube: Working on making the changes :)

Also, I forgot that I'm supposed to use get_widget_derived()

A bit of roadblock but all is fine and clear now

Keyikedalube: Made the changes to CanvasProperties like this:
```c++
...
canvas_properties->get_widget_derived("toggle_wh_r" , toggle_wh_r , "Link width and height"  , "Unlink width and height");
canvas_properties->get_widget_derived("toggle_res_r", toggle_res_r, "Link x and y resolution", "Unlink x and y resolution");
...
```

Also corrected the property_link part like this:
```c++
...
Widget_Link::Widget_Link():
	Glib::ObjectBase("widget_link"),
	property_tooltip_active_(*this, "tooltip_active", ""),
	property_tooltip_inactive_(*this, "tooltip_inactive", "")
{
	init();
}

Widget_Link::Widget_Link(const std::string &tlt_inactive, const std::string &tlt_active):
	Glib::ObjectBase("widget_link"),
	property_tooltip_active_(*this, "tooltip_active", ""),
	property_tooltip_inactive_(*this, "tooltip_inactive", "")
{
	init(tlt_inactive, tlt_active);
}

Widget_Link::Widget_Link(BaseObjectType *cobject):
	Glib::ObjectBase("widget_link"),
	Gtk::ToggleButton(cobject),
	property_tooltip_active_(*this, "tooltip_active", ""),
	property_tooltip_inactive_(*this, "tooltip_inactive", "")
{
	init();
}

Widget_Link::Widget_Link(BaseObjectType *cobject, const std::string &tlt_inactive, const std::string &tlt_active):
	Glib::ObjectBase("widget_link"),
	Gtk::ToggleButton(cobject),
	property_tooltip_active_(*this, "tooltip_active", ""),
	property_tooltip_inactive_(*this, "tooltip_inactive", "")
{
	init(tlt_inactive, tlt_active);
}
...
```

Got this error:
Screenshot from 2020-07-24 09-52-31

Deriving widgets is new to me so I'm finding your method a bit complicated :(

I have an alternative solution though:
Manually set the tooltip for toggle_wh_r as "width" and toggle_res_r as "resolution" and during the construction of Wdget_Link do string comparison and ultimately set the Link/Unlink phrase based on get_active() status. Of course, the string (four to be exact) has to be stored as constants inside the Widget_Link class.
Simple and easy

I'm proceeding with my suggestion

rodolforg: It'

> Deriving widgets is new to me so I'm finding your method a bit complicated :(

I can't say I master it too. :/
Take a look at this example I made for you: https://gist.github.com/rodolforg/802222eb41287da744646529615d91e2

Interesting enough, you can inspect and change the widget properties via GtkInspector via Ctrl+Shift+D or Ctrl+Shift+I ;)
(make sure you have gtkinspector installed :P )

Keyikedalube: Custom alternative solution implemented here commit f3bf66278432e7db35dbe8db23d879c8084ebeda

It works

Keyikedalube: Your glib::property stuff is also done in completion. All you have to do is call that `ref_builder->get_custom_widgets(...)` from canvas_properties.cpp file for `toggle_wh_r` and `toggle_res_r`

rodolforg: But did you understand the purpose?
It's about the UI designer to provide the tooltips him(her)self via Glade (file).

However, if you are not comfortable in dig into it, just forget it XD

rodolforg:
By the way, where's synfig.gresource.xml ? :)

synfig-studio/ui/synfig-catalog.xml
Keyikedalube: Still too shady :(

I'm sticking with my implementation
It's already clean and eliminated code duplication from canvas_properties.cpp

You can enhance it in the future after merging

Keyikedalube:
@rodolforg In the `top_srddir/synfig-studio/ui/` directory

rodolforg:
Is it on this PR?

Keyikedalube:
Hohoho..! I missed that file ;P

It's not my fault though because when I do `git status` my synfig.gresource.xml isn't listed

This line is the reason
https://github.com/synfig/synfig/blob/8b93aa1c2baf37ca687010c617db988632df79fb/.gitignore#L16

Pushing the changes...

synfig-studio/src/gui/widgets/widget_link.h
Keyikedalube: This is also related to Glib::Property I think. I won't be fixing this too.

You can enhance it in the future :)

Keyikedalube:
This PR is done
- [X] Working canvas properties dialog
- [x] Successful build using CMake
- [ ] Autotools :(

synfig-studio/src/gui/dialogs/canvasproperties.h
```diff
@@ -27,14 +27,25 @@
...
+#include <gtkmm/builder.h>
+#include <gtkmm/checkbutton.h>
 #include <gtkmm/dialog.h>
-#include <gtkmm/tooltip.h>
-#include <gtkmm/table.h>
 #include <gtkmm/entry.h>
+#include <gtkmm/label.h>
+#include <gtkmm/scale.h>
+#include <gtkmm/spinbutton.h>
+#include <gtkmm/table.h>
```
rodolforg, line 37: not used here with your changes AFAIK (and it's deprecated)

Keyikedalube:
I've tried updating the Autotools script...
Added these lines to `synfig-studio/configure.ac`:
```
GLIB_COMPILE_RESOURCES=`$PKG_CONFIG --variable=glib-compile-resources`
AC_SUBST(GLIB_COMPILE_RESOURCES)
```
And these lines to `synfig-studio/src/gui/Makefile.am`
```make
nodist_synfigstudio_SOURCES = resource.c

synfigstudio.$(OBJEXT): resource.c

synfig_gresource_xml=$(top_srcdir)/ui/synfig.gresource.xml
resource_files=$(shell $(GLIB_COMPILE_RESOURCES) --sourcedir=$(top_srcdir)/ui --generate-dependencies $(synfig_gresource_xml))
resource.c: $(top_srcdir)/ui/synfig.gresource.xml $(resource_files)
	$(shell $(GLIB_COMPILE_RESOURCES) --sourcedir=$(top_srcdir)/ui --generate-source --target=$@ $(synfig_gresource_xml))

CLEANFILES = \
  $(BUILT_SOURCES) \
  $(NULL)
```

File (resource.c) is generated in the build dir but make is trying to build it in advance and I get errors about gio.h no such file or directory.
resource.c is also supposed to be linked with gtkmm libraries. Thought adding as a manual dependency with synfigstudio would help. But make wants to generate object file first before synfigstudio does (which is linked with the required libs)
https://github.com/synfig/synfig/blob/8b93aa1c2baf37ca687010c617db988632df79fb/synfig-studio/src/gui/Makefile.am#L121-L136

This demands more investment and I don't have the time to prioritize learning more. CMake's too darn complicated already ;P

synfig-studio/src/gui/dialogs/canvasproperties.cpp
```diff
@@ -30,20 +30,16 @@
...
-#include "dialogs/canvasproperties.h"
-#include <gtkmm/frame.h>
-#include <gtkmm/table.h>
-#include <gtkmm/label.h>
-#include <gtkmm/alignment.h>
-#include <synfigapp/canvasinterface.h>
-#include "trees/metadatatreestore.h"
-#include <gtkmm/treeview.h>
-#include <gtkmm/scrolledwindow.h>
+//#include <gtkmm/treeview.h>
```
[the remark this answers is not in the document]
Keyikedalube: Check the implementation file .cpp

There's a bunch of commented of code that I left intact about meatreestore(?)

synfig-studio/src/gui/dialogs/canvasproperties.cpp
```diff
@@ -56,87 +52,114 @@ using namespace studio;
...
+		error("BuilderError: " + exception.what());
+	}
 
-	Gtk::Alignment *dialogPadding = manage(new Gtk::Alignment(0, 0, 1, 1));
-	dialogPadding->set_padding(12, 12, 12, 12);
-	get_vbox()->pack_start(*dialogPadding, false, false, 0);
+	canvas_properties->get_widget("inner_box", inner_box);
+	bool expand = true;
+	bool fill   = true;
+	outer_box->pack_start(*inner_box, expand, fill);
```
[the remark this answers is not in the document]
Keyikedalube: It's for outer_box. We got it with get_content_area()
There's no way we can see it on glade

Because our implementation starts from inner_box

synfig-studio/src/gui/dialogs/canvasproperties.h
Keyikedalube: Updating...

Keyikedalube: Table is used by create_meta_data_view()

synfig-studio/src/gui/dialogs/canvasproperties.cpp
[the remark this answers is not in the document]
rodolforg: Isn't this the implementation file?

synfig-studio/src/gui/dialogs/canvasproperties.cpp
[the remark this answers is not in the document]
Keyikedalube: Also, I implemented only the main content for Canvas Properties inside the glade file.
That's why we are packing our inner_box to the content_area of Gtk::Dialog

synfig-studio/src/gui/dialogs/canvasproperties.h
rodolforg: But it shouldn't be in .cpp? It's not used in .h :)

synfig-studio/src/gui/dialogs/canvasproperties.cpp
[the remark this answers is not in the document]
rodolforg: Oh I see now. Sorry.

synfig-studio/src/gui/dialogs/canvasproperties.cpp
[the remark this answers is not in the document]
rodolforg: Ok :)

synfig-studio/src/gui/dialogs/canvasproperties.cpp
```diff
@@ -56,87 +52,114 @@ using namespace studio;
...
-	add_action_widget(*cancel_button,0);
-	cancel_button->signal_clicked().connect(sigc::mem_fun(*this, &studio::CanvasProperties::on_cancel_pressed));
-
-	//set_default_response(1);
-
-	get_vbox()->show_all();
+void
+CanvasProperties::set_rend_desc(const RendDesc &rend_desc)
+{
+	if (update_lock) return;
```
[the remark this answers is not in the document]
Keyikedalube: I think it would

I tried removing update_lock previously when I started out working on this PR
Encountered runtime crash
Didn't want to dig deeper so I kept the working code instead

synfig-studio/src/gui/dialogs/canvasproperties.h
Keyikedalube: Got it. Moving it. :)

rodolforg:
Sorry I delayed too long here with you.

There's something 'bothering' me in this PR: it does two 'unrelated' things: migrate to g_resource and reimplement dialog.
I guess it would be better split into two PRs.
My reasoning: 
- it would be better to have only one way to ui loading/storing in code. Synfig is complicated enough to provide two different ways to do something 'basic' 
- MAYBE it would be better if we change the folder tree structure: move the "ui" folder into "gui" folder. Or move it into gui > resources (new folder) > ui.  Gedit does it this way (https://github.com/GNOME/gedit/tree/master/gedit/resources) – I couldn't find another one who does it this way though lol  So let it as it is.

synfig-studio/src/gui/dialogs/canvasproperties.cpp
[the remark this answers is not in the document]
rodolforg: Ok but I think this check could be placed anyway as it is:
```
if (rend_desc_ == rend_desc)
	return;
```

rodolforg:
Ops. I didn't have the intention to push that! How did it happen???

rodolforg:
>     * it would be better to have only one way to ui loading/storing in code. Synfig is complicated enough to provide two different ways to do something 'basic'

What I mean is: let's migrate the existent ones first or right after this PR merge :)

rodolforg:
Could you please rebase?

synfig-studio/ui/synfig-catalog.xml
rodolforg: Ok :) Sorry I couldn't explain it better to you :(

synfig-studio/src/gui/dialogs/canvasproperties.cpp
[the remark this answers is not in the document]
Keyikedalube: Testing it. Looks legit and efficient too

Keyikedalube:
> resources (new folder) > ui

Hmm... That will make it easier for build scripts to use the relative path of resources from the current source directory instead of recursing backward like I did on CMake

Keep it for future enhancement too. I don't want to touch the autoconfigure and automake files. Might mishandle them :(

synfig-studio/src/gui/dialogs/canvasproperties.cpp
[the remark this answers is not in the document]
Keyikedalube: Semantic issue turns up on IDE
> Invalid operands to binary expression

Still building though. Too many updates so I have 700 files to build... [74/704]

rodolforg:
> I don't want to touch the autoconfigure and automake files.

I did the autotools changes here. I'll add to your PR as soon it's rebased.

synfig-studio/src/gui/dialogs/canvasproperties.cpp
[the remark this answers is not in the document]
rodolforg: Oh. rend_desc doesn't have a comparison operator? :o

Keyikedalube:
Still building [322/704]

Rebase in 20 min...

Sorry, I've got this slow hardware -> 2.0 GHz

rodolforg:
How many cores do you have? :)

Keyikedalube:
2 physical cores and 2 virtual cores

I use j -2 only because I don't want the build to swap out my firefox session 

Shameful hardware for a big project but this is my first computer so no regrets

synfig-studio/src/gui/dialogs/canvasproperties.cpp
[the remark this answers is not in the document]
Keyikedalube: Build failed
Bunch of errors about synfig::RendDesc not derived from std::pair<...>

The other way is to compare individual property members but that'd be too long

Proceeding with rebase

Keyikedalube:
Rebase done :)

Have a great day!

Keyikedalube:
@rodolforg And thanks for looking into this PR. You made it so much better with reviews :+1: :+1: :+1:

rodolforg:
> And thanks for looking into this PR

You're welcome - I'm always afraid I demand too much XD

Great job here :)

So if the checks run fine, we could squash it all, right? Or does leaving those separated commits make sense?

Keyikedalube:
squashing is better

review commits are dirty once they get merged on master branch

rodolforg:
ok. I'll do something here before.

rodolforg:
@Keyikedalube does CMake version rebuild resource.c if the .glade changes?

Keyikedalube:
Yes. After all, generating resource.c is also a target so make is intelligent enough to know when input file ie., canvas_properties.glade changes

I've changed glade design multiple times and noticed the output being regenerated in my ide output pane before :)

rodolforg:
Oh I see you put the .glade file as a dependence explicitly.

Would it be feasible to autodetect the deps in a next PR?
`glib-compile-resources --generate-dependencies synfig.gresource.xml [--sourcedir=   optional] `

---

I miss some translations in the dialog too ("Gamma Correction tab label and Time entry fields)

Keyikedalube:
IDK
I'm thinking of calling another add_custom_command(..) but this time with --generate-dependencies
However, the output part is a bit complicated :(

rodolforg:
Ok, my part here is done. :)

Thank you. Great job! :D

If @ice0  or @morevnaproject approves (specially the change to embed the resource in the app instead of using external files), just squash and merge :)  (a lot of emoticons…)

Next steps (PRs):
- migrate the other 3 glade files to gresource (and then remove the ResourceHelper methods related to UI files)
- provide the glib properties to widget_link
- maybe move ui into a new gui/resources folder

rodolforg:
> I miss some translations in the dialog too ("Gamma Correction tab label and Time entry fields)

You changed some strings (upper/lower cases). I'm ok with them.

Keyikedalube:
> Oh I see you put the .glade file as a dependence explicitly.
> 
> Would it be feasible to autodetect the deps in a next PR?
> `glib-compile-resources --generate-dependencies synfig.gresource.xml [--sourcedir= optional] `

Figured how to achieve that autodetection behavior on CMake but it's hack-y

We have to rename our `synfig.gresource.xml` file to `synfig.gresource.xml.in` as a configuration file for CMake to trigger synfig dependencies and then simply send the output to `add_custom_command(... DEPENDS)`

I could update the script in this PR if you want :)

ice0:
@Keyikedalube that will be awesome!

Keyikedalube:
OK, making the changes ;)

Keyikedalube:
There is a slight problem with synfig CMake. The variable `GRESOURCE_DEPENDENCIES` doesn't seem to be persistent when building multiple times even though the variable is set to be.

Keyikedalube:
For reference, I made a post on [StackOverflow](https://stackoverflow.com/questions/63664809/use-list-as-dependencies-on-add-custom-command)

The sample code runs as expected. Output variable `GRESOURCE_DEPENDENCIES` remains persistent across multiple builds.

Keyikedalube:
> There is a slight problem with synfig CMake. The variable `GRESOURCE_DEPENDENCIES` doesn't seem to be persistent when building multiple times even though the variable is set to be.

Oh wait! The reason could probably be due to old generated build files.

Deleting my old build-dir and rebuilding fresh.

Will post the updates when done :)

synfig-studio/src/gui/CMakeLists.txt
ice0, line 34: Maybe better rename it to `custom_widgets.c`?

Keyikedalube: Name sounds nice but in project structure sense it's actually a resource file even though it's C

Keyikedalube: But since it's generated outside synfig source, I think it's OK to name it that way. Do I proceed with the change?

Keyikedalube:
Something's not right with synfig's CMake. The variable isn't persistent after second build. It was in the first build. Updating .glade file doesn't trigger the build at all.

synfig-studio/src/gui/CMakeLists.txt
ice0: `custom_widgets_resource` maybe?

ice0: Ok, I will look further, maybe I misunderstood the purpose of this file

Keyikedalube: > custom_widgets_resource maybe?

Sounds better. Committing the changes...

ice0:
If I understand correctly `glib-compile-resources` is used to compile `canvas_properties.glade` and add it to synfigstudio executable. And if we change `canvas_properties.glade` we need to recompile the binaries.

If so, why not use `GtkBuilder` to load the `.glade` at runtime? We can then hook into the file change event and reload `.glade` even without restarting Synfig Studio.

Keyikedalube:
> And if we change canvas_properties.glade we need to recompile the binaries.

Yes, however, recompiling the binary is not much performance overhead. It only needs to be re-linked with synfigstudio.

> If so, why not use GtkBuilder to load the .glade at runtime? We can then hook into the file change event and reload .glade even without restarting Synfig Studio.

Ok, there is speed :)

But the purpose of compiling and embedding our resource is for easing maintenance. That way maintainers packaging synfig need to only worry about the libraries and the executables he/she needs but not the program resources. Plus, we don't ship UI updates, right? Mostly bug fixes, performance improvement updates, and feature updates. So I'm in the opinion that embedding program resources with our target executable is a wise step forward.

On the other side... for developers and (especially) designers, hooking with the file change event and experimenting with the glade file is more efficient. But isn't our WYSIWYG Glade program enough? We already know what our UI will look like even before integrating it with our target project. There is no `edit->compile->run->test-UI` process. Design once and if the preview is good enough then proceed with building our target project. And we don't even have to worry our resulting UI output will turn out different from what we previewed.

However, since most of the synfig UIs are loaded directly from the filesystem, I understand if you want to keep that as synfig standard. In case synfig has no plan to migrate to GResource system, it's wise to implement loading `canvas_properties.glade` from the file system instead.

Keyikedalube:
> Something's not right with synfig's CMake

I was wrong.

Long story short, **Ninja** was the issue.

I just changed CMake build generator to **Unix Makefiles** and all good news. Our target triggers the build when glade file changes.

Experimenting with Ninja again to see if a solution can be achieved.

Keyikedalube:
I guess this bug report is relevant: https://cmake.org/Bug/view.php?id=14747
And this answer sounds like our probable solution: https://cmake.org/Bug/view.php?id=14747

Keyikedalube:
Ninja needs an explicit file path for setting resource dependency. Fixed CMake issue. Test is successful; build is now triggered when glade file changes. Pushing the changes...

ice0:
> But the purpose of compiling and embedding our resource is for easing maintenance. That way maintainers packaging synfig need to only worry about the libraries and the executables he/she needs but not the program resources. Plus, we don't ship UI updates, right? Mostly bug fixes, performance improvement updates, and feature updates. So I'm in the opinion that embedding program resources with our target executable is a wise step forward.

Agree. But Synfig still needs translations, icons, sounds and other resources.
Anyway, this behavior can be easily made customizable. I also like the idea that a designer doesn't need to install build tools (and understand how they work), but can improve the user experience and share that improvement with others.

Keyikedalube:
[GResource doc](https://developer.gnome.org/gio/stable/GResource.html)
> Applications and libraries often contain binary or textual data that is really part of the application, rather than user data.

Numix theme, Papirus icons, Materia theme, etc are actually user data so it makes sense to install them in our filesystem. They affect all GTK applications.

Synfig icons and sounds too can be compiled as gresource because they're only needed by synfig itself and not other GTK apps. Translation files I'm not sure because I've never touched it.

Keyikedalube:
> maybe move ui into a new gui/resources folder

Created gui/resources folder for this issue https://github.com/synfig/synfig/issues/1733

This PR will have to wait for the above issue to be fixed first for rebasing purposes.

Keyikedalube:
> There's something 'bothering' me in this PR: it does two 'unrelated' things: migrate to g_resource and reimplement dialog.
I guess it would be better split into two PRs.

Now it does one thing only :)

GResource is delayed because of old CMake version from Travis.

synfig-studio/src/gui/dialogs/canvasproperties.h
```diff
@@ -43,39 +53,131 @@
...
 
+	mutable int update_lock;
+
+	struct UpdateLock
+	{
+		int &locked;
+		UpdateLock(int &locked):locked(locked){locked++;}
+		~UpdateLock(){locked--;}
+	};
+
```
rodolforg, line 132: Are these lines useful?

Keyikedalube: I think so... The last time I removed that portion of code synfig crashed when I modified any value in the canvas properties dialog under Notebook tab. I don't remember exactly how.

I'll go back and see what the behavior is again.

Keyikedalube: > synfig crashed when I modified any value in the canvas properties dialog under Notebook tab.

I was wrong. Synfig crashed when launching.

Screenshot from 2020-10-17 06-22-18

Maybe the lock is to prevent the dialog contents to be set more than once based on what I observed while commenting out those lines in widgets signal call
```c++
void
CanvasProperties::on_widget_changed()
{
    if (update_lock) return;
    UpdateLock lock(update_lock);
    ...
}
```

rodolforg:
Are you sure Widget_RendDesc should be deprecated? It's used in Render dialog too…

Keyikedalube:
> Are you sure Widget_RendDesc should be deprecated? It's used in Render dialog too…

For this #721 I'm thinking @morevnaproject wants the `Gtk::FileChooserDialog` for:
1. Image render
2. Video render

And if my guess is right because to me that dialog is simple and user-friendly... Widget_RendDesc has to go away. There's no way we can attach a gamma pattern or notebook tab in filechooserdialog. It's possible but that's bad design and the dialog is gonna be so big.

rodolforg:
@Keyikedalube  please rebase it to fix the CMakeLists.txt conflict and add the missing License notices to your newly created files:

- synfig-studio/src/gui/widgets/widget_gammapattern.cpp
- synfig-studio/src/gui/widgets/widget_gammapattern.h

Please copy the text from #1793 , as it will be in all file headers. ;)

Keyikedalube:
I think this PR should be closed and instead implement new glade files with this issue #777

Because the current dialog implementation looks too biggish with too many widgets packed together in one dialog window
Screenshot from 2020-10-31 09-33-40

rodolforg:
Maybe it's just a matter of limiting char length for entry widgets?

ice0:
@Keyikedalube I think it might be better to merge this PR and continue on top of these changes?

The only reason why it still not merged - is because I need to check is all whose signals is really required, or we can do without them.
```
	entry_width ->signal_value_changed().connect(sigc::mem_fun(*this, &CanvasProperties::on_width_changed));
	entry_height->signal_value_changed().connect(sigc::mem_fun(*this, &CanvasProperties::on_height_changed));
	entry_xres  ->signal_value_changed().connect(sigc::mem_fun(*this, &CanvasProperties::on_xres_changed));
```

Keyikedalube:
Don't merge yet. I'm still testing rodolforg's suggestion...

Keyikedalube:
> Maybe it's just a matter of limiting char length for entry widgets?

Left width/height untouched because there's no way we can know what user resolution preferences might be.
Could be big or small.

- Not maximized
Screenshot from 2020-11-01 15-02-48

- Maximized
Screenshot from 2020-11-01 15-02-54

Using width chars breaks the UI when dialog is maximized. Look at **X/YRes** and **Physical ...**. The latter sticks to width size specified but not the former. Using `expand` CENTER property is not a good alternative because if a user enters larger number maybe he wants to work at 16K resolution, he might resize the dialog window to make the widget's length bigger. In that case, he'll notice **Physical Width/Height** is broken and report it as UI a bug.

Keyikedalube:
> I think it might be better to merge this PR and continue on top of these changes?

@ice0 I'm thinking merging and working on top of these changes is inefficient now.
- I've implemented CSS in `synfig-studio/src/gui/resources/css` folder
- This PR uses old `synfig-studio/ui` folder

@rodolforg has plans to migrate those ui contents to resources folder like CSS

So if someone contributes to issue #777 he has to migrate ui stuff and strip down my glade files for each dialogs. Too much work.
Instead implement new ones in one go and save more time for all of us :)
And keep git history changes clean.
More efficient, right?

I'm absolutely positive this PR needs to be closed and instead implement new glade files from scratch.

Keyikedalube:
Closed in favor of #777

*Read above comments for reason*

rodolforg:
> > Maybe it's just a matter of limiting char length for entry widgets?
> 
> Left width/height untouched because there's no way we can know what user resolution preferences might be.
> Could be big or small.
> 
>     * Not maximized
>       Screenshot from 2020-11-01 15-02-48
> 
>     * Maximized
>       Screenshot from 2020-11-01 15-02-54
> 
> 
> Using width chars breaks the UI when dialog is maximized. Look at **X/YRes** and **Physical ...**.
> The latter sticks to width size specified but not the former.
> 
> Using `expand` CENTER property is not a good alternative because if a user enters larger number maybe he wants to work at 16K resolution, he might resize the dialog window to make the widget's length bigger.
> In that case, he'll notice **Physical Width/Height** is broken and report it as a UI bug.

Well, we don't need to let the internal grid expands together with window maximization. We don't have to let it be maximized at all :)

Keyikedalube:
@rodolforg I proposed a new re-ordered layout for Image tab over here 👉🏽  https://github.com/synfig/synfig/issues/777#issuecomment-733680841. Also, proposed removal of **Image Area** tab in the next comment and how those widgets values can be handled internally based on Image width and height values so users don't have to get frustrated with more canvas tweaking.

I'd love your suggestions :)

If the new proposal is agreed then it'd be easier for me to rebase this PR and resubmit again under a new branch name LOL.

PR #1582, opened 2020-07-24
CPack NSIS Installer.

AYESDIE:
Provides CPack configuration to make a Windows NSIS Installer.

image

image

image


AYESDIE:
@morevnaproject Are there any Synfig banners that can be used with NSIS installer? I couldn't find one so I made *HEADER* and *WELCOMEFINISH* banners myself. 

HEADER


WELCOMEFINISH

morevnaproject:
There is "installer_logo.sif" file in "synfig-studio/images" dir.
But they are dated from 2007 and ones you made are look better ^__^!

cmake/InstallMSYS2.cmake
```diff
@@ -82,6 +83,7 @@ if(WIN32)
             ERROR_QUIET
             OUTPUT_STRIP_TRAILING_WHITESPACE
         )
+        string(REPLACE "\\" "/" MLT_DIRECTORY "${MLT_DIRECTORY}")
```
ice0, line 86: You can use `cygpath -m <path>` and get Windows paths with slashes, and you won't need to replace them..

ice0:
Merged. Thank you!

Issue #280, opened 2016-11-04
Crash on "rotate order" action on spline


caryoscelus:
```
synfigstudio: actions/valuenodedynamiclistinsert.cpp:204: virtual void synfigapp::Action::ValueNodeDynamicListInsert::perform(): Assertion `list_entry.value_node.rcount()==1' failed.
```

genete:
Can't reproduce on 1.3.0 revision 20170114 on Windows XP 32 bits

caryoscelus:
Can't reproduce crash in latest version either, but rotation doesn't seem to work when attempting to do it from the first element. Rotating from other elements seems to work fine.

morevnaproject:
Rotation always performs _relatively_ to element you clicking on. THis is why you see no change when clicking on first element.

caryoscelus:
Oh well, the crash happens to me again (also tried 1.0.2 - it's present there as well). Should investigate what are the specific conditions of it some time.

As for what it does: does it mean that it places selected element first in the list? If so, i think it should be clarified in the action name.

morevnaproject:
> As for what it does: does it mean that it places selected element first in the list? 

It seems so.

>If so, i think it should be clarified in the action name.

Agree.

rodolforg:
This problem was introduce by commit 790e124bea30a2e4b23166559c9a7582760eddda.

https://github.com/synfig/synfig/blob/790e124bea30a2e4b23166559c9a7582760eddda/synfig-studio/src/synfigapp/actions/valuenodedynamiclistinsert.cpp#L204

As I don't understand `rcount` ("number of reversible instances" [??] of `rhandle` "Replaceable Object Handle" [???]), I don't know if it should be always equal to 1 as  `assert` requires. If it should be simply greater than 1, code doesn't crash and appears to work fine.

rodolforg:
@blackwarthog could you please review this assert?

blackwarthog:
> As I don't understand rcount ("number of reversible instances" [??] of rhandle "Replaceable Object Handle" [???]), I don't know if it should be always equal to 1 as assert requires. If it should be simply greater than 1, code doesn't crash and appears to work fine via context menu in work area.

etl::handle - is a pointer with ref-count, like Glib::RefPtr. Contained object must be subclass of etl::shared_object and has methods ref(), unref() and count()

[...]

PR #1583, opened 2020-07-25
don't crash debug version when rotating vertex order

rodolforg:
As stated here https://github.com/synfig/synfig/issues/280#issuecomment-591782338 , I don't know if I can just remove the assert,
so let do my work without crash :)


ice0:
Merged. Thank you!

PR #1585, opened 2020-07-25
Add button to stop rendering process

### LINKED ISSUE #1366: Allow user to stop the Rendering process in Synfig Studio
Master: ccf39ec

**Issue description**:
1. Open any animation project
2. Go to: `File > Render...` and start to render your animation to the output target

Once the rendering process is started, it is not possible to stop it from Synfig Studio.

It would be very useful to have a button somewhere in the `Info` Panel near to the progress percentage bar to stop it:
Stop_rendering_button

This feature would be used for example in case a rendering is started by mistake or with wrong options.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91175541-allow-user-to-stop-the-rendering-process-in-synfig-studio?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

FirasH:
Fix #1366

### DIFF
--- a/synfig-studio/src/gui/asyncrenderer.cpp
+++ b/synfig-studio/src/gui/asyncrenderer.cpp
@@ -723,6 +723,7 @@ public:
 
 AsyncRenderer::AsyncRenderer(etl::handle<synfig::Target> target_,synfig::ProgressCallback *cb):
 	status(RENDERING_UNDEFINED),
+	interaction(UNDEFINED),
 	cb(cb),
 	start_clock(0),
 	finish_clock(0),
@@ -800,23 +801,45 @@ AsyncRenderer::stop()
 			//Glib::MainContext::get_default()->iteration(false);
 
 			std::string error_message;
-			if(status == RENDERING_SUCCESS)
-				signal_success_();
+
+			if(interaction == UNDEFINED)
+			{
+				if(status == RENDERING_SUCCESS)
+					signal_success_();
+				else
+					error_message = _("Animation couldn't be rendered");
+
+				target=0;
+				render_thread=0;
+				lock.release();
+
+				if(status == RENDERING_ERROR)
+				{
+					if(ProgressLogger *logger = dynamic_cast<ProgressLogger*>(cb))
+						error_message += "\n" + logger->get_error_message();
+				}
+			}
 			else
-				error_message = _("Animation couldn't be rendered");
+			{
+				target=0;
+				render_thread=0;
+				lock.release();
 
-			target=0;
-			render_thread=0;
-
-			lock.release();
-
-			if (status == RENDERING_ERROR) {
-				if (ProgressLogger *logger = dynamic_cast<ProgressLogger*>(cb))
-					error_message += "\n" + logger->get_error_message();
+				error_message = _("Rendering stopped by user");
 			}
 
 			signal_finished_(error_message);
 		}
+	}
+}
+
+void
+AsyncRenderer::user_stop()
+{
+	if(target)
+	{
+		interaction = BY_USER;
+		stop();
 	}
 }
 
--- a/synfig-studio/src/gui/asyncrenderer.cpp
+++ b/synfig-studio/src/gui/asyncrenderer.cpp
@@ -723,7 +723,6 @@ public:
 
 AsyncRenderer::AsyncRenderer(etl::handle<synfig::Target> target_,synfig::ProgressCallback *cb):
 	status(RENDERING_UNDEFINED),
-	interaction(UNDEFINED),
 	cb(cb),
 	start_clock(0),
 	finish_clock(0),
@@ -779,7 +778,7 @@ AsyncRenderer::~AsyncRenderer()
 }
 
 void
-AsyncRenderer::stop()
+AsyncRenderer::stop(AsyncRenderer::Interaction interaction)
 {
 	if(target)
 	{
@@ -802,7 +801,7 @@ AsyncRenderer::stop()
 
 			std::string error_message;
 
-			if(interaction == UNDEFINED)
+			if(interaction == INTERACTION_UNDEFINED)
 			{
 				if(status == RENDERING_SUCCESS)
 					signal_success_();
@@ -828,16 +827,6 @@ AsyncRenderer::stop()
 
 			signal_finished_(error_message);
 		}
-	}
-}
-
-void
-AsyncRenderer::user_stop()
-{
-	if(target)
-	{
-		interaction = BY_USER;
-		stop();
 	}
 }
 
@@ -926,7 +915,9 @@ AsyncRenderer::render_target()
 #ifdef GLIB_DISPATCHER_BROKEN
 		done_connection=Glib::signal_timeout().connect(
 			sigc::bind_return(
-				mem_fun(*this,&AsyncRenderer::stop),
+				sigc::bind(mem_fun(*this,&AsyncRenderer::stop),
+						 AsyncRenderer::INTERACTION_UNDEFINED
+				),
 				false
 			)
 			,0
--- a/synfig-studio/src/gui/asyncrenderer.cpp
+++ b/synfig-studio/src/gui/asyncrenderer.cpp
@@ -824,8 +824,6 @@ AsyncRenderer::stop()
 				target=0;
 				render_thread=0;
 				lock.release();
-
-				error_message = _("Rendering stopped by user");
 			}
 
 			signal_finished_(error_message);
--- a/synfig-studio/src/gui/asyncrenderer.h
+++ b/synfig-studio/src/gui/asyncrenderer.h
@@ -60,10 +60,10 @@ public:
 		RENDERING_SUCCESS
 	};
 
-        enum Interaction {
-            UNDEFINED,
-            BY_USER
-        };
+	enum Interaction {
+		INTERACTION_UNDEFINED,
+		INTERACTION_BY_USER
+	};
 
 private:
 	//! Signal emitted when target has been stopped or has finished
@@ -82,9 +82,6 @@ private:
 	//! Current rendering status
 	//! Undefined if not finished
 	Status status;
-        
-	//! Handles user interaction on the rendering process
-	Interaction interaction;
 
 	synfig::ProgressCallback *cb;
 	//! Signal to be emitted when the target is requested to stop
@@ -109,8 +106,7 @@ public:
 	virtual ~AsyncRenderer();
 
 	void start();
-	void stop();
-	void user_stop();
+	void stop(Interaction interaction = INTERACTION_UNDEFINED);
 	void pause();
 	void resume();
 
--- a/synfig-studio/src/gui/asyncrenderer.h
+++ b/synfig-studio/src/gui/asyncrenderer.h
@@ -60,6 +60,11 @@ public:
 		RENDERING_SUCCESS
 	};
 
+        enum Interaction {
+            UNDEFINED,
+            BY_USER
+        };
+
 private:
 	//! Signal emitted when target has been stopped or has finished
 	//! An error message is passed as argument in case of error;
@@ -77,6 +82,9 @@ private:
 	//! Current rendering status
 	//! Undefined if not finished
 	Status status;
+        
+	//! Handles user interaction on the rendering process
+	Interaction interaction;
 
 	synfig::ProgressCallback *cb;
 	//! Signal to be emitted when the target is requested to stop
@@ -102,6 +110,7 @@ public:
 
 	void start();
 	void stop();
+	void user_stop();
 	void pause();
 	void resume();
 
--- a/synfig-studio/src/gui/docks/dock_info.cpp
+++ b/synfig-studio/src/gui/docks/dock_info.cpp
@@ -179,7 +179,7 @@ studio::Dock_Info::Dock_Info()
 
 	Gtk::Button *stop_button = manage(new Gtk::Button());
 	stop_button->set_label("Stop");
-	stop_button->signal_clicked().connect(sigc::mem_fun(*this, &studio::Dock_Info::stop_renderer));
+	stop_button->signal_clicked().connect(sigc::mem_fun(*this, &studio::Dock_Info::on_stop_button_clicked));
 
 	table->attach_next_to(*stop_button, render_progress, Gtk::POS_RIGHT, 1, 1);
 
@@ -203,7 +203,7 @@ studio::Dock_Info::~Dock_Info()
 {
 }
 
-void studio::Dock_Info::stop_renderer()
+void studio::Dock_Info::on_stop_button_clicked()
 {
 	async_renderer->stop();
 }
@@ -218,9 +218,9 @@ void studio::Dock_Info::changed_canvas_view_vfunc(etl::loose_handle<CanvasView> 
 	}
 }
 
-void studio::Dock_Info::set_async_render(etl::handle<studio::AsyncRenderer> *ar)
+void studio::Dock_Info::set_async_render(etl::handle<studio::AsyncRenderer> ar)
 {
-	async_renderer = *ar;
+	async_renderer = ar;
 }
 
 void studio::Dock_Info::set_n_passes_requested(int value)
--- a/synfig-studio/src/gui/docks/dock_info.cpp
+++ b/synfig-studio/src/gui/docks/dock_info.cpp
@@ -105,7 +105,6 @@ studio::Dock_Info::Dock_Info()
 :Dock_CanvasSpecific("info",_("Info"),Gtk::StockID("synfig-info"))
 {
 	set_use_scrolled(false);
-	enable_stop_button(false);
 
 	Gtk::Grid *table = manage(new Gtk::Grid);
 	table->set_column_spacing(3);
@@ -205,18 +204,16 @@ studio::Dock_Info::~Dock_Info()
 
 void studio::Dock_Info::on_stop_button_clicked()
 {
-	if(async_renderer && App::dialog_message_2b(
+	if(!async_renderer)
+            return;
+
+	if(App::dialog_message_2b(
 		_("Stop rendering"),
 		_("The rendering process will be stopped. Are you sure?"),
 		Gtk::MESSAGE_QUESTION,
 		_("Cancel"),
 		_("Stop")))
 		async_renderer->stop();
-}
-
-void studio::Dock_Info::enable_stop_button(bool status)
-{
-	stop_button.set_sensitive(status);
 }
 
 void studio::Dock_Info::changed_canvas_view_vfunc(etl::loose_handle<CanvasView> canvas_view)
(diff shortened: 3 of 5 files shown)


FirasH:
@rodolforg what do you think about this initial commit?

It's still needed to add a `RENDERING_STOPPED` Status and handle warning messages etc.

synfig-studio/src/gui/docks/dock_info.cpp
```diff
@@ -177,6 +177,12 @@ studio::Dock_Info::Dock_Info()
 	render_progress.set_text(strprintf("%.1f%%", 0.0));
 	render_progress.set_fraction(0.0);
 
+	Gtk::Button *stop_button = manage(new Gtk::Button());
+	stop_button->set_label("Stop");
+	stop_button->signal_clicked().connect(sigc::mem_fun(*this, &studio::Dock_Info::stop_renderer));
```
rodolforg, line 182: Particularly I prefer the gtk naming convention for callbacks: on_stop_button_clicked. But just my taste.

synfig-studio/src/gui/docks/dock_info.h
```diff
@@ -63,6 +67,7 @@ class Dock_Info : public Dock_CanvasSpecific
 
 	//! Current render progress - 0.0 to 1.0
 	//  depends on n_passes_requested and current_pass
+	void set_async_render(etl::handle<AsyncRenderer> *ar);
```
rodolforg, line 70: you don't have to pass it as pointer (of pointer). Pass by value or reference.

synfig-studio/src/gui/docks/dock_info.cpp
FirasH: Done, better to keep things consistent

synfig-studio/src/gui/docks/dock_info.h
FirasH: Yup!

synfig-studio/src/gui/docks/dock_info.cpp
```diff
@@ -197,6 +203,11 @@ studio::Dock_Info::~Dock_Info()
 {
 }
 
+void studio::Dock_Info::on_stop_button_clicked()
+{
+	async_renderer->stop();
```
rodolforg, line 208: Two things related to stop button:
- check if async_renderer is valid before defer it (`if (async_renderer)`)
- it should be sensitive (= enabled) only when something is being rendered :)

FirasH: Committed the changes!
I've moved the button under the progress bar to keep the symmetry of the dialog:
button_placement

synfig-studio/src/gui/docks/dock_info.cpp
```diff
@@ -197,6 +203,22 @@ studio::Dock_Info::~Dock_Info()
 {
 }
 
+void studio::Dock_Info::on_stop_button_clicked()
+{
+	if(async_renderer && App::dialog_message_2b(
```
rodolforg, line 208: Please improve the readability, by spliting this if statement.

```c++
if (!async_renderer)
return;

bool ok = App::dialog......
```

synfig-studio/src/gui/render.cpp
```diff
@@ -485,6 +485,7 @@ RenderSettings::submit_next_render_pass()
 
 		App::dock_info_->set_n_passes_pending(render_passes.size()); //! Decrease until 0
 		App::dock_info_->set_render_progress(0.0); //For this pass
+		App::dock_info_->enable_stop_button(true);
```
rodolforg, line 488: Let the Dock_Info do its job by itself here.
Update the button sensitivity status inside the set_render_progress (if any value below 1.0, disable it) and set_async_render (if invalid, disable it)

synfig-studio/src/gui/render.cpp
```diff
@@ -571,6 +554,7 @@ RenderSettings::on_finished(std::string error_message)
 			App::sound_render_done->set_playing(true);
 		}
 		App::dock_info_->set_render_progress(1.0);
+		App::dock_info_->enable_stop_button(false);
```
rodolforg, line 557: Same as said before: let dockinfo handle button status

FirasH:
At this point the button is working consistently, but still there are some warnings displayed when rendering is stopped:
warning_dialog

Need to change `stop()` method now:
https://github.com/synfig/synfig/blob/8b93aa1c2baf37ca687010c617db988632df79fb/synfig-studio/src/gui/asyncrenderer.cpp#L780-L821
here things start to get more delicate and complicated...

rodolforg:
So we just need to improve the message, as it shouldn't be set as an error, right?
I'm busy these days, but I'll try to take a look on thursday, ok?

FirasH:
> So we just need to improve the message, as it shouldn't be set as an error, right?

Yes, I think so.

> I'm busy these days, but I'll try to take a look on thursday, ok?

No problems, I'm also busy at work during the week 👍

rodolforg:
@FirasH Ok, I have a suggestion:
What about changing the `AsyncRenderer::stop()` declaration to include a new parameter: `reason`, that would be an enum: `REASON_UNKNOWN` (default value) and `REASON_CANCELLED_BY_USER`.

If cancelled by user (via your new stop button), ignore error message in callback. Or prepend a message like. Cancelled by user.

@morevnaproject @ice0

morevnaproject:
@rodolforg I think this is good solution!

P.S. Sorry for delays with 1.3.15 release, also got a busy week here. ^__^"

rodolforg:
May

> ignore error message in callback.

Maybe it's better : store `logger->get_error_message()` before `AsyncRenderer::stop()` emits `signal_stop_`  and display it.

rodolforg:
> P.S. Sorry for delays with 1.3.15 release, also got a busy week here. ^__^"

:)

FirasH:
> @FirasH Ok, I have a suggestion:
> What about changing the `AsyncRenderer::stop()` declaration to include a new parameter: `reason`, that would be an enum: `REASON_UNKNOWN` (default value) and `REASON_CANCELLED_BY_USER`.
> 
> If cancelled by user (via your new stop button), ignore error message in callback. Or prepend a message like. Cancelled by user.
> 
> @morevnaproject @ice0

I was thinking about something similar, I'll look into it.

Anyway I suggest to name the parameter `Interaction` with enum `UNDEFINED` and `BY_USER` so we could use it for handling user interaction on the rendering process.

synfig-studio/src/gui/asyncrenderer.cpp
```diff
@@ -800,26 +801,48 @@ AsyncRenderer::stop()
...
 
-			lock.release();
+			if(interaction == UNDEFINED)
+			{
+				if(status == RENDERING_SUCCESS)
+					signal_success_();
+				else
+					error_message = _("Animation couldn't be rendered");
+
+				target=0;
```
FirasH, line 812: Honestly, not sure what those three lines do...

rodolforg: `error_message = _("Animation couldn't be rendered");` ?
It "prepends" the error message that would be filled via line 819
`				error_message += "\n" + logger->get_error_message();`

`target = 0`is used to free the memory of `target` (it should be nullptr instead of 0 or NULL since C++11).

synfig-studio/src/gui/asyncrenderer.h
```diff
@@ -102,6 +110,7 @@ class AsyncRenderer : public etl::shared_object, public sigc::trackable
 
 	void start();
 	void stop();
+	void user_stop();
```
rodolforg, line 113: No need for it: just change previous line `stop()` to `stop(Interaction interaction = UNDEFINED)`.
When it's a user interaction, call `stop(BY_USER)` ;)

synfig-studio/src/gui/asyncrenderer.h
```diff
@@ -77,6 +82,9 @@ class AsyncRenderer : public etl::shared_object, public sigc::trackable
 	//! Current rendering status
 	//! Undefined if not finished
 	Status status;
+        
+	//! Handles user interaction on the rendering process
+	Interaction interaction;
```
rodolforg, line 87: See below Why this should be deleted.

synfig-studio/src/gui/asyncrenderer.h
rodolforg: Or `stop(AsyncRender::BY_USER)` if there's a namespace problem.

synfig-studio/src/gui/asyncrenderer.cpp
FirasH: This part here:
https://github.com/synfig/synfig/blob/d750262352a6bfb66b5fcc9ec092d008c41249f7/synfig-studio/src/gui/asyncrenderer.cpp#L812-L814

So I didn't touch it as for now.

synfig-studio/src/gui/asyncrenderer.h
FirasH: I cannot get this to work...

synfig-studio/src/gui/render.cpp
```diff
@@ -522,28 +522,10 @@ RenderSettings::submit_next_render_pass()
 
 		canvas_interface_->get_ui_interface()->task(strprintf(_("Rendering %s"), pass_filename.c_str()));
 
-		/*
```
rodolforg, line 525: I wonder why this was commented out…

synfig-studio/src/gui/asyncrenderer.cpp
rodolforg: Got it.
I think only the 
`if(status == RENDERING_ERROR) { ...} ` below should be guarded by `if (interaction ....)`.

No matter why render should be stopped, there should be no target and no render_thread anymore (and lock must be releases). `target` means the exporter/encoder.

synfig-studio/src/gui/asyncrenderer.h
rodolforg: What's happening?

FirasH: When I define `Interaction` with it's default `UNDEFINED` value, the compiler gives an error about the member function definition...
I don't have the code on this machine, but I suspect it's related to calling a member variable of a class within its method, later I'll report here the compiler error.

FirasH:
I was wondering that if we stop the rendering of a single file e.g. and mp4 or avi, the file should be deleted as well.
But we will think about that later (in a separate issue/pull request)...

synfig-studio/src/gui/asyncrenderer.h
```diff
@@ -60,6 +60,11 @@ class AsyncRenderer : public etl::shared_object, public sigc::trackable
 		RENDERING_SUCCESS
 	};
 
+        enum Interaction {
```
rodolforg, line 63: Please fix identation (tab instead of space)

synfig-studio/src/gui/asyncrenderer.h
```diff
@@ -60,6 +60,11 @@ class AsyncRenderer : public etl::shared_object, public sigc::trackable
...
+            UNDEFINED,
```
rodolforg, line 64: It's better to prepend `INTERACTION_` in enum items

synfig-studio/src/gui/asyncrenderer.h
FirasH: Here you can find my compilation output: https://pastebin.com/xjaBNZL7

synfig-studio/src/gui/asyncrenderer.h
FirasH: Here you can find my compilation output: https://pastebin.com/xjaBNZL7

FirasH: Here you can find the changes applied with the previous output: https://pastebin.com/XKAaZuUi

rodolforg: Hm… Apparently sigc connections don't apply the default value for callback arguments :p

Two possible solutions IMO:
- adopt the java overloading style: two methods: stop() and stop(Interaction) methods. The former calls the last filling the argument with default value.
- make use of sigc::bind, as described here: https://developer.gnome.org/gtkmm-tutorial/stable/sec-binding-extra-arguments.html.en https://developer.gnome.org/libsigc++-tutorial/stable/ch04.html https://developer.gnome.org/gtkmm-tutorial/stable/sec-connecting-signal-handlers.html.en )

FirasH: mmm, I should use `sigc::bind(function, value_to_pass)` right?

Tried with this:
`stop_button.signal_clicked().connect(sigc::bind(sigc::mem_fun(*this, &studio::Dock_Info::on_stop_button_clicked)), studio::AsyncRenderer::INTERACTION_BY_USER);`

but still getting compiling error... honestly no idea why...

FirasH: My diff so far: https://pastebin.com/kmWw02BP

rodolforg: You would bind a AsyncRenderer::stop() call, not the callback ;)
Anyway, the problem isn't at this line.

In your previous compilation error report, you can (hardly) find that the problem in in line 922 of asyncrenderer.cpp (on some version of it…).

Captura de tela de 2020-08-07 19-54-16

Here's my tip for those cases GCC spills out a lot of strange error messages. ;) select the whole error block, copy an pasted in a text editor and search for path or namespaces it's exclusively in your code (not in those #included or linked): I used "gui/" on this case. "required from here" is good too. With regex it's easier to filter out what of your source code is triggering this mess.
Then you look at the line/column it describes and... have hope lol

```
#ifdef GLIB_DISPATCHER_BROKEN
		done_connection=Glib::signal_timeout().connect(
			sigc::bind_return(
				mem_fun(*this,&AsyncRenderer::stop),   // HERE
				false
			)
			,0
		);
```

The AsynRenderer::stop() signature changed to AsynRenderer::stop(enum), right? So that call misses the argument now.

Sorry I didn't guide you properly to the problematic line before

FirasH: No problems, you're very helpful 👍 
I found that error message (console output colors errors and warnings) but the error message itself does not give information because it indicates the last line of the edited file.

btw gnome documentantion isn't very easy to follow, at least for me it's missing lots of stuff...

I should bind directly the method like this as reference:
`sigc::bind(async_renderer.stop(), studio::AsyncRenderer::INTERACTION_BY_USER);`

Regaring the example you pointed out, I should use `bind()` instead of `bind_return()`, right? This way I bind the "input" values.

I looked at some other examples in the source code of Synfig and I'm starting to get a clearer idea of what should be done.

I'll try later to apply those changes and report here my updates.

FirasH: Followed your tips and now compiles fine :smile:
Still needs some tuning though.

FirasH: Here you find the changes: https://pastebin.com/YSDwk240

At this point compilation is fine, button is enabled, BUT rendering isn't stopped after confirmation.

rodolforg: This is the reason:
@@ -213,7 +214,10 @@ void studio::Dock_Info::on_stop_button_clicked()
        Gtk::MESSAGE_QUESTION,
        _("Cancel"),
        _("Stop")))
-       async_renderer->user_stop();
+       sigc::bind(
+           sigc::mem_fun(*async_renderer, &studio::AsyncRenderer::stop),
+           studio::AsyncRenderer::INTERACTION_BY_USER
+       );
 }

neither `sigc::bind()` or `sigc::mem_fun()` call a method: they just a way to wrap or to point to a class method like it was a C-style function (i.e., make a class method be called as it was a regular function). ;)

FirasH: I've seen many samples in the source code where the binded method is connected to a signal, like this:
`menu->signal_hide().connect(sigc::bind(sigc::ptr_fun(&delete_widget), menu));`

but in our case `stop()` does not have a `signal_something()` to `connect()` to and call the method...

I don't have experience with this stuff, how would you do that?

rodolforg: Exactly. It's meant to be used for signal connections (slots): what function should signal calls when it's emitted. The problem is that in C++, we usually want to call an object method, that depends on... the object ;) 
Gtkmm use a similar approach as Qt (now) does (https://wiki.qt.io/New_Signal_Slot_Syntax)

Anyway,  we just need here a regular call.
`stop(INTERACTION_BY_USER)` :)

FirasH: Finally done :yum:

synfig-studio/src/gui/docks/dock_info.cpp
```diff
@@ -213,7 +214,12 @@ void studio::Dock_Info::on_stop_button_clicked()
 		Gtk::MESSAGE_QUESTION,
 		_("Cancel"),
 		_("Stop")))
-		async_renderer->user_stop();
+	{
+		sigc::bind(
+			sigc::mem_fun(*async_renderer, &studio::AsyncRenderer::stop),
+			studio::AsyncRenderer::INTERACTION_BY_USER);
```
rodolforg, line 220: This is useless :) You're not using it: `sigc::bind` is used (as argument) for `signal_xxx.connect()` ;)

FirasH: ok, this is used only in case of a `signal` connection.

synfig-studio/src/gui/docks/dock_info.cpp
```diff
@@ -207,6 +232,11 @@ void studio::Dock_Info::changed_canvas_view_vfunc(etl::loose_handle<CanvasView>
 	}
 }
 
+void studio::Dock_Info::set_async_render(etl::handle<studio::AsyncRenderer> ar)
+{
+	async_renderer = ar;
```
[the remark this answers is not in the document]
FirasH: As of now the button is enabled only as there is something that is getting rendered:
https://github.com/synfig/synfig/blob/086ae7d1a854220a829915b08166b25a12333618/synfig-studio/src/gui/docks/dock_info.cpp#L254-L257

Even if someone calls this with a null pointer, it would still be not sensitive anyway.
Did I miss something or you agree?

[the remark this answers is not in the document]
rodolforg: Code changes…
For now, only one place calls the Dock_Info::set_async_renderer(), but who knows?
Better let it "future-proof" :) Simple to solve.
`if (!ar)` for example.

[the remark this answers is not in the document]
FirasH: Right!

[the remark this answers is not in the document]
FirasH: Anyway thank you @rodolforg !
I'm learning a lot of interesting stuff thanks to your tips :+1:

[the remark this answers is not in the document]
rodolforg: I'm glad to help you :)

rodolforg:
Note for dev before merging pull request: squash it ;)

ice0:
Merged. Thank you so much, guys!

PR #1600, opened 2020-08-02
make autotools use imagemagick++ cflags from pkg-config

### LINKED ISSUE #1599: Missing include path for MagickCore/magick-baseconfig.h
synfig 1.3.15 Linux Debian unstable amd64

MagickCore/magick-baseconfig.h isn't stored in /usr/include but in 
/usr/include/x86_64-linux-gnu/ImageMagick-7
or for imagemagick 6
/usr/include/x86_64-linux-gnu/ImageMagick-6

Maybe the best is to use pkgconfig to set CFLAGS accordingly ?

Here in MagickCore-7.Q16HDRI.pc

````
Cflags: -I${includearchdir} -I${includedir} -fopenmp -DMAGICKCORE_HDRI_ENABLE=1 -DMAGICKCORE_QUANTUM_DEPTH=16

````
Build error message 
````
[100%] Building CXX object src/modules/mod_magickpp/CMakeFiles/mod_magickpp.dir/main.cpp.o
cd /src/synfig-1.3.15/obj-x86_64-linux-gnu/src/modules/mod_magickpp && /usr/lib/ccache/c++  -DHAVE_CONFIG_H -DIMAGEMAGICK_CONFIG -Dmod_magickpp_EXPORTS -I/src/synfig-1.3.15/obj-x86_64-linux-gnu/src/modules/mod_magickpp -I/src/synfig-1.3.15/obj-x86_64-linux-gnu/src -I/src/synfig-1.3.15/synfig-core/src -isystem /usr/include/sigc++-2.0 -isystem /usr/lib/x86_64-linux-gnu/sigc++-2.0/include -isystem /usr/include/cairo -isystem /usr/include/glib-2.0 -isystem /usr/lib/x86_64-linux-gnu/glib-2.0/include -isystem /usr/include/pixman-1 -isystem /usr/include/uuid -isystem /usr/include/freetype2 -isystem /usr/include/libpng16 -isystem /usr/include/glibmm-2.4 -isystem /usr/lib/x86_64-linux-gnu/glibmm-2.4/include -isystem /usr/include/giomm-2.4 -isystem /usr/lib/x86_64-linux-gnu/giomm-2.4/include -isystem /usr/include/libmount -isystem /usr/include/blkid -isystem /usr/include/ImageMagick-7 -isystem /usr/include/libxml++-2.6 -isystem /usr/lib/x86_64-linux-gnu/libxml++-2.6/include -isystem /usr/include/libxml2 -isystem /usr/include/pango-1.0 -isystem /usr/include/fribidi -isystem /usr/include/harfbuzz -isystem /usr/include/mlt++ -isystem /usr/include/mlt  -g -O2 -fdebug-prefix-map=/src/synfig-1.3.15=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC   -std=gnu++11 -o CMakeFiles/mod_magickpp.dir/main.cpp.o -c /src/synfig-1.3.15/synfig-core/src/modules/mod_magickpp/main.cpp
In file included from /usr/include/ImageMagick-7/Magick++/Include.h:14,
                 from /usr/include/ImageMagick-7/Magick++.h:10,
                 from /src/synfig-1.3.15/synfig-core/src/modules/mod_magickpp/trgt_magickpp.h:38,
                 from /src/synfig-1.3.15/synfig-core/src/modules/mod_magickpp/main.cpp:40:
/usr/include/ImageMagick-7/MagickCore/magick-config.h:25:10: fatal error: MagickCore/magick-baseconfig.h: No such file or directory
   25 | #include "MagickCore/magick-baseconfig.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
````

### PR DESCRIPTION
**Pull request by rodolforg**
related to #1599 , but it was about the CMake building system

### DIFF
--- a/synfig-core/src/modules/mod_magickpp/Makefile.am
+++ b/synfig-core/src/modules/mod_magickpp/Makefile.am
@@ -7,6 +7,10 @@ AM_CPPFLAGS = \
 	-I$(top_builddir) \
 	-I$(top_srcdir)/src
 
+AM_CXXFLAGS = \
+	@CXXFLAGS@ \
+	@MAGICKPP_CFLAGS@
+
 
 moduledir = @MODULE_DIR@
 
@@ -26,8 +30,8 @@ libmod_magickpp_la_LDFLAGS = \
 	-avoid-version
 
 libmod_magickpp_la_CXXFLAGS = \
-	@SYNFIG_CFLAGS@ \
-	@MAGICKPP_CFLAGS@
+	@MAGICKPP_CFLAGS@ \
+	@SYNFIG_CFLAGS@
 
 libmod_magickpp_la_LIBADD = \
 	../../synfig/libsynfig.la \

### CODE REVIEW

**Comment by rodolforg** [comment]
@ice0 do you think this could really fix the macOS building in Catalina?
I just remove it in https://github.com/synfig/synfig/pull/1486

**Comment by ice0** [comment]
I will take a look.

**Comment by ice0** [comment]
Ok, here what's going on:

Currently Synfig compilation on MacOS fails with error:
`clang: error: unsupported option '-fopenmp'`

This is happening because Clang from XCode tools is compiled without OpenMP support.
We can use `gcc` or `clang` installed from Homebrew to resolve this.

But ImageMagick uses different approach.
As described in this article (https://iscinumpy.gitlab.io/post/omp-on-high-sierra/) we can hide `-fopenmp` argument from compiler and pass it directly to preprocessor, so that should fix the problem. And if we look at `ImageMagick.pc` file, we can see that they use the `-Xpreprocessor` flag.

Also when I pass `-Xpreprocessor -fopenmp` to `clang` from the command line everything works fine. But Synfig building still fails.

After that, I noticed that the arguments were shuffled on the command line, and I found this email on the libtool mailing lists discussing libtool argument reordering.

https://www.mail-archive.com/libtool@gnu.org/msg10260.html

and here: 

https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/7BRVA3N6C7GCWLY6LT72MPNSIHXBMBQ2/

After that, I tried using different flags to fix this and still no success.

So I looked inside the libtool code to see why this doesn't work, and found what that flags is directly imported from `libMagick++-7.Q16HDRI.la` file. O_o

```
# Linker flags that cannot go in dependency_libs.
inherited_linker_flags='  -pthread -fopenmp'
```

and code which collects dependencies:
```
      for deplib in $libs; do
        lib=
        found=false
        case $deplib in
        -mt|-mthreads|-kthread|-Kthread|-pthread|-pthreads|--thread-safe \
        |-threads|-fopenmp|-openmp|-mp|-xopenmp|-omp|-qsmp=*)
          if test prog,link = "$linkmode,$pass"; then
            compile_deplibs="$deplib $compile_deplibs"
            finalize_deplibs="$deplib $finalize_deplibs"
          else
            func_append compiler_flags " $deplib"
            if test lib = "$linkmode"; then
                case "$new_inherited_linker_flags " in
                    *" $deplib "*) ;;
                    * ) func_append new_inherited_linker_flags " $deplib" ;;
```

So this looks like hardcoded behavior of libtool. Anyway we can patch ImageMagick's `.la` files (at least on build server).

@rodolforg @morevnaproject any ideas?

rodolforg:
Ugh over complicated :( I dunno what to do

morevnaproject:
I think the easiest way is to make our own modification of ImageMagick's brew formula and explicitly build and install it - https://github.com/Homebrew/brew/issues/1468

ice0:
Merged. Thank you!

PR #1607, opened 2020-08-05
fix old timetrack crashing on waypoint right-clicking

rodolforg:
Reported by tawatoons in https://forums.synfig.org/t/synfig-1-3-15-crash-on-right-clicking-waypoint/11252

It's probably due to a commit reverting for fixing another issue in Windows


rodolforg:
Somehow it wasn't setting the internal param_tree_store_ ever.

rodolforg:
Maybe we should release another version as soon as it's possible, as 1.3.15 has this terrible error :( so easy/common to trigger.

ice0:
Merged. Thank you!

morevnaproject:
> Maybe we should release another version as soon as possible, as 1.3.15 has this terrible error :( so easy/common to trigger.

Agree!

rodolforg:
@morevnaproject  wait for another fix :D

rodolforg:
> @morevnaproject wait for another fix :D

Done via #1608

PR #1614, opened 2020-08-08
canvas end-time could be set by wrong reason

rodolforg:


rodolforg:
It looks like a copy'n paste mistake.
Line 975 takes care about gamma value, and end-time should be set due to end-time differences.

ice0:
Merged. Thank you!

Issue #384, opened 2017-06-30
Synfig does not add audio layers to MPG file when rendering.


CoryAlbrecht:
Synfig 1.0.2 on Ubuntu GNOME 16.04.2

Synfig does not add audio layers to rendered MPG file. From the console output it looks like only a video stream is passed to ffmpeg but no audio stream.

```
cory@movingpictures[13:28:49]~/Documents/Synfig Projects/QGIS Backgrounder 1$ synfig --fps 25  -t ffmpeg -o qb1.mpg QGIS\ Backgrounder\ 1.sif 
synfig(31191) [01:30:01 PM] info: Loading modules from /etc/synfig/synfig_modules.cfg
/usr/bin/synfig: warning: Unable to find module "mod_libavcodec" (file not found)
synfig(31191) [01:30:01 PM] info: Loading file: QGIS Backgrounder 1.sif
synfig(31191) [01:30:01 PM] info: Render broken up into 1 block 781 pixels tall, and a final block 299 pixels tall
ffmpeg version 3.2-1001 Copyright (c) 2000-2016 the FFmpeg developers
  built with gcc 5.4.0 (Ubuntu 5.4.0-6ubuntu1~16.04.4) 20160609
[...]

PR #1623, opened 2020-08-10
Export video files with sound

morevnaproject:
Now sound is exported too when user renders to mp4/avi (ffmpeg target).

### DIFF
--- a/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
+++ b/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
@@ -127,7 +127,7 @@ ffmpeg_trgt::~ffmpeg_trgt()
 	delete [] color_buffer;
 
 	// Remove temporary sound file
-	struct stat buf;
+	GStatBuf buf;
 	if (g_stat(sound_filename.c_str(), &buf) != -1) {
 		if( g_remove(sound_filename.c_str()) != 0 ) {
 			synfig::warning("Error deleting temporary sound file.");
@@ -182,7 +182,7 @@ ffmpeg_trgt::init(ProgressCallback *cb=NULL)
 	soundProcessor.set_infinite(false);
 	get_canvas()->fill_sound_processor(soundProcessor);
 	// Generate random filename here
-	struct stat buf;
+	GStatBuf buf;
 	do {
 		synfig::GUID guid;
 		sound_filename = String(filename)+"."+guid.get_string().substr(0,8)+".wav";
--- a/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
+++ b/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
@@ -260,7 +260,11 @@ ffmpeg_trgt::init(ProgressCallback *cb=NULL)
 	vargs.push_back(ffmpeg_binary_path);
 	if (with_sound) {
 		vargs.push_back("-i");
+#if defined(WIN32_PIPE_TO_PROCESSES)
+		vargs.push_back("\"" + sound_filename + "\"");
+#else
 		vargs.push_back(sound_filename);
+#endif
 	}
 	vargs.push_back("-f");
 	vargs.push_back("image2pipe");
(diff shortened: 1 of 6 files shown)


morevnaproject:
TODO: Need to check how this function works on Windows with multibyte non-Latin symbols.

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
```diff
@@ -164,6 +175,26 @@ ffmpeg_trgt::set_rend_desc(RendDesc *given_desc)
 bool
 ffmpeg_trgt::init(ProgressCallback *cb=NULL)
 {
+
+	if(!SoundProcessor::subsys_init())
+		throw std::runtime_error(_("Unable to initialize subsystem \"Sound\""));
```
rodolforg, line 180: Wouldn't be better to report error in `cb` and return false, instead of throwing an exception?

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
```diff
@@ -122,6 +125,14 @@ ffmpeg_trgt::~ffmpeg_trgt()
 	file=NULL;
 	delete [] buffer;
 	delete [] color_buffer;
+
+	// Remove temporary sound file
+	struct stat buf;
+	if (g_stat(sound_filename.c_str(), &buf) != -1) {
+		if( g_remove(sound_filename.c_str()) != 0 ) {
+			synfig::warning("Error deleting temporary sound file.");
```
ice0, line 133: I suggest adding the filename to the message so that the user knows which one they are trying to delete.

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
```diff
@@ -122,6 +125,14 @@ ffmpeg_trgt::~ffmpeg_trgt()
 	file=NULL;
 	delete [] buffer;
 	delete [] color_buffer;
+
+	// Remove temporary sound file
+	struct stat buf;
```
ice0, line 130: Use GStatBuf: https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-stat

morevnaproject: Got it! Thank you for the hint. ^__^

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
morevnaproject: Done.

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
morevnaproject: Absolutely! Did the changes now. ^__^

morevnaproject:
I have found that this feature doesn't works on Windows (when building through MSYS2). This is because MLT doesn't have avformat (ffmpeg) module enabled.

To fix that we need to:
1. Install ffmpeg package via MSYS2.
2. Recompile MLT with ffmpeg.

morevnaproject:
> TODO: Need to check how this function works on Windows with multibyte non-Latin symbols.

It is working on WIndows now. Even with Russian/Arabic/OtherNonLatin filenames.

morevnaproject:
I wonder, if we can include this for 1.4.0?

For now I am doing a test build as 1.3.17 to give people a test.  Still not sure if this will be in new dev release or it is better to postpone this for 1.4.1...

rodolforg:
I tested it here and it worked fine in two different codecs with ffmpeg target (debian testing, master branch)

rodolforg:
I wonder if the new font widget could be embed too lol

Fahad-Alsaidi:
thank you for this feature. It makes exporting  synfig project much easier

rodolforg:
@morevnaproject So is this issue solved? #384

morevnaproject:
Backported to 1.4.x branch - 4fa1db268a78fb395a6a27f9d61cd705eca64ac9

morevnaproject:
> @morevnaproject So is this issue solved? #384

@rodolforg No, the issue #384 is solved by #1830. Will be backported soon. ^__^

PR #1632, opened 2020-08-12
Remove weird sleep() call

morevnaproject:


ice0:
It looks like a hack to get ffmpeg to complete work. Do you have any issues after remove it?

morevnaproject:
I actually didn't tested :rofl:

PR #1633, opened 2020-08-13
Do not use Glib::locale_to_utf8() in synfig::get_binary_path()

morevnaproject:
Glib::locale_to_utf8() doesn't give correct result when path contains multibyte characters (i.e. Arabic symbols).


ice0:
Merged. Thank you!

Issue #1624, opened 2020-08-11
[Skeleton Tool] Enable/Disable Skeleton Deformation layer when tool switched


morevnaproject:
Need to implement two cases:

1. If ((user switches to Skeleton Tool) and (current selected layer is "Skeleton Deformation")), then:
set this "Skeleton Deformation" as invisible/disabled.

2. If ((user switches from Skeleton Tool to any other tool) and (current selected layer is "Skeleton Deformation")), then:
set this "Skeleton Deformation" as visible/enabled.

Aa20475:
Hey! I've made this change but the `dock_layers` is not getting upgraded.Do I have to trigger a signal for this? Or is there a function to update it?

rodolforg:
From `layer.cpp`:
```c++
void
Layer::set_active(bool x)
{
	if(active_!=x)
	{
		active_=x;

		Node::on_changed();
		signal_status_changed_();
	}
}
```

rodolforg:
And theoretically it's updated in `layertreestore.cpp`:

```c++
LayerTreeStore::LayerTreeStore(etl::loose_handle<synfigapp::CanvasInterface> canvas_interface_):
	Gtk::TreeStore			(ModelHack()),
	queued					(false),
	canvas_interface_		(canvas_interface_)
{
	layer_icon=Gtk::Button().render_icon_pixbuf(Gtk::StockID("synfig-layer"),Gtk::ICON_SIZE_SMALL_TOOLBAR);

	// Connect Signals to Terminals
	canvas_interface()->signal_layer_status_changed().connect(sigc::mem_fun(*this,&studio::LayerTreeStore::on_layer_status_changed));
```

Aa20475:
Hey! Are active and enable the same?

rodolforg:
Whoops I don't know:) @morevnaproject ?

morevnaproject:
Yes, active and enable are the same. See `synfig-studio/src/synfigapp/actions/layeractivate.cpp`. ^__^

rodolforg:
So, it would automatically update the layer tree view :)

Aa20475:
Ohh thanks! @morevnaproject  @rodolforg I triggered the `signal_layer_status_changed` in `CanvasInterface`. :)

Aa20475:
I think this issue is closed! ^__^

PR #1635, opened 2020-08-14
Skeleton tool : Conditional disable/enable of Deformation layer on tool switch

Aa20475:
Issue #1624

### DIFF
--- a/synfig-studio/src/gui/states/state_bone.cpp
+++ b/synfig-studio/src/gui/states/state_bone.cpp
@@ -437,7 +437,6 @@ StateBone_Context::StateBone_Context(CanvasView *canvas_view) :
 	if(etl::handle<Layer_SkeletonDeformation>::cast_dynamic(layer)){
 		get_work_area()->set_type_mask(get_work_area()->get_type_mask()-Duck::TYPE_TANGENT|Duck::TYPE_WIDTH);
 		get_canvas_view()->toggle_duck_mask(Duck::TYPE_NONE);
-		cout<<"Hi"<<endl;
 		layer->disable();
 		get_canvas_interface()->signal_layer_status_changed()(layer,false);
 	}else{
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -437,6 +437,9 @@ StateBone_Context::StateBone_Context(CanvasView *canvas_view) :
 	if(etl::handle<Layer_SkeletonDeformation>::cast_dynamic(layer)){
 		get_work_area()->set_type_mask(get_work_area()->get_type_mask()-Duck::TYPE_TANGENT|Duck::TYPE_WIDTH);
 		get_canvas_view()->toggle_duck_mask(Duck::TYPE_NONE);
+		cout<<"Hi"<<endl;
```
rodolforg, line 440: ops ;)

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -437,6 +437,9 @@ StateBone_Context::StateBone_Context(CanvasView *canvas_view) :
...
+		layer->disable();
+		get_canvas_interface()->signal_layer_status_changed()(layer,false);
```
rodolforg, line 442: it should not be necessary at all!
`void disable() { set_active(false); }` calls `set_active()`

And `set_active()` calls the signal
```c++
Layer::set_active(bool x)
{
	if(active_!=x)
	{
		active_=x;

		Node::on_changed();
		signal_status_changed_();
	}
```

Aa20475: Hey! For some reason, without the  `get_canvas_interface()->signal_layer_status_changed()(layer,false);` , the `dock_layers` is not being updated. :thinking:

morevnaproject:
It works well. Just make sure to remove redundant comment. ^__^

synfig-studio/src/gui/states/state_bone.cpp
rodolforg: Ah I got it.
The layer panel doesn't monitor it indeed. And it makes sense somehow (due to switching between open documents) :)
So the solution is the way you did or calling an synfigapp action as Konstantin pointed.

morevnaproject:
Merged now. Thank you!

PR #1637, opened 2020-08-15
Attempt to fix Travis build on OSX 10.10

morevnaproject:


morevnaproject:
It seems like build was fixed remotely (pip guys pushed a fix?).

PR #1638, opened 2020-08-16
Make position of dialog buttons consistent

### LINKED ISSUE #1165: Cancel/Confirm buttons position is not consistent across all dialogs
Master 09220e

**Issue description**:
In most cases the dialogs have the button to confirm on the right and the button to cancel on the left. For example:
example_left-right

There are some dialogs which have the opposite design. For example:
other-example

I would suggest to make all dialogs consistent. What do you think?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/88816227-cancel-confirm-buttons-position-is-not-consistent-across-all-dialogs?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

FirasH:
Fix #1165 

The final order of buttons will be fixed (then we'll deal with different OSes):
Other | Cancel/Close | Ok/Confirm

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -2946,11 +2946,11 @@ App::dialog_open_file_with_history_button(const std::string &title, std::string 
 
 	dialog->set_transient_for(*App::main_window);
 	dialog->set_current_folder(prev_path);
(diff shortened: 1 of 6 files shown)


ice0:
Merged. Thank you!

Issue #1627, opened 2020-08-11
[Skeleton Tool] Child bone placed incorrectly on Skeleton Deformation layer 


morevnaproject:
At certain circumstances child bone appears in wrong place when added using Skeleton Tool in Skeleton Deformation Layer.

Steps to reproduce:
1. Create Skeleton Deformation Layer using Skeleton Tool, add 1 bone.
2. Modify any point of that bone.
3. Create next bone. -> it appears in wrong place.

Video - https://youtu.be/dzTPdNbqS3U

Hint: This is about dealing with first/second bones.

Issue #1626, opened 2020-08-11
[Skeleton Tool] Edit Second bone too when Skeleton Deformation Layer edited


morevnaproject:
Let's suppose user created several bones in Skeleton Deformation Layer (using Skeleton Tool).
Then he realized that some bones are not placed correctly, so he changed their points.
With current behavior, when user edits first bone (initial position) the position of second bone is not changed.
This leads to problems, shown in this video -https://youtu.be/obN94tO6xGM

Solution:

When user edits first bone of Skeleton Deformation Layer, then the same modification should be applied to second bone.

It is important to note that we should apply delta value.

I.e. let's suppose first bone has angle 30 degrees and it was changed to value of 45 degrees. Then we should add 15 degrees to second bone.

Aa20475:
Hey!
I've been thinking about how to solve this issue. But I couldn't come up with any way. I thought,
- Add the second bone's corresponding duck to selected ducks so that the drag applies to it too but I couldn't find a way to get access to it as it's hidden? Also, other thing is that, when the user is selecting a duck to modify the bone, the click handler which I write is not being used. (any tips on this?)
- Save the values of offsets and other parameters during click event and then update the second bone accordingly. Here also, the click handler which I write is not being used. 

What do you think? Also, I think issue #1627 is related to this.

morevnaproject:
I think it's better to go with a second approach.

> Here also, the click handler which I write is not being used.

Can you please push a code you have and show which part you talking about?

Aa20475:
>Can you please push a code you have and show which part you talking about?

I was talking about this handler. When I was solving some other issue, I found out that this handler is not being called when the user clicks on a duck. So, now I'm confused bout how to get the initial offsets and parameters of the bone on click.
https://github.com/synfig/synfig/blob/master/synfig-studio/src/gui/states/state_bone.cpp#L523

morevnaproject:
Maybe modify SetValueNode action? Possibly, you can check which valuenode was modified and if it is a bone of Skeleton Deformation Layer, then do changes  for second bone?

![screenshot_002](https://user-images.githubusercontent.com/332868/90337286-b427f100-e00b-11ea-947e-c4430b307728.png)

Aa20475:
Heya that's a nice idea! Will try! 😁

morevnaproject:
We still need this issue resolved, it is blocking 1.6.0 release.

[...]

PR #1641, opened 2020-08-17
[Skeleton Tool] Child bone placed incorrectly on Skeleton Deformation layer

Aa20475:
Hey!

Issue #1627 

I fixed this. It is basically coz I was giving the second bone as the reference to the calculations of the parameters of the new child bone. In the video, after you changed the first bone, as pointed out in #1626 , the second bone and first bone are not in the same position. So this caused the wrong placement. I guess it works fine now as I'm using first bone. But #1626  still remains!


morevnaproject:
> In the video, after you changed the first bone, as pointed out in #1626 , the second bone and first bone are not in the same position.

Exactly! ^__^

morevnaproject:
The issue is fixed, thank you!

P.S. BTW, if you put "Closes #1627" into description of PR, then it will close related issue automatically when PR is merged. ^__~

PR #1651, opened 2020-08-23
fix memory leak on thread pool

rodolforg:
ThreadPools::threads is filled with threads allocated in ThreadPool::wakeup() . They're never destroyed.

They can be safely deleted after joined.
https://stackoverflow.com/questions/30090299/delete-stdthread-after-calling-join


ice0:
Merged. Thank you!

PR #1652, opened 2020-08-23
fix memory leak with Widget_Timetrack: row_info pointer to value

rodolforg:
Widget_Timetrack::rebuild_param_info_list() fills param_info_map with parameter panel a newly allocated row info that would override previous value. Previous param_info_map::clear() call wouldn't deallocate them properly.

Therefore, for avoiding this problem, don't use unneeded data via pointers,
but store them directly in param_info_map.


ice0:
Merged. Thank you!

PR #1653, opened 2020-08-23
Some cleanup on Layer_Freetype

rodolforg:


rodolforg:
Yes, I know probably all this layer will be replaced by Pango rendering instead of direct usage of FreeType library.
However, I'm having some size/position difference issues with this replacement. So, for now, let me improve this layer…

rodolforg:
@ice0 could this be a CMake specific issue?

ice0:
@rodolforg sure. CMake still could have different compilation flags and defines. I will take a look.

ice0:
The problem is in that place:
```
// special_family.option3 = nullptr
possible_families.push_back(special_family.option3);
```

This does not happen when FontConfig is installed, but it may need to be considered.

ice0:
Now it fails because of this:

```
// i = 6; (but special_font_family_db has only 5 items)
for (int i = 0; special_font_family_db[i].alias; i++) {
```

I changed it to:
```
int count = sizeof(special_font_family_db) / sizeof(special_font_family_db[0]);
for (int i = 0; i < count && special_font_family_db[i].alias; i++) {
```

ice0:
Merged. Thank you!

PR #1655, opened 2020-08-24
speed up Layer_Freetype: don't load face file on every rendering

rodolforg:


rodolforg:
Depends on #1653

ice0:
Merged. Thank you!

PR #1662, opened 2020-08-24
[Skeleton Tool] Active bone not set for Skeleton Deformation

### LINKED ISSUE #1656: [Skeleton Tool] Active bone not set for Skeleton Deformation
With Skeleton Deformation Layer:

When user clicks **on any control point** of bone, then this bone is highlighted as "active" but it doesn't really becomes active. As result, next created bone is not parented to this bone.

Note: when user clicks on bone itself (**not** its control point), then everything works as expected.

Video - https://youtu.be/hTLZSWxRzUo

Aa20475:
Closes #1656


morevnaproject:
It works now. Thank you!

PR #1663, opened 2020-08-24
[Skeleton Tool] Remove "Create Child Bone" action menu

### LINKED ISSUE #1558: [Skeleton Tool] Remove "Create Child Bone" action menu
I suggest to remove this action menu after work on Skeleton Tool will be completed.

So, it is supposed that user should use Skeleton Tool to add new bones and this makes consistent workflow.

Aa20475:
Closes #1558


morevnaproject:
Looks good. Thank you!

PR #1667, opened 2020-08-25
Removed pip installation from MacOS install script (it is already installed by python formula)

ice0:
It trying to install pip, but pip3 is already installed.


morevnaproject:
I remember on old version of OSX (10.9)  there were no pip... Maybe something changed...

PR #1688, opened 2020-09-01
autotools: make module flags and libs for the module itself only

rodolforg:


rodolforg:
Magick++ and pango where referred are a general dependency for synfig core… Just a small cleanup.
(maybe it'd speed up compilation and linking)

ice0:
Performance test (clean build using ./2-build-debug.sh):
standard:
```
real    57m9,729s
user    6m27,269s
sys     9m10,986s
```

with optimizations:
```
real    55m50,220s
user    6m10,636s
sys     8m58,068s
```

ice0:
Merged. Thank you!

morevnaproject:
Woah! This actually can fix another issue - we had problem when flags from Magick++ prevented building of Synfig core binary. I need to check that.. Thank you!

Issue #1058, opened 2019-12-29
"unsafe srcdir value" error trying to build ETL


FirasH:
**Synfig version & platform**:
Synfig master (47d5d4c)
Windows 10 Pro 64-bit, MINGW64_NT-10.0-18363

**Issue description**:
On my machine the directory path to my home contains a space.
(in my case the home directory is called "Postazione 2")

This issue does not happen on my Linux machine where there isn't any space in the path.

Following the steps for building this is what happens:
1) Execute 1-setup-windows-msys2.sh and it works fine (64452ec).
2) Execute 2-build-production.sh and the script reports this error:
"unsafe srcdir value: path_with_blank_space"

Looking at the code the issue happens because the relative path contains a blank space, indeed executing ./configure while locating inside /synfig/ETL does not cause the error.

If we could change the script to execute ./configure while it is located in /synfig/ETL the relative path does not include the home directory which is causing the issue with the blank space. Is that a big issue?

FirasH:
With cmake I could compile the master branch easily, so this issue could be resolved.

Only problem I noticed is that some icons are missing around the gui and the starting splash screen does not have an image.

@ice0, @AYESDIE do you know if this issue is already known?
I checked the issue tracked and found #1511 where seems resolved. In my case I just compiled with cmake and nothing else, this is my build:
![cmake_build](https://user-images.githubusercontent.com/6705690/91710772-185ac100-eb85-11ea-865d-4d8f6cb1f8d9.png)

ice0:
It looks like the missing icons are standard GTK icons.
Please check if you have `mingw-w64-x86_64-adwaita-icon-theme` installed.

I'm missing a splash screen too, @AYESDIE, can you check it?

ice0:
ok, found the problem: splash_screen is in './share/synfig/icons/classic/' but should be in './share/synfig/images/' directory.

FirasH:
@ice0 
> It looks like the missing icons are standard GTK icons.
> Please check if you have `mingw-w64-x86_64-adwaita-icon-theme` installed.

`mingw-w64-x86_64-adwaita-icon-theme-3.36.0-1` was already installed before running the compilation.

AYESDIE:
Okay, I think it's because those icons are not being moved to the right place either. I'll take a look at that and push those changes soon.

ice0:
@FirasH Did it prints something interesting in console when you run Synfig Studio?

ice0:
[...]

PR #1689, opened 2020-09-01
[CMake] Fixed path for splash screen image and added Adwaita icons for MSYS2 builds.

AYESDIE:
Fixes the issue raised [here](https://github.com/synfig/synfig/issues/1058#issuecomment-683697028) in #1058.

### DIFF
--- a/cmake/InstallMSYS2.cmake
+++ b/cmake/InstallMSYS2.cmake
@@ -132,4 +132,7 @@ if(WIN32)
     )
     file(COPY ${LIB_DIRECTORIES} DESTINATION ${SYNFIG_BUILD_ROOT}/lib)
     install(DIRECTORY ${LIB_DIRECTORIES} DESTINATION lib)
+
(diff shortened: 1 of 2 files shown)


FirasH:
I've completely reinstalled msys and rebuilt your branch, but still the interface has the same missing icons and missing splash screen.

I have no idea if there's an issue on my side, but my system is vanilla (just the programs I use, not even the antivirus)

AYESDIE:
It doesn't have anything to do with the antivirus. After you run the cmake command, can you go and check whether there's /cmake-build-msys/output/share/icons/Adwaita` folder in there. It should be there after you run the cmake command. Also, in this case, if there's no source (which means you've not installed that dependency, it should fail at the cmake part). 

For me, on running the `./2-build-msys-cmake.sh`, those icons are placed in the correct folder once the cmake part is complete, and the splash screen is placed where it should be as well when it's built.

AYESDIE:
Even on the appveyor build, the splash screen is generated into the correct folder [here](https://ci.appveyor.com/project/Synfig/synfig/builds/34965655/job/rppv0uph5s6y0u55?fullLog=true#L2649).

ice0:
Merged. Thank you!

PR #1697, opened 2020-09-05
[Skeleton Tool] Hide "Make Parent To Active Bone" menu item when not using Skeleton Tool

### LINKED ISSUE #1669: [Skeleton Tool] Hide "Make Parent To Active Bone" menu item when not using Skeleton Tool
The "Make Parent To Active Bone" menu item appears when user right-clicks on any control point of bone.
This menu item should be visible only when Skeleton Tool is active.

Aa20475:
Closes #1669

### DIFF
--- a/synfig-studio/src/synfigapp/actions/valuedescbonesetparent.cpp
+++ b/synfig-studio/src/synfigapp/actions/valuedescbonesetparent.cpp
@@ -103,10 +103,14 @@ Action::ValueDescBoneSetParent::is_candidate(const ParamList &x)
 	if (!candidate_check(get_param_vocab(),x))
 		return false;
 
+	i = x.find("canvas_interface");
+	etl::handle<CanvasInterface> ci = i->second.get_canvas_interface();
+
 	return value_desc.parent_is_value_node()
 		&& ValueNode_Bone::Handle::cast_dynamic(value_desc.get_parent_value_node())
 		&& child
-		&& ValueNode_Bone::Handle::cast_dynamic(child);
+		&& ValueNode_Bone::Handle::cast_dynamic(child)
+		&& ci->get_state()=="bone";
 }
 
 bool
--- a/synfig-studio/src/gui/states/state_bone.cpp
+++ b/synfig-studio/src/gui/states/state_bone.cpp
@@ -367,6 +367,8 @@ StateBone_Context::StateBone_Context(CanvasView *canvas_view) :
 {
 	egress_on_selection_change=true;
 
+	get_canvas_interface()->set_state("bone");
+
 	/*setting up the tool options menu*/
 
 	// 0, title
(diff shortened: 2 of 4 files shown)


morevnaproject:
I wonder, is there any better way to identify current active tool, instead of adding `get_canvas_interface()->set_state()` method?

@rodolforg @ice0 any advice here? ^__^

ice0:
@morevnaproject you can use something like:
```
canvas_view->get_smach().get_state_name() == "bline"
```

Aa20475:
But we can't access `canvas_view` in actions right?

ice0:
I think you can use `App::get_selected_canvas_view()` for that.

morevnaproject:
> I think you can use `App::get_selected_canvas_view()` for that.
> 
> 
> 
> Also I think maybe add a method to Toolbox to get it state, as it looks more intuitive for me. Something like:
> 
> 
> 
> ```
> 
> App::get_toolbox()->get_state_name(); // returns current state name
> 
> ```
> 
> 
> 
> What do you think guys?
> 
> 

Maybe even simpler? ^__^ -
'''
App::get_state_name()
'''

morevnaproject:
> I think you can use `App::get_selected_canvas_view()` for that.

@ice0 Unfortunately, we can't use `App::` in actions (as they are part of `synfigapp`, not `gui`).

ice0:
@morevnaproject `CanvasInterface` is a part of `synfigapp` so anyway you need it.

morevnaproject:
@ice0 So, do you suggest to keep the current solution as is?

ice0:
The only improvement I can suggest is to remove `CanvasInterface::set_state` and `CanvasInterface::get_state` and use `smach` for that.

morevnaproject:
I agree with that, but how to access `smach` data from action of `synfigapp`?

rodolforg:
Hi.
AFAIK `synfigapp` is focused on change the "synfig document" behind the curtains– I mean, the GUI ;)
Change the current tool is not such an action, as well as "select specific layer" is not too.

So _maybe_ `synfigapp` is not the right place?

ice0:
@rodolforg @morevnaproject I think you are right guys. So I need a time to think about it. Maybe someone will suggest a better solution.

rodolforg:
Well, PR #1698 was merged and it has this same `CanvasInterface::set_state` and `CanvasInterface::get_state` on master branch...
So I think we can merge this PR and decide about it later .

synfig-studio/src/synfigapp/actions/valuedescbonesetparent.cpp
```diff
@@ -103,10 +103,14 @@ Action::ValueDescBoneSetParent::is_candidate(const ParamList &x)
...
 
+	i = x.find("canvas_interface");
+	etl::handle<CanvasInterface> ci = i->second.get_canvas_interface();
+
 	return value_desc.parent_is_value_node()
 		&& ValueNode_Bone::Handle::cast_dynamic(value_desc.get_parent_value_node())
 		&& child
-		&& ValueNode_Bone::Handle::cast_dynamic(child);
+		&& ValueNode_Bone::Handle::cast_dynamic(child)
+		&& ci->get_state()=="bone";
```
rodolforg, line 113: maybe check `&& ci`  before `&& ci->get_state()=="bone";`

rodolforg:
Superseded by #2064

PR #1699, opened 2020-09-07
Don't hold layers selection when creating a new layer

### LINKED ISSUE #1666: Creating a new Layer, selection of other Layers is held
**Synfig version**:
master 617fd4b

**Issue description**:
1) Create a new animation
2) Select the `Rectangle Tool` (or any other tool that creates a new `Layer`)
3) Draw a `Rectangle` on Canvas
      Now a new `Layer` is added to the animation
4) Draw another `Rectangle` on Canvas
      Now a second `Layer` is added to the animation

At this point both `Layers` are selected.

I'm not sure if this is the right behavior, but for my usage I would prefer to select only the last created `Layer` (to avoid unselection and selection again each time).

If this is the expected behavior, maybe there should be an option to disable it?

FirasH:
Fix #1666


ice0:
Merged. Thank you!

PR #1716, opened 2020-09-29
Fixed #1715: Updated to fix warning message

### LINKED ISSUE #1715: Don't show warning message if config file is found
Synfig master branch

**Issue description**:

When you run the `synfig` binary to render, it shows a warning on startup.

```
synfig(13472) [19:19:06] warning: Cannot open ./synfig_modules.cfg
synfig(13472) [19:19:06] warning: Cannot open C:\msys64\home\user/.local/share/synfig/synfig_modules.cfg
````

but this warning is incorrect because the `synfig_modules.cfg` file still can be found in the` ../etc/` folder.

So we need to display warning only if `synfig_modules.cfg` file is not found in any known locations.
Known locations is:
- `.` (Current directory)
- `../etc` (relative from `synfig` binary location directory)
- `%USER_PROFILE%/.local/share/` (for Windows)
- `INSTALL_PREFIX` (for example /usr/local/...) for Linux

If file is found when we need to show message with location where it is found. For example:
```
Using settings from `c:\synfig\etc\synfig_modules.cfg`
```

If file is not found then display message like this:
```
The configuration file `synfig_modules.cfg` was not found. Searched in the following directories (<list of directories>)
```

warlockz:
Updated to fix warning message.

Please can you review and merge my PR.

Thanks :)

### DIFF
--- a/autobuild/osx/app-template/Contents/Info.plist
+++ b/autobuild/osx/app-template/Contents/Info.plist
@@ -4,7 +4,8 @@
 <dict>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>English</string>
-	
+	<key>CFBundleIdentifier</key>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
 	<key>CFBundleGetInfoString</key>
 	<string>_VERSION_, © 2001-2008 The Synfig Team</string>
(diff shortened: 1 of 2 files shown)


ice0:
It did not builds. So I close it now.

PR #1720, opened 2020-10-01
Update 1-setup-windows-msys2.sh

gajerachintan9:


Keyikedalube:
Please try to solve actual issues from the issue board. There are tons :) When you help close an issue by contributing quality code/documentation/content you're helping Synfig community a lot.

Here's a similar and useful issue that needs content reordering: https://github.com/synfig/synfig-docs-dev/issues/65
It's very easy. You don't need to learn code or write new content at all... it's a WYSIWYG document type in reStructuredText, just familiarize with the layout and you can do it :)

PR #1739, opened 2020-10-05
Fixed GCC warnings (different signedness comparison)

rahmanuh:
I saw some more warning messages (as you had said), will try to fix them as well. But I will study this project first :)

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_soundwave.cpp
+++ b/synfig-studio/src/gui/widgets/widget_soundwave.cpp
@@ -198,7 +198,7 @@ bool Widget_SoundWave::on_draw(const Cairo::RefPtr<Cairo::Context>& cr)
 		if (t >= sound_delay) {
 			synfig::Time dt = t - sound_delay;
 			unsigned long index = int(dt * stride) + channel_idx;
-			if (index >= 0 && index < buffer.size())
+			if (index < buffer.size())
 				std::copy(buffer.begin() + index, buffer.begin() + index + bytes_per_sample, &value);
 		}
 		int y = time_plot_data->get_pixel_y_coord(value);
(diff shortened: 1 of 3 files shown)


synfig-studio/src/gui/widgets/widget_soundwave.cpp
```diff
@@ -197,7 +197,7 @@ bool Widget_SoundWave::on_draw(const Cairo::RefPtr<Cairo::Context>& cr)
 		int value = middle_y;
 		if (t >= sound_delay) {
 			synfig::Time dt = t - sound_delay;
-			int index = int(dt * stride) + channel_idx;
+			unsigned long index = int(dt * stride) + channel_idx;
 			if (index >= 0 && index < buffer.size())
```
Keyikedalube, line 201: Just remove `index >= 0` since unsigned long can never be negative. We're unnecessarily comparing an always +ve value with +ve

Helpful post 👇
https://stackoverflow.com/questions/5066362/why-is-comparing-if-an-unsigned-int-0-a-pointless-comparison

Keyikedalube (changes requested):

Keyikedalube:
> I saw some more warning messages (as you had said), will try to fix them as well. But I will study this project first :)

Please don't fix other warnings in this PR. Stay within the scope of the issue you are already fixing OK ie., just the data type for now.

Also please feel free to open an issue in the issue board if you want to fix more warning messages :)

synfig-studio/src/gui/widgets/widget_soundwave.cpp
ice0: I am not sure this should be removed. Maybe it can get negative values, so we need to dig deeper here.

Keyikedalube: The post said if a -ve is returned, `index` will just be defined with the largest unsigned number instead. However, deeper investigation could be done on what each variable of this statement `int(dt * stride) + channel_idx` does.

rahmanuh: Did a little investigation:
- `dt = t - sound_delay`. --> I think it will never be negative since it is a diff.
- `stride = frequency * bytes_per_sample * n_channels` --> I think it will also never be negative.
- `channel_idx`--> I am not sure about this, but I think this will never be negative. 

That being said `index = int(dt * stride) + channel_idx` will always positive.
I will dig deeper to make sure.

ice0:
Merged. Thank you!

PR #1740, opened 2020-10-05
Remove unused headers (should reduce build time)

ice0:


Keyikedalube:
@ice0 I see you are working hard on improving build time... :)

Why not just wait for C++20? Modules feature is coming soon. Messy #include headers issue will be solved with modules.

ice0:
@Keyikedalube here 2 problems:
1) Some decisions were made to support older OS and compilers (which have no support for C++20).
2) Unfortunately, modules are not a silver bullet. It can even slow down the build of complex projects like Synfig. You can check here: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1441r0.pdf

Keyikedalube:
Yup, the argument is correct. Experts all agree with a common reason being the way current and existing build tools and compilers are designed. There is a simlar debate about modules disadvantages for CMake (specifically buildsystems) topic arguing that the buildsystem would have to adapt to accomodate modules because right now it's incompatible with C++20 http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1427r0.pdf

I do agree with your stated arguments :) I only look/contribute at the GUI end... the rest is a shady area51 for me LOL

ice0:
No problem here. And your attention to such things is very helpful.

Issue #1754, opened 2020-10-09
Do not add "/usr/local/etc/" to the search locations for "synfig_modules.cfg" on Windows


ice0:
**Synfig version & platform**:
Latest master & Windows

**Issue description**:
```
synfig(12920) [10:01:07] warning: Cannot find 'synfig_modules.cfg', trying to load default modules
synfig(12920) [10:01:07] warning: Searched in the following directories:
synfig(12920) [10:01:07] warning:     ./synfig_modules.cfg
synfig(12920) [10:01:07] warning:     C:\Users\user\Downloads\SynfigStudio-1.3.16-testing-2020-10-08-win64-ec8d4/etc/synfig_modules.cfg
synfig(12920) [10:01:07] warning:     /usr/local/etc/synfig_modules.cfg
```

As you may have noticed, there is a *nix path (`/usr/local/etc/`) on Windows. This path is redundant and can be easily disabled using the `# ifdef` statement.

PR #1758, opened 2020-10-09
Don't search for linux paths on Windows for synfig_modules.cfg

AnshitaVishwa:
Fix issue #1754 

Synfig currently searches for` synfig_modules.cfg` in various paths at startup. It is also searching for it on a linux path 
`/usr/local/etc `even on a windows system. This PR adds an `#ifdef`  guard to disable the path on windows.


AYESDIE (approved):
Looks good to me. Nice job.

ice0:
Merged. Thank you!

Issue #1757, opened 2020-10-09
Resolve dangling else warning message


Keyikedalube:
**Issue description**:
https://github.com/synfig/synfig/blob/cde6d0a960a6bee84473fddcaa425fe6c67c9d64/synfig-studio/src/synfigapp/vectorizer/centerlinepolygonizer.cpp#L416-L421

The above outer if body is not scoped properly. Notice the inner if has its else part.
Without braces, the else could also belong to the outer if too depending on indentation level.

The compiler suggests explicitly adding `braces to avoid ambiguous 'else' [-Wdangling-else]`

Keyikedalube:
👉 Easy and Hacktoberfest 😃

DhairyaBahl:
@AYESDIE @Keyikedalube I would like to work on this issue but my previous 3 PRs are also pending. Kindly Review them.

Thank you :)

Keyikedalube:
@DhairyaBahl you already earned the Contributor status in this repo. Since this issue is too easy... please wait for this week and give chance to a newbie to take on this issue.

If there is no activity by next Monday, feel free to make a PR :)

DhairyaBahl:
@Keyikedalube Okay I got it :) I would also like to work on tougher issues as well. but first i have to deal with the previous PRs. I am getting quite confused between all the PRs.

Thank you :)

JoshEvan:
Hi @Keyikedalube, thank you for the opportunity,
I would like to try to solve this issue. Can I work on it?

Keyikedalube:
@JoshEvan Yes, go and make a PR :)

Happy hacking!

PR #1771, opened 2020-10-11
Add braces to avoid ambiguous else on extractBorders (fixes [-Wdangling-else] warning)

JoshEvan:
Changes: Adding braces on `if`-`else` statement on `extractBorders`
Why: To remove dangling else warning message according to issue #1757 
Impact: No dangling else warning message


ice0:
Merged. Thank you!

PR #1774, opened 2020-10-13
Replaced gtk::HBox and gtk::VBox with gtk::Grid in synfig-studio CanvasVeiw

### LINKED ISSUE #1764: Migrate from Gtk::H/VBox to Gtk::Grid for CanvasView
Building from source
Debian sid

**Issue description**:
CanvasView class uses `Gtk::HBox` and `Gtk::VBox` containers which are already deprecated long ago.

Migrate from old deprecated container box to `Gtk::Grid` which is pretty much stable.

Screenshot from 2020-10-10 10-45-31

The reason I'm not recommending `Gtk::Box` is because [packing API names](https://developer.gnome.org/gtk4/stable/ch40s02.html#id-1.7.4.4.21) have changed in Gtk4. And using Box over Grid would just add another maintenance flag to look out for in the future when synfig upgrades to gtkmm4.

With Grid implemented, these files **synfig/synfig-studio/src/gui/canvasview[.h,.cpp]** would be free of `Gtk::Box` API and improve synfigstudio stability.

fardragon:
Closes #1764

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -1997,7 +1997,7 @@ CanvasView::create_tab_label()
 	label->show();
 
 	closebutton = manage(new Gtk::Button());
-	closebutton->set_margin_right(4);
+	closebutton->set_margin_end(4);
 	grid->attach(*closebutton, 1, 0, 1, 1);
 	Gtk::Image* closebutton_image(manage(new Gtk::Image(
 			Gtk::StockID("gtk-close"),
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -1997,6 +1997,7 @@ CanvasView::create_tab_label()
 	label->show();
 
 	closebutton = manage(new Gtk::Button());
+	closebutton->set_margin_right(4);
 	grid->attach(*closebutton, 1, 0, 1, 1);
 	Gtk::Image* closebutton_image(manage(new Gtk::Image(
 			Gtk::StockID("gtk-close"),
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -49,7 +49,6 @@
 #include <gtkmm/separator.h>
 #include <gtkmm/eventbox.h>
 #include <gtkmm/label.h>
-#include <gtkmm/box.h>
 #include <gtkmm/grid.h>
 #include <gtkmm/menu.h>
 #include <gtkmm/menuitem.h>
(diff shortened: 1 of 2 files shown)


synfig-studio/src/gui/canvasview.cpp
```diff
@@ -905,10 +904,13 @@ CanvasView::create_time_bar()
 	//time_window_scroll->set_can_focus(true); // Uncomment this produce bad render of the HScroll
 	time_window_scroll->show();
 
-	timetrack = manage(new class Gtk::VBox());
-	timetrack->pack_start(*widget_kf_list);
-	timetrack->pack_start(*timeslider);
-	timetrack->pack_start(*time_window_scroll);
+	timetrack = manage(new Gtk::Grid());
+	time_window_scroll->set_hexpand(true);
+	timetrack->attach(*time_window_scroll, 0, 0, 1, 1);
```
Keyikedalube, line 909: We're not changing the value of the last two `attach(..., 1, 1)` parameters. Just remove those two parameters and let default argument values be used from the API instead.

That way code would be more readable :)

Keyikedalube: Same for all other `grid->attach()` functions

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1985,7 +1997,7 @@ CanvasView::create_tab_label()
 	label->show();
 
 	closebutton = manage(new Gtk::Button());
```
Keyikedalube, line 1999: While you are at this I'd appreciate if you could improve `closebutton` right margin (`set_margin_end()`) to value 4

Because we have this incomplete button display
Screenshot from 2020-10-14 07-01-27

synfig-studio/src/gui/canvasview.cpp
rodolforg: Actually @Keyikedalube , for now on Windows version, we need to set those last two parameters. Otherwise, it won't build :(

synfig-studio/src/gui/canvasview.cpp
fardragon: Fixed. Didn't happen on my gtk theme.

synfig-studio/src/gui/canvasview.cpp
fardragon: Unfortunately I don't have Windows machine handy to test this.

Keyikedalube: Understood :)

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1997,6 +1997,7 @@ CanvasView::create_tab_label()
...
+	closebutton->set_margin_right(4);
```
Keyikedalube, line 2000: `set_margin_right()` is [deprecated](https://developer.gnome.org/gtkmm/stable/classGtk_1_1Widget.html#ac9e58b824148641fea1fa41bfe257efb)... use
`set_margin_end()` instead

Keyikedalube (changes requested):

synfig-studio/src/gui/canvasview.cpp
Keyikedalube: Resolved...

Keyikedalube (approved):
All OK from my side :)

synfig-studio/src/gui/canvasview.cpp
Keyikedalube: Resolved...

ice0:
Merged. Thank you!

Issue #1775, opened 2020-10-13
Perf measurement tools (and other things)


define-private-public:
I want to move over some of the discussion from this ticket [this ticket](https://github.com/synfig/synfig/issues/1712) over here.
---

I've been poking around the Synfig source code and have found a few spots where instructions can be reordered & rewritten to better take advantage of SIMD optimizations.  To verify that it actually is the case, I wrote two python scripts.

The first does a few render passes on all of the .sif files from the [synfig-tests](https://gitlab.com/synfig/synfig-tests) repo, then dumps the results into a CSV file.  The second script will look at two CSV files and then chart them (using Pyplot) to do a basic comparison (e.g. mean run time, error variance, etc).  The idea is that you do a run first which acts as the "reference run".  Then you do some code changes and then do a comparison.  

Would you like me to submit a PR for these scripts or do you just want a gist of them?  I think they could be helpful for anyone else who wants to do perf profiling.  If you want to merge them, where should they be placed?

Right now the "measurement" script needs to be run from the `build` directory.  And the `synfig-tests` repo needs to be cloned into that directory as well.  The "comparison graph" script just needs to be supplied two CSV output files from the measurement script.

There are more changes that I want to add in to the scripts, but for an initial version, I'd like to get them in.
---

I did also make one (very) miniscule perf adjustment that I'd like to submit.  I'll put that in as it's own PR.  Do you need me to provide any proof (such as a chart) showing the performance measurement?

ice0:
Hi!

> If you want to merge them, where should they be placed?

I think you can make a folder `perf/scripts` and put them there.

> Do you need me to provide any proof (such as a chart) showing the performance measurement?

Yes, this will be awesome. Also this is very useful for future references.

morevnaproject:
Cool!

Keyikedalube:
Script needs to be updated for CMake because it's no longer `output/bin/synfig` but `output/CMAKE_BUILD_TYPE/bin/synfig`

PR #1778, opened 2020-10-15
Perf imps

define-private-public:
As mentioned in #1775, the perf improvement here is very miniscule, but I'd like to push in the two scripts I wrote that should help with measurement.  I plan on adding some more PRs that should be more significant.

So show you how tiny the change has been, here is some output from the `view_comparison_graph.py` script:
Figure_1

### DIFF
--- a/perf/scripts/requirements.txt
+++ b/perf/scripts/requirements.txt
@@ -1,0 +1,10 @@
+certifi==2020.6.20
+cycler==0.10.0
+kiwisolver==1.2.0
+matplotlib==3.3.2
+numpy==1.19.2
+Pillow==8.0.0
+pkg-resources==0.0.0
+pyparsing==2.4.7
+python-dateutil==2.8.1
+six==1.15.0
--- a/perf/scripts/test_render_all_perf.py
+++ b/perf/scripts/test_render_all_perf.py
@@ -1,0 +1,111 @@
+#!/usr/bin/python3
+#
+# This is a script that will run the synfig rendering multiple times on a set of
+# .sif files.  Record the time it took each pass to render, and then store those
+# readings into a .csv file.  To properly run this:
+#
+# 1. This script needs to be placed in the root of the build directory, and run
+#    from that directory
+# 2. The `synfig-tests` repo needs to be cloned (in the build directory as well.
+#    The repo is found here: https://gitlab.com/synfig/synfig-tests
+# 3. Don't have any applications running at the same time.  It can mess with the
+#    performance measurements.
+#
+# It can take an extremely long time to run (.e.g 25 mintues right now).  Before
+# trying to do any perf tweaking in the rendering code, make sure you generate
+# a reference/baseline to go off from.  This should be the latest commit on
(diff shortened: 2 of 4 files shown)


ice0:
Hi!
Nice to see you start from ColorMatrix class. Fixing it should improve performance greatly.

In my opinion, the main slowdown factor is the lack of batch processing of pixel data. Since it uses pointers to color functions, this results in an expensive function call for **every** pixel processed.

I can illustrate this with a small example:
**test.cpp**
```
#include <random>
#include <ctime>

inline int val2(int r, int g, int b) {
  return r + g + b*2;
}


inline int val(int r, int g, int b) {
  return r*2 + g + b;
}

int main() {
  const int count = 1024*1024*100;
  std::srand(std::time(nullptr)); // use current time as seed for random generator

  std::uint32_t out = 0;
  int r = std::rand();
  int g = std::rand();
  int b = std::rand();

  typedef int (*funcptr)(int, int, int);  // Declare typedef
  funcptr func = nullptr;    // Define function-pointer variable, and initialize
  
  // without std::rand compiler can do optimization here 
  if (std::rand() > 100) {
    func = &val;
  } else {
    func = &val2;
  }

  for (int i = 0; i < count; i++) {
  #ifdef PTR
     out += func(r, g, b);
  #else
     out += val(r, g, b);
  #endif
  }
}
```

test results:
**call a function for each pixel**
```
$ g++ ./test.cpp -O3 -DPTR
$ time ./a.exe

real    0m0.134s
user    0m0.000s
sys     0m0.000s
```

**function call now inlined**
```
$ g++ ./test.cpp -O3
$ time ./a.exe

real    0m0.032s
user    0m0.015s
sys     0m0.015s
```

**3x** speedup!

P.S. 
1) Maybe we need `requirements.txt` here (for python scripts)
2) Can you change spaces to tabs? (what's not critical, but whole project uses tabs)

ice0 (approved):

define-private-public:
So far I've only been making cursory glances on instructions/lines that could be moved around to take better advantage of of auto-vectorization.  I plan on doing more substantial changes, but I want to learn a bit more about the project before going a little more into adjusting the architecture of things.  I'm well aware of how pointer chasing and ruin performance.

1. I'll get you a requirements.txt soon,
2. I'll put the spaces into tabs for the C++ code.  Do you want it for the Python scripts as well?  Most people try to follow PEP8 and they recommend spaces (with an indent of 4) over tabs.

ice0:
> I'll put the spaces into tabs for the C++ code. Do you want it for the Python scripts as well? Most people try to follow PEP8 and they recommend spaces (with an indent of 4) over tabs.

Only for C++ code.

define-private-public:
Requirements file added and spaces->tabs fixed.

ice0:
Please check LGTM warning about `num_runs` variable. Is it used somewhere?

define-private-public:
whoops, that was a leftover form a previous iteration.  It's gone now.

ice0:
Merged. Thank you!

PR #1783, opened 2020-10-17
MSVC build improvements #2

ice0:
Continuing to fix MSVC build errors (Still)


rodolforg:
Well, the Widget_Sound depends on the MLT, and without it, it has no purpose. So maybe disable the widget_sound and its dock on MSVC building entirely instead of putting the MLT guard on every method?

ice0:
`soundprocessor.cpp` is used in bunch of places: `layer_sound.cpp`, `trgt_ffmpeg.cpp`, `gui/render.cpp`, etc. 
So I think it's better to disable MLT using `#define` as a temporary solution. Just in terms of changed lines of code :)

PR #1786, opened 2020-10-19
Remove function pointers from ColorMatrix

define-private-public:
In #1778 , @ice0  mentioned that removing function pointers from the `ColorMatrix` class could lead to some perf benefits.  I'll report that there have been some very miniscule improvements from this.  (I can post a comparison chart if requested)

This mostly simplifies the code (e.g. reducing unnecessary member functions in `ColorMatrix::BatchProcessor`).  The internal `transform()` function has also been rewritten to be a bit more SIMD friendly.  Though as stated before, these perf bumps have been miniscule.  I think it might be best to start looking at other objects to optimized.


ice0:
I think you are moving in right direction. For example, this is the rendering performance of the [example.zip](https://github.com/synfig/synfig/files/5399523/example.zip) file.

(I rendered in high resolution on purpose to see bottlenecks more clearly)
```
sudo perf record -g ./synfig ./example.sifz -w 4000 -h 4000 -t ffmpeg
```


**sudo perf report -G**
```
Samples: 217K of event 'cpu-clock:pppH', Event count (approx.): 54345750000
  Children      Self  Command  Shared Object             Symbol
+   82,85%     0,00%  synfig   synfig                    [.] process_job                                                                       ◆
-   63,40%    63,39%  synfig   libsynfig.so              [.] synfig::blendfunc_COMPOSITE<synfig::Color>                                        ▒
     63,38% process_job                                                                                                                        ▒
        synfig::blendfunc_COMPOSITE<synfig::Color>  
```

As you can see, it spends 63% of the time in the blending function.

P.S. I don't remember why, but it seemed to me that the color matrix had something to do with this code. Perhaps I just got confused with it.
Anyway, this is exactly the part I was talking about.

define-private-public:
Sure.  I can take a look at those functions next.  Are you interested in merging in these changes?  If not, I can close the PR.

ice0:
I suggest to leave it open for a time. 
First, I want to make a better example to test the speed of the color matrix.
Second, ColorMatrix is currently at the top of build time (due to templates), so I still have an idea for refactoring it.

define-private-public:
What you're idea on refactoring out the templates?  It is something that I noticed is that it does take quite a while to compile.

ice0:
I think maybe we can completely abandon templates here.

At least Skia doesn't use templates for they own ColorMatrix implementation (https://github.com/google/skia/blob/master/src/effects/SkColorMatrix.cpp).

define-private-public:
Okay.  But their ColorMatrix is a bit different than ours (e.g. no `o` field).

Is there some other part you think I should look into optimizing?  I've been meaning to do a Hotspot run and see where a bottleneck may me.

ice0:
> Is there some other part you think I should look into optimizing?

I think `synfig::blendfunc_XXX` is a good next target because it looks like it's used everywhere.

PR #1789, opened 2020-10-19
Improved css

### LINKED ISSUE #1061: Inconsistent Synfig UI on GNOME desktop
Building from master branch
Debian sid

**Issue description**:
On GNOME desktop Synfig UI and native UI looks quite inconsistent. Below is an example screenshot taken for *About program's window*:
Screenshot from 2019-12-31 22-22-12
Screenshot from 2019-12-31 22-26-12
Notice how the sizes of header buttons **About | Credits | Licence** for Synfig Studio doesn't match with the GNOME Terminal one
BTW the second screenshot is how all GTK apps look normally on GNOME desktop

Same is observed for file open dialog box:
Screenshot from 2019-12-31 22-21-58
Screenshot from 2019-12-31 22-44-16

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/86251218-inconsistent-synfig-ui-on-gnome-desktop?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1390: Some labels in dialog buttons are not correctly aligned
Synfig 1.3.13 on all platforms

**Issue description**:
In some dialogs the text labels are not correctly aligned.
For example in `Help > About Synfig Studio`, the button `Close` is aligned to the top:
button_label_alignment

Keyikedalube:
Fixes #1061
Fixes #1390

- Compile OS X specific code only in its environment
- Use classes to apply changes to multiple widgets
- Use ids to apply changes to specific widgets
- Don't style dialogs

Use [Windows](https://stackoverflow.com/questions/37035936/how-to-get-native-windows-decorations-on-gtk3-on-windows-7-and-msys2) gtk3 theme to have a proper global look and feel instead of modifying UI by hand.

In my opinion, synfigstudio main window toolbar buttons need to be scaled down for every platform for space efficiency. But not the other parts of GUI ie., dialogs and their buttons. These things should be left alone. Or let third-party themes do the trick. Like this [Windows 8](https://www.gnome-look.org/p/1013223/) theme.

<strike>I need help with styling timeslider. Couldn't figure out how to increase its height with CSS. Or is it even possible?</strike>

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -67,6 +67,8 @@
 #include <glibmm/thread.h>
 #include <glibmm/timer.h>
 
+#include <giomm/file.h>
+
 #include <gtk/gtk.h>
 #include <gdk/gdk.h>
 
@@ -2329,48 +2331,24 @@ App::apply_gtk_settings()
 	// enable menu icons
 	g_object_set (G_OBJECT (gtk_settings), "gtk-menu-images", TRUE, NULL);
 
-	// fix CSS
-	Glib::ustring data;
-	// Fix GtkPaned (big margin makes it hard to grab first keyframe))
-	data += "GtkPaned { margin: 2px; }\n";
-	// Fix #348: Synfig's Interface went Too Thick
-	// following css works in gtk since 3.14:
-	data += ".button                            { padding-left: 4px; padding-right: 4px; }\n";
-	data += ".button                            { padding-top: 0px; padding-bottom: 0px; }\n";
-	data += ".button *                          { padding-top: 4px; padding-bottom: 4px; }\n";
-	data += ".button > GtkBox                   { padding-top: 0px; padding-bottom: 0px; }\n";
-	data += ".button > GtkBox > *               { padding-top: 4px; padding-bottom: 4px; }\n";
-	data += ".button > GtkLabel                 { padding-top: 0px; padding-bottom: 0px; }\n";
-	data += "GtkComboBox > .button > GtkBox > * { padding-top: 0px; padding-bottom: 0px; }\n";
-	data += ".entry                             { padding-top: 0px; padding-bottom: 0px; }\n";
-	data += "progress, trough 					{ min-height: 20px; }\n";
-#if GTKMM_MAJOR_VERSION < 3 || (GTKMM_MAJOR_VERSION == 3 && GTKMM_MINOR_VERSION < 22)
-	// following css works in old versions of gtk
-	data += "button { padding: 0px; }\n";
-#else
-	// following css works for gtk 3.22:
-	data += "entry, spinbutton { min-height: 16px; }\n";
-	data += "button { min-height: 16px; min-width: 16px; padding: 0px; }\n";
-#endif
-	data += "button > box { padding: 5px; }\n";
-	data += "button > image { padding: 5px; }\n";
-	data += "combobox > box > button > box { padding-top: 0px; padding-bottom: 0px; }\n";
-	// Fix #810: Insetsetive context menus on OSX
-	g_object_get (G_OBJECT (gtk_settings), "gtk-theme-name", &theme_name, NULL);
-	if ( String(theme_name) == "Adwaita" )
-		data += ".window-frame, .window-frame:backdrop { box-shadow: none; margin: 0; }\n";
-	g_free(theme_name);
+	auto provider = Gtk::CssProvider::create();
+	auto screen   = Gdk::Screen::get_default();
+	auto css_file = Gio::File::create_for_path(ResourceHelper::get_css_path("synfig.css"));
 
-	if (!data.empty()) {
-		Glib::RefPtr<Gtk::CssProvider> css = Gtk::CssProvider::create();
-		try {
-			css->load_from_data(data);
-		} catch (Gtk::CssProviderError &e) {
-			synfig::warning("Failed to load css rules. %s", e.what().c_str());
-		}
-		Glib::RefPtr<Gdk::Screen> screen = Gdk::Screen::get_default();
-		Gtk::StyleContext::add_provider_for_screen(screen,css, GTK_STYLE_PROVIDER_PRIORITY_USER);
+#ifdef __APPLE__
+		g_object_get (G_OBJECT (gtk_settings), "gtk-theme-name", &theme_name, NULL);
+		if ( String(theme_name) == "Adwaita" )
+			css_file = Gio::File::create_for_path(ResourceHelper::get_css_path("synfig.mac.css"));
+		g_free(theme_name);
+#endif
+
+	try {
+		provider->load_from_file(css_file);
+	} catch (Glib::Error &e) {
+		synfig::warning("%s", e.what().c_str());
 	}
+
+	Gtk::StyleContext::add_provider_for_screen(screen, provider, GTK_STYLE_PROVIDER_PRIORITY_USER);
 }
 
 std::string
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -59,6 +59,7 @@
 #include <gtkmm/toggletoolbutton.h>
 #include <gtkmm/separatortoolitem.h>
 #include <gtkmm/menutoolbutton.h>
+#include <gtkmm/stylecontext.h>
 
 #include <glibmm/uriutils.h>
 #include <glibmm/convert.h>
@@ -596,6 +597,10 @@ CanvasView::CanvasView(etl::loose_handle<Instance> instance,etl::handle<CanvasIn
 	keyframe_dialog          (*App::main_window,canvas_interface_),
 	preview_dialog           ()
 {
+	// Make this toolbar small for space efficiency
+	auto style_context = this->get_style_context();
+	style_context->add_class("synfigstudio-efficient-workspace");
+
 	canvas_options = CanvasOptions::create(*App::main_window, this);
 
 	layer_tree=0;
@@ -944,7 +949,6 @@ CanvasView::create_time_bar()
 		Gtk::Image *icon = manage(new Gtk::Image(Gtk::StockID("synfig-animate_mode_off"), iconsize));
 		animatebutton = Gtk::manage(new class Gtk::ToggleButton());
 		animatebutton->set_tooltip_text(_("Turn on animate editing mode"));
-		icon->set_padding(0,0);
 		icon->show();
 		animatebutton->add(*icon);
 		animatebutton->signal_toggled().connect(sigc::mem_fun(*this, &CanvasView::toggle_animatebutton));
@@ -957,7 +961,6 @@ CanvasView::create_time_bar()
 		Gtk::Image *icon = manage(new Gtk::Image(Gtk::StockID("synfig-timetrack"), iconsize));
 		timetrackbutton = Gtk::manage(new class Gtk::ToggleButton());
 		timetrackbutton->set_tooltip_text(_("Toggle timebar"));
-		icon->set_padding(0,0);
 		icon->show();
 		timetrackbutton->add(*icon);
 		timetrackbutton->signal_toggled().connect(sigc::mem_fun(*this, &CanvasView::toggle_timetrackbutton));
@@ -1178,7 +1181,6 @@ CanvasView::create_display_bar()
 
 	{ // Setup render options dialog button
 		Gtk::Image *icon = Gtk::manage(new Gtk::Image(Gtk::StockID("synfig-render_options"), iconsize));
-		icon->set_padding(0, 0);
 		icon->show();
 
 		render_options_button = Gtk::manage(new class Gtk::ToolButton());
@@ -1194,7 +1196,6 @@ CanvasView::create_display_bar()
 
 	{ // Setup preview options dialog button
 		Gtk::Image *icon = Gtk::manage(new Gtk::Image(Gtk::StockID("synfig-preview_options"), iconsize));
-		icon->set_padding(0, 0);
 		icon->show();
 
 		preview_options_button = Gtk::manage(new class Gtk::ToolButton());
@@ -1233,7 +1234,6 @@ CanvasView::create_display_bar()
 
 	{ // Set up the show grid toggle button
 		Gtk::Image *icon = manage(new Gtk::Image(Gtk::StockID("synfig-toggle_show_grid"), iconsize));
-		icon->set_padding(0, 0);
 		icon->show();
 
 		show_grid = Gtk::manage(new class Gtk::ToggleToolButton());
@@ -1250,7 +1250,6 @@ CanvasView::create_display_bar()
 
 	{ // Set up the snap to grid toggle button
 		Gtk::Image *icon = manage(new Gtk::Image(Gtk::StockID("synfig-toggle_snap_grid"), iconsize));
-		icon->set_padding(0, 0);
 		icon->show();
 
 		snap_grid = Gtk::manage(new class Gtk::ToggleToolButton());
@@ -1270,7 +1269,6 @@ CanvasView::create_display_bar()
 
 	{ // Setup refresh button
 		Gtk::Image *icon = Gtk::manage(new Gtk::Image(Gtk::StockID("gtk-refresh"), iconsize));
-		icon->set_padding(0, 0);
 		icon->show();
 
 		refreshbutton = Gtk::manage(new class Gtk::ToolButton());
@@ -1305,7 +1303,6 @@ CanvasView::create_display_bar()
 
 	{ // Set up the background rendering button
 		Gtk::Image *icon = manage(new Gtk::Image(Gtk::StockID("synfig-toggle_background_rendering"), iconsize));
-		icon->set_padding(0, 0);
 		icon->show();
 
 		background_rendering_button = Gtk::manage(new class Gtk::ToggleToolButton());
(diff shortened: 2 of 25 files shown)


synfig-studio/src/gui/canvasview.cpp
```diff
@@ -596,6 +597,10 @@ CanvasView::CanvasView(etl::loose_handle<Instance> instance,etl::handle<CanvasIn
 	keyframe_dialog          (*App::main_window,canvas_interface_),
 	preview_dialog           ()
 {
+	// Make this toolbar small for space efficiency
+	auto style_context = this->get_style_context();
+	style_context->add_class("synfigstudio-efficient-workspace");
```
rodolforg, line 602: Do we need to add a class to the canvas view as a whole or just a specific toolbar?

synfig-studio/src/gui/app.cpp
```diff
@@ -2308,34 +2308,49 @@ App::apply_gtk_settings()
...
-	data += "combobox > box > button > box { padding-top: 0px; padding-bottom: 0px; }\n";
-	// Fix #810: Insetsetive context menus on OSX
+//	data += "entry, spinbutton { min-height: 16px; }\n";
+//	data += "button { min-height: 16px; min-width: 16px; padding: 0px; }\n";
+//#endif
+//	data += "button > box { padding: 5px; }\n";
+//	data += "button > image { padding: 5px; }\n";
+//	data += "combobox > box > button > box { padding-top: 0px; padding-bottom: 0px; }\n";
+
+	data += "#render-progress * { min-height: 20px; }\n";
```
rodolforg, line 2335: Why do we need to set this progress bar with a different style?

synfig-studio/src/gui/app.cpp
```diff
@@ -2308,34 +2308,49 @@ App::apply_gtk_settings()
...
+
+	data += ".synfigstudio-efficient-workspace button { padding: 0px; }\n";
+	data += ".synfigstudio-efficient-workspace button > image { padding: 5px; }\n";
+
+	data += "#low-resolution label { padding-left: 5px; padding-right: 5px; }\n";
```
rodolforg, line 2340: Why do we need to set this specific label with a different style?

synfig-studio/src/gui/app.cpp
rodolforg: Oh, ok. You just wrote in a different way what was already done.

synfig-studio/src/gui/canvasview.cpp
Keyikedalube: It's safe to set it as a whole. Otherwise, I'd have to duplicate these lines down at least two times inside CanvasView:
1. For top toolbar
2. For bottom toolbar
3. And additionally the tab bar (Wasn't styled earlier...)

Remember DRY? I think setting class as a whole is much better. We're styling only buttons and it's image.
```css
class .synfigstudio-efficient-workspace {
    button        : <value>;
    button > image: <value>;
}
```

synfig-studio/src/gui/app.cpp
Keyikedalube: When we set button padding to 0px, its text and icons would touch the edge of the button. Just look at GitHub "Resolve conversation" button below. Try to imagine it's text label without left/right padding. Now that looks bad because the information looks like it was cramped inside.

Anyway, what the previous code did was set the button padding to 0px and its image ie., button icons to 5px. And whoever wrote that code left or forgot to set padding for the text. So I took it up and fixed that :)

ice0:
I am thinking maybe making the css loadable from a file? 
This will allow anyone to edit the design without knowing the code.
In addition, it will allow you to switch design/theme, for example, from regular to compact.

synfig-studio/src/gui/app.cpp
rodolforg: Couldn't be done by removing the first line of your previous comment?
```
class .synfigstudio-efficient-workspace {
    button        : <value>;   # removing this line?
    button > image: <value>;
}
```

Keyikedalube: If we remove this line we're going to default back to GTK padding which is kinda too big :(
Plus we're unnecessarily adding more padding to button's image.

I'll send you the screenshot...

Keyikedalube:
Yeah, but there's OS X specific code about insensitive context menus. And I don't want to waste more time on build scripts like in my other PRs.

I'm think I'd have to modify both build scripts to intelligently insert OS X stylesheet code using configuration file (just like my previous attempt with `*.xml.in` file). I'm not really sure but I'm trying to save more time on my side here too. Hacking synfig is not the only thing I'm doing 24/7 :(

Anyway, if you think it's simple then I could try implementing loadable css file...

synfig-studio/src/gui/app.cpp
Keyikedalube: BTW. screenshots were taken with SynfigStudio window maximized

# Default GTK/Adwaita theme without our custom CSS
01 default gtk
# `button : <value>` removed
02 rodolforg
# Our current custom CSS applied
03 css

ice0:
I think for the first iteration, we can just do #ifdef for file loading function. For example:
```
#ifdef _WIN32
  load_style("synfig.win.css");
#elseif _APPLE
  load_style("synfig.mac.css");
#else 
  load_style("synifg.css");
#endif
```

I am not sure, but it looks like there aren't many CSS styles out there, so you don't have to worry about duplicating code. In total there will be 3 almost identical css files with some OS-specific elements.

synfig-studio/src/gui/app.cpp
rodolforg: Oh, I said it wrong :P 
I meant remove/reduce the padding only for image 'inside' button.
Well, anyway, there should be more issues :P Thank you for clarifying me.

ice0: Looks nice and clean!

rodolforg:
@ice0 , What about:
```c++
  load_style("synfig.css");
#ifdef _WIN32
  load_style("synfig.win.css");
#elseif _APPLE
  load_style("synfig.mac.css");
#else 
  load_style("synfig.lin.css");
#endif
```

Load a generic one, and maybe an OS-specific.

And, @Keyikedalube , you can add it in a new `share/css` folder, just like ui, sound, etc. (and provide an easy way to use it with that helper functions for path).

synfig-studio/src/gui/app.cpp
rodolforg: > Why do we need to set this specific label with a different style?

Why don't use a generic way like `button > label`?
I'm asking because I don't know, ok? :) Not to complain.

Keyikedalube:
Guys, I was reading this post

https://www.geeksforgeeks.org/how-to-include-one-css-file-in-another/

Write once and import :)

No need to duplicate code by hand multiple times...

:+1: :+1: :+1:  your suggestions. Thanks!

ice0:
Looks like Gnome CSS also support this feature (https://developer.gnome.org/gtk3/stable/chap-css-overview.html)
Awesome!

synfig-studio/src/gui/app.cpp
Keyikedalube: Wait I didn't realize I made a mistake here
```css
class .synfigstudio-efficient-workspace {
    button        : <value>;
    button > image: <value>;
}
```
LOL it's
```css
.synfigstudio-efficient-workspace  button  {
    padding: <value>;
```
```css
.synfigstudio-efficient-workspace button > image {
    padding: <value>;
}
```

Sorry, I was a little burned out :P I didn't review my own comments

Hope our conversation doesn't get any more complicated :smile: :rofl:

Keyikedalube:
Autotools build might fail... And if it does I need help LOL

Keyikedalube:
Need help with why OS X build is failing...

synfig-studio/src/gui/app.cpp
```diff
@@ -2332,36 +2332,32 @@ App::apply_gtk_settings()
...
 #if __APPLE__
-	// Fix #810: Insensitive context menus on OSX
-	g_object_get (G_OBJECT (gtk_settings), "gtk-theme-name", &theme_name, NULL);
-	if ( String(theme_name) == "Adwaita" )
-		data += ".window-frame, .window-frame:backdrop { box-shadow: none; margin: 0; }\n";
-	g_free(theme_name);
+		g_object_get (G_OBJECT (gtk_settings), "gtk-theme-name", &theme_name, NULL);
+		if ( String(theme_name) == "Adwaita" )
+			auto css_file = Gio::File::create_for_path(ResourceHelper::get_css_path("synfig.mac.css"));
+		g_free(theme_name);
```
Keyikedalube: Maybe it's because of those lines

My guess is Adwaita condition check is getting skipped resulting in css_file being undeclared

Should I remove the theme code?

ice0: The problem is that `css_file` gets destroyed when it goes out of scope from `if` condition. So you need to declare it before `if`.

rodolforg: So .mac.css would be called only if theme is Adwaita?

Keyikedalube: > So .mac.css would be called only if theme is Adwaita?

Yes. But I don't have a Mac with me so someone should test it on other themes and not only on Adwaita.

@ice0 Thanks for the scope :) Me personally I don't want to keep that if guard because the toolbar buttons might not be the same for other themes. And Apple users might complain synfig UI has become too thick again when other themes are applied.

@morevnaproject wrote that code I think. I read his post on [gtk mailing lists](http://gtk.10911.n7.nabble.com/Gtk3-MacOS-OSX-context-menu-issues-td86576.html)

My opinion... test if OS X has insensitive context menus issue for other GTK themes. If so then remove the if part and let OS X apply the generic theme like all other platforms.

ice0: I agree. As temporary solution, I suggest to leave this check as-is, so we can merge your changes, and add an issue to check and fix that problem.

For out-of-scope variable. You can make a reorder, to fix this:
```
auto css_file = Gio::File::create_for_path(ResourceHelper::get_css_path("synfig.css"));

#ifdef __APPLE__
		g_object_get (G_OBJECT (gtk_settings), "gtk-theme-name", &theme_name, NULL);
		if ( String(theme_name) == "Adwaita" )
			css_file = Gio::File::create_for_path(ResourceHelper::get_css_path("synfig.mac.css"));
		g_free(theme_name);
#endif
```

Keyikedalube: Won't the variable `css_file` names conflict in case Adwaita check turns out correct?
Something like this compiler error
> Variable already declared here...

Keyikedalube: @ice0 What about like this?
 ```
#ifdef __APPLE__
		Glib::RefPtr<Gio::File css_file;
		g_object_get (G_OBJECT (gtk_settings), "gtk-theme-name", &theme_name, NULL);
		if ( String(theme_name) == "Adwaita" )
			css_file = Gio::File::create_for_path(ResourceHelper::get_css_path("synfig.mac.css"));
		g_free(theme_name);
#endif
```

Comment (Keyikedalube) [reply synfig-studio/src/gui/app.cpp]: Proceeding with your suggestion anyway :) Mine's still flawed in some cases...

Comment (ice0) [reply synfig-studio/src/gui/app.cpp]: > Won't the variable css_file names conflict in case Adwaita check turns out correct?

No, it just replace value. As you can notice I removed the ʻauto` keyword.

Comment (Keyikedalube) [reply synfig-studio/src/gui/app.cpp]: Noticed OS X build succeeded on Travis :)

Now, pending tasks left are:
- padding timeslider
- removing commented code after final review's done

File: synfig-studio/src/gui/resources/css/CMakeLists.txt [hunk 5]
```diff
@@ -0,0 +1,49 @@
...
+# Prefix all CSS_FILES list with destination directory for
+#
+# add_custom_command(OUTPUT ...)
+#
+set(CSS_FILES_OUTPUT)
+foreach(CSS_FILE ${CSS_FILES})
+	list(APPEND CSS_FILES_OUTPUT ${SYNFIG_BUILD_ROOT}/${INSTALL_DESTINATION}/${CSS_FILE})
+endforeach()
+
+# file(COPY..) has limitations
```
ice0, line 25: You can use `configure_file` here with COPYONLY (https://cmake.org/cmake/help/latest/command/configure_file.html)

Keyikedalube: Doesn't work :(

Tried this
```cmake
configure_file(${CSS_FILES} ${SYNFIG_BUILD_ROOT}/${INSTALL_DESTINATION}/${CSS_FILES} COPYONLY)
```
and this...
```cmake
configure_file(${CSS_FILES_INPUT} ${CSS_FILES_OUTPUT} COPYONLY)
```

*No such file or directory* exception was caught when building and running

ice0: This example works for me:
```
set(INSTALL_DESTINATION "share/synfig/css")
set(CSS_FILES
    synfig.css
    synfig.mac.css
)

foreach(CSS_FILE ${CSS_FILES})
    configure_file(${CSS_FILE} ${INSTALL_DESTINATION}/${CSS_FILE} COPYONLY)
endforeach()
```

Keyikedalube: Thanks! The earlier didn't work because I tried to copy a list. Looping each item is much better than my lengthy custom target style :P

synfig-studio/src/gui/docks/dock_layergroups.cpp
```diff
@@ -67,6 +69,10 @@ Dock_LayerGroups::Dock_LayerGroups():
 	action_group_group_ops(Gtk::ActionGroup::create("action_group_dock_layergroups")),
 	group_action_manager(new GroupActionManager)
 {
+	// Make Sets toolbar buttons small for space efficiency
+	auto context = get_style_context();
```
ice0, line 73: Since the context is not used further, it can be simplified (just a suggestion):
```
get_style_context()->add_class("synfigstudio-efficient-workspace");
```

ice0 (approved):

synfig-studio/src/gui/docks/dock_layergroups.cpp
Keyikedalube: Sounds legit... Tried it and encountered this
Screenshot from 2020-10-23 06-28-42
Maybe we're suppose to wait for the object to be returned and assigned to the variable first...

Keyikedalube:
If I'm not able to figure out how to fix the time slider padding by this week (today and tomorrow)
I'm marking this PR as "Ready for review"
It's draining all my time away

BTW, here are some things I tried:
- Manually set the height from widget_timeslider (was 20, tried 40)
*But this is risky because the last time I tried to fix this issue #1145 I broke other widgets; graph and sound*

- `Gtk::DrawingArea` inherits `Gtk::Widget` so I tried manipulating min-height using CSS

Both of them didn't work

### Anyway, here's how the time slider looks now after this PR:
Screenshot from 2020-10-23 06-44-53

### And before this PR:
Screenshot from 2020-10-23 06-45-24

synfig-studio/src/gui/docks/dock_layergroups.cpp
rodolforg: Why are you assigning the (void) returned by add_class to a variable? :)

rodolforg:
> If I'm not able to figure out how to fix the time slider padding by this week (today and tomorrow)
> I'm marking this PR as "Ready for review"
> It's draining all my time away
> 
> BTW, here are some things I tried:
> 
>     * Manually set the height from widget_timeslider (was 20, tried 40)
>       _But this is risky because the last time I tried to fix this issue #1145 I broke other widgets; graph and sound_
> 
>     * `Gtk::DrawingArea` inherits `Gtk::Widget` so I tried manipulating min-height using CSS
> 
> 
> Both of them didn't work
> ### Anyway, here's how the time slider looks now after this PR:
> 
> Screenshot from 2020-10-23 06-44-53
> ### And before this PR:
> 
> Screenshot from 2020-10-23 06-45-24

AS far as I remember, the height of timeslider is related to the height of the header of Parameters panel treeview.

rodolforg:
And that is reason indeed. 
Captura de tela de 2020-10-23 17-06-00
 See? They're aligned.

The treeview header has buttons (for column sort) and somehow they style were affected.
Captura de tela de 2020-10-23 17-08-02
(picture from GtkInspector)

rodolforg:
Will all commits be squashed?

Keyikedalube:
> Will all commits be squashed?

Yes

And if I remember correctly, Telegram repo has a policy of keeping only one commit per PR. Contributors are supposed to squash their commits.
I prefer their style too. [One PR one commit](https://github.com/telegramdesktop/tdesktop/blob/dev/.github/CONTRIBUTING.md#squash-your-commits-to-a-single-commit).
But that depends on how other projects want to organize their changes :)

Keyikedalube:
> And that is reason indeed.
> Captura de tela de 2020-10-23 17-06-00
> See? They're aligned.
> 
> The treeview header has buttons (for column sort) and somehow they style were affected.
> Captura de tela de 2020-10-23 17-08-02
> (picture from GtkInspector)

And I thought OOP design was to compartmentalize...

Those two are connected and I would never have suspected that strange behavior.

Anyway thanks for the heads up! Investigating right away...

rodolforg:
They are related to sync scroll of parameters panel with timetrack

Em sex, 23 de out de 2020 21:01, Keyikedalube Ndang <
notifications@github.com> escreveu:

ice0:
Merged. Thank you!

Issue #1785, opened 2020-10-18
[MSYS/CMake] Build fails if msys64 is not installed to c:\msys64


Svarov-RZM:
**Issue description**
Because of hard-coded path "c:/msys64/usr/bin/perl" in "synfig.git\cmake\SynfigIntltool.cmake build fails if MSYS is installed to a different place (drive H: in my case).

I made a quick patch to address this problem. It determines current WIN32 path to msys64 directory in 2-build-msys-cmake.sh script and then passes it to cmake through '-D' argument. I will send a pull request if this is a correct way to address this issue (my understanding of cmake is very basic).

ice0:
```
# When MinGW is used without MSYS environment (for example, Ninja
# generator uses the standard Windows `cmd` shell for building), `intltool-merge`
# cannot be started, because it is just a Perl script, and `cmd` does not know how
# to run it. In this case, we need to explicitly add the interpreter to the command.
if(INTLTOOL_MERGE_EXECUTABLE AND MSYS)
    message(STATUS "Fixing intltool-merge command...")
    set(INTLTOOL_MERGE_EXECUTABLE c:/msys64/usr/bin/perl ${INTLTOOL_MERGE_EXECUTABLE})
endif()
```

I think there should be `NOT MSYS`. This way it will patch the intltool path only if MSYS is not found.
I broke it in this commit: https://github.com/synfig/synfig/commit/42dd20a6bfa683c84787bb0e286d9816d55b37da

Svarov-RZM:
@ice0 
I think the current logic is correct. The problem is that ninja itself calls cmd.exe from MSYS environment for processing external commands instead of using bash or call it directly. Cmd.exe, in turn, know nothing about perl scripts, so we need to specify full path to the interpreter.

I sent patch I was talking about that addresses this issue.

ice0:
Yes, you are right.
In this situation, is better to detect MSYS path by using environment variables. They are already set, if you run CMake from MSYS shell.

So here the place where you can define MSYS_ROOT:

**synfig/CMakeLists.txt**
```
# MSYS is defined for MSYS Makefiles generator and undefined for Ninja (for example)
# But if you run CMake from MSYS shell, we can still detect MSYS by environment variables (MINGW_PREFIX)
# https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-To-Write-Platform-Checks
if(MINGW AND NOT MSYS)
[...]

PR #1790, opened 2020-10-19
[MSYS/CMake] Fix hard-coded path to MSYS directory

Svarov-RZM:
Fixes https://github.com/synfig/synfig/issues/1785
The command 'pwd -W' used in script should be able to return proper path from any place inside of MSYS environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ endif()
 if(MSVC)
 	add_definitions(-D_USE_MATH_DEFINES) # <cmath> header need this define to add M_PI define
 	add_definitions(-DNOMINMAX) # windows.h defines min/max and brokes std::min/max
+	add_definitions(-DWITHOUT_MLT) # Disable MLT for MSVC build (as there are no such package yet)
 
 	## /MP Enable multi-processor compilation (faster)
 	add_compile_options(/MP /W0 /bigobj)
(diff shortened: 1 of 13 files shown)


2-build-msys-cmake.sh
```diff
@@ -1,7 +1,12 @@
 #!/usr/bin/env bash
 
+# Determine WIN32 path to msys64 directory
+# This is needed in case msys64 is installed to a different
+# place rather than usual c:\msys64, say H:\Projects\msys64
+MSYS_ROOT=`pwd -W|grep -o .*/msys64`
```
ice0, line 6: This will not work, if you install MSYS for example to `c:\msys2` folder.

AYESDIE:
I think this issue can be easily resolved using `MINGW_PREFIX`, which can be set for Appveyor CI as [here](https://github.com/synfig/synfig/blob/7538ae5b9de8f0fa9e880512f1910dbcda924d17/appveyor.yml) and can be used inside CMake as [here](https://github.com/synfig/synfig/blob/ff4053293e4593865cb983238ce0be3108c8cbf8/cmake/InstallMSYS2.cmake). Although, I'm unsure why it was changed back to that.

ice0:
@AYESDIE Agree. I commented here: https://github.com/synfig/synfig/issues/1785#issuecomment-712569609

PR #1792, opened 2020-10-19
remove old CVS ID field from layers, importers,

rodolforg:
targets and actions.

This breaks ABI, but we don't let user link Studio with
old library versions anyway.
Any non-embeded dll/so module must be rebuild...

$Id$ was replaced in CVS/SVN checkouts with revision identifier,
but this doesn't work on git and alike. Well, it could work,
but not how it was intended (with ident keyword in .gitattributes).

We could set a script to replace those $Id$ with current commit hash,
but AFAIK it isn't done for years now.

Anyway, we have SYNFIG_LAYER_SET_VERSION(), ACTION_SET_VERSION(),
SYNFIG_TARGET_SET_VERSION(), etc. to define resource versioning.
I think it's enough.


ice0 (approved):

ice0:
Merged. Thank you!

PR #1811, opened 2020-10-28
Improved redo shortcut for Linux

### LINKED ISSUE #1005: Ctrl+Shift+Z is redo command on Linux but on Synfig it's Zoom to Fit
1.3.11

**Issue description**:
On Linux the keyboard combination Ctrl+Shift+Z is for redo action.

Got used to redo-ing previous undo actions using that keyboard shortcut after using a GNOME Linux distro *(Some GNOME apps register only Ctrl+Shift+Z as the redo command and not Ctrl+Y)*
This shortcut works on XFCE desktop too.

Don't you think Synfig Zoom to Fit keyboard shortcut is conflicting with the Linux standard redo shortcut key?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/83574004-ctrl-shift-z-is-redo-command-on-linux-but-on-synfig-it-s-zoom-to-fit?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Keyikedalube:
Also changed old redo shortcut from `Ctrl+r` to `Ctrl+y`

Closes #1005

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -1363,8 +1363,12 @@ DEFINE_ACTION("keyframe-properties", _("Properties"));
 	ACCEL("<Control>8",								"<Actions>/canvasview/quality-08"					);
 	ACCEL("<Control>9",								"<Actions>/canvasview/quality-09"					);
 	ACCEL("<Control>0",								"<Actions>/canvasview/quality-10"					);
-	ACCEL("<Control>z",								"<Actions>/action_group_dock_history/undo"				);
-	ACCEL("<Control>r",								"<Actions>/action_group_dock_history/redo"				);
+	ACCEL("<Primary>z",        "<Actions>/action_group_dock_history/undo");
+#if _WIN32
+	ACCEL("<Control>y",        "<Actions>/action_group_dock_history/redo");
+#else
+	ACCEL("<Primary><Shift>z", "<Actions>/action_group_dock_history/redo");
+#endif
 	ACCEL2(Gtk::AccelKey(GDK_KEY_Delete,Gdk::ModifierType(),				"<Actions>/action_group_layer_action_manager/action-LayerRemove"	));
 	ACCEL2(Gtk::AccelKey('(',Gdk::CONTROL_MASK,					"<Actions>/canvasview/decrease-low-res-pixel-size"			));
 	ACCEL2(Gtk::AccelKey(')',Gdk::CONTROL_MASK,					"<Actions>/canvasview/increase-low-res-pixel-size"			));


synfig-studio/src/gui/app.cpp
```diff
@@ -1364,7 +1364,11 @@ DEFINE_ACTION("keyframe-properties", _("Properties"));
 	ACCEL("<Control>9",								"<Actions>/canvasview/quality-09"					);
 	ACCEL("<Control>0",								"<Actions>/canvasview/quality-10"					);
 	ACCEL("<Control>z",								"<Actions>/action_group_dock_history/undo"				);
-	ACCEL("<Control>r",								"<Actions>/action_group_dock_history/redo"				);
+#if __gnu_linux__
+	ACCEL("<Control><Shift>Z",                      "<Actions>/action_group_dock_history/redo");
```
ice0, line 1368: Is any difference here between lower-case 'z' and upper-case 'Z'? Asking because as I can see lower-case is used before.

Keyikedalube: Uh no...

I used upper case because of my habit with newer GTKMM versions. Lower case doesn't work but in this PR both works OK. I'm compiling against newer GTKMM 3.24 so...

But if you want consistency I could change that :)

ice0: I think that's okay then, because we'll come back to this code again. (For Cmd/Ctrl key for MacOS)

rodolforg: Wouldn't be better to use `#if defined(_WIN32) || defined(__APPLE__)`?
I don't know if `__gnu_linux__` is for clang compiler as well.
A random article says `__linux__` would be enough, but I don't know :P
https://blog.kowalczyk.info/article/j/guide-to-predefined-macros-in-c-compilers-gcc-clang-msvc-etc..html

ice0: @rodolforg You are right. For MacOS redo is Cmd+Shift+Z (https://osxdaily.com/how-undo-redo-mac/), so using `#ifdef _WIN32` is a way  better.

ice0: @Keyikedalube maybe also check `<Primary>` key here? (for Cmd/Ctrl+Shit+Z) If you make a change, I can test it on MacOS.

Keyikedalube: @rodolforg I referenced header macro from this website https://sourceforge.net/p/predef/wiki/OperatingSystems/

Keyikedalube: @ice0 <Primary> works on my side. It's mapped to Control key for Linux

Keyikedalube:
Build passed. Squashing my commits.

synfig-studio/src/gui/app.cpp
rodolforg: @Keyikedalube your link also says `__linux__` is valid XD.
Anyway, I believe it's better to use synfig 'standard' way: first the windows case, else (if necessary) apple case, else anything else (linux, openbsd, etc.)

@ice0 why `<Primary>` here only? Mac user will be confused: `ctrl+z` undo, `ctrl+y` or `ctrl+shift+z` does not anything.

ice0: @rodolforg just for testing purposes. Just checked - all works as expected (on MacOS <Primary> will become Command key). So I suggest 2 changes to this PR.
1) Change ifdef as suggested by Rodolfo:
```
#if _WIN32
	ACCEL("<Control>y",                             "<Actions>/action_group_dock_history/redo");
#else
	ACCEL("<Primary><Shift>z",                      "<Actions>/action_group_dock_history/redo");
#endif
```
2. Replace <Control> to <Primary> for Undo (Ctrl+Z)

After that we can switch all other shortcuts.
File:
- New (Ctrl+N)
- Open (Ctrl+O)
- Save (Ctrl+S)
- Save As (Shift+Ctrl+S)
- Import (Ctrl+I)
- New (Ctrl+N)

Edit:
- Select All Layers (Shift+Ctrl+A)
- Unselect All Layers (Shift+Ctrl+D)
- Select All Handles (Ctrl+A)
- Unselect All Handles (Ctrl+D)

etc...

Keyikedalube: Working on it...

ice0:
Merged. Thank you!

PR #1813, opened 2020-10-28
Lose entry focus on mouse leave event 

### LINKED ISSUE #1808: Make toolbar spinbuttons and entries lose focus on mouse leave event
Building from source
Debian sid

**Issue description**:
Screenshot from 2020-10-26 17-10-05

When these buttons grab focus, single key canvas navigation (`.` one frame forward and `,` one frame backward) shortcuts become a bit difficult to access.

With focus on and off feature using mouse enter/leave event, users don't have to worry their keyboard shortcuts losing focus to another widget.

I tried implementing this behavior in my sample program [over here](https://stackoverflow.com/questions/64535307/turn-widgets-focus-on-and-off-based-on-mouse-enter-leave-event). Hope this behavior is achievable for synfig too :)

Keyikedalube:
Closes #1808

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -975,6 +975,8 @@ CanvasView::create_time_bar()
 	current_time_widget->signal_value_changed().connect(
 		sigc::mem_fun(*this,&CanvasView::on_current_time_widget_changed)
 	);
+	current_time_widget->signal_leave_notify_event().connect(
+		sigc::mem_fun(*this, &CanvasView::on_leave_notify_event));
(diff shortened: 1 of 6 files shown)


rodolforg:
What if user is typing in some parameter or another text entry/area and accidentally (or not) move mouse around and pass over those spinbuttons?

Keyikedalube:
We could turn off enter event then and just keep the leave event

I'll test your expected behavior first before that

Keyikedalube:
*Reason explained [here](https://github.com/synfig/synfig/issues/1808#issuecomment-719505586)*

PR #1815, opened 2020-10-30
[Autotools] Verbose mode is disabled by default

ice0:
Build output is much cleaner now when using autotools


ETL/configure.ac
```diff
@@ -5,6 +5,8 @@ AC_PREREQ(2.60)
 AC_INIT([Extended Template Library],[1.5.0],[https://github.com/synfig/synfig/issues],[ETL])
 AC_REVISION
```
ice0, line 6: No brackets, very strange

PR #1821, opened 2020-11-06
restore ctrl-c ctrl+v ctrl+a in Color and Gradient Dialogs

### LINKED ISSUE #1713: Copy paste keyboard UX issue
Building from source
Debian sid

**Issue description**:
1. Draw a rectangle & change its color to red
2. Try to copy the hex color value using **Ctrl+C**
3. Draw a circle and bring up its color properties dialog box
4. Paste **Ctrl+V**

Another rectangle layer gets pasted. I think this is due to bad input handling design.
When a user highlights the color code and hits Ctrl+C, the color should be copied because user's focus is on the dialog box now not the canvas workspace.

*Right-clicking copy and paste works but that's tedious. Keyboard saves a lot more time than mouse*

rodolforg:
fix #1713

As a consequence, application shortcuts/hotkeys will only work
when those dialogs are not on focus, but this behaviour
should be expected by user.

### DIFF
--- a/synfig-studio/src/gui/dialogs/dialog_color.cpp
+++ b/synfig-studio/src/gui/dialogs/dialog_color.cpp
@@ -69,7 +69,6 @@ Dialog_Color::Dialog_Color():
 			sigc::mem_fun(*this, &Dialog_Color::on_set_fc_pressed));
 	create_close_button();
 
-	add_accel_group(App::ui_manager()->get_accel_group());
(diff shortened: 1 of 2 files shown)


rodolforg:
If approved, please cherry-pick to 1.3.x branch

ice0:
Merged. Thank you!
ping @morevnaproject

Issue #1794, opened 2020-10-20
Improve tool items keyboard shortcuts


Keyikedalube:
**Issue description**:
- New single key shortcuts proposal
[draft document](https://synfig-docs-dev.readthedocs.io/en/latest/projects/shortcuts.html)

- Keynames code for reference
[gdk keynames](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gdk/keynames.txt)

- File to modify
https://github.com/synfig/synfig/blob/master/synfig-studio/src/gui/app.cpp

Using menu mnemonics for tool items is causing conflicting keyboard shortcuts. Some were discussed here #1336

Single key shortcuts are based on Inkscape model and some proposed shortcuts are already resolved:
- **Zoom in and Zoom out**
New improved zoom shortcuts implemented here #1769

While some need to be adapted to newer shortcuts:
- **Scale and Rotate**
New navigation shortcuts use those keys now #1776

You can propose yours below and could be discussed with project contributors :)
Though the final approval would have to be made by @morevnaproject

Keyikedalube:
:point_right: Easy and Hacktoberfest :smile:

Keyikedalube:
I guess it's OK now to remove the Hacktoberfest label. The event's over in 2 hrs (over here).

BTW, I feel this issue is deserving of "good first issue" :)

ice0:
Done :)

DhairyaBahl:
@ice0 I would like to work on this issue but should I ?? its good first issue so i ideally i should leave it for the new contributors. I dunno, plz tell me. 

Thanks:)

ice0:
@DhairyaBahl Sure, just take it. There is a lot of interesting issues for a new contributors and we can add more later :)

DhairyaBahl:
@ice0 @keyikedalube 

I referred to [draft document](https://synfig-docs-dev.readthedocs.io/en/latest/projects/shortcuts.html) and there is no mention of use of the alt key. So, Do I have to use the alt key or just key ???

Kindly tell me Thanks :)

Keyikedalube:
The goal is to implement single key behavior for synfig tool items just like Inkscape and GIMP.

Yes, you are going to use just the key.

Alt key is reserved for menu and button mnemonics.

DhairyaBahl:
> The goal is to implement single key behavior for synfig tool items just like Inkscape and GIMP.
> 
> Yes, you are going to use just the key.
> 
[...]

PR #1823, opened 2020-11-07
Improving Keyboard Shortcuts for the toolbox items

DhairyaBahl:
@ice0 This is the PR for the issue #1794 and in this i almost fixed every keyboard shortcut which i could with inkscape except few which i couldn't find or couldn't understand like mirror vertical and mirror horizontal. they must be named as something else. Kindly tell me that and I will fix that also. or you can check these changes first and fully merge them into code base  then we will move to the ones which were left. Your Call :)

Kindly review this PR . Thank You

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -1305,23 +1305,23 @@ DEFINE_ACTION("keyframe-properties", _("Properties"));
 	// the toolbox
 	ACCEL("s",								"<Actions>/action_group_state_manager/state-normal"			);
 	ACCEL("m",								"<Actions>/action_group_state_manager/state-smooth_move"		);
-	ACCEL("<Mod1>s",								"<Actions>/action_group_state_manager/state-scale"			);
-	ACCEL("<Mod1>t",								"<Actions>/action_group_state_manager/state-rotate"			);
-	ACCEL("<Mod1>m",								"<Actions>/action_group_state_manager/state-mirror"			);
+	ACCEL("l",								"<Actions>/action_group_state_manager/state-scale"			);
+	ACCEL("a",								"<Actions>/action_group_state_manager/state-rotate"			);
+	ACCEL("i",								"<Actions>/action_group_state_manager/state-mirror"			);
 	ACCEL("e",								"<Actions>/action_group_state_manager/state-circle"			);
 	ACCEL("r",								"<Actions>/action_group_state_manager/state-rectangle"			);
 	ACCEL("asterisk",								"<Actions>/action_group_state_manager/state-star"			);
 	ACCEL("g",								"<Actions>/action_group_state_manager/state-gradient"			);
-	ACCEL("<Mod1>p",								"<Actions>/action_group_state_manager/state-polygon"			);
+	ACCEL("o",								"<Actions>/action_group_state_manager/state-polygon"			);
 	ACCEL("b",								"<Actions>/action_group_state_manager/state-bline"			);
-	ACCEL("<Mod1>o",								"<Actions>/action_group_state_manager/state-bone"			);
+	ACCEL("n",								"<Actions>/action_group_state_manager/state-bone"			);
 	ACCEL("t",								"<Actions>/action_group_state_manager/state-text"			);
 	ACCEL("u",								"<Actions>/action_group_state_manager/state-fill"			);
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/app.cpp
```diff
@@ -1303,21 +1303,21 @@ DEFINE_ACTION("keyframe-properties", _("Properties"));
...
+	ACCEL("r",								"<Actions>/action_group_state_manager/state-rectangle"			);
+	ACCEL("asterisk",								"<Actions>/action_group_state_manager/state-star"			);
+	ACCEL("g",								"<Actions>/action_group_state_manager/state-gradient"			);
 	ACCEL("<Mod1>p",								"<Actions>/action_group_state_manager/state-polygon"			);
-	ACCEL("<Mod1>b",								"<Actions>/action_group_state_manager/state-bline"			);
+	ACCEL("b",								"<Actions>/action_group_state_manager/state-bline"			);
 	ACCEL("<Mod1>o",								"<Actions>/action_group_state_manager/state-bone"			);
-	ACCEL("<Mod1>x",								"<Actions>/action_group_state_manager/state-text"			);
+	ACCEL("t",								"<Actions>/action_group_state_manager/state-text"			);
 	ACCEL("<Mod1>f",								"<Actions>/action_group_state_manager/state-fill"			);
```
Keyikedalube, line 1319: state-fill is bucket 👉🏽 `u`

Keyikedalube (changes requested):
Change the fill accel key

Keyikedalube:
Tools to be discussed:
- Mirror
- Draw
- Cutout
- Width
- Sketch
- Scale
- Rotate
- Polygon

DhairyaBahl:
@Keyikedalube I changed the fill accel key but what should we do about those To Be Discussed Items that you mentioned above. 

Kindly tell me so that I can work further if required. Thanks :)

Keyikedalube:
> what should we do about those To Be Discussed Items

Well, those tool items still need to be changed :) Look at the draft document again and try to come up with new accel keys for those items. Some were already proposed. Since this PR already implements keyboard shortcuts behavior similar to Inkscape, you can fire up Inkscape on your computer and tinker around with it tools (left sidebar) just to get some idea 😄

DhairyaBahl:
@Keyikedalube Okay I am on it :) 

Thanks :)

DhairyaBahl:
@Keyikedalube I was working with inkscape and
1 we have to correct the documentation I guess because d key starts the drawing tool. 
2.i think rest of the shortcuts that are proposed in the documentation are usable and we should go with them.
although i you guys want, I can try to make my own proposal of changes, no issues in that. 
3.as i can see in inkscape, polygons and stars both have same shortcut may be ! its written stars and polygons
4. Kindly tell me what should i do next. apply those keys from the documentation,or propose my own. 

Thanks :)

Keyikedalube:
> we have to correct the documentation

Don't worry about documentation because once this PR is done, that page has to be deleted anyway.

> p key starts the drawing tool

Thanks for noticing that :) Synfig draw tool can use `p`

> i think rest of the shortcuts that are proposed in the documentation are usable and we should go with them

Yes, but try to reserve non-letter characters for canvas operation OK. Because now we use period `.` and comma `,` for frame seeking and then there's bracket left `[` for keyframe prev and bracket right `]` for next keyframe. See? non-letter characters are predominantly used for canvas operation already.
Try to come up shortcut key `a-z` for those remaining **to be discussed** lists

Keyikedalube:
Here's my proposal for:
- scale `l` *<-- that's small L*
- rotate `a`
- mirror `i`

Proposal for keys not mentioned in the doc:
- Zoom `z`
- Skeleton `n`

The rest can use the keys already proposed.

DhairyaBahl:
@Keyikedalube Kindly review the changes. I was also thinking about alternatives for those non-keyboard keys but, according do me , for frame seeking comma and period are best because of "<" and ">" in the key which is relatable with backward and forward. 

Thanks :)

Keyikedalube (approved):

Keyikedalube:
@DhairyaBahl Great job and thanks for taking up the issue and making a PR :)

Now it's up to maintainers to have the final decision.

DhairyaBahl:
@ice0 @morevnaproject  Kindly review this PR. And suggest if I have to make any changes. Thank you :)

ice0:
Merged. Thank you!

Issue #658, opened 2018-10-23
bool values inconvenient interpolation


zozorg:
**Synfig version & platform**:
1.3.10

**Issue description**:
Values of type "bool" seem to be inconveniently interpolated, as in, to be "false" it takes BOTH the previous and the next waypoint to be false, forcing you to use 2 waypoints to switch value.
This could be interpreted as "false waypoint=0, true waypoint=1, anything > 0 is interpreted as true", but this behavior is retained even when "constant" interpolation is used.

rodolforg:
This issue is already fixed in v1.4.1.

PR #1826, opened 2020-11-08
fix animation interpolation for boolean values

### LINKED ISSUE #1825: Improper interpolation of bool type
**Issue description**

The issue is that 'bool' values take three waypoints to change their state. Let's describe the following animation: the circle layer has its color wrapped into a 'switch' converter. We want the color to be blue at 0s and then change it to green at 1s. Should be easily accomplished with 2 waypoints: 'switch-off' at 0s and 'switch-on' at 1s. But that's not going to work because as soon as time changes to 1f the boolean value will change to 'true'. As a workaround we need three waypoints: 'false' at 0s, 'false' at 23f and 'true' at 1s which is counter-intuitive and cumbersome.
BoolInterpolationBug

**Note**: The interpolation actually shouldn't matter for boolean values, because, you know, they only have two states. So any interpolation should be treated as constant.

**Affected versions**

Seems to be all. Moreover, there's a duplicate issue here: https://github.com/synfig/synfig/issues/658
It didn't get any attention, so I am making a version with more detailed description and example file: 
[BoolInterpolationBug.zip](https://github.com/synfig/synfig/files/5506222/BoolInterpolationBug.zip)

I hope that it at least gets 'confirmed' status so it will be addressed somewhere in the future.

rodolforg:
The boolean value set in a given timepoint is kept until
a new waypoint.

As before, it ignores any set of interpolation type.

fix #1825 and #658

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_animatedinterface.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_animatedinterface.cpp
@@ -835,25 +835,17 @@ public:
 				return animated.waypoint_list_.front().get_value(t);
 			if(animated.waypoint_list_.empty())
 				return false;
-			if(t<r)
+			if(t<=r)
 				return animated.waypoint_list_.front().get_value(t);
-			if(t>s)
+			if(t>=s)
 				return animated.waypoint_list_.back().get_value(t);
 
-			WaypointList::const_iterator iter;
-			WaypointList::const_iterator next;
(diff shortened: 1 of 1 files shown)


rodolforg:
Although it occurs in 1.2.2, I think it should be cherrypicked to 1.3.x branch.
Current behavior just change the value in the exact instant of the waypoint and change back to other value…

ice0:
Merged. Thank you!

PR #1829, opened 2020-11-09
fix Duplicate layer ignoring all layers but the first one

### LINKED ISSUE #1590: Duplicate ignores all layers except the first one
### Issue description
Say you have Duplicate layer over two others layers such as Translate and Circle. Each of layers have special conversions: Translate shifts X value with each iteration, Circle changes color with each iteration. The problem is that only Translate layer properly calculates duplicate conversion, the other one is ignored.

This picture illustrates the problem:
DupBug1315

Pay attention that in 1.0.2 version not only circle shifts its value, but also changes color.
Test sif file is also attached.
[DupBug.zip](https://github.com/synfig/synfig/files/4983629/DupBug.zip)

### Affected versions
Synfig 1.2.2 and higher.

rodolforg:
fix #1590

### DIFF
--- a/synfig-core/src/synfig/context.cpp
+++ b/synfig-core/src/synfig/context.cpp
@@ -522,8 +522,12 @@ Context::build_rendering_task() const
 
 	// TODO: apply z_range and z_blur (now applies in Canvas::optimize_layers)
 
-	return *context
-		 ? (*context)->build_rendering_task(context.get_next())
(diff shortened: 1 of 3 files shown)


rodolforg:
As it is a regression to 1.0.x, I think this one should be cherry-picked to 1.3.x release :)

ice0:
Merged. Thank you!

Issue #1731, opened 2020-10-02
Crash when adding converters to 'list' type (splines)


Svarov-RZM:
**Issue description**
In some specific cases Synfig can no longer process converters on splines parameters. For example, we have two shapes: rectangle and circle. We want a dot to follow 'rectangle' spline and then half-through the animation jump to 'circle' shape. It's easy to achieve using 'Spline Vertex' converter, we convert origin of a dot and then convert spline parameter to a switch and that's it. 
And this scheme works in Synfig 1.0.2 but doesn't in all newer versions starting from 1.2.x.

**To reproduce**

1. Create any layer with 'Origin' parameter, let's say Circle.
2. Convert 'Origin' to 'Spline Vertex'.
3. Convert 'Spline' parameter of 'Spline Vertex' converter to 'Switch' and Synfig crashes.

It seems to be a global issue and probably affect other converters that can work with splines such as 'Time Loop' for instance but I haven't checked that. Additionally, I attached an example file of the animation described above (it crashes on loading).
[VertexListBug.zip](https://github.com/synfig/synfig/files/5317738/VertexListBug.zip)

rodolforg:
Confirmed :(

Svarov-RZM:
I started learning C++ not so long ago and was trying to solve this issue because it hinders my workflow. I was able to solve it, but not entirely sure I did it correctly.

Please check out this commit: https://github.com/Svarov-RZM/synfig/commit/baae32c80ef91960306cff76ac746ec4fc498afc
I don't know why, but for some reason ValueNode_BLineCalcVertex::operator()(Time t)const in valuenode_blinecalcvertex.cpp was trying to access its own class through pointer (`->` operator) instead of accessing it directly (`.` operator). In short, there were some unnecessary pointer magic and in some cases it would crash in `assert` call.

I reworked it to use direct access to ValueNode_BLineCalcVertex class and program doesn't crash anymore. I also did some testing and didn't find any regressions.

@rodolforg @ice0 what's your opinion on this? Is it appropriate fix?

ice0:
Hi! Thanks for the investigating the problem.

 I haven't dug deep on ValueNodes, so this is an assumption. But as I see, when you convert the Spline parameter to Switch, it cannot convert it to the `ValueNode_BLine` class. The reason it wants to cast is just to check if the spline is looped (Bline is a spline in this context). And it fails because the conversion failed. In this situation, we can assume that the shape not looped.

[...]

PR #1836, opened 2020-11-12
Fix crash in Spline Tangent converter

Svarov-RZM:
Part of https://github.com/synfig/synfig/issues/1731
This fixes crash in 'Spline Tangent' converter similar to one that was addressed in 'Spline Vertex' previously.


ice0:
Merged. Thank you!

PR #1839, opened 2020-11-13
[ETL] Removed `etl::trivial` class. Refactored `Param` class.

ice0:


rodolforg:
Class in unions are really tricky.
Runtime check is needed here :P With valgrind included too

ice0:
The only problem I see here - is passing derived class to Param. In that situation, we call only parent destructor, and memory will be leaked.
Anyway, I think this part can be refactored to use `shared_ptr` to store values,

Issue #1843, opened 2020-11-14
1.4.0 fail to build with autotools


marillat:
synfig 1.4.0 Debian unstable amd64

```
[...]

PR #1845, opened 2020-11-14
opengl is now buildable again - but doesn't render :(

rodolforg:


synfig-core/configure.ac
```diff
@@ -347,7 +347,7 @@ AC_ARG_WITH(opengl,
 ],[
 	with_opengl="no"
 ])
-if test $with_opengl != "no" ; then {
+if test $with_opengl = "no" ; then {
```
ice0, line 350: Looks like this will break logic?
It looks for `gl.pc`, only if you say `--with_opengl`

rodolforg: Actually, no.
https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/External-Software.html

ice0:
```
./configure --without-opengl
```

```
OpenGL ---------------------------> yes


OpenCL ---------------------------> no
libdv ----------------------------> yes
OpenEXR --------------------------> yes
```

Definitely not the behavior the user expects.

rodolforg: Well, I believe you :)
This commit was made on May 24 :P I just rebased and pushed now.
Maybe I changed in a bad way to autobuild include opengl support without touch it and forgot to not commit back then.
Well, I'll remove those changes. Thank you for check it (and sorry for that)

rodolforg: Done

ice0:
Merged. Thank you!

rodolforg:
Partial fix #1843

PR #1856, opened 2020-11-17
Removed unused `ETL/stringf` headers. Preparing to move stringf from ETL to core

ice0:


rodolforg:
Nice!

When/if you move stringf.h, please split into .h and .cpp and eliminate the fake header :P

ice0:
> Nice!
> 
> When/if you move stringf.h, please split into .h and .cpp and eliminate the fake header :P

Ok

PR #1861, opened 2020-11-22
[Synfig Studio] Fixed crash when frame rendering throws exception while animating

rodolforg:
It now displays the exception message inside WorkArea when the current frame throws an exception.

It also displays an error dialog while animating in WorkArea

**Description**

If a frame throws an exception/error, Synfig shows its message:

- in console/prompt
- in work area at problematic frame:
a782b3880b4612b31c89fa472be27af07976bffd

- via error message dialog, while animating:
f976d23fbb9c6bdd9e550f2b1003bbc5a7da4743

fix crash-bug reported in forums by lgolden:
https://forums.synfig.org/t/lst-file-from-papagayo-crashes-synfig/11540

### DIFF
--- a/synfig-studio/src/gui/asyncrenderer.cpp
+++ b/synfig-studio/src/gui/asyncrenderer.cpp
@@ -889,6 +889,7 @@ AsyncRenderer::render_target()
 {
 	etl::handle<Target> target(AsyncRenderer::target);
 
+	std::string error_str;
 	try {
 		if(target && target->render(cb))
 		{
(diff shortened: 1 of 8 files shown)


morevnaproject:
>     * via error message dialog, while animating

I see it displays error message while playback. Maybe we should display workarea warning instead (so playback willnot be interrupted)?

Thank you for the fix! ^__^

rodolforg:
Well, I thought it would be good to draw user attention while on playback :)

I'll try to make the required changes

ice0:
Don't know if it might be worth stopping playback if an error occurs?

morevnaproject:
> Don't know if it might be worth stopping playback if an error occurs?

If this exception breaks rendering operation (I mean when doing export via CLI or GUI), then, yes - it is better to show error message and stop playback.

rodolforg:
> If this exception breaks rendering operation (I mean when doing export via CLI or GUI),

When rendering the final animation, the CLI and GUI (via render dialog) stop at broken frame (already with a proper error report).

morevnaproject:
> When rendering the final animation, the CLI and GUI (via render dialog) stop at broken frame (already with a proper error report).

In this case I agree - during playback it is better to show a warning dialog AND stop the playback. ^__^

rodolforg:
And I wasn't able to show the frame error in workarea while on playback :P
It should be rendering, but it's not :P

rodolforg:
Done :)

rodolforg:
> (already with a proper error report).

Well, the GUI version only showed a generic string… so I detailed it better in last commit

morevnaproject:
I have tested latest commit and found one issue: it is not possible to playback anymore after warning appears for the first time.
Video - https://youtu.be/lRCAPifsWqo

rodolforg:
> it is not possible to playback anymore after warning appears for the first time.

fixed

morevnaproject:
Testing now! ^__^

morevnaproject (approved):
Works perfectly now! Thank you!

ice0 (approved):

PR #1866, opened 2020-11-24
[CI] Added GitHub Actions

ice0:


synfig-studio/src/gui/CMakeLists.txt
```diff
@@ -89,7 +89,7 @@ add_subdirectory(resources)
 target_include_directories(synfigstudio PRIVATE ${MLT_INCLUDE_DIRS}) # required for widget_soundwave
 
 target_link_libraries(synfigstudio
-	${FONT_CONFIG_LIBRARIES}
+	${FONT_CONFIG_LDFLAGS}
```
rodolforg, line 92: Is this correct?

ice0: This works, but it is not correct. Thanks for pointing out.

PR #1870, opened 2020-11-25
Fixed GitHub Actions build (MacOS 11)

ice0:


ice0:
@rodolforg I was wrong, we can pass libraries paths using `target_link_libraries`:

> Specify libraries or flags to use when linking a given target and/or its dependents.

https://cmake.org/cmake/help/latest/command/target_link_libraries.html

PR #1878, opened 2020-11-29
[Synfig Studio] Added ability to select available Sets when adding Layers to Sets

### LINKED ISSUE #1452: Add ability to select available Sets when adding Layers to Sets
**Synfig version**:
master 4c519a9

**Issue description**:
1. Open any animation or create a new one
2. In the `Layers Panel` right-click on a Layer and select `Add Layers to Set`

A dialog will pop up asking for the name of the `Set`:
add_set

At this point if the user wants to add a Layer to an already existing `Set`, it's needed to assign the same name of that `Set`.

It would be useful to add the functionality to select from the already created `Sets` using something like a drop-down menu listing available options.

FirasH:
Fix #1452

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -453,6 +453,16 @@ public:
 	}
 };
 
+class SetsModel : public Gtk::TreeModel::ColumnRecord
+{
+	public:
+		SetsModel() {
+			add(entry_set_name);
+		}
+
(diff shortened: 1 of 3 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

MiLoCasellas:
Thanks!!! ^_^

PR #1879, opened 2020-11-30
"This tool has no options" label now centered vertically

FirasH:
When `Tools` without user editable options are selected, the label in the `Tool Options` window is not vertically centered.
This is unpleasant to look at.

This merge request centers the label vertically:

Before:
before

After:
no-options-label


ice0 (approved):

ice0:
Merged. Thank you!

PR #1890, opened 2020-12-06
Fix crash in Spline Width and return of improper type in Spline converters

Svarov-RZM:
Fix https://github.com/synfig/synfig/issues/1731
This is the final part. Feel free to close the linked issue after merging.

The pull request contains two commits:
* First fixes crash in 'Spline Width' converter when you try to wrap it in 'Switch'. Similar to previously addressed in 'Spline Vertex' and 'Spline Tangent' converters.
* Second addresses the regression introduced in this commit: a90d014c56f0cbee21c2ab0ef68936b9e84c49b1
It was no longer possible to convert real values to 'Spline Tangent/Width' and angle to 'Spline Tangent'. The reason is that upon creation of converter the size of spline would be zero and code will always return ValueBase of vector type so 'Bad Connection' error would be thrown (expected real, got vector). See linked issue for screenshot of the error.


ice0 (approved):

ice0:
Merged. Thank you!

Issue #1898, opened 2020-12-07
[Synfig Studio] `lst` files incorrectly imported when path contains non-Latin characters


ice0:
**Synfig version & platform**:
Latest master
Windows

**Issue description**:
https://forums.synfig.org/t/synfig-cant-read-files-from-list-in-some-machines/11553/7

rodolforg:
This issue is already fixed in v1.4.1.

PR #1924, opened 2020-12-13
[Synfig Studio] Fixed incorrect import of `lst` files when path contains non-Latin characters

ice0:
Fix https://github.com/synfig/synfig/issues/1898

### DIFF
--- a/synfig-core/src/synfig/listimporter.cpp
+++ b/synfig-core/src/synfig/listimporter.cpp
@@ -30,8 +30,6 @@
 #	include <config.h>
 #endif
 
-#include <fstream>
-
 #include "listimporter.h"
 
 #include "general.h"
@@ -45,8 +43,6 @@
 
 /* === U S I N G =========================================================== */
 
-using namespace std;
-using namespace etl;
 using namespace synfig;
 
 /* === M A C R O S ========================================================= */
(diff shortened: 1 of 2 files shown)


synfig-core/src/synfig/listimporter.cpp
```diff
@@ -71,10 +67,11 @@ Importer(identifier)
 {
 	fps=15;
 
-	ifstream stream(identifier.filename.c_str());
+	FileSystem::ReadStream::Handle stream = identifier.get_read_stream();
 
-	if(!stream)
-	{
+	// TODO(ice0): There is no way to report about error to GUI. We can't
+	//  throw error from constructor (bad practice), and can't return error code.
```
rodolforg, line 73: Agree. Should we refactor the Importer class to have a `load()` method – that should be called only once?

rodolforg (commented):
I can't test it because I barely have access to Windows on those pandemic days (as I'm home office'ing due to social distancing).
Besides, my language, Portuguese, uses latin-only characters XD

synfig-core/src/synfig/listimporter.cpp
ice0: I think yes, this is a good idea. We can back to this problem after 1.4.1 release.

ice0:
> I can't test it because I barely have access to Windows on those pandemic days (as I'm home office'ing due to social distancing).
> Besides, my language, Portuguese, uses latin-only characters XD

No problem here, I already tested it and all works fine. Just need someone who can check the code too.

rodolforg (approved):
Code looks ok to me

PR #1929, opened 2020-12-13
[Synfig Studio] Added ability to undock panel using the context menu command

rodolforg:
Now user doesn't need to drag panel out to undock it.

Some users couldn't figure it out easily how to do it, so
here we give them an alternative way.

Discussed here: https://forums.synfig.org/t/moving-panels-in-layout/11634

And move the undocked panel near to the mouse position

### DIFF
--- a/synfig-studio/src/gui/docks/dockable.cpp
+++ b/synfig-studio/src/gui/docks/dockable.cpp
@@ -31,7 +31,11 @@
 
 #include "docks/dockable.h"
 
-#include <gtkmm/stock.h>
+#if GTK_CHECK_VERSION (3,20,0)
+#include <gdkmm/seat.h>
+#else
+#include <gdkmm/devicemanager.h>
+#endif
 
 #include <gui/app.h>
 #include <gui/docks/dockmanager.h>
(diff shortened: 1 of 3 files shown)


synfig-studio/src/gui/docks/dockable.cpp
```diff
@@ -156,6 +161,30 @@ Dockable::attach_dnd_to(Gtk::Widget& widget)
 	widget.signal_drag_data_received().connect( sigc::mem_fun(*this, &Dockable::on_drag_data_received ));
 }
 
+void Dockable::dettach()
```
ice0, line 164: Maybe better to rename it to `Undock` instead of `Detach`?

rodolforg: I chose `dettach` because the class has `attach_dnd_to()` . And now I saw made a typo XD
But I can change it to `undock` without any complaining.

ice0: Ok, when I think better to rename method to `detach` (to save attach/detach naming) and `Gtk::MenuItem(_("Undock panel")`

rodolforg: Done

ice0:
@rodolforg please rebase

ice0 (approved):

ice0:
Merged. Thank you!

Issue #1885, opened 2020-12-04
1.4.0 Cant render saved files: No such file or directory


OrangNeedl4:
![Screenshot](https://user-images.githubusercontent.com/75488669/101176802-c1520580-3625-11eb-88a1-61e3100de8fa.jpg)

![Error](https://user-images.githubusercontent.com/75488669/101178676-0d9e4500-3628-11eb-8bad-26dff0c9e2dc.png)

On the new stable version of Synfig, every time i try to render a secuence of a saved file (doesnt matter if its from an older version or new) it just appears an error that says **"unable to create file C:/User/Pictures/*name of the file*: No such file directory"** this also happens with an image sequence, i tried to reinstall synfig but it happens exactly the same, My system uses Windows 10

BobSynfig:
Try to render it in a folder without a "non-latin" character in the path.
The tilde is surely the (well-known) problem.

OrangNeedl4:
Wow, thanks Bob, it looks that i just had to take off the ñ of my file, (also, in the older version this doesnt happen)thank u again

rodolforg:
This issue is already fixed in v1.4.1.

PR #1940, opened 2020-12-21
[Synfig Studio/Windows] Fixed issue with rendering to file when using non-Latin filenames

ice0:
Linked issue: https://github.com/synfig/synfig/issues/1885

### DIFF
--- a/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
+++ b/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
@@ -207,7 +207,7 @@ ffmpeg_trgt::init(ProgressCallback *cb=NULL)
 		if (binary_path != "")
 			binary_path = etl::dirname(binary_path)+ETL_DIRECTORY_SEPARATOR;
 		binary_path += *iter+".exe";
-		if( access( binary_path.c_str(), F_OK ) != -1 ) {
+		if (g_access( binary_path.c_str(), F_OK ) != -1 ) {
 			binary_path = "\"" + binary_path + "\"";
 			ffmpeg_binary_path = binary_path;
 			break;
--- a/synfig-core/src/tool/joblistprocessor.cpp
+++ b/synfig-core/src/tool/joblistprocessor.cpp
@@ -63,14 +63,8 @@
 #include "joblistprocessor.h"
 
 #include <giomm/file.h>
+#include <glib/gstdio.h>
 
-#endif
-
-#ifdef _MSC_VER
-#include  <io.h>
-#define access _access_s
-#define W_OK 2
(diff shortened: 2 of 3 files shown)


ice0:
@rodolforg I've already checked it on Windows, so only code review.

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
```diff
@@ -195,25 +195,21 @@ ffmpeg_trgt::init(ProgressCallback *cb=NULL)
...
 
 	String ffmpeg_binary_path;
-	std::list< String > binary_choices;
-	binary_choices.push_back("ffmpeg");
-	binary_choices.push_back("avconv");
-	std::list< String >::iterator iter;
-	for(iter=binary_choices.begin();iter!=binary_choices.end();iter++)
-	{
+	std::list<std::string> binary_choices = {"ffmpeg", "avconv"};
+	for(const auto& iter : binary_choices) {
```
rodolforg, line 199: It isn't an iterator but the element "itself", so I think `iter` isn't a good variable name.

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
```diff
@@ -195,25 +195,21 @@ ffmpeg_trgt::init(ProgressCallback *cb=NULL)
...
-		if (binary_path != "")
+		if (!binary_path.empty()) {
 			binary_path = etl::dirname(binary_path)+ETL_DIRECTORY_SEPARATOR;
-		binary_path += *iter+".exe";
-		if( access( binary_path.c_str(), F_OK ) != -1 ) {
-			binary_path = "\"" + binary_path + "\"";
-			ffmpeg_binary_path = binary_path;
+		}
+		binary_path.append(iter).append(".exe");
+		if (g_access(binary_path.c_str(), F_OK ) != -1 ) {
```
rodolforg, line 207: I imagine this is the issue solver? `g_access` instead of `access` ?

rodolforg (changes requested):

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
ice0: Yes.

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
ice0: Ok

ice0:
I decided to split this PR to avoid mixing code refactoring with fixing the problem.

rodolforg (dismissed):

PR #1942, opened 2020-12-21
[Synfig Studio] Disabled resizing of the Preview Options dialog

FirasH:
With current master it's possible to increase the size of `Preview Options` dialog:
resize-preview-dialog

which has no benefit for user and is not pleasant to look at.


ice0 (approved):

ice0:
Merged. Thank you!

PR #1944, opened 2020-12-22
[Synfig Core] Removed unused headers

ice0:


synfig-core/src/modules/mod_mng/trgt_mng.h
```diff
@@ -26,12 +26,7 @@
 /* === H E A D E R S ======================================================= */
 
 #include <synfig/target_scanline.h>
-#include <synfig/string.h>
-#include <synfig/targetparam.h>
-#include <cstdio>
-
-// #include <png.h>
-// #include <jpeglib.h>
+#include <cstdio> // FILE
```
rodolforg, line 29: We should later make an interface or guidelines for targeters/importers (and modules/addons/plugins) properly deal with: file names and file opening/handling.

rodolforg (approved):

PR #1947, opened 2020-12-24
[Core] Refactored trgt_ffmpeg

ice0:


synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
```diff
@@ -195,49 +182,40 @@ ffmpeg_trgt::init(ProgressCallback *cb=NULL)
...
 		}
-		pclose(pipe);
-		synfig::info(strprintf("\"%s\" --> %s", command.c_str(), result.c_str()));
-		if (result.substr(0,1) == "/" || result.substr(0,2) == "./" ){
-			ffmpeg_binary_path = binary_path;
+		/* `which` exit status
+		 * 0      if all specified commands are found and executable
+		 * 1      if one or more specified commands is nonexistent or not executable
+		 * 2      if an invalid option is specified */
+		synfig::info(etl::strprintf("\"%s\" --> %s", command.c_str(), result.c_str()));
```
rodolforg, line 210: no need to use `strprintf` here: `synfig::info()` accepts format + values.

ice0: fixed

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
```diff
@@ -195,49 +182,40 @@ ffmpeg_trgt::init(ProgressCallback *cb=NULL)
...
-			binary_path = etl::dirname(binary_path)+ETL_DIRECTORY_SEPARATOR;
-		binary_path += *iter+".exe";
-		if( access( binary_path.c_str(), F_OK ) != -1 ) {
-			binary_path = "\"" + binary_path + "\"";
-			ffmpeg_binary_path = binary_path;
-			break;
-		}
+	// Windows always have ffmpeg
+	std::string binary_path = etl::dirname(synfig::get_binary_path(".")) + "/ffmpeg.exe";
+	if (g_access(binary_path.c_str(), F_OK ) != -1 ) { // File found
```
rodolforg, line 188: Maybe we should use `X_OK` ? It checks for existence and execution permission.

ice0: Windows file systems do not have an execution bit for the file (https://docs.microsoft.com/en-us/windows/win32/fileio/file-attribute-constants). So it looks like this is not necessary.

rodolforg: Oh, Ok.
I just thought it because, at my work, the IT crew setup somehow some folders contents can't be executed :) 
But I don't know if this function would catch those cases.

ice0: Very interesting. I will take a look.

ice0: @rodolforg found details here: https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-access

> On Windows, the file protection mechanism is not at all POSIX-like, and the underlying function in the C library only checks the FAT-style READONLY attribute, and does not look at the ACL of a file at all. This function is this in practise almost useless on Windows. Software that needs to handle file permissions on Windows more exactly should use the Win32 API.

PR #1952, opened 2020-12-26
[Synfig Studio] fix file path on rendering file with relative path

rodolforg:
Issue:
1. User launches a file via Synfig Studio from command line with a relative path
(examples:
`$ synfigstudio mydir/myfile.sif`
 or
`$ synfigstudio myfile.sif`
)

2. Synfig Studio opens and loads the file normally (as always)

3. User opens Render dialog.

4. The filename entry text points to home dir + relative path!
(examples:
/home/user/mydir/myfile.sif // Problem: does mydir exist?
 or
/home/user/myfile.sif // It may be ok, but shouldn't it point to same folder of sif file?
)

### DIFF
--- a/synfig-studio/src/gui/render.cpp
+++ b/synfig-studio/src/gui/render.cpp
@@ -231,7 +231,12 @@ RenderSettings::set_entry_filename()
 	}
 
 	if (!is_absolute_path(filename))
-		filename = Glib::get_home_dir() + ETL_DIRECTORY_SEPARATOR + filename;
+	{
+		if (filename.find(App::custom_filename_prefix) == 0)
+			filename = Glib::get_home_dir() + ETL_DIRECTORY_SEPARATOR + filename;
+		else
+			filename = etl::absolute_path(filename);
+	}
 	
 	try
 	{


rodolforg:
The problem comes from `synfig::open_canvas_as(const FileSystem::Identifier &identifier,const String &as,String &errors,String &warnings )`, that doesn't store the absolute filepath in `Canvas` object.

Another solution would be:
```c++
/// File: loadcanvas.cpp
Canvas::Handle
synfig::open_canvas_as(const FileSystem::Identifier &identifier,const String &as,String &errors,String &warnings)
{
	String filename = FileSystem::fix_slashes(as);
	// Add these lines below:
	if (!etl::is_absolute_path(filename))
		filename = etl::absolute_path(filename);

```

but I was afraid that could break another stuff due to `etl::absolute_path()` trusting on `getcwd()`, that doesn't handle the famous encoding problems.

ice0:
> but I was afraid that could break another stuff due to `etl::absolute_path()` trusting on `getcwd()`, that doesn't handle the famous encoding problems.

This can be resolved by using [g_path_is_absolute (const gchar *file_name)](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-path-is-absolute) instead of `etl::absolute_path()`

ice0:
@rodolforg 
Is that still an issue on current master? I can't reproduce.

rodolforg:
Neither do I. How was it fixed? Mistery? :D

ice0:
> Neither do I. How was it fixed? Mistery? :D

I think it maybe was fixed after switching to `Gio::Application`

rodolforg:
Right now I can't test on 1.4.x though... Does it work there?
AppImage packages have an issue with relative paths...

rodolforg:
Closed as it works fine on master branch.

Issue #402, opened 2017-08-14
Synfig Studio cannot open files it has just recently saved


CoryAlbrecht:
Synfig Studio 1.0.2 on Ubuntu GNOME 16.04.2

Saved as a *.sif file then later tried to reload and continue working on the project.

![image](https://user-images.githubusercontent.com/8172909/29289348-13065dc4-810a-11e7-9147-a2cbe3802122.png)

The .sif file that produces the error along with the referenced PNG files.
[synfig-error.zip](https://github.com/synfig/synfig/files/1223326/synfig-error.zip)

rodolforg:
This issue is fixed in v1.4.1.

PR #1955, opened 2020-12-27
[Synfig Core] Added support for loading `average` values

ice0:
Linked issue: https://github.com/synfig/synfig/issues/402

Test file: [average.zip](https://github.com/synfig/synfig/files/5744310/average.zip)

### DIFF
--- a/synfig-core/src/synfig/loadcanvas.cpp
+++ b/synfig-core/src/synfig/loadcanvas.cpp
@@ -88,6 +88,7 @@
 #include "valuenodes/valuenode_twotone.h"
 #include "valuenodes/valuenode_weightedaverage.h"
 #include "valuenodes/valuenode_wplist.h"
+#include "valuenodes/valuenode_average.h"
 
 #endif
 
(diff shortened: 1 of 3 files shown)


synfig-core/src/synfig/loadcanvas.cpp
```diff
@@ -2362,6 +2365,13 @@ CanvasParser::parse_dynamic_list(xmlpp::Element *element,Canvas::Handle canvas)
 			String loop=element->get_attribute("loop")->get_value();
 			weightedaverage_value_node->set_loop(is_true(loop));
 		}
+	} else if (element->get_name()=="average") {
+		average_value_node = new synfig::ValueNode_Average(type);
```
rodolforg, line 2369: Shouldn't it pass `canvas` to constructor as `weighted_average` and `bline` in above lines do?

rodolforg (changes requested):

synfig-core/src/synfig/loadcanvas.cpp
ice0: It should! Thanks for the hint!

PR #1956, opened 2020-12-27
[Core] A little refactoring of loadcanvas.cpp

ice0:
Based on top of this PR (https://github.com/synfig/synfig/pull/1955)

### DIFF
--- a/synfig-core/src/synfig/loadcanvas.cpp
+++ b/synfig-core/src/synfig/loadcanvas.cpp
@@ -53,7 +53,6 @@
 
 #include "blur.h"
 #include "boneweightpair.h"
-#include "boneweightpair.h"
 #include "exception.h"
 #include "importer.h"
 #include "gradient.h"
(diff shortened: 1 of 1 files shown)


synfig-core/src/synfig/loadcanvas.cpp
```diff
@@ -2284,6 +2284,13 @@ CanvasParser::parse_static_list(xmlpp::Element *element,Canvas::Handle canvas)
 	return value_node;
 }
 
+bool read_bool_attribute(xmlpp::Element* element, const char* name) {
```
rodolforg, line 2287: - prepend `static`

- Maybe rename to `is_bool_attribute_true()` or `evaluate_bool_attribute()`
When I read `if (read_bool_attribute)` I have the impression I'm checking if it could read the attribute, not exactly evaluating its value.

rodolforg (changes requested):

synfig-core/src/synfig/loadcanvas.cpp
ice0: > I have the impression I'm checking if it could read the attribute

This is the correct impression, because it tries to read the attribute and returns `false` if it cannot read it (for example, the attribute was not found).

> Maybe rename to is_bool_attribute_true()

Agree. This is more correct.

PR #1957, opened 2020-12-28
[Synfig Studio] remove unused headers

rodolforg:
and do some sorting


rodolforg:
It's an endless job! :P

rodolforg:
Ouch. So CMake doesn't set the config.h file with the same #define Autotools does? :P

ice0:
> Ouch. So CMake doesn't set the config.h file with the same #define Autotools does? :P

Some defines maybe missing. I will take a look.

rodolforg:
Missing #defines (as per ipc.cpp):

- HAVE_SYS_TYPES_H 
- HAVE_SYS_STAT_H 
- HAVE_SYS_ERRNO_H
- HAVE_UNISTD_H 
- HAVE_FCNTL_H

ice0 (approved):

ice0:
Merged. Thank you!

PR #1958, opened 2020-12-28
[Synfig Studio] Split top horizontal toolbar in two toolbars (horizontal+vertical)

### LINKED ISSUE #1951: [Synfig Studio] Reorganize buttons around canvas area
master

**Issue description**:
Successor to discussion started in #1451.

The key idea is to reduce used screen estate by reorganizing the buttons/tools most used.
There are some mockups in the linked issue.

Here you can see an actually working test I made right now by splitting the top bar in two.
On top there are file/export/preview options, while on the right drawing options like ducks and guides/grid (all in a single row are too wide).

canvas-buttons-reorder

Another step step could be to move tools out of the toolbox to the left side of canvas (Inkscape style interface) and move somewhere the color editor in the toolbox.

Let me know what you think!

FirasH:
Fix #1951

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -584,11 +584,13 @@ CanvasView::CanvasView(etl::loose_handle<Instance> instance,etl::handle<CanvasIn
 	Gtk::Widget *widget_work_area = create_work_area();
 	init_menus();
 	Gtk::Widget *widget_top_bar = create_top_toolbar();
+	Gtk::Widget *widget_stopbutton = create_stop_button();
 	Gtk::Widget *widget_right_bar = create_right_toolbar();
 	Gtk::Widget *widget_time_bar = create_time_bar();
 	
 	Gtk::Grid *layout_grid = manage(new Gtk::Grid());
-	layout_grid->attach(*widget_top_bar,     0, 0, 2, 1);
+	layout_grid->attach(*widget_top_bar,     0, 0, 1, 1);
+	layout_grid->attach(*widget_stopbutton,  1, 0, 1, 1);
 	layout_grid->attach(*widget_right_bar,   1, 1, 1, 2);
 	layout_grid->attach(*widget_space,       0, 1, 1, 1);
 	layout_grid->attach(*widget_work_area,   0, 2, 1, 1);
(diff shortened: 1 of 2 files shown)


FirasH:
This is how it look with last commits:
issue-1951-current-state

@Keyikedalube I don't think the toolbar on the right is following the `synfig.css` because of:
`displaybar->set_toolbar_style(Gtk::TOOLBAR_ICONS);`

if I remove that line the size of the icons is exactly the same as the top toolbar (but with text also).

The look of the default toolbar IMHO is better than forcing smaller icons.
As you can see here:
icon-size
the icon is as big as the button (on Windows should be like the right toolbar IIRC).

Maybe removing completely all paddings and using the stock icon size could be a good compromise between space and UI buttons consistency? (also now the top toolbar has less buttons)

Keyikedalube:
Someone made that toolbar content small (before I did) because there were too many buttons
But since you're splitting into two toolbars and maybe more, your judgment is correct :)
We don't need to style them with `synfig.css`
Default size looks proper already

FirasH:
There are some different styles in my KDE then...
Indeed you're right style has its effect also on the right toolbar.

Here is the same on Windows 10 (with `synfig.css`):
toolbar-split-windows

FirasH:
With the last commit I think the request is ready for merge.

morevnaproject:
Cool, I like this!

morevnaproject (approved):
Looks like this feature is ready! 

P.S. I didn't looked at the code, just tested how it works. ^__^

ice0 (approved):

ice0:
Before:
Screenshot_80

After:
Screenshot_82

This is how it looks on Full HD monitor. As you can see there is no reason to move toolbar right, for such monitors. 

And I think maybe add the ability to customize the location of this toolbar?

morevnaproject:
@ice0 I think it's better to have same layout, independently from screen size/resolution.

synfig-studio/src/gui/dials/toggleducksdial.cpp
```diff
@@ -55,12 +55,12 @@ using namespace studio;
...
 ToggleDucksDial::ToggleDucksDial(const Gtk::IconSize &size)
 {
-	init_label_button(ducks_position, size, "synfig-toggle_duck_position", _("Position handles"), _("Toggle position handles"));
-	init_label_button(ducks_vertex,   size, "synfig-toggle_duck_vertex",   _("Vertex handles"),   _("Toggle vertex handles"));
-	init_label_button(ducks_tangent,  size, "synfig-toggle_duck_tangent",  _("Tangent handles"),  _("Toggle tangent handles"));
-	init_label_button(ducks_radius,   size, "synfig-toggle_duck_radius",   _("Radius handles"),   _("Toggle radius handles"));
-	init_label_button(ducks_width,    size, "synfig-toggle_duck_width",    _("Width handles"),    _("Toggle width handles"));
-	init_label_button(ducks_angle,    size, "synfig-toggle_duck_angle",    _("Angle handles"),    _("Toggle angle handles"));
+	init_label_button(ducks_position, size, "synfig-toggle_duck_position", _("Position handles"), _("Toggle Position Handles"));
+	init_label_button(ducks_vertex,   size, "synfig-toggle_duck_vertex",   _("Vertex handles"),   _("Toggle Vertex Handles"));
```
rodolforg, line 59: Why did you uppercase these tooltips but not previous ones? Like those in (new) line 1453 in canvasview.cpp:
```c++
	action_group->add( Gtk::Action::create("save", Gtk::StockID("synfig-save"), _("Save"), _("Save document")),
		hide_return(sigc::mem_fun(*get_instance().get(), &Instance::save))
	);
	action_group->add( Gtk::Action::create("save-as", Gtk::StockID("synfig-save_as"), _("Save As..."), _("Save document As")),
```

rodolforg (changes requested):
Some (in)consistencies ;)

synfig-studio/src/gui/dials/toggleducksdial.cpp
FirasH: This change is up to the style we want to use.
In this case I followed the style used in the menus.

Once the style we want to use is determined I can go through all the menu entries and tooltips to make them consistent 👍 

Let's say we have something like:
- "Toggle snap to Grid" it becomes "Toggle Snap to Grid"
- "Window > Workspace > Save workspace..." it becomes "Window > Workspace > Save Workspace..."

Here I've set all in uppercase except adverbs/conjunctions.
I'm not a fan of all uppercase text, but I've noticed that quite common in English UIs.

FirasH:
As @ice0 noticed, maximizing the interface on large screens leads to some unused space in the toolbars, but still some things are potentially missing in the toolbars, like moving the `Toolbox Tools` on the right side of the canvas view or adding some other buttons like `Export` (if those makes sense to others).

Also it's most likely that vertical space is more critical than horizontal space.
In my case I animate only vertical stuff, but that's my personal use case.

synfig-studio/src/gui/dials/toggleducksdial.cpp
rodolforg: Oh, that Menu Item text then. Sorry :) I thought it was a tooltip. ;)

FirasH: It's a tooltip IIRC but I copied the string used in the menu, anyway I'll check and make sure it's consistent.

FirasH:
@rodolforg I've cleaned the tooltips following your observations, now it's more consistent :+1:

ice0 (approved):

ice0:
Merged. Thank you!

PR #1959, opened 2020-12-28
[Synfig Studio] fix states accelerator shortcut in tooltips

rodolforg:
Previously, it could show the word "asterisk" instead of
character * for Star tool, for example.


ice0 (approved):

ice0:
Merged. Thank you!

PR #1968, opened 2021-01-03
[Synfig Studio] Waypoint Editor didn't properly support enums

### LINKED ISSUE #1965: [Synfig Studio] Waypoint Editor "Value" could set not feasible Parameter values
master 4b7f89b

**Issue description**:
1. Create a new animation
2. Draw a `Star` on Canvas (works with any Layer)
3. Enable Animate mode and create a waypoint for `Blend Method` parameter
4. Double-click on the waypoint created in step 3 to open `Waypoint Editor` dialog
waypoint-editor-value
`Value` is represented as a spin box, not as a dropdown.

5. Scroll on `Value` spin box to change its value e.g. -20
waypoint-editor-out-of-bounds
The spin box hasn't a min/max value and the value is set.

This happens for all values set with dropdowns in `Parameters` panels.

rodolforg:
fix #1965

### DIFF
--- a/synfig-studio/src/gui/dialogs/dialog_waypoint.cpp
+++ b/synfig-studio/src/gui/dialogs/dialog_waypoint.cpp
@@ -132,8 +132,7 @@ void
 Dialog_Waypoint::set_value_desc(synfigapp::ValueDesc value_desc)
 {
 	value_desc_=value_desc;
-	if(value_desc.get_value_node() && value_desc.get_value_node()->get_parent_canvas())
-		waypointwidget->set_canvas(value_desc.get_value_node()->get_parent_canvas());
+	waypointwidget->set_valuedesc(value_desc_);
 
 	if (value_desc.is_value_node())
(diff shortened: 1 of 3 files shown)


rodolforg:
Maybe backport it?

FirasH:
Tested and works perfectly!


Noticed that with enums type of variables when some kind of interpolation is set, those will get through all variables (this happened even before)
For example if you interpolate from `Blend Method` `Composite` to `Alpha` Synfig will get through each one of them like from 0 to 20.

If interpolation is set to Constant it's fine, but I don't know if this is a plus or an issue...

rodolforg:
Maybe it should use "Constant" interpolation by default. Perhaps all "enum" parameters.

ice0 (approved):

ice0:
> Maybe backport it?

I think yes.

ice0:
Merged. Thank you!

PR #1970, opened 2021-01-04
[Synfig Studio] Implement Ctrl+Shift+Scroll to zoom horizontally in panels

### LINKED ISSUE #1326: Add ability to Zoom Timeline directly in the Timetrack
**Synfig version**:
Master: 81dc3b6

**Issue description**:
Currently to `Zoom In on Timeline` and `Zoom Out on Timeline` it is necessary to go through the top `View` menu.

Sometimes it is useful to zoom in/out rapidly to edit the waypoints (e.g. when having many closely placed waypoints)

It would be convenient to have the ability to do that directly in the Timetrack regardless of the keyboard shortcuts (that need to be known as a prerequisite).

Some possibilities could be:
1. Add buttons somewhere in the new Timetrack
2. Personally I would suggest to use `Ctrl + Scroll` because (at least for me) it seems very intuitive and also the mouse is already in use for editing on the Timetrack.

What do you think about it?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/90865866-add-ability-to-zoom-timeline-directly-in-the-timetrack?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

FirasH:
Fix #1326

### DIFF
--- a/synfig-studio/src/gui/selectdraghelper.h
+++ b/synfig-studio/src/gui/selectdraghelper.h
@@ -772,11 +772,11 @@ bool SelectDragHelper<T>::process_scroll_event(GdkEventScroll* event)
 		case GDK_SCROLL_RIGHT: {
 			if ((event->state & GDK_CONTROL_MASK) && zoom_enabled) {
 				if ((event->state & GDK_SHIFT_MASK) == GDK_SHIFT_MASK)
-					// Ctrl+Shift+scroll: horizontal zoom in
-					signal_zoom_horizontal_in_requested().emit();
-				else
-					// Ctrl+scroll: vertical zoom in
+					// Ctrl+Shift+Scroll: vertical zoom in
 					signal_zoom_in_requested().emit();
+				else
+					// Ctrl+Scroll: horizontal zoom in
+					signal_zoom_horizontal_in_requested().emit();
 			} else {
 				if (!scroll_enabled)
 					return false;
@@ -794,11 +794,11 @@ bool SelectDragHelper<T>::process_scroll_event(GdkEventScroll* event)
 		case GDK_SCROLL_LEFT: {
 			if ((event->state & GDK_CONTROL_MASK) && zoom_enabled) {
 				if ((event->state & GDK_SHIFT_MASK) == GDK_SHIFT_MASK)
-					// Ctrl+Shift+Scroll: horizontal zoom out
-					signal_zoom_horizontal_out_requested().emit();
-				else
-					// Ctrl+scroll: vertical zoom out
+					// Ctrl+Shift+Scroll: vertical zoom out
 					signal_zoom_out_requested().emit();
+				else
+					// Ctrl+Scroll: horizontal zoom out
+					signal_zoom_horizontal_out_requested().emit();
 			} else {
 				if (!scroll_enabled)
 					return false;
--- a/synfig-studio/src/gui/selectdraghelper.h
+++ b/synfig-studio/src/gui/selectdraghelper.h
@@ -782,11 +782,11 @@ bool SelectDragHelper<T>::process_scroll_event(GdkEventScroll* event)
 					return false;
 
 				if ((event->state & GDK_SHIFT_MASK) == GDK_SHIFT_MASK)
-					// Shift+Scroll: scroll right
-					signal_scroll_right_requested().emit();
-				else
-					// Scroll: scroll up
+					// Shift+Scroll: scroll up
 					signal_scroll_up_requested().emit();
+				else
+					// Scroll: scroll right
+					signal_scroll_right_requested().emit();
 			}
 			return true;
 		}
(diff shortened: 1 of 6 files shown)


rodolforg (dismissed):

ice0:
Anyone use vertical zoom?

rodolforg:
Curves/Graph and Sound panel does zoom vertically.

ice0:
> Curves/Graph and Sound panel does zoom vertically.

Ok. I think maybe better to change Ctrl+Scroll to Horizontal zoom (since it used more often) and Shift+Ctrl+Scroll to Vertical zoom?

FirasH:
@ice0 I’d like to add also horizontal scrolling sometime in the future and we would have:
* `Scroll`: move vertically
* `Ctrl+Scroll`: zoom vertically
* `Shift+Scroll`: move horizontally
* `Ctrl+Shift+Scroll`: zoom horizontally

I think this is how other softwares usually manage this type of things. I’ll investigate how others do this stuff.

What do you think?

FirasH:
On Canvas we have:
* `Scroll`: move vertically
* `Ctrl+Scroll`: zoom
* `Shift+Scroll`: move horizontally

ice0:
> On Canvas we have:
> 
> * `Scroll`: move vertically
> * `Ctrl+Scroll`: zoom
> * `Shift+Scroll`: move horizontally

Yes, this is the main reason why I suggest using the same combination for the main zoom action. And for the timeline, the primary zoom for me is the horizontal zoom. Maybe I'm wrong and someone uses it differently.

FirasH:
@ice0 I was thinking about `Shift` for horizontal zoom because it's used for horizontal movements.

So the logic would be
`...+Shift+Scroll`: apply function on horizontal axis

If we swap horizontal and vertical zooming we break this logic and it's less intuitive at least for me.
Let's say I want to move horizontally (press `Shift`) and change zoom always horizontally (release `Shift` and press `Ctrl` instead of toggling `Ctrl` while `Shift` is pressed) it becomes more complicated. Same for the vertical part.

At the same time generally zooming is performed at the same time horizontally and vertically (and that's probably the reason why Canvas is like this)

but we can do either way.
@rodolforg @morevnaproject what do you think?

FirasH:
Managed to complete the `Shift+Scroll` for moving timebar back and forth.

Let me know what you think!

morevnaproject:
> And for the timeline, the primary zoom for me is the horizontal zoom

I agree with @ice0 - horizontal zoom is more common for timeline (I couldn't remember myself zooming vertically at all).
I think it is better to have Ctrl+Scroll for horizontal zoom.
Also Ctrl+Scroll is more common for horizontal zoom in other apps. ^__^

FirasH:
Ok!

I'll update the shortcuts 👍

synfig-studio/src/gui/selectdraghelper.h
```diff
@@ -763,26 +771,44 @@ bool SelectDragHelper<T>::process_scroll_event(GdkEventScroll* event)
 		case GDK_SCROLL_UP:
 		case GDK_SCROLL_RIGHT: {
```
rodolforg, line 772: Should we respect axis then?
If users has a device with a horizontal scroll, it should scroll horizontally without `Shift` :P

FirasH: * `Scroll`: move horizontally
* `Ctrl+Scroll`: zoom horizontally
* `Shift+Scroll`: move vertically
* `Ctrl+Shift+Scroll`: zoom vertically

This is your idea?

rodolforg: For GDK_SCROLL_UP and GDK_SCROLL_DOWN cases, leave as PR does.

For GDK_SCROLL_RIGHT and GDK_SCROLL_LEFT, do as you proposed now (I actually thought only about the horizontal case, but I guess it's fine to `shift` mean axis-changing :) The Mirror tool uses the same concept)

FirasH: Done!

rodolforg: Hmm… I just meant to separate the case `GDK_SCROLL_UP` from case `GDK_SCROLL_RIGHT`, and same for `GDK_SCROLL_DOWN` from `GDK_SCROLL_LEFT`.
The UP and DOWN would behave as you had implemented.
The RIGHT and LEFT would behave scrolling right/left respectively, but as horizontal zoom if with `ctrl` and vertically with `shift`

rodolforg: For clarification:
```c++
		case GDK_SCROLL_UP:
		case GDK_SCROLL_RIGHT: {
		/// code
		}
```
would become
```c++
		case GDK_SCROLL_UP: {
		/// your code until your penultimate commit
		case GDK_SCROLL_RIGHT: {
		/// code similar to above, but swapping axis
		}
```

And analogous to to `DOWN`/`LEFT`

FirasH:
@rodolforg made the changes (hopefully correct)

In the end we got this compromise:
* `Scroll`: move vertically
* `Ctrl+Scroll`: zoom horizontally
* `Shift+Scroll`: move horizontally
* `Ctrl+Shift+Scroll`: zoom vertically

which copies the logic of the canvas area.

rodolforg:
> which copies the logic of the canvas area.

Well, the work/canvas area actually zoom both axis simultaneously XD

rodolforg (approved):
It now follows the logic proposed by @ice0  and seconded by @morevnaproject .
It also do a more natural scroll direction for those (privileged) ones that have horizontal-scroll-wheel.

ice0 (approved):

ice0 (approved):

ice0:
Merged. Thank you!

Issue #1921, opened 2020-12-12
[Synfig Studio] Duplicating Layers where Duplicate is used, creates a new index but doesn't update Connected value


FirasH:
**Synfig version & platform**:
master e84f6cb

**Issue description**:

1) Create a new animation
2) Add a Layer on `Canvas`, for example a `Star` (no need for outlines etc.)
3) Add a `Duplicate` Layer
Now in the `Library` window a new value called `duplicate` is added.

4) `Group` the `Star` and the `Duplicate` Layers
5) Select the `Star` and in the `Parameters` window right-clic on `Origin` and `Convert > Composite`
Now `Origin` has two sub-parameters: `X-Axis` and `Y-Axis`

6) Select `X-Axis` and connect to the `duplicate` value in the `Library` window
Now on Canvas there's a `Star` duplicated three times horizontally.

7) Duplicate the `Group` created in step 4
In the `Library` window a new value called `Index 1` is added. This value should be the parameter used by the duplicated `Duplicate` Layer.

If you drag the duplicated `Group` you can see it contains only one star. That's because it's still connected to the old `duplicate` value.
From a logical point of view it makes sense, but from a user perspective I expect to duplicate everything and get a whole copy of the duplicated Layer.

8) Connect the `Star` to `Index 1`
The duplicated group now has three stars

I think Step 8 should be the expected behavior of Synfig at Step 7, what do you think?

rodolforg:
This issue is fixed in v1.4.1.

Issue #453, opened 2017-11-05
Copy of skeleton not linked


zozorg:
**Synfig version & platform**:
linux 1.3.4

**Issue description**:
When copying a skeleton and its linked splines, by copy/pasting or embedding an imported canvas, the resulting skeleton isn't linked to the splines.
I'm willing to post a bounty when confirmed, I assume this one must be somewhat hard to fix. What $$$ would sound reasonable?

morevnaproject:
> Just create a skeleton with 3 bones, duplicate it, and move the bones around.

Yes, I see what you mean. When I duplicate skeleton layer (within the same document) it creates copy of each bone. But those copies still reference parents from old skeleton.

morevnaproject:
In general any copy operations related with bones are dangerous at the moment. For example this one - #455.

zozorg:
Posted a [$50 bounty](https://www.bountysource.com/issues/51184738-cannot-copy-a-skeleton)

zozorg:
+10$ bounty, total $60

src-r-r:
I'm taking a look at this issue now.

src-r-r:
I've been looking through the source. I think part of the problem is that the design of the bone/skeleton relationship is not very structured. Bones are just there: they don't belong to anything. Is this by intention?

Taking a cue from Blender, I propose skeletons take on a more active role, so that a user can select a skeleton, double click, and enter the skeleton group to modify the bones.

Does this work for the design?

rodolforg:
@src-r-r The possible cause of this problem is a wrong duplication of the ValueNode that build the relationships among bones themselves.
Maybe LayerSkeleton should override Layer::clone() method to fix the bones' parenting after duplication.

locky504:
I'm working on this issue. PR #3702 submitted.

PR #1971, opened 2021-01-05
[Synfig Studio] Fix action LayerDuplicate with layers "duplicate" o.O

rodolforg:
Two problems fixed:
- #1921
- User wants to duplicate two layers "duplicate" in different groups (user would receive a error dialog and he'd need to duplicate again, but one layer at a time in order to get work done)

And it implements a test for LayerDuplicate action (in `synfig-studio/test`folder). For Autotools, build and run it with `make check` ;)

### DIFF
--- a/synfig-studio/src/synfigapp/actions/layerduplicate.cpp
+++ b/synfig-studio/src/synfigapp/actions/layerduplicate.cpp
@@ -61,6 +61,33 @@ ACTION_SET_VERSION(Action::LayerDuplicate,"0.0");
 
 /* === P R O C E D U R E S ================================================= */
 
+/// Search for all duplicates of explicit layers (those set as action parameter) and implicit layers too (contents of layer of group type)
+static void
+traverse_layers(synfig::Layer::Handle layer, synfig::Layer::Handle cloned_layer, std::map<synfig::Layer::Handle, synfig::Layer::Handle>& cloned_layer_map);
+
+/// Get value nodes that are special cases when duplicating
+static etl::rhandle<ValueNode>
+get_special_layer_valuenode(synfig::Layer::Handle layer)
+{
+	if (layer->get_name() == "duplicate")
+		return layer->dynamic_param_list().find("index")->second;
+	return etl::rhandle<ValueNode>();
+}
+
+/// Scan a LinkableValueNode parameter tree and replaces special valuenodes with their respective duplicates (clones)
+static void
+do_replace_valuenodes(LinkableValueNode::Handle link_vn, const std::pair<etl::rhandle<synfig::ValueNode>, etl::rhandle<synfig::ValueNode>>& vn_pair)
+{
+	const int link_count = link_vn->link_count();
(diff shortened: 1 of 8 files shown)


synfig-studio/src/synfigapp/actions/layerduplicate.cpp
```diff
@@ -114,15 +114,26 @@ Action::LayerDuplicate::is_ready()const
...
-	std::list<synfig::Layer::Handle>::const_iterator i;
+	// pair (original layer, cloned layer)
+	std::map<synfig::Layer::Handle,synfig::Layer::Handle> cloned_layer_map;
+	// pair (original special valuenode, cloned special valuenode)
+	std::map<etl::rhandle<ValueNode>, etl::rhandle<ValueNode>> cloned_valuenode_map;
+	// pair (canvas, last exported valuenode "Index #" -> Layer_Duplicate parameter: index )
+	std::map<Canvas::LooseHandle, int> last_index;
 
-	for(i=layers.begin();i!=layers.end();++i)
+	for(auto i=layers.cbegin();i!=layers.cend();++i)
```
ice0, line 136: `for (const auto i : layers)`?

rodolforg:
And maybe this is the way to properly solve this #453

rodolforg:
I think we should backport it.

By the way, the test file will fail at last test: it's the crash I reported recently #1983

rodolforg:
@ice0 do you know why MacOS 11.0 action keeps failing?

rodolforg:
I just found out that the Paste (layer) command (available in LayerActionManager) code is almost a "copy'n-paste" of LayerDuplicate code. So we need to fix it there too. Or, even better, avoid this code duplication (oh my, so many 'duplicate' words)

ice0:
> @ice0 do you know why MacOS 11.0 action keeps failing?

Looks like GitHub problem: https://github.com/actions/virtual-environments/issues/1814#issuecomment-757479694

ice0:
> I just found out that the Paste (layer) command (available in LayerActionManager) code is almost a "copy'n-paste" of LayerDuplicate code. So we need to fix it there too. Or, even better, avoid this code duplication (oh my, so many 'duplicate' words)

Agree!

ice0 (approved):

rodolforg:
> Agree!

Ok, so I'll try to discover what could I do to avoid the duplication… lol
I thought about adding a new action parameter: "to-canvas"; if set, then it means it's a copy'n paste action, not a regular 'duplicate'.

rodolforg:
@ice0 in another PR or this one? (I still didn't start to work on this copy'n paste stuff)

ice0:
> @ice0 in another PR or this one? (I still didn't start to work on this copy'n paste stuff)

I think better to do it in another PR, so less changes need to be backported.

ice0 (approved):

ice0:
Merged. Thank you!

PR #1973, opened 2021-01-06
[Synfig Studio] fix mainwindow not showing when is offscreen

### LINKED ISSUE #1871: Minimization bug
(we're more likely to help if you're using latest development builds)
Latest normal build
**Issue description**:
Minimizing the window and then closing it causes the app to never get maximized. This includes when it takes time to build and if we minimize it, it seems to crash sometimes and then never gets maximized. Opening the app causes it to be minimized. Uninstalling and reinstalling doesn't seem to work.
**OS so far used and tried on**
Windows
**Success rate of triggering this issue**
One hundred percent so far.

rodolforg:
Synfig Studio restores the last position and size of Main Window
on launching.

If user had multiple monitors when closing Synfig last time he used it,
Main Window could be moved to a no-more existent monitor/area
and not be visible.

Some Window Managers (like on MS Windows case) does not prevents
opening a window on a invisible area.

On MS Windows, closing Synfig Studio app in a minified state
moves the window to position (-32000,-32000).

fix #1871

### DIFF
--- a/synfig-studio/src/gui/docks/dockmanager.cpp
+++ b/synfig-studio/src/gui/docks/dockmanager.cpp
@@ -35,6 +35,9 @@
 #include <stdexcept>
 
 #include <gdkmm/general.h>
+#if GTK_CHECK_VERSION(3,22,0)
+#include <gdkmm/monitor.h>
+#endif
 
 #include <gtkmm/paned.h>
 #include <gtkmm/window.h>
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/docks/dockmanager.cpp
```diff
@@ -535,6 +602,8 @@ Gtk::Widget* DockManager::read_widget(std::string &x)
 
 		if (!widget) return NULL;
 
+		check_and_fix_window_position(left, top, width, height);
```
ice0, line 605: There is a strange problem here. The coordinates of the top left corner of the Synfig main window are 0, 217. At first I thought it was a Windows 10 issue or a GTK bug, but this example works fine and setting the coordinates to 0, 0 moves the window to the top left corner:

```
#include <gtkmm.h>

int main(int argc, char *argv[])
{
	Glib::RefPtr<Gtk::Application> app = Gtk::Application::create(argc, argv, "org.gtkmm.examples.base");

	Gtk::Window window;
	window.set_default_size(200, 200);
	window.move(0, 0);

	return app->run(window);
}
```

But for some reason the Synfig window works differently and the top corner goes out of the visible area.

rodolforg: Sorry, I didn't understand.
I moved the synfig window, positioned the mouse cursor at top left corner, pressed Ctrl+Q to close it, and without moving the mouse pointer, I relaunched Synfig. The window opened in the exact position that mouse pointer marked.

I tested to close minimized window and it shows up at (0,0) on relaunch.

ice0: This is how it looks with 0,0 coordinates.

P.S. This is possible GTK bug.

diLSVY7xPW

rodolforg: That wasn't like that when I tested on Windows o.O

Anyway, try to move the window after `present()` it.

ice0: > Anyway, try to move the window after present() it.

Bug is reproducible when you try to `move()` after `present()`, lol )

And yes, the problem exists because this code is being called twice. I did not dig why this is happening, I will look later.

At the first time window positioned correctly and shows. If you call `move()` again it positions incorrectly. Reproduced only on multi-monitor configuration, if monitors have different DPI.

synfig-studio/src/gui/docks/dockmanager.cpp
```diff
@@ -420,6 +423,70 @@ bool DockManager::read_bool(std::string &x)
 	return read_string(x) == "true";
 }
 
+static void check_and_fix_window_position(int &x, int &y, int &width, int &height)
+{
+	// At least magic_mininum_length x magic_mininum_length pixels of window must be shown on display
+	const int magic_mininum_length = 50;
```
ice0, line 429: constexpr?

synfig-studio/src/gui/docks/dockmanager.cpp
ice0: @rodolforg please add temporary hack here:
```
		// FIXME(ice0): hack to fix incorrect positioning https://github.com/synfig/synfig/issues/1975
		if (App::main_window->is_visible()) App::main_window->hide();
		App::main_window->move(left, top);
		App::main_window->set_default_size(width, height);
```

synfig-studio/src/gui/docks/dockmanager.cpp
rodolforg: I don't get constexpr, but I believe you :)
Done

synfig-studio/src/gui/docks/dockmanager.cpp
rodolforg: Done

rodolforg:
Besides Main Window, I added the fix to floating dialogs too.

ice0 (approved):

ice0:
Merged. Thank you!

PR #1974, opened 2021-01-10
[Synfig Studio] Set Enum parameters to Static by default

### LINKED ISSUE #1972: Use "Constant" interpolation by default for all "enum" parameters
Problem description:

> Noticed that with enums type of variables when some kind of interpolation is set, those will get through all variables (this happened even before)
For example if you interpolate from `Blend Method` `Composite` to `Alpha` Synfig will get through each one of them like from 0 to 20.

> If interpolation is set to Constant it's fine, but I don't know if this is a plus or an issue...

_Originally posted by @FirasH in https://github.com/synfig/synfig/issues/1968#issuecomment-753650346_

FirasH:
Fix #1972


ice0 (approved):

ice0:
Merged. Thank you!

PR #1992, opened 2021-01-21
[Synfig Studio] Fix parameter values of several layers shown as distance metrics when they are not

rodolforg:
It now sets an existent parameter hint (`set_is_distance`) to `synfig::Vector`/`synfig::Point` data type too, not only for `synfig::Real`.

Users are often confused because of it, like here https://forums.synfig.org/t/how-is-scale-calculated/11348


ice0 (approved):

ice0:
Merged. Thank you!

rodolforg:
If @Svarov-RZM can check all layer parameters and valuenodes types, what do you think about backporting to 1.4.x @ice0 ?

ice0:
@rodolforg Sure!

PR #1994, opened 2021-01-22
[Core] WIP - first frame is now numbered as 1, not 0; fix #1990

### LINKED ISSUE #1990: Start time line from frame 1
Following on from discussion here: https://forums.synfig.org/t/how-is-scale-calculated/11348/12 can the start frame be changed from 0 to 1? 

It would bring it in line with other animation software such as Pencil, Opentoonz, Blender, Natron etc

rodolforg:


rodolforg:
I'll create some test cases :p
And maybe some layer parameters must be investigated (those of Time group, for example)

rodolforg:
To fix this issue, I have to create a new class TimeSpan (or Time::Span) and a new ValueBase type (type_timespan).
May I do it, @ice0 @morevnaproject ?

PR #1995, opened 2021-01-23
[Core] create stringhelper

rodolforg:


rodolforg:
@ice0 what do you think about this PR? Useful?
Should we move strprintf and alike from etl to it?

ice0:
> Should we move strprintf and alike from etl to it?

I was planning to replace `strprintf` with `fmt::format` during GSoC 2021 (https://github.com/fmtlib/fmt).
Any other functions can be moved to string_helper.

synfig-core/src/synfig/string_helper.cpp
```diff
@@ -0,0 +1,104 @@
...
+{
+	text.erase(text.begin(),
+			   std::find_if(text.begin(), text.end(),
+							[](int chr) { return !std::isspace(chr);})
+			   );
+	return text;
+}
+
+std::string
+synfig::right_trim(std::string text)
```
ice0, line 87: I suggest use `const std::string& text` if you want non-destructive trimming or `std::string&& text` for destructive.
Just to understand what's happening with the original string.

rodolforg:
Oh the dependecy hell...
Would it work fine with gettext and transifex?
Would it give us any advantage?

Anyway, what else should I move to this PR? Besides strprintf/strscanf, there are only filepath-related functions in stringf.h

ice0:
> Oh the dependecy hell...

Agree. The only reason to switch to `fmt` is because `fmt` is implementation of C++20 std::format. Oh, and speed. https://github.com/fmtlib/fmt#speed-tests

> Would it work fine with gettext and transifex?

Sure. There is no changes here.

> Would it give us any advantage?

Not so much. But better speed, better documentation, errors in format strings at compile time (most important part) and also it's already in C++20 standard.

> Anyway, what else should I move to this PR? Besides strprintf/strscanf, there are only filepath-related functions in stringf.h

I think better to merge this PR, and do other stuff in other PRs. 
Move strprintf/strscanf as separate PR as it's mostly copy-paste work.
Move filepath-related functions to FileSystem class?

synfig-core/src/synfig/string_helper.cpp
rodolforg: As with C++11, it would automatically use `move`, I thought it wouldn't be necessary.
But I'll change it.

rodolforg: Done. Changed all function signatures to `const std::(w)string&`.

ice0 (approved):

ice0:
Merged. Thank you!

Issue #1873, opened 2020-11-27
Installed fonts are missing in Font Family list


FirasH:
**Synfig version & platform**:
master bbb2dd7
Windows 10

**Issue description**:

1) Install a new font (.otf or .ttf)
2) Launch Synfig Studio
3) Insert a `Text Layer` on `Canvas`
4) Go to the `Parameters` panel and change `Font Family` value to select the font installed in step 1

The font is not available in the `Font Family` list.

FirasH:
Works fine on openSUSE

rodolforg:
Maybe fontconfig is not updated automatically…

FirasH:
It has something to do with the last commits (a couple of months).

I have some animations made with previous versions of 1.5.0 on Windows where fonts were correctly listed.
Now the same fonts on the same machine are not listed anymore.

rodolforg:
As the related code didn't change since I created the Widget_FontFamily, I don't know how to inspect it.
Maybe something about building? Autotools x CMake? I don't know :(

ice0:
Just checked. This is FontConfig problem (https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues/144).

As a temporary solution - right-click on your font file and choose "Install for all users".

PR #1997, opened 2021-01-24
[Core/Studio] Added support for user-installed fonts on Windows 10 (1809)

ice0:
Fix https://github.com/synfig/synfig/issues/1873

### DIFF
--- a/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
+++ b/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
@@ -61,6 +61,7 @@
 #endif
 
 #include <algorithm>
+#include <glibmm.h>
 
 #endif
 
@@ -169,6 +170,14 @@ private:
 	FontConfigWrap()
 	{
 		config = FcInitLoadConfigAndFonts();
+#ifdef _WIN32
+		// Windows 10 (1809) Added local user fonts installed to C:\Users\%USERNAME%\AppData\Local\Microsoft\Windows\Fonts
+		std::string localdir = Glib::getenv("LOCALAPPDATA");
+		if (!localdir.empty()) {
+			localdir.append("\\Microsoft\\Windows\\Fonts\\");
+			FcConfigAppFontAddDir(config, (const FcChar8 *)localdir.c_str());
+		}
+#endif
 	}
 	~FontConfigWrap() {
 		FcConfigDestroy(config);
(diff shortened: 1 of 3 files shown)


synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
```diff
@@ -675,7 +684,18 @@ Layer_Freetype::new_face(const String &newfont)
...
 
 #ifdef _WIN32
-	possible_font_directories.push_back("C:\\WINDOWS\\FONTS\\");
+	// All users fonts
+	std::string windir = Glib::getenv("windir");
+	if (windir.empty()) {
+		possible_font_directories.emplace_back("C:\\WINDOWS\\FONTS\\");
+	} else {
+		possible_font_directories.emplace_back(windir + "\\Fonts\\");
+	}
```
rodolforg, line 693: What about:
```c++
	if (windir.empty())
		windir = "C:\\WINDOWS;
	possible_font_directories.emplace_back(windir + "\\Fonts\\");

File: synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp [hunk 2]
```diff
@@ -675,7 +684,18 @@ Layer_Freetype::new_face(const String &newfont)
...
+	// Windows 10 (1809) Added local user fonts installed to C:\Users\%USERNAME%\AppData\Local\Microsoft\Windows\Fonts
+	std::string localdir = Glib::getenv("LOCALAPPDATA");
+	if (!localdir.empty()) {
+		possible_font_directories.emplace_back(localdir + "\\Microsoft\\Windows\\Fonts\\");
+	}
```
rodolforg, line 698: As the code is duplicated, would it worth to create a simple static (='local', not in the sense of a class method) function near file header?
```c++
#ifdef _WIN32
// Windows 10 (1809) Added local user fonts installed to C:\Users\%USERNAME%\AppData\Local\Microsoft\Windows\Fonts
static std::string get_user_font_folder() {
...
}
#endif 
```

synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
ice0: As for me, it's just faster, but I can change it if it's look more clear

synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
ice0: Any idea where to put it? As I can see where is no shared headers between `lyr_freetype` and `widget_fontfamily`.

synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
rodolforg: It's a matter of taste, actually. But I think it's error prone.

synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
rodolforg: I don't know.
I think Synfig library should provide those system and resource paths :P 
Anyway, not important.

ice0:
Can't backport to 1.4 because it doesn't have required changes to text layer

Issue #1410, opened 2020-05-15
[Future request] Allow user to limit number of cores the render can utilize


Svarov-RZM:
**Description**
The rendering engine hogs all available CPU cores by default. I think it would be good to allow user to limit the amount of cores Synfig uses. For example, I usually listen to various playlists on YouTube while I'm working and in some cases those CPU spikes to 100% when Synfig renders current frame is noticeable.

It would be good to have such setting in "Preferences->Render" dialog, something like "Limit render to number of cores".

ice0:
I agree. As a temporary solution, you can limit the number of rendering threads using the environment variable `SYNFIG_GENERIC_THREADS`.

Just run "SYNFIG_GENERIC_THREADS = 2 ./synfigstudio" on Linux or set it in the settings on Windows.

SushantAA:
I want to work on this issue 
pls assign me 
is there any chat Chanel

ice0:
Hi and welcome here!
Assigned to you.

P.S.
Unfortunately there is no chat here.
Btw, you can ask here or at https://forums.synfig.org

SushantAA:
thanks :smile:

DhairyaBahl:
@ice0 PR related to the issue is merged ! Is it resolved ? If yes then kindly close it :) If not, then kindly assign it to me.

Thanks and Regards :)

PR #1999, opened 2021-01-24
[Synfig Studio] Allow to set number of rendering threads from Preferences dialog

SushantAA:
fix - #1410 
I am new to GTK :turtle: 
pls help me in verifying the logic 
currently, it only changes the core to 3 ( I don't know how to check [ how many cores is running ] )

### DIFF
--- a/synfig-core/src/synfig/threadpool.cpp
+++ b/synfig-core/src/synfig/threadpool.cpp
@@ -277,17 +277,15 @@ ThreadPool::enqueue(const Slot &slot) {
 
 
 void 
-ThreadPool::set_threads(int no_core){
+ThreadPool::set_num_threads(int num_cores){
 	max_running_threads = g_get_num_processors();
-	if(no_core!=0){
-		max_running_threads = no_core;
+	if(num_cores!=0){
+		max_running_threads = num_cores;
 	}
 	if (const char *s = getenv("SYNFIG_GENERIC_THREADS"))
 		max_running_threads = atoi(s) + 1;
 
 	if (max_running_threads < 2) max_running_threads = 2;
-	if (max_running_threads > 2) --max_running_threads;
-	++running_threads;
 }
 
 void
--- a/synfig-core/src/synfig/threadpool.cpp
+++ b/synfig-core/src/synfig/threadpool.cpp
@@ -275,6 +275,21 @@ ThreadPool::enqueue(const Slot &slot) {
 	wakeup();
 }
 
+
+void 
+ThreadPool::set_threads(int no_core){
+	max_running_threads = g_get_num_processors();
+	if(no_core!=0){
+		max_running_threads = no_core;
+	}
+	if (const char *s = getenv("SYNFIG_GENERIC_THREADS"))
+		max_running_threads = atoi(s) + 1;
+
+	if (max_running_threads < 2) max_running_threads = 2;
+	if (max_running_threads > 2) --max_running_threads;
+	++running_threads;
+}
+
 void
 ThreadPool::wait(std::condition_variable &cond, std::unique_lock<std::mutex> &lock) {
 	if (--running_threads < max_running_threads)
--- a/synfig-core/src/synfig/threadpool.cpp
+++ b/synfig-core/src/synfig/threadpool.cpp
@@ -277,10 +277,10 @@ ThreadPool::enqueue(const Slot &slot) {
 
 
 void 
-ThreadPool::set_num_threads(int num_cores){
+ThreadPool::set_num_threads(int num_threadsd){
 	max_running_threads = g_get_num_processors();
-	if(num_cores!=0){
-		max_running_threads = num_cores;
+	if(num_threadsd!=0){
+		max_running_threads = num_threadsd;
 	}
 	if (const char *s = getenv("SYNFIG_GENERIC_THREADS"))
 		max_running_threads = atoi(s) + 1;
--- a/synfig-core/src/synfig/threadpool.cpp
+++ b/synfig-core/src/synfig/threadpool.cpp
@@ -277,10 +277,10 @@ ThreadPool::enqueue(const Slot &slot) {
 
 
 void 
-ThreadPool::set_num_threads(int num_threadsd){
+ThreadPool::set_num_threads(int num_threads){
 	max_running_threads = g_get_num_processors();
-	if(num_threadsd!=0){
-		max_running_threads = num_threadsd;
+	if(num_threads!=0){
+		max_running_threads = num_threads;
 	}
 	if (const char *s = getenv("SYNFIG_GENERIC_THREADS"))
 		max_running_threads = atoi(s) + 1;
--- a/synfig-core/src/synfig/threadpool.h
+++ b/synfig-core/src/synfig/threadpool.h
@@ -43,6 +43,7 @@
 
 /* === C L A S S E S & S T R U C T S ======================================= */
 
+using namespace synfig;
 namespace synfig {
 
 class ThreadPool {
@@ -97,6 +98,8 @@ public:
 
 	void enqueue(const Slot &slot);
 	void wait(std::condition_variable &cond, std::unique_lock<std::mutex>& lock);
+
+	void set_threads(int no_core);
 
 	int get_max_threads() const
 		{ return max_running_threads; }
--- a/synfig-core/src/synfig/threadpool.h
+++ b/synfig-core/src/synfig/threadpool.h
@@ -43,7 +43,6 @@
 
 /* === C L A S S E S & S T R U C T S ======================================= */
 
-using namespace synfig;
 namespace synfig {
 
 class ThreadPool {
@@ -99,7 +98,7 @@ public:
 	void enqueue(const Slot &slot);
 	void wait(std::condition_variable &cond, std::unique_lock<std::mutex>& lock);
 
-	void set_threads(int no_core);
+	void set_num_threads(int num_cores);
 
 	int get_max_threads() const
 		{ return max_running_threads; }
--- a/synfig-core/src/synfig/threadpool.h
+++ b/synfig-core/src/synfig/threadpool.h
@@ -98,7 +98,7 @@ public:
 	void enqueue(const Slot &slot);
 	void wait(std::condition_variable &cond, std::unique_lock<std::mutex>& lock);
 
-	void set_num_threads(int num_cores);
+	void set_num_threads(int num_threadsd);
 
 	int get_max_threads() const
 		{ return max_running_threads; }
--- a/synfig-core/src/synfig/threadpool.h
+++ b/synfig-core/src/synfig/threadpool.h
@@ -98,7 +98,7 @@ public:
 	void enqueue(const Slot &slot);
 	void wait(std::condition_variable &cond, std::unique_lock<std::mutex>& lock);
 
-	void set_num_threads(int num_threadsd);
+	void set_num_threads(int num_threads);
 
 	int get_max_threads() const
 		{ return max_running_threads; }
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -285,7 +285,7 @@ std::set< String > studio::App::brushes_path;
 String studio::App::image_editor_path;
 
 String studio::App::sequence_separator(".");
-String studio::App::number_of_cores("" + ('0'+g_get_num_processors()) );
+int studio::App::number_of_cores(g_get_num_processors());
 String studio::App::navigator_renderer;
 String studio::App::workarea_renderer;
 
@@ -733,7 +733,7 @@ public:
 			}
 			if(key=="number_of_cores")
 			{
-				App::number_of_cores=value;
+				App::number_of_cores=atoi(value.c_str());
 				return true;
 			}
 			if(key=="navigator_renderer")
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -285,7 +285,7 @@ std::set< String > studio::App::brushes_path;
 String studio::App::image_editor_path;
 
 String studio::App::sequence_separator(".");
-int    studio::App::number_of_cores = g_get_num_processors();
+int    studio::App::number_of_threads = g_get_num_processors();
 String studio::App::navigator_renderer;
 String studio::App::workarea_renderer;
 
@@ -543,9 +543,9 @@ public:
 				value=App::sequence_separator;
 				return true;
 			}
-			if(key=="number_of_cores")
+			if(key=="number_of_threads")
 			{
-				value=App::number_of_cores;
+				value=App::number_of_threads;
 				return true;
 			}
 			if(key=="navigator_renderer")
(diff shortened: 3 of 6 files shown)


synfig-core/src/synfig/threadpool.h
```diff
@@ -98,6 +99,8 @@ class ThreadPool {
 	void enqueue(const Slot &slot);
 	void wait(std::condition_variable &cond, std::unique_lock<std::mutex>& lock);
 
+	void set_threads(int no_core);
```
rodolforg, line 102: Please rename the method to `set_num_threads` ('num' from 'number'): let the method name descriptive in order to help code reading.

And rename `no_core` to `num_cores`or something like that: 'no' means 'without' here

synfig-studio/src/gui/dialogs/dialog_setup.cpp
```diff
@@ -535,7 +542,7 @@ Dialog_Setup::create_render_page(PageInfo pi)
 	preview_background_color_button.signal_color_set().connect(
 		sigc::mem_fun(*this, &studio::Dialog_Setup::on_preview_background_color_changed) );
 
-}
+}
```
rodolforg, line 545: trim whitespace

synfig-studio/src/gui/app.h
```diff
@@ -51,6 +51,7 @@
 #include <synfig/layers/layer_bitmap.h>
 #include <synfig/string.h>
 #include <synfig/time.h>
+#include <synfig/threadpool.h>
```
rodolforg, line 54: Why did you add it?
Neither `app.h` nor `app.cpp` need it.

synfig-studio/src/gui/app.cpp
```diff
@@ -285,6 +285,7 @@ std::set< String > studio::App::brushes_path;
 String studio::App::image_editor_path;
 
 String studio::App::sequence_separator(".");
+String studio::App::number_of_cores("" + ('0'+g_get_num_processors()) );
```
rodolforg, line 288: use standard c++ function : `std::to_string()` instead of ASCII tricks.
But why you store an integer number as text?!

synfig-core/src/synfig/threadpool.h
```diff
@@ -43,6 +43,7 @@
 
 /* === C L A S S E S & S T R U C T S ======================================= */
 
+using namespace synfig;
```
rodolforg, line 46: [Never use `using namespace` in a header file](https://stackoverflow.com/questions/5849457/using-namespace-in-c-headers).

synfig-studio/src/gui/dialogs/dialog_setup.cpp
```diff
@@ -669,6 +676,8 @@ Dialog_Setup::on_restore_pressed()
 		fps_template_combo->set_active_text(DEFAULT_PREDEFINED_FPS);
 		adj_pref_fps->set_value(24.0);
 		image_sequence_separator.set_text(".");
+		core_render.set_text("" + ('0'+g_get_num_processors()));
```
rodolforg, line 679: as already said in this review: use `std::to_string()`

synfig-studio/src/gui/dialogs/dialog_setup.cpp
```diff
@@ -904,6 +916,12 @@ Dialog_Setup::on_autobackup_changed()
 	auto_backup_interval.set_sensitive(toggle_autobackup.get_active());
 }
 
+void
+Dialog_Setup::on_core_change()
```
rodolforg, line 920: give a better name to method: the 'core' didn't change. ;)

synfig-core/src/synfig/threadpool.cpp
```diff
@@ -275,6 +275,21 @@ ThreadPool::enqueue(const Slot &slot) {
...
+	max_running_threads = g_get_num_processors();
+	if(no_core!=0){
+		max_running_threads = no_core;
+	}
+	if (const char *s = getenv("SYNFIG_GENERIC_THREADS"))
+		max_running_threads = atoi(s) + 1;
+
+	if (max_running_threads < 2) max_running_threads = 2;
+	if (max_running_threads > 2) --max_running_threads;
+	++running_threads;
```
rodolforg, line 290: I couldn't understand why did you do it…

synfig-core/src/synfig/threadpool.cpp
```diff
@@ -275,6 +275,21 @@ ThreadPool::enqueue(const Slot &slot) {
...
+}
+
```
rodolforg, line 292: Maybe the `max_running_threads`should be bounded to num_processors, even if 'no_core' is greater than it.

rodolforg (changes requested):

synfig-studio/src/gui/dialogs/dialog_setup.cpp
```diff
@@ -501,6 +501,13 @@ Dialog_Setup::create_render_page(PageInfo pi)
 	pi.grid->attach(image_sequence_separator, 1, row, 1, 1);
 	image_sequence_separator.set_hexpand(true);
 
+	// render - cores
+	attach_label(pi.grid, _("Limit render to number of cores"), ++row);
+	pi.grid->attach(core_render, 1, row, 1, 1);
```
rodolforg, line 506: `core_render` is really a bad name choice :P It doesn't mean what it stores/is.

SushantAA: I know they are bad names [ Sorry I forgot to write it :disappointed:   ] but I just wanted to verify logic. because I am new here.
Sorry for the trouble in pointing out the bad name.

synfig-studio/src/gui/app.h
```diff
@@ -227,6 +228,7 @@ class App : public Gtk::Main, private IconController
 	static synfig::String predefined_fps;
 	static float preferred_fps;
 	static synfig::String sequence_separator;
+	static synfig::String number_of_cores;
```
ice0, line 231: Better to use integer value here

synfig-studio/src/gui/dialogs/dialog_setup.cpp
ice0: No problem here. Logic is looks correct. Also you need to check CI builds. Currently they fails with error.

rodolforg: @SushantAA I'm sorry if it sounded harsh, it wasn't my intention. It was my last comment about variable name and I explained it better in previous comments, so I thought I could be more concise. I was wrong :(

SushantAA: @rodolforg nothing was harsh, it's fine  :smiley:

synfig-studio/src/gui/app.h
SushantAA: to change the number of cores by calling a function 
https://github.com/SushantAA/synfig/blob/issue-1410/synfig-core/src/synfig/threadpool.cpp#L280

synfig-studio/src/gui/dialogs/dialog_setup.h
```diff
@@ -163,6 +164,7 @@ class Dialog_Setup : public Dialog_Template
 	Gtk::Entry        image_sequence_separator;
 	Gtk::ComboBoxText workarea_renderer_combo;
 	Gtk::Switch       toggle_play_sound_on_render_done;
+	Gtk::Entry		  number_of_core_select;
```
SushantAA, line 167: ah, i will correct that space

SushantAA:
Basically pls 
* verify the logic and the implementation
* I have no idea how to check the distribution of core ( I use OS: Linux mint 20)
* I am not able to set the initial display of the number of cores on the Gtk::Entry

synfig-studio/src/gui/app.cpp
```diff
@@ -285,6 +285,7 @@ std::set< String > studio::App::brushes_path;
 String studio::App::image_editor_path;
 
 String studio::App::sequence_separator(".");
+int studio::App::number_of_cores(g_get_num_processors());
```
rodolforg, line 288: cosmetic: it isn't related to surrounding variables. Move it to below ` String studio::App::workarea_renderer; ` separated by a blank line :)

rodolforg (changes requested):
As ice0 said, your logic seems ok :)

ice0:
> * verify the logic and the implementation

Looks good. 

> * I have no idea how to check the distribution of core ( I use OS: Linux mint 20)

Can you explain this with more details?

> * I am not able to set the initial display of the number of cores on the Gtk::Entry

You've already done it. But there is a bug in your code. Please check `~/.config/synfig/settings-1.4` to see why it doesn't work.

synfig-studio/src/gui/app.h
SushantAA: I tried removing it but it showed a build error

rodolforg: You should read the build error to check what file misses this header and fix it properly.

synfig-studio/src/gui/dialogs/dialog_setup.cpp
```diff
@@ -1064,6 +1080,9 @@ Dialog_Setup::refresh()
 	// Refresh the sequence separator
 	image_sequence_separator.set_text(App::sequence_separator);
 
+	// Refresh the number of core
+	number_of_core_select->set_value(int(adj_number_of_core->get_value()));
```
rodolforg, line 1084: This isn't right: `refresh()` is to update the form fields, and not set the field with its "own" value.
You could keep the App "member" variable you created commits ago, but now removed and used it here :) Something like:
`number_of_core_select->set_value(App::num_threads);`

synfig-studio/src/gui/dialogs/dialog_setup.h
```diff
@@ -163,6 +164,8 @@ class Dialog_Setup : public Dialog_Template
 	Gtk::Entry        image_sequence_separator;
 	Gtk::ComboBoxText workarea_renderer_combo;
 	Gtk::Switch       toggle_play_sound_on_render_done;
+	Glib::RefPtr<Gtk::Adjustment> adj_number_of_core;
+	Gtk::SpinButton*	number_of_core_select;
```
rodolforg, line 168: Answering your comment here: you can align line 168 with lines 164-166, but with spaces, not tab.

synfig-studio/src/gui/dialogs/dialog_setup.h
```diff
@@ -99,6 +99,7 @@ class Dialog_Setup : public Dialog_Template
 	void on_def_background_type_changed(); //bound on clicked
 	void on_def_background_color_changed();
 	void on_def_background_image_set();
+	void on_number_of_core_select();
```
rodolforg, line 102: Looking to other signals slots/callbacks name: prefer `on_number_of_core_changed()`.
In earlier commits you used `on_core_changed()`, and it led to a confusion: did the core changed? No, the number of core (or threads).

synfig-studio/src/gui/dialogs/dialog_setup.cpp
```diff
@@ -904,6 +914,12 @@ Dialog_Setup::on_autobackup_changed()
 	auto_backup_interval.set_sensitive(toggle_autobackup.get_active());
 }
 
+void
+Dialog_Setup::on_number_of_core_select()
+{
+	ThreadPool::instance().set_num_threads(int(adj_number_of_core->get_value()));
```
rodolforg, line 920: Gtk/gtkmm has a method here that help you: `Gtk::SpinButton::get_value_as_int()` :D

synfig-studio/src/gui/dialogs/dialog_setup.cpp
```diff
@@ -76,6 +77,7 @@ Dialog_Setup::Dialog_Setup(Gtk::Window& parent):
 	adj_pref_x_size(Gtk::Adjustment::create(480,1,10000,1,10,0)),
 	adj_pref_y_size(Gtk::Adjustment::create(270,1,10000,1,10,0)),
 	adj_pref_fps(Gtk::Adjustment::create(24.0,1.0,100,0.1,1,0)),
+	adj_number_of_core(Gtk::Adjustment::create(g_get_num_processors(),1,g_get_num_processors(),1,10,0)),
```
rodolforg, line 80: Reading the code of `ThreadPool::set_num_threads(int num_cores)`, the minimum value should be 2, not 1.

rodolforg (changes requested):
@ice0 do you think it should use the word `core` instead of `thread`?

synfig-studio/src/gui/dialogs/dialog_setup.h
SushantAA: is `on_number_of_thread_changed()`  fine

synfig-studio/src/gui/dialogs/dialog_setup.cpp
SushantAA: get_value_as_int() shows error as it is not a member of Gtk::Adjustment
in other places also it is not used
https://github.com/synfig/synfig/blob/master/synfig-studio/src/gui/dialogs/dialog_setup.cpp#L760

synfig-studio/src/gui/dialogs/dialog_setup.h
SushantAA: wrong_align
I don't know why it's happening
https://github.com/SushantAA/synfig/blob/issue-1410/synfig-studio/src/gui/dialogs/dialog_setup.h#L168

synfig-studio/src/gui/dialogs/dialog_setup.cpp
rodolforg: It's for `Gtk::SpinButton`, not for its `Gtk::Adjustment`. But ok.

synfig-studio/src/gui/dialogs/dialog_setup.h
ice0: Agree. But better on_number_of_thread*s*_changed()

synfig-studio/src/gui/dialogs/dialog_setup.cpp
ice0: @rodolforg I think threads is better. We can't change CPU cores :)

ice0:
Looks fine. The only suggestion is to rename cores to threads, because we changing number of rendering threads.

SushantAA:
@rodolforg @ice0 I think its complete

ice0 (approved):
Looks good to me

rodolforg (approved):

ice0:
Merged. Thank you!

synfig-studio/src/gui/app.cpp
```diff
@@ -542,6 +543,11 @@ class Preferences : public synfigapp::Settings
 				value=App::sequence_separator;
 				return true;
 			}
+			if(key=="number_of_threads")
+			{
+				value=App::number_of_threads;
```
rodolforg, line 548: @SushantAA  maybe you should convert this integer into string before assign to value?

SushantAA: It should be a string :grimacing:
@rodolforg should the initial setting of threads should be 2 or 12. currently, it's 2 but I think 12 will be better

rodolforg: Look few lines above: use `etl::strprintf()` to 'convert' the `App::number_of_threads`.

IIRC, you make it to be the maximum threads as default.

SushantAA: https://github.com/synfig/synfig/pull/2023

PR #2011, opened 2021-02-06
[ETL] current_working_directory() function now works correctly with non-Latin filenames

ice0:
This problem mentioned by @rodolforg here: https://github.com/synfig/synfig/pull/1952, so I fixed it.

> etl::absolute_path() trusting on getcwd(), that doesn't handle the famous encoding problems.

### DIFF
--- a/ETL/ETL/_stringf.h
+++ b/ETL/ETL/_stringf.h
@@ -37,6 +37,7 @@
 #include <string>
 #include <cstdarg>
 #include <cstdlib>
+#include <glibmm/miscutils.h>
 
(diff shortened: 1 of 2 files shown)


rodolforg (approved):
I really don't like to make libsynfig (more) dependent on glibmm, it prevents to use libsynfig on simpler projects (like embed it on Godot Engine or maybe some Android projects).

Maybe use iconv to convert utf-16 to utf-8 (on windows) would solve this issue.

Anyway, it isn't a reality right now. So, ok.

PR #2029, opened 2021-02-18
[Core] assure uniqueness for new node guid

rodolforg:


rodolforg:
Not probable, but not impossible to happen…

Maybe backport to 1.4.1?

synfig-core/src/synfig/node.cpp
```diff
@@ -202,8 +204,11 @@ Node::get_guid()const
 	std::lock_guard<std::mutex> lock(guid_mutex_);
 	if(!guid_)
 	{
-		guid_.make_unique();
-		global_node_map().add(guid_, const_cast<Node*>(this));
+		bool added = false;
```
ice0, line 207: Maybe shorten this code?
```
do {
	guid_.make_unique();
} while (!global_node_map().add(guid_, const_cast<Node*>(this)));
```

rodolforg: Yeah, I thought it this way, but maybe not that readable.
Compiler would throw that variable away anyway, right?

ice0: > Compiler would throw that variable away anyway, right?
Yes.

ice0 (approved):

ice0:
Merged. Thank you!

Issue #540, opened 2018-05-08
100% crash when drag and drop layer while renaming


Jengall:
**Synfig 1.2.1 & pc x64**:

**Issue description**:
When I drag and drop layer (to move them up down or inside group, it occur that the layer go in "rename mode" than it s a 100% crash if I move it.

Jengall:
win 10 64b sorry

morevnaproject:
Cannot reproduce it with the latest master branch.

Jose-Moreno:
@morevnaproject I've been able to reproduce this in the past, but I'd like to attempt to reproduce this with the master branch. Would I need to compile Synfig for Windows then? or do you have any automated build for Windows that can be used for testing?

ice0:
@Jose-Moreno i can't reproduce it on latest master.
Can you check it also?

https://coolbug.org/downloads/morevna/builds/SynfigStudio-1.3.12-testing-2020.01.25-win64-f6e20.zip

Jose-Moreno:
@ice0 Hi, sorry I've been migrating to a new computer recently. I'll test this as soon as I'm able. Thanks!

Jose-Moreno:
@ice0 Alright I tested this, here's a video you can watch with the problem. Before recording I experienced two previous crashes. Right now it's much more difficult to make the program crash (it now crashes silently) but it can still happen. 

I noticed there's a lock icon when trying to drag a layer that is being edited, however you can still drag layers that have their "rename" state enabled, and while this no longer seems to crash the program 100%, it could be still contributing to it.

https://youtu.be/AVxZhBJwaF8

The crash occurs a little bit after 3:20 timestamp but I showed the other behavior during the video. I hope this helps.

https://youtu.be/AVxZhBJwaF8

rodolforg:
I could reproduce it once only. Here's the backlog:
```
(gdb) bt
#0  0x00007ffff505f5a0 in _int_malloc (av=av@entry=0x7ffff5193c40 <main_arena>, bytes=bytes@entry=24) at malloc.c:3683
#1  0x00007ffff5060997 in __GI___libc_malloc (bytes=24) at malloc.c:3073
#2  0x00007fffe3b3f208 in  () at /usr/lib/x86_64-linux-gnu/librsvg-2.so.2
#3  0x00007fffe3b7bae6 in  () at /usr/lib/x86_64-linux-gnu/librsvg-2.so.2
[...]

PR #2043, opened 2021-02-23
[Synfig Studio] avoid dereference of invalid TreeIter on Layer TreeView

rodolforg:
related to #540

### DIFF
--- a/synfig-studio/src/gui/trees/layertree.cpp
+++ b/synfig-studio/src/gui/trees/layertree.cpp
@@ -775,7 +775,10 @@ LayerTree::on_edited_value(const Glib::ustring&path_string,synfig::ValueBase val
 {
 	Gtk::TreePath path(path_string);
 
-	const Gtk::TreeRow row = *(param_tree_view().get_model()->get_iter(path));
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/trees/layertree.cpp
```diff
@@ -775,7 +775,10 @@ LayerTree::on_edited_value(const Glib::ustring&path_string,synfig::ValueBase val
 {
 	Gtk::TreePath path(path_string);
 
-	const Gtk::TreeRow row = *(param_tree_view().get_model()->get_iter(path));
+	const Gtk::TreeIter iter = param_tree_view().get_model()->get_iter(path);
+	if (!iter)
+		return;
+	const Gtk::TreeRow row = *iter;
```
ice0, line 781: Maybe better to use reference here?
```
const Gtk::TreeRow& row = *iter;
```

rodolforg: It seems it works :)
Done

ice0 (approved):

ice0:
Merged. Thank you!

PR #2044, opened 2021-02-23
[Core] Fixed warnings

ice0:


synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -1080,8 +1080,10 @@ StateBone_Context::find_bone(Point point,Layer::Handle layer,int lay)const
 
 		}
 		if(abs(close_line)<=0.2){
-			if(ret<list.size())return ret;
-			else return -1;
+			if (ret < static_cast<int>(list.size()))
```
rodolforg, line 1083: you could simply use `int(list.size())`, don't you?

ice0: Yeah.

rodolforg (approved):

PR #2050, opened 2021-02-27
[Synfig Studio] don't let inner layer be duplicated twice

rodolforg:
If action LayerDuplicate has among the layers to be duplicated a
layer named "a" that is 'inside' of a group layer (pastecanvas)
named "g". In case of "g" is a layer included in the action layer
list too, don't duplicate the "a" twice, ie. the group and
its contents - including "a" - are duplicated, but "a" won't
be duplicated again.

It would clash the GUID of both "a" duplicates, and that leads
to crash.
A way to avoid it in a general way would be too troublesome :P
Many hierarchy levels and special GUIDs to handle.

Regular (and advanced) user normally won't need it, anyway.

Full description and minor discussion here:
https://forums.synfig.org/t/proposal-of-change-of-behavior-duplicating-both-group-and-an-inner-layer/11823

### DIFF
--- a/synfig-studio/src/synfigapp/actions/layerduplicate.cpp
+++ b/synfig-studio/src/synfigapp/actions/layerduplicate.cpp
@@ -30,18 +30,18 @@
 #	include <config.h>
 #endif
 
-#include <synfig/general.h>
-
 #include "layerduplicate.h"
-#include "layeradd.h"
-#include <synfig/context.h>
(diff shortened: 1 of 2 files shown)


rodolforg:
fixed the test to reflect the new behavior

ice0 (approved):

ice0:
Merged. Thank you!

PR #2060, opened 2021-03-09
[Snapcraft] Added Lottie dependency (lxml)

ice0:
Also removed slang dependency
and added `readme`

### DIFF
--- a/autobuild/snap-stable/snapcraft.yaml
+++ b/autobuild/snap-stable/snapcraft.yaml
@@ -1,6 +1,5 @@
 name: synfigstudio
 title: 'Synfig Studio'
-base: core18
 version: '1.4.0'
 summary: Create and edit 2D animations and compositions
 description: |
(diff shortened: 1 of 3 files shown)


rodolforg:
\o/
Does AppImage need this fix too?
So, now the most problematic issue on 1.4.1 milestone is AppImage "freezing"/slowness? :)
But I want to fix copy-n-paste skeleton and time-loop as valuenode problem before 1.4.1 release :D

ice0:
> Does AppImage need this fix too?

I am not sure, so I need to check this.

> So, now the most problematic issue on 1.4.1 milestone is AppImage "freezing"/slowness? :)

Yes. This problem is related to FontConfig. It creates a font cache at startup, so we need to check here - can we show some message box before it starts working, or it starts right after the library is loaded. 

> But I want to fix copy-n-paste skeleton and time-loop as valuenode problem before 1.4.1 release :D

No problem. As @morevnaproject is still busy we have a lot of time here :)

rodolforg:
As @morevnaproject is still busy we have a lot of time here :)

Happy and sad at same time... lol

morevnaproject (approved):

PR #2064, opened 2021-03-15
[Studio] Hide "Make Parent To Active Bone" menu item when not using Skeleton Tool

ice0:
Fix https://github.com/synfig/synfig/issues/1669

### DIFF
--- a/synfig-studio/src/gui/instance.cpp
+++ b/synfig-studio/src/gui/instance.cpp
@@ -1230,7 +1230,7 @@ Instance::make_param_menu(Gtk::Menu *menu,synfig::Canvas::Handle canvas, synfiga
 	}
 
 	if (std::string("bone") == canvas_view->get_smach().get_state_name()) {
-		param_list.add("child",canvas_view->get_work_area()->get_active_bone_value_node());
+		param_list.add("active_bone", canvas_view->get_work_area()->get_active_bone_value_node());
 	}
 
 	if(param_list2.empty())
--- a/synfig-studio/src/gui/instance.cpp
+++ b/synfig-studio/src/gui/instance.cpp
@@ -1229,7 +1229,9 @@ Instance::make_param_menu(Gtk::Menu *menu,synfig::Canvas::Handle canvas, synfiga
 			param_list.add("selected_value_desc",selected_value_desc);
 	}
 
-	param_list.add("child",canvas_view->get_work_area()->get_active_bone_value_node());
+	if (std::string("bone") == canvas_view->get_smach().get_state_name()) {
+		param_list.add("child",canvas_view->get_work_area()->get_active_bone_value_node());
+	}
 
 	if(param_list2.empty())
 		add_actions_to_menu(&parammenu, param_list,categories);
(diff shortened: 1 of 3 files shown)


synfig-studio/src/gui/instance.cpp
```diff
@@ -1218,7 +1218,9 @@ Instance::make_param_menu(Gtk::Menu *menu,synfig::Canvas::Handle canvas, synfiga
 			param_list.add("selected_value_desc",selected_value_desc);
 	}
 
-	param_list.add("child",canvas_view->get_work_area()->get_active_bone_value_node());
+	if (std::string("bone") == canvas_view->get_smach().get_state_name()) {
```
rodolforg, line 1221: I don't think you need to explictly cast here.

rodolforg (commented):
What do you think about remove that already merged method get_state on this PR too?

synfig-studio/src/gui/instance.cpp
ice0: `get_state_name` returns `const char *` so to compare it, you need to cast. Or use something like `strcmp` which I don't want to use.

ice0:
> What do you think about remove that already merged method get_state on this PR too?

Ok, I will take a look.

synfig-studio/src/gui/instance.cpp
rodolforg: Oh, ok. I thought it would return std::string. Sorry.

synfig-studio/src/gui/instance.cpp
```diff
@@ -1218,7 +1218,9 @@ Instance::make_param_menu(Gtk::Menu *menu,synfig::Canvas::Handle canvas, synfiga
...
+		param_list.add("child",canvas_view->get_work_area()->get_active_bone_value_node());
```
rodolforg, line 1222: what do you think about renaming `"child"` to `"active_bone"` ?

ice0:
> What do you think about remove that already merged method get_state on this PR too?

The problem is in this method:
```
Action::ValueDescSet::prepare()
{
	clear();

	get_canvas_interface()->signal_value_desc_set()(value_desc,value);

	// If current tool is Bone Tool, changing a bone of a Skeleton Deformation
	// Layer should affect its pair too. One follow each other with Bone Tool.
	// Just remembering: Skeleton Deformation a list of bone pairs:
	//   * the first element of every pair compounds the resting pose
	//   * the second element of every pair compounds the current animated pose
	if (get_canvas_interface()->get_state()=="bone") {
```

It uses the `get_state()` method to get the current instrument. And I'm not sure about it.
Can we move this check out of this method? Maybe pass some parameter here?

synfig-studio/src/gui/instance.cpp
ice0: This would require renaming this parameter in `valuedescbonesetparent.cpp` and maybe somewhere else, so I suggest leaving it as is for now.

rodolforg: Only those two files.
It was the GSoC student that created it: just for ValueDescBoneSetParent action he created. 0948ac4b5883d4d10563a98fd0b38459ff7b6a38

rodolforg (dismissed):

synfig-studio/src/gui/instance.cpp
ice0: Done

rodolforg (approved):

PR #2072, opened 2021-03-22
[Appveyor] Temporarily disable artifact storing (run out of space)

ice0:


morevnaproject (approved):

rodolforg:
By the way, any idea why it suddenly increased the artifact size?

ice0:
> By the way, any idea why it suddenly increased the artifact size?

The artifacts stored for six months, so we just exceeded our limits. The problem is that we cannot delete them, so we just need to wait for a while until they expire.

Issue #1557, opened 2020-07-13
[Skeleton Tool] Use single parameter to define bone width


morevnaproject:
I have played a bit with current Skeleton Tool and I found that it will be more convenient to have one single parameter for bone width, instead of two separate parameters for origin and tip.

![screenshot_005](https://user-images.githubusercontent.com/332868/87326522-097c6880-c55d-11ea-8fca-e48b8567ea1a.png)

Can you change that, please? ^__^

morevnaproject:
After thinking again on that, I got another idea:

It would be nice to have 2 parameters, but one will define width (tip and origin) for bones created in Regular Skeleton Layer, and second parameter will define width  (tip and origin) for Skeleton Deformation layer.

Also, it would be nice to temporary hide one parameter, depending on which layer type we are working with. ^__^

Aa20475:
hey! Please test this in the latest commit!

morevnaproject:
Hi! I have tested last commit. One correction:

> it would be nice to temporary hide one parameter, depending on which layer type we are working with

Currently it switches visible  parameter depending on layer type selected in Tool Options Panel (1).
But instead it is expected to switch visibility depending on active layer (2). ^__^

 
![screenshot_012](https://user-images.githubusercontent.com/332868/87678534-46d33700-c7a5-11ea-9ec5-251303c6f3f6.png)

Aa20475:
Can you verify this? I tried something :sweat_smile:

morevnaproject:
Hi! I have tested, but behavior is really weird... For example, I can't set Bone Width value to  0.10 or 1.50 - only integer values accepted. Also, at some moments it resets width to other weird values - like 0.0u or 7.0u...

Aa20475:
Ohkay! Will fix it!

morevnaproject:
I think you can implement it in the following way:

You can have two widgets in Tool Options Panel - one will define width (tip and origin) for bones created in Regular Skeleton Layer, and second parameter will define width (tip and origin) for Skeleton Deformation layer.

And all you have to do is just show/hide particular widget depending on the type of selected layer.

Aa20475:
Alright! Will try that too! 😁

Aa20475:
Hey! I've made some changes. I just am switching the sensitivity of the widget here. But, please let me know if the functionality is fine. Also, I had to change the `layer_change_handler`, so please verify issue #1603

morevnaproject:
Hi! 
[...]

PR #2074, opened 2021-03-22
[Synfig Studio] Improve Skeleton toolbox layout

FirasH:
Original layout:
skeleton_tool_current

Proposed layout:
skeleton_tool_update

We need to update also other tools to use Gtk::Grid instead of Gtk::Table which is now deprecated.

### DIFF
--- a/synfig-studio/src/gui/states/state_bone.cpp
+++ b/synfig-studio/src/gui/states/state_bone.cpp
@@ -109,7 +109,7 @@ class studio::StateBone_Context : public sigc::trackable
 	synfigapp::Settings& settings;
 
 	// holder of optons
-	Gtk::Grid options_table;
+	Gtk::Grid options_grid;
 
 	// title
 	Gtk::Label title_label;
@@ -121,6 +121,7 @@ class studio::StateBone_Context : public sigc::trackable
 
 	//  bone width
 	Gtk::Label bone_width_label;
+	Gtk::HBox skel_box;
 	Widget_Distance skel_bone_width_dist;
 	Widget_Distance skel_deform_bone_width_dist;
 
@@ -363,7 +364,7 @@ StateBone_Context::StateBone_Context(CanvasView *canvas_view) :
 	/*setting up the tool options menu*/
 
 	// 0, title
-	title_label.set_label(_("Skeleton Creation"));
+	title_label.set_label(_("Skeleton Tool"));
 	Pango::AttrList list;
 	Pango::AttrInt attr = Pango::Attribute::create_attr_weight(Pango::WEIGHT_BOLD);
 	list.insert(attr);
@@ -373,6 +374,9 @@ StateBone_Context::StateBone_Context(CanvasView *canvas_view) :
 	// 1, layer name label and entry
 	id_label.set_label(_("Name:"));
 	id_label.set_alignment(Gtk::ALIGN_START,Gtk::ALIGN_CENTER);
+	id_entry.set_hexpand();
+	id_box.pack_start(id_label, Gtk::PACK_SHRINK);
+	id_box.pack_start(id_entry);
 
 
 	// 2, Bone width
@@ -385,44 +389,44 @@ StateBone_Context::StateBone_Context(CanvasView *canvas_view) :
 	skel_deform_bone_width_dist.set_range(0,10000000);
 	skel_deform_bone_width_dist.set_sensitive(true);
 
+	skel_box.pack_start(skel_bone_width_dist);
+	skel_box.pack_start(skel_deform_bone_width_dist);
+
 	load_settings();
 
 	// 4, Layer choice
-	layer_label.set_label(_("Layer to Create:"));
+	layer_label.set_label(_("Layer"));
 	layer_label.set_alignment(Gtk::ALIGN_START, Gtk::ALIGN_CENTER);
-	layer_label.set_attributes(list);
-	layer_label.set_sensitive(true);
 	
 	create_layer.set_label(_("Create Layer"));
 	create_layer.set_alignment(Gtk::ALIGN_CENTER, Gtk::ALIGN_CENTER);
 	create_layer.set_sensitive(true);
 
 
-	// pack all options to the options_table
+	// pack all options to the options_grid
 
 	// 0, title
-	options_table.attach(title_label,0, 0,1,1);
+	options_grid.attach(title_label, 0, 0, 2, 1);
 	// 1, name
-	options_table.attach(id_label,0, 1,1,1);
-	options_table.attach(id_entry,2, 1,3,1);
+	options_grid.attach(id_box, 0, 1, 2, 1);
 	// 2, default bone width
-	options_table.attach(bone_width_label,0,2,1,1);
-	options_table.attach(skel_bone_width_dist,2, 2,3,1);
-	options_table.attach(skel_deform_bone_width_dist,2, 2,3,1);
+	options_grid.attach(bone_width_label, 0, 2);
+	options_grid.attach(skel_box, 1, 2);
 	// 3, Layer choice
-	options_table.attach(layer_label,0, 4,1,1);
-	options_table.attach(radiobutton_skel,0, 5,2,1);
-	options_table.attach(radiobutton_skel_deform,2, 5,2,1);
-	options_table.attach(create_layer,2,6,2,1);
+	options_grid.attach(layer_label, 0, 3, 2, 1);
+	options_grid.attach(radiobutton_skel, 0, 4, 2, 1);
+	options_grid.attach(radiobutton_skel_deform, 0, 5, 2, 1);
+	options_grid.attach(create_layer, 0, 6, 2, 1);
 
 	create_layer.signal_clicked().connect(sigc::mem_fun(*this,&StateBone_Context::make_layer));
 	radiobutton_skel.signal_toggled().connect(sigc::mem_fun(*this,&StateBone_Context::update_layer));
 
 	// fine-tune options layout
-	options_table.set_border_width(GAP*2); // border width
-	options_table.set_row_spacing(GAP); // row gap
-	options_table.set_margin_bottom(0);
-	options_table.show_all();
+	options_grid.set_hexpand();
+	options_grid.set_border_width(GAP*2); // border width
+	options_grid.set_row_spacing(GAP); // row gap
+	options_grid.set_margin_bottom(0);
+	options_grid.show_all();
 
 	skel_deform_bone_width_dist.hide();
 
@@ -485,7 +489,7 @@ void
 StateBone_Context::refresh_tool_options()
 {
 	App::dialog_tool_options->clear();
-	App::dialog_tool_options->set_widget(options_table);
+	App::dialog_tool_options->set_widget(options_grid);
 	App::dialog_tool_options->set_local_name(_("Skeleton Tool"));
 	App::dialog_tool_options->set_name("bone");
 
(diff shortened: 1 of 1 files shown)


ice0:
The original looks better to me. 
@FirasH maybe just update it to Gtk::Grid without changing?
Also can you check the build errors?

FirasH:
> 
> 
> The original looks better to me.
> @FirasH maybe just update it to Gtk::Grid without changing?
> Also can you check the build errors?

@ice0, my idea was to change the layout so that it would match the layout of the other tools like this one:
cutout-tool

I also think that both the original and the proposal are still **not consistent** with the overall interface logic. I had to play a bit with the current layout to understand what it does because it seems very quickly sorted.


Maybe it's better to apply a more structural change for this toolbox layout.
For example instead of using those radio buttons for layer type selection:
radio-button-skeleton
we could use something like for the `Gradient Tool` with two buttons:
layer-type-gradient
and move the `Create Layer` button to the bottom like it's done for the `Spline Tool`?
spline-tool

What do you think about this idea?

FirasH:
This is getting a bit closer to the idea:
skeleton-layout-2

ice0:
Yes. This looks better.

rodolforg:
> skeleton-layout-2

I suggest a minor padding/margin/distance between "Name:" label and entry box.

FirasH:
Here you can see the idea visually complete, still need to connect the `Layer Type:` button though:
skeleton-tool-final-proposal

Let me know what you think.

FirasH:
Only in this toolbox the settings you choose for one of the Layer Types is hold separately and saved.
If you select e.g. `Skeleton Layer` and change `Bone Width` value, then select `Skeleton Deform Layer` for `Bone Width` another spinbox is used that holds another value.

I can't find a reason for this behavior... but I didn't use this tool that much to grasp the advantages. Anyhow the logic used is not consistent with the other tools and makes it more complicated to manage in the code. Maybe we can simplify it?

ice0:
ping @morevnaproject

FirasH:
@Keyikedalube what do you think about the proposed layout?

Keyikedalube:
Your proposed layout looks consistent with the other tool properties :)

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -73,9 +73,10 @@ using namespace synfigapp;
 	name->set_size_request(px)
 #endif
```
Keyikedalube, line 87: https://developer.gnome.org/gtkmm/3.24/classGtk_1_1Alignment.html#details

> Use Widget alignment and margin properties instead.

Can you update this part also, please?

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -361,29 +362,32 @@ StateBone_Context::StateBone_Context(CanvasView *canvas_view) :
...
+	title_label.set_halign(Gtk::ALIGN_START);
+	title_label.set_valign(Gtk::ALIGN_CENTER);
 
-	// Layer name
 	id_label.set_label(_("Name:"));
-	id_label.set_alignment(Gtk::ALIGN_START,Gtk::ALIGN_CENTER);
-	id_entry.set_hexpand();
+	id_label.set_halign(Gtk::ALIGN_START);
+	id_label.set_valign(Gtk::ALIGN_CENTER);
+	SPACING(id_gap, GAP);
```
Keyikedalube, line 421: https://developer.gnome.org/gtkmm/3.24/classGtk_1_1Alignment.html#details

> Use Widget alignment and margin properties instead.

Can you update this part also, please?

Keyikedalube: The `SPACING` part

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -126,17 +140,13 @@ class studio::StateBone_Context : public sigc::trackable
...
+	// Layer Type
 	Gtk::Label layer_label;
-	Gtk::RadioButton::Group radiogroup;
-	Gtk::RadioButton radiobutton_skel;
-	Gtk::RadioButton radiobutton_skel_deform;
-	Gtk::Button create_layer;
-
+	Gtk::ToggleButton layer_skel_togglebutton;
+	Gtk::ToggleButton layer_skel_deform_togglebutton;
+	Gtk::HBox layer_types_box;
```
Keyikedalube: Do you plan to replace deprecated `HBox` in another PR?

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -66,6 +66,19 @@ using namespace synfigapp;
...
+	{ \
+		Gtk::Image *icon = manage(new Gtk::Image(Gtk::StockID(stockid), \
+			Gtk::ICON_SIZE_SMALL_TOOLBAR)); \
+		button.add(*icon); \
+	} \
+	button.set_relief(Gtk::RELIEF_NONE); \
+	button.set_tooltip_text(tooltip); \
+	button.signal_toggled().connect(sigc::mem_fun(*this, \
+		&studio::StateBone_Context::fun))
+#endif
```
Keyikedalube, line 80: I think this can be templated. For users who debug, macro functions will cause them a lot of headaches.

From [C++ Advantages/Disadvantages of Implementing functions as macros](https://stackoverflow.com/questions/16305844/c-advantages-disadvantages-of-implementing-functions-as-macros)
> The closest rule of thumb you can get is whether something needs to happen or change during compilation. If so, consider a macro, otherwise just code it up. Remember, templates count as code and can do quite a few things you might be tempted to use a macro for, but can be somewhat safer.

There is no performance overhead for this class so it's safe to template your macro function.

Keyikedalube: Or a function since your argument data types are same.

synfig-studio/src/gui/states/state_bone.cpp
FirasH: This should be fixed in the last commit 👍

synfig-studio/src/gui/states/state_bone.cpp
FirasH: This `SPACING` approach is used a lot in the Toolboxes, maybe we should replace it everywhere as a separate issue?

synfig-studio/src/gui/states/state_bone.cpp
FirasH: Even `LAYER_CREATION` as for `SPACING` is used a lot in Toolboxes and as for the latter probably the best way to deal with this separately?

synfig-studio/src/gui/states/state_bone.cpp
Keyikedalube: 👌

synfig-studio/src/gui/states/state_bone.cpp
Keyikedalube: OK 👌

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -1282,3 +1338,47 @@ StateBone_Context::_on_signal_value_desc_set(ValueDesc value_desc,ValueBase valu
...
+		save_settings();
+		c_layer = 0;
+		update_tool_options(0);
+		load_settings();
+	}
+
+	if(get_layer_skel_flag() +
+	   get_layer_skel_deform_flag() == 0)
+		set_layer_skel_flag(true);
+}
```
Keyikedalube, line 1360: I think instead of declaring new variables  `layer_skel_flag` and `layer_skel_deform_flag` and their setters and getters, `toggle_button.set_active()` and `toggle_button.get_active()` can do the trick since we're only dealing with boolean values.

Something like this:
```c++
void
StateBone_Context::toggle_layer_skel()
{
	if(layer_skel_togglebutton.get_active())
	{
		layer_skel_deform_togglebutton.set_active(false);

		save_settings();
		c_layer = 0;
		update_tool_options(0);
		load_settings();
	}

	if(layer_skel_togglebutton.get_active() +
	   layer_skel_deform_togglebutton.get_active() == 0)
		layer_skel_togglebutton.set_active(true);
}
```

FirasH: I don't remember why I choose to keep flags as used currently for other tools, but looking at the code as of now as you suggested I don't think they're necessary... Checking now, will report ASAP.

FirasH: @Keyikedalube I've removed them, works fine with less stuff :wink:

FirasH:
This is how the Skeleton toolbox looks in the latest commit:
latest-skeleton-toolbox

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -287,14 +264,14 @@ StateBone_Context::load_settings()
 		}
 
 		if(settings.get_value("bone.layer_skel",value) && value=="0")
-			set_layer_skel_flag(false);
+			layer_skel_togglebutton.set_active(false);
 		else
-			set_layer_skel_flag(true);
+			layer_skel_togglebutton.set_active(true);
```
Keyikedalube: Instead of if-else why not just convert string `value` to int directly:
```c++
layer_skel_togglebutton.set_active(std::stoi(value));
```

Keyikedalube: That way you call `set_active()` only once ;)

Keyikedalube: Same for deform_skel toggle button too

FirasH: `load_settings()` is used both for initializing and loading of users' settings.
If we are on the very first run of Synfig Studio this value still doesn't exist and I dont't think that will work.

Keyikedalube (dismissed):

FirasH:
@morevnaproject can this be merged before releasing the latest development version?

morevnaproject:
> Only in this toolbox the settings you choose for one of the Layer Types is hold separately and saved.
> If you select e.g. `Skeleton Layer` and change `Bone Width` value, then select `Skeleton Deform Layer` for `Bone Width` another spinbox is used that holds another value.
> 
> I can't find a reason for this behavior...

This behavior is requested by me [here](https://github.com/synfig/synfig/issues/1557#issuecomment-657648536).
The reason for that is following: For ordinary `Skeleton Layer` user generally wants thin points (with small Width  thickness), but for  `Skeleton Deform Layer` this value should be big, because Width plays an important role to define influence area of deformation. ^__^

morevnaproject:
> This is getting a bit closer to the idea:
> skeleton-layout-2

I like this version - with radio buttons for layer type selection. They are easier to understand, because they have text caption.

And I agree about padding after `Name:` label. I would make it even bigger. Ideally it would be nice to make text entry for `Name` to be the same width as `Bone Width` entry below -
112472995-18fbb480-8d6e-11eb-86a7-839a487b262c

FirasH:
@morevnaproject
What about designing clearer icons for `Layer Type`?

Speaking of this, to keep a note, here are some of my thoughts about the toolboxes:
In the toolboxes we can reduce used space by using icons.

For example in the `Draw Tool` toolbox using specific icons in place of binary checkboxes and avoiding very large spinboxes (maybe placing more than one on a single line?) would reduce used screen space and IMHO make it would look better and professional.

large-toolbox

morevnaproject:
> @morevnaproject What about designing clearer icons for `Layer Type`?
> 
> Speaking of this, to keep a note, here are some of my thoughts about the toolboxes: In the toolboxes we can reduce used space by using icons.
> 
> For example in the `Draw Tool` toolbox using specific icons in place of binary checkboxes and avoiding very large spinboxes (maybe placing more than one on a single line?) would reduce used screen space and IMHO make it would look better and professional.
> 
> large-toolbox

That would be nice to design an icon for Skeleton Deformation layer (we do not have special icon for it). ^__^
Still, I think it is better to have text labels in parameters panel of Bone tool, as they are more clear for understanding. ^__^

morevnaproject:
Can someone implement changes requested here? ^__^ - https://github.com/synfig/synfig/pull/2074#issuecomment-905599871

rodolforg:
> That would be nice to design an icon for Skeleton Deformation layer (we do not have special icon for it). ^**^**

@morevnaproject  Maybe a bendy/curved bone? XD

Issue #2014, opened 2021-02-13
[Synfig Studio] Delete event doesn't reset CanvasProperties rend_desc values


Keyikedalube:
**Synfig version & platform**:
Building from source
Fedora 33

**Issue description**:
1. Open Canvas Properties dialog
2. Change width value
3. Close dialog using the close button from the title bar
4. Open Canvas Properties dialog again
5. Width value is not reset

rChen10:
I'd like a go at this issue. Can someone assign it to me?

PR #2075, opened 2021-03-23
Resolve issue #2014 (CanvasProperties did not properly reset on delete)

rChen10:
Refresh was being called by CanvasProperties directly rather than being connected to signal_show() as demonstrated in CanvasOptions.

This is my first pull request, so any feedback is appreciated.


rChen10:
Forgot to branch; going to fix the repo and make a new PR.

PR #2076, opened 2021-03-23
[Synfig Studio] The canvas properties dialog now displays values correctly after being closed with the window close button.

rChen10:
Refresh was being called by CanvasProperties directly rather than being connected to signal_show() as demonstrated in CanvasOptions.

This is my first pull request, so any feedback is appreciated.

### DIFF
--- a/synfig-studio/src/gui/dialogs/canvasproperties.cpp
+++ b/synfig-studio/src/gui/dialogs/canvasproperties.cpp
@@ -128,7 +128,8 @@ CanvasProperties::CanvasProperties(Gtk::Window& parent,etl::handle<synfigapp::Ca
 	ok_button->signal_clicked().connect(sigc::mem_fun(*this, &studio::CanvasProperties::on_ok_pressed));
 
 	get_vbox()->show_all();
-	refresh();
+	//refresh();
+	signal_show().connect(sigc::mem_fun(*this, &studio::CanvasProperties::refresh));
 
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/dialogs/canvasproperties.cpp
```diff
@@ -128,7 +128,8 @@ CanvasProperties::CanvasProperties(Gtk::Window& parent,etl::handle<synfigapp::Ca
 	ok_button->signal_clicked().connect(sigc::mem_fun(*this, &studio::CanvasProperties::on_ok_pressed));
 
 	get_vbox()->show_all();
-	refresh();
+	//refresh();
```
Keyikedalube, line 131: You can remove that commented code.

rChen10: Alright! Thanks for the advice. I was a bit hesitant to delete code, but I suppose that is what git is for. I've updated the PR.

Keyikedalube (approved):

synfig-studio/src/gui/dialogs/canvasproperties.cpp
Keyikedalube: You can resolve this review using the Resolve button :)

ice0 (approved):

ice0:
Merged. Thank you!

PR #2077, opened 2021-03-23
Resolves Issue#2018 (Improve Vectorizer Dialog)

rChen10:
This is what the new dialog looks like:
2021-03-23 19_37_44-_Synfig Animation 1 - Synfig Studio

I added padding to the dialog which was originally missing. Rather than making the dialog a fixed size, I decided to scale the widgets within. To do so, I made the grid's columns non-homogeneous and I set the entry widget's hexpand to true. Please let me know if this will have any unintended side effects.

A better way to do this would be to set hexpand where the widgets are created. I can make this change if this layout is approved.


rChen10:
I'm going to change the UI to use Builder instead.

PR #2079, opened 2021-03-25
[Synfig Studio] Toolbox migration from Gtk::Table to Gtk::Grid

FirasH:
Migrate toolbox layouts from `Gtk::Table` (deprecated) to `Gtk::Grid`.

### DIFF
--- a/synfig-studio/src/gui/states/state_normal.cpp
+++ b/synfig-studio/src/gui/states/state_normal.cpp
@@ -283,7 +283,9 @@ StateNormal_Context::StateNormal_Context(CanvasView* canvas_view):
 	Pango::AttrInt attr = Pango::Attribute::create_attr_weight(Pango::WEIGHT_BOLD);
 	list.insert(attr);
 	title_label.set_attributes(list);
-	title_label.set_alignment(Gtk::ALIGN_START, Gtk::ALIGN_CENTER);
+	title_label.set_hexpand();
+	title_label.set_halign(Gtk::ALIGN_START);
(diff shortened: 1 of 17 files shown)


synfig-studio/src/gui/states/state_normal.cpp
```diff
@@ -46,6 +46,7 @@
 #include <synfig/angle.h>
 
 #include <synfigapp/main.h>
+#include <gtkmm-3.0/gtkmm/widget.h>
```
Keyikedalube, line 49: Do we have to specify gtkmm version on include header now?

We already told the compiler we are using gtkmm-3.0 through build script (Autotools and CMake)
Specifying gtkmm version again on headers kinda look redundant

@ice0 @rodolforg

FirasH: Missed that 😉 
Thanks for checking, I’ll remove it.

synfig-studio/src/gui/states/state_smoothmove.cpp
```diff
@@ -47,6 +47,8 @@
 #include <synfig/general.h>
 
 #include <synfigapp/main.h>
+#include <gtkmm-3.0/gtkmm/widget.h>
```
ice0, line 50: Here too.

FirasH:
`state_lasso.cpp` needs some cleaning... but we'll deal with that in a separate pull request.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2098, opened 2021-04-04
[Synfig Studio] Replace deprecated Gtk::HScale with Gtk::Scale

FirasH:
Replace deprecated `Gtk::HScale` used in toolboxes with `Gtk::Scale`.

Also fine-tuned step increment so that with mouse scroll the step is of 0.1, with arrows 0.01 and with Ctrl+arrows is 0.1 instead of some predefined step increment.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2102, opened 2021-04-06
[Synfig Studio] Sometimes Widget_Vector didn't use locale for decimal separator

rodolforg:
Example:
 In Canvas Properties dialog, we have the Image Area delimited by two points:
the top left corner and the bottom right corner.
Both points use Widget_Vector to edit it.
There, you could only type real numbers by using 'dot' (.) to separate
integer from the fractional part (e.g. 4.5). However, this character
depends on locale. Brazil and Germany, for example, use 'comma' (,).
 Another case: on the same dialog, the Focus Point field.

This commit also fixes synfig::remove_trailing_zeroes() - I accidentally
let two chars to be summed instead of appending them to string

### DIFF
--- a/synfig-core/src/synfig/string_helper.cpp
+++ b/synfig-core/src/synfig/string_helper.cpp
@@ -36,7 +36,7 @@
 #endif
 
 std::string
-synfig::remove_trailing_zeroes(const std::string& text)
+synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point)
 {
 	std::string result(text);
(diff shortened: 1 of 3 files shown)


rodolforg:
I suggest to backport it.
This PR depends on #1995.

ice0:
@rodolforg Could you add to the description how and where it is used?

rodolforg:
Is it better now?

ice0:
> Is it better now?

Yes, thank you!

ice0 (approved):

ice0:
Merged. Thank you!

PR #2104, opened 2021-04-10
MSVC compilation fixes (04-2021)

ice0:
- Fixed incorrect CMake build output path in multi-config builds (e.g. MSVC). 
  old path: `output\bin\Debug`, `output\lib\Debug`, new path: `output\Debug\bin`, `output\Debug\lib`, etc.
- Fixed `mod_ffmpeg` build
- Fixed `mod_png` build
- Fixed `mod_filter` build
- Fixed `mod_imagemagick` build
- Fixed `lyr_freetype` build
- Fixed `mod_geometry` build
- Fixed `mod_gif` build
- Fixed `mod_gradient` build
- Fixed `mod_svg` build


synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -2033,160 +2000,171 @@ Svg_parser::new_guid(){
...
-	COLOR_NAME("yellowgreen",154, 205, 50)
-	return 0;
+Svg_parser::getColor(String name, int position) {
+	if (position < 1 || position > 3) return 0;
+
+	struct RGB {
+		std::uint8_t color[3];
+	};
+
+	static std::unordered_map<std::string, RGB> color_names = {
```
rodolforg, line 2013: Why unordered?

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
```diff
@@ -70,7 +75,6 @@ SYNFIG_TARGET_SET_VERSION(ffmpeg_trgt,"0.1");
 /* === M E T H O D S ======================================================= */
 
 ffmpeg_trgt::ffmpeg_trgt(const char *Filename, const synfig::TargetParam &params):
-	pid(-1),
```
rodolforg, line 73: is it ok to leave it uninitialized for non-MSVC?

synfig-core/src/modules/mod_svg/svg_parser.cpp
ice0: In this case, it's just faster. Benchmark:
**unordered_map**
```
Elapsed time: 383 ms
Elapsed time: 382 ms
Elapsed time: 384 ms
```

**map**
```
Elapsed time: 588 ms
Elapsed time: 586 ms
Elapsed time: 583 ms
```

**benchmark.cpp (compiled with -O3)**
```
    const std::string colors[] = {"yellow", "springgreen", "white", "tomato", "mediumspringgreen"};
    int colors_count = sizeof(colors) / sizeof(colors[0]);

    std::srand(std::time(nullptr)); // use current time as seed for random generator

    milliseconds ms = duration_cast< milliseconds >(system_clock::now().time_since_epoch());

    long result = 0;

    for (int i=0; i < 1000*1000*10; ++i) {
      result = getColor(colors[rand() % colors_count], 1);
    }

    milliseconds end = duration_cast< milliseconds >(system_clock::now().time_since_epoch());

    std::cout << "Elapsed time: " << (end - ms).count() << " ms" << std::endl;
```

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
ice0: Yes. MSVC does not have `fork()` function, so this field is not used. I disabled it in the header file:

**trgt_ffmpeg.h**
```
#ifdef HAVE_FORK
	pid_t pid = -1;
#endif
```

ice0: I moved the initialization into a header file: 
```
#ifdef HAVE_FORK
	pid_t pid = -1;
#endif
```

ice0:
@rodolforg please approve PR if all issues are resolved

synfig-core/src/modules/mod_png/trgt_png_spritesheet.cpp
```diff
@@ -479,6 +478,7 @@ png_trgt_spritesheet::write_png_file()
     png_write_info_before_PLTE(png_ptr, info_ptr);
     png_write_info(png_ptr, info_ptr);
 
+	unsigned char* buffer = new unsigned char[4 * sheet_width];
```
rodolforg, line 481: Maybe unique_ptr?
It would avoid problems with `delete[] buffer`.

synfig-core/src/modules/mod_svg/svg_parser.cpp
rodolforg: Oh, I wondered sorted `map` would be faster to lookup :O (slower on inserting data though)

synfig-core/src/modules/mod_png/trgt_png_spritesheet.cpp
ice0: Done.

rodolforg (approved):
I can't anything about CMakeLists.txt files, but the rest seems ok.

PR #1802, opened 2020-10-24
Fix crash in Spline Vertex converter (https://github.com/synfig/synfig/issues/1731)

Svarov-RZM:

PR #2118, opened 2021-04-20
ChangeLog for version 1.4.1

morevnaproject:
Please consider the following rule when writing ChangeLog:
- If changelog entry is affecting **ETL**, then it is necessary to append " [ETL]" at the end of the line
- If changelog entry is affecting **synfig-core**, then it is necessary to append " [core]" at the end of the line
- If changelog entry is affecting **synfig-studio**, then it is necessary to append " [studio]" at the end of the line

This is because we have separate ChangeLog entries in `ETL/NEWS`, `synfig-core/NEWS`, `synfig-studio/NEWS`.

The existence of those separate ChangeLogs is present for historical reasons - when Synfig is built with Autotools the build system generate separate source tarballs for ETL, core and studio. Obviously, each source package should have its own changelog. 

Thus, previously we used to write 3 separate changelogs, which was a pain.

This is why starting from 1.4.1 we are introducing single changelog in the root of Synfig's source. We will write it according to rules mentioned in the beginning (adding "[ETL]" and/or "[core]" and/or "[studio]" tags at the end of each line) and then will use [ChangeLog-split.sh script](https://github.com/synfig/synfig/blob/73f4540df5291c20a6981f81e30347fdb621c005/ChangeLog-split.sh) to separate this single changelog into `ETL/NEWS`, `synfig-core/NEWS` and `synfig-studio/NEWS`.

I understand this is a bit cumbersome, but it's a burden of legacy.

When we will migrate to CMake and remove Autotools support, then we can stay with a single ChangeLog.


morevnaproject:
Also, can someone help me with writing user-friendly Release Notes from this changelog? ^__^ - https://github.com/synfig/synfig-docs/issues/25

rodolforg:
I saw there are some differences Changelog x Branch commits:
- It's missing the #1182 in Changelog
- In changelog, the PR #1802 is labeled as "19112f1 Fix crash in Spline Vertex converter (https://github.com/synfig/synfig/issues/1731) [core]"

morevnaproject:
@rodolforg Good catch! I added them now. Thank you! ^__^

rodolforg (approved):

rodolforg:
> s - when Synfig is built with Autotools the build system generate separate source tarballs for ETL, core and studio.

@morevnaproject  actually, Autotools could be generate a single source tarball and build everything at once ;)

morevnaproject:
Yes, it can, but that will require to do big rewrite of Autotools scripts (i.e. merge all configure.ac files into one, etc.).
I think it's better to wait for the moment when CMake will be ready and drop Autotools. ^__^

PR #2126, opened 2021-04-26
[Synfig Studio] Replacing Gtk::HBox with Gtk::Box

FirasH:
Replaced the following deprecated stuff:
- `Gtk::HBox`

and some comments cleaning as well.

### DIFF
--- a/synfig-studio/src/gui/docks/dock_info.cpp
+++ b/synfig-studio/src/gui/docks/dock_info.cpp
@@ -178,7 +178,7 @@ studio::Dock_Info::Dock_Info()
 	stop_button.set_valign(Gtk::ALIGN_CENTER);
 	stop_button.signal_clicked().connect(sigc::mem_fun(*this, &studio::Dock_Info::on_stop_button_clicked));
 
-	render_grid->attach(*overlay, 0, 0);
+	render_grid->attach(*overlay, 0, 0, 1, 1);
 	render_grid->attach_next_to(stop_button, Gtk::POS_RIGHT, 1, 1);
 	grid->attach_next_to(*render_progress_label, *separator2, Gtk::POS_BOTTOM, 8, 1);
 	grid->attach_next_to(*render_grid, *render_progress_label, Gtk::POS_BOTTOM, 7, 1);
--- a/synfig-studio/src/gui/docks/dock_info.cpp
+++ b/synfig-studio/src/gui/docks/dock_info.cpp
@@ -155,7 +155,7 @@ studio::Dock_Info::Dock_Info()
 	grid->attach(*separator2, 0, 4, 8, 1);
 
 	// Render Progress Bar
-	Gtk::HBox *render_box = manage(new Gtk::HBox());
+	Gtk::Grid *render_grid = manage(new Gtk::Grid());
 	Gtk::Label *render_progress_label = manage(new Gtk::Label());
 	Gtk::Overlay *overlay = manage(new Gtk::Overlay());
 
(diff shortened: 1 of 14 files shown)


rodolforg:
Hm… I don't think it always good to replace the `HBox` with `Grid`. `HBox` is deprecated, but `Box` is not, as they're still available in Gtk 4 / gtkmm4 :
https://developer.gnome.org/gtk4/stable/GtkBox.html
https://developer.gnome.org/gtkmm-tutorial/stable/sec-multi-item-containers.html.en#sec-boxes

`Grid` doesn't works the same as `Box` when programmer codes `Box` with `pack_start()` and `pack_end()`. AFAIK, `Grid` can't reproduce the `Box` behavior when both methods are used.

FirasH:
In the current code `pack_start` and `pack_end` are used in case of checkbuttons to place them on the right edge of the `HBox`. To achieve the same visual result I had to expand the label on the left and the UI cannot be distinguished from the previous version.
So yes, you’re right the way `Box` and `Grid` work is not the same but I don’t think any user will suffer this change because, as I noted before, unless you read the code you can’t notice the difference.

I think that for this specific case there isn’t any practical difference as the layout is very simple but at least we avoid using stuff that is likely to be removed anyway in the near future or at least this is what’s suggested in Gtk documentation. I know we need to reference an older version of Gtk but if some changes can be done future proofing the code it’s even better IMHO, less work for the same piece of code.

rodolforg:
I read somewhere else that they would be deprecated, but they were not. And I hope not: Grid is over complicated for simple single rows or columns layout, and don't provide that feature widgets on start and on end in the same way.
So… I don't see why don't simply replace H/VBox with Box only.

Anyway, not really important if you changed while keeping the same layout/resize behavior.

FirasH:
mmm, let's see what other developers suggest...
For me both are equallly viable solutions 👍

Keyikedalube:
> likely to be removed anyway in the near future or at least this is what’s suggested in Gtk documentation

I remember last year making a similar PR with `Gtk::Grid` as a replacement for `Gtk::H/VBox`. Because I thought the docs meant it too

> I read somewhere else that they would be deprecated, but they were not.

I've been digging where I read that too, couldn't find it, but found this post instead:
[Is GtkBox going to be deprecated?](https://discourse.gnome.org/t/is-gtkbox-going-to-be-deprecated/2332)

I agree with rodolforg's suggestion for `Gtk::Box` as a container for single row/column and `Gtk::Grid` for table-like container. But this PR is fine too. I can understand reading the code placing the first widget using `attach()` and the remaining widgets using `attach_next_to()`.

Not related to this PR but in Gtk 4 `Gtk::Toolbar` has been removed and Gtk community also suggest using `Gtk::Box` as a replacement. The user was trying to implement single row toolbar:
https://discourse.gnome.org/t/hope-gtktoolbar-back-to-gtk4/6201

FirasH:
To me `Gtk::Box` feels more appropriate and less complicated for this specific purpose as @rodolforg suggested and chose `Gtk::Grid` just for the chances of having `Gtk::Box` replaced soon but doesn't appear to be the case.


Probably the best solution for now is using `Gtk::Box` and keep this decision consistent across Synfig Studio code.

Keyikedalube:
Would you squash this? Git diff shows `Gtk::Box` replacing `Gtk::Grid` because of your initial commit. If you squash it the diff between old and new (your PR)  would appear correct.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2128, opened 2021-04-27
[CMake] Bump minimum required CMake version to 3.5

ice0:
Actually we need version 3.1 at the moment

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.5)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
 project(synfig-studio)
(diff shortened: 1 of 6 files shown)


Keyikedalube:
I'm curious why synfig has `cmake_minimum_required()` set both in the root and subdirectory? Shouldn't the root one be enough?
```
(root) synfig-core/CMakeLists.txt
    |___ synfig-core/src/CMakeLists.txt

(root) synfig-studio/CMakeLists.txt
    |___ synfig-studio/src/CMakeLists.txt
```

ice0:
> I'm curious why synfig has `cmake_minimum_required()` set both in the root and subdirectory? Shouldn't the root one be enough?

Agree. Removed extra lines.

ice0:
@Keyikedalube @FirasH @rodolforg 

Guys, please approve the PRs, which you think is ok to be merged. This will help a lot.

Keyikedalube (approved):

FirasH:
I honestly have no idea about how `cmake` works and how should be set up.
Anyway looking at the code seems fine to me and better than before as we're referencing the same `cmake` version everywhere in the code.

FirasH (approved):

Keyikedalube:
@FirasH I think ice0 meant not only this but the other PRs also. More eyes scanning the code would greatly help project maintainers :)

rodolforg (approved):

PR #2133, opened 2021-04-30
[Synfig Core] Set Random and Derivative valuenodes enum parameters to static by default

### LINKED ISSUE #2131: [Synfig Studio] Interpolation should be static by default for Random converted values
master e6bbcab

**Issue description**:
1) Create a new animation
2) Add a `Layer` (e.g. a `Star`)
3) Select the `Layer` and go to the `Parameters` window
4) Right-click on one of the Values and `Convert > Random`

The `Interpolation` Value should be set as Static by default.

FirasH:
Fix #2131

### DIFF
--- a/synfig-core/src/modules/mod_noise/valuenode_random.cpp
+++ b/synfig-core/src/modules/mod_noise/valuenode_random.cpp
@@ -65,10 +65,14 @@ ValueNode_Random::ValueNode_Random(const ValueBase &value):
 	init_children_vocab();
 	random.set_seed(time(nullptr));
 
+	ValueNode_Const::Handle val_smooth;
+	val_smooth = ValueNode_Const::Handle::cast_static(ValueNode_Const::create(int(RandomNoise::SMOOTH_CUBIC)));
+	val_smooth->set_static(true);
+
 	set_link("radius",ValueNode_Const::create(Real(1)));
 	set_link("seed",ValueNode_Const::create(random.get_seed()));
 	set_link("speed",ValueNode_Const::create(Real(1)));
-	set_link("smooth",ValueNode_Const::create(int(RandomNoise::SMOOTH_CUBIC)));
+	set_link("smooth",val_smooth);
 	set_link("loop",ValueNode_Const::create(Real(0)));
 
 	Type &type(get_type());
@@ -269,7 +273,6 @@ ValueNode_Random::get_children_vocab_vfunc()const
 		.add_enum_value(RandomNoise::SMOOTH_SPLINE,"spline",_("Spline"))
 		.add_enum_value(RandomNoise::SMOOTH_CUBIC,"cubic",_("Cubic"))
 	);
-
 
 	ret.push_back(ParamDesc(ValueBase(),"loop")
 		.set_local_name(_("Loop Time"))
--- a/synfig-core/src/synfig/valuenodes/valuenode_derivative.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_derivative.cpp
@@ -149,9 +149,18 @@ ValueNode_Derivative::ValueNode_Derivative(const ValueBase &value):
 	LinkableValueNode(value.get_type())
 {
 	init_children_vocab();
+
+	ValueNode_Const::Handle val_acc;
+	val_acc = ValueNode_Const::Handle::cast_static(ValueNode_Const::create((int)(NORMAL)));
+	val_acc->set_static(true);
+
+	ValueNode_Const::Handle val_order;
+	val_order = ValueNode_Const::Handle::cast_static(ValueNode_Const::create((int)(FIRST)));
+	val_order->set_static(true);
+
 	set_link("interval",      ValueNode_Const::create(Real(0.01))); // Default interval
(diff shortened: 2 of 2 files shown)


synfig-core/src/modules/mod_noise/valuenode_random.cpp
```diff
@@ -66,10 +66,14 @@ ValueNode_Random::ValueNode_Random(const ValueBase &value):
 	set_children_vocab(ret);
 	random.set_seed(time(NULL));
 
+	ValueNode_Const::Handle val_smooth;
+	val_smooth = ValueNode_Const::Handle::cast_dynamic(ValueNode_Const::create(int(RandomNoise::SMOOTH_CUBIC)));
```
rodolforg, line 70: No need to (slow) dynamic cast here: We know it IS a `ValueNode_Const`, because it was created by `ValueNode_Const::create()`.

If a type cast is needed, use `static_cast` .
Same for other cases in this PR.

rodolforg (changes requested):

synfig-core/src/modules/mod_noise/valuenode_random.cpp
FirasH: Right :+1:

synfig-core/src/modules/mod_noise/valuenode_random.cpp
```diff
@@ -66,10 +66,14 @@ ValueNode_Random::ValueNode_Random(const ValueBase &value):
...
 	random.set_seed(time(NULL));
 
+	ValueNode_Const::Handle val_smooth;
+	val_smooth = ValueNode_Const::Handle::cast_static(ValueNode_Const::create(int(RandomNoise::SMOOTH_CUBIC)));
+	val_smooth->set_static(true);
+
 	set_link("radius",ValueNode_Const::create(Real(1)));
 	set_link("seed",ValueNode_Const::create(random.get_seed()));
 	set_link("speed",ValueNode_Const::create(Real(1)));
-	set_link("smooth",ValueNode_Const::create(int(RandomNoise::SMOOTH_CUBIC)));
```
ice0, line 72: @rodolforg @FirasH I think better to add optional argument to `ValueNode_Const::create`. Something like:
```
ValueNode_Const() = default;
ValueNode_Const(const ValueBase &x, etl::loose_handle<Canvas> canvas = 0, bool is_static = false);

static ValueNode* create(const ValueBase &x=ValueBase(), etl::loose_handle<Canvas> canvas = 0, bool is_static = false);
```

usage:
```
set_link("smooth",ValueNode_Const::create(int(RandomNoise::SMOOTH_CUBIC), nullptr, true));
```

FirasH: @ice0 how would you set `ValueBase` to static if it's `const`?
Not sure if I'm missing something...

ice0: https://github.com/synfig/synfig/compare/master...ice0:issue-2131?expand=1

rodolforg: >  I think better to add optional argument to `ValueNode_Const::create`.

Hm… I don't know. On other hand the `nullptr, true` as arguments are less 'readable'.

ice0: > > I think better to add optional argument to `ValueNode_Const::create`.
> 
> Hm… I don't know. On other hand the `nullptr, true` as arguments are less 'readable'.

You can always write something like:
```
ValueNode_Const::create((int)(NORMAL), nullptr, true)); // value, canvas, is_static
```

or 
```
constexpr auto canvas = nullptr;
constexpr auto is_static = true;
ValueNode_Const::create((int)(NORMAL), canvas, is_static));
```

or 
```
ValueNode_Const::create_static((int)(NORMAL));
```

As for me, the real problem is that the method does not return what it should return. It can return `ValueNode_BoneWeightPair` or `ValueNode_Bone` or `ValueNode_Composite` or `ValueNode_Const`.

FirasH: Not sure what's the best way to deal with this...
@ice0 and @rodolforg you're much more experienced than me, I'll let you pick the best solution.

ice0: I think we can merge this change as is, and back to improvement later. @rodolforg your opinion?

rodolforg: > constexpr auto canvas = nullptr;
> constexpr auto is_static = true;
> ValueNode_Const::create((int)(NORMAL), canvas, is_static));

I prefer this versionm if you're using any changes.



> I think we can merge this change as is, and back to improvement later

Ok

rodolforg (commented):
Sorry I didn't see I was blocking this PR :(

rodolforg (approved):

rodolforg:
@FirasH  could you please solve the conflict?

rodolforg:
@ice0  is this `feat`, `fix` or `refactor`?

PR #2135, opened 2021-05-01
[CMake] Fixed ETL tests

ice0:
Now tests can be run with `make test` command.

Tests cannot be built at this time due to the lack of GLib dependency.
This will be fixed later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,10 @@ endif(MSVC)
 
 message("-- CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
 
+if (${ENABLE_TESTS})
+	enable_testing()
+endif()
+
 include(SynfigFindGit)
 
 add_subdirectory(ETL)
--- a/ETL/CMakeLists.txt
+++ b/ETL/CMakeLists.txt
@@ -4,17 +4,6 @@ project(ETL)
 
 option(ENABLE_TESTS "Enable tests" OFF)
 
-find_package(Threads)
-if (CMAKE_USE_PTHREADS_INIT)
-    set(HAVE_LIBPTHREAD ON)
-endif()
-
-## for tests
-include_directories(${CMAKE_SOURCE_DIR})
-
-## for test to find generated etl_profile.h
-include_directories(${CMAKE_CURRENT_BINARY_DIR})
-
 add_subdirectory(ETL)
 
 if (${ENABLE_TESTS})
--- a/ETL/test/CMakeLists.txt
+++ b/ETL/test/CMakeLists.txt
@@ -1,35 +1,42 @@
 # for test to find generated etl_profile.h
 include_directories(${PROJECT_BINARY_DIR})
 
+# for ETL headers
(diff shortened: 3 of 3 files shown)


ice0:
`find_package(Threads)` removed because we are using C++11 threads.

CMakeLists.txt
```diff
@@ -64,6 +64,10 @@ endif(MSVC)
 
 message("-- CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
 
+if (${ENABLE_TESTS})
```
rodolforg, line 67: As I don't understand CMake, maybe I'm making a stupid question. Anyway: Can this variable be called before the `option(ENABLE_TESTS …)` be declared?

I only saw the option in CMakeLists.txt of the subdirs ETL, synfig-core and synfig-studio.

ice0: > Can this variable be called before the option(ENABLE_TESTS …) be declared?

Yes. You can set CMake variable from outside using command-line `-D` option. For example:
```
cmake -DENABLE_TESTS=1
```

`option` is used to add an option to the CMake GUI. 
This part also needs to be improved, but I want to back to it later. 

For example: do we need to add separate `tests` options for `core` and `studio`? Same for `docs` target, etc.


cmake-gui

rodolforg: What I most miss in the CMake is features like `configure --help` and `configure --with-x` lol
I hope we get it some day.

rodolforg: Is it easy to track all these variables for a later helper?

ice0: You can use `cmake -L ../synfig` to see available variables:
```
Boost_INCLUDE_DIR:PATH=/usr/include
Boost_SYSTEM_LIBRARY_RELEASE:STRING=/usr/lib/x86_64-linux-gnu/libboost_system.so.1.71.0
CMAKE_BUILD_TYPE:STRING=Debug
CMAKE_INSTALL_PREFIX:PATH=/usr/local
DATA_PREFIX:PATH=/usr/local
ENABLE_TESTS:BOOL=0
GETTEXT_MSGFMT_EXECUTABLE:FILEPATH=/usr/bin/msgfmt
GETTEXT_MSGMERGE_EXECUTABLE:FILEPATH=/usr/bin/msgmerge
IMG_EXT:STRING=png
ImageMagick_EXECUTABLE_DIR:PATH=/usr/bin
ImageMagick_Magick++_ARCH_INCLUDE_DIR:PATH=/usr/include/x86_64-linux-gnu/ImageMagick-6
ImageMagick_Magick++_INCLUDE_DIR:PATH=/usr/include/ImageMagick-6
ImageMagick_Magick++_LIBRARY:FILEPATH=/usr/lib/x86_64-linux-gnu/libMagick++-6.Q16.so
MSGFMT_EXECUTABLE:FILEPATH=/usr/bin/msgfmt
WITHOUT_MAGICPP:BOOL=OFF
WITH_MAGICPP:BOOL=OFF
```

rodolforg (approved):

PR #2095, opened 2021-04-03
[Synfig Studio] new color editor

rodolforg:

PR #2144, opened 2021-05-13
[Studio] Set the minimum height of the toolbox

ice0:
So that the color widget cannot be hidden by mistake or after undocking the toolbox.

**Before**
Toolbox before
Toolbox before

**After**
Toolbox after

### DIFF
--- a/synfig-studio/src/gui/docks/dock_toolbox.cpp
+++ b/synfig-studio/src/gui/docks/dock_toolbox.cpp
@@ -90,10 +90,11 @@ Dock_Toolbox::Dock_Toolbox():
 	scrolled_window->show();
 
 	Widget_Defaults* widget_defaults(manage(new Widget_Defaults()));
+	widget_defaults->set_size_request(-1, 100);
 
 	tool_box_paned = manage(new Gtk::Paned(Gtk::ORIENTATION_VERTICAL));
-	tool_box_paned->pack1(*scrolled_window, Gtk::PACK_EXPAND_WIDGET|Gtk::PACK_SHRINK, 3);
-	tool_box_paned->pack2(*widget_defaults, Gtk::PACK_EXPAND_WIDGET|Gtk::PACK_SHRINK, 3);
+	tool_box_paned->pack1(*scrolled_window, Gtk::PACK_EXPAND_WIDGET|Gtk::PACK_SHRINK, false);
+	tool_box_paned->pack2(*widget_defaults, Gtk::PACK_EXPAND_WIDGET|Gtk::PACK_SHRINK, false);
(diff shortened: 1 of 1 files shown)


rodolforg:
Nice.

But this will change the other standard workarea layouts (compositing and animating).
I don't know if it's a problem though

ice0:
> But this will change the other standard workarea layouts (compositing and animating).
> I don't know if it's a problem though

Can you show how it change them?

Keyikedalube:
- **Compositing layout**
We cannot resize the properties dock further up.
Screenshot_2021-05-13_12-37-12
We could dock toolbox and toolbox properties in the same row instead of isolating them above and below
- **Animating layout**
Screenshot_2021-05-13_12-36-49
Hide the toolbox? Since the primary focus for this layout is the timetrack panel.

ice0:
> We could dock toolbox and toolbox properties in the same row instead of isolating them above and below

> Hide the toolbox? Since the primary focus for this layout is the timetrack panel.

Agree with both. In any case, we can try this approach and, if difficulties arise, switch to another solution.

FirasH:
I suggest to split the tools from the color selector/editor, maybe not immediately, but implement it somehow with the new implementation #2095 if it's dockable?

rodolforg:
@ice0 what about a larger Panel handle and a smaller minimum height for color toolbox?

> But this will change the other standard workarea layouts (compositing and animating).
> I don't know if it's a problem though

@morevnaproject  any thoughts?

ice0:
> @ice0 what about a larger Panel handle and a smaller minimum height for color toolbox?

Is it possible to increase handle size? Just checked it and I don't see how this can be achieved.

morevnaproject:
> Can you show how it change them?

You can change them through Window -> Workspace menu.

> @morevnaproject any thoughts?

I agree with changes proposed in https://github.com/synfig/synfig/pull/2144#issuecomment-840369710 ^__^

rodolforg:
> Is it possible to increase handle size?

`void Gtk::Paned::set_wide_handle()`

How wide it is depends on theme, I guess.

ice0:
> > Is it possible to increase handle size?
> 
> [void Gtk::Paned::set_wide_handle()](https://developer.gnome.org/gtkmm/3.18/classGtk_1_1Paned.html#ae910c0cafd04f347d2f371994607182f)
> 
> How wide it is depends on theme, I guess.

Screenshot_155a

Right is wide :)
I did not see much difference.

rodolforg:
Here on my system: :)
Captura de tela de 2021-06-03 03-53-59 Captura de tela de 2021-06-03 03-53-53

PR #2147, opened 2021-05-14
[Studio] Fixed issue with incorrect window size setting when undocking/docking a window

ice0:
Also it now correctly sets the window title for a undocked window.

**Title before**
Title before

**Title after**
Title after

**Before**
https://user-images.githubusercontent.com/5604544/118217832-72e04700-b4a0-11eb-9c77-e1b9508d8121.mp4

**After**

https://user-images.githubusercontent.com/5604544/118217859-82f82680-b4a0-11eb-8a27-090eb1c5c57e.mp4

### DIFF
--- a/synfig-studio/src/gui/docks/dockable.cpp
+++ b/synfig-studio/src/gui/docks/dockable.cpp
@@ -135,6 +135,7 @@ Dockable::on_drag_data_get(const Glib::RefPtr<Gdk::DragContext>&, Gtk::Selection
 {
 	Dockable* tmp(this);
 	dnd_success_ = true;
+	saved_widget_size_ = {get_width(), get_height()};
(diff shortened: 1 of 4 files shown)


rodolforg:
I tested it here, and I saw a (minor) issue:

1. I undock Parameters Panel: Great, the floating window now has a proper type "Parameters"
2. I undock Keyframes Panel and dock it on that same floating window: Whoops - the window title keeps the same name "Parameter".

Do you think this can be fixed?

ice0:
> I tested it here, and I saw a (minor) issue:
> 
> 1. I undock Parameters Panel: Great, the floating window now has a proper type "Parameters"
> 2. I undock Keyframes Panel and dock it on that same floating window: Whoops - the window title keeps the same name "Parameter".
> 
> Do you think this can be fixed?

Sure. I will take a look.

PR #2153, opened 2021-05-16
Update Changelog

morevnaproject:


ice0:
Some thoughts about `ChangeLog-development.md`.
It currently looks like `git log`, so I'm not sure if anyone will read it. So what is the purpose of doing it? 

This is what a good changelog should look like in my opinion: 
https://github.com/angular/angular.js/blob/master/CHANGELOG.md

I think we need to fill the gap where the changes committed directly, without PR, and use the PR header in the changelog. At least it will be something readable.

morevnaproject:
Hi! As I have explained earlier, the purpose of this changelog is to keep track which PRs were backported to stable branch and which are not. We need it as long as we deal with segment of commits in git history where changes were committed directly to master branch (without PRs).

Also I use that file to generate Weekly Reports. ^__^

ice0 (approved):

PR #2157, opened 2021-05-17
Fixed 'locale::facet::_S_create_c_locale name not valid' error when compiling with MinGW compiler

ice0:
This error appears when starting Synfig:

Screenshot_146

MinGW C++ std::locale accepts "C" and "POSIX" and
it does not support other locales.

### DIFF
--- a/synfig-core/src/synfig/string_helper.cpp
+++ b/synfig-core/src/synfig/string_helper.cpp
@@ -33,17 +33,26 @@
 
 #include <algorithm>
 #include <locale>
+#include "general.h"
 #endif
 
 std::string
 synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point)
 {
 	std::string result(text);
+// TODO(ice0): move all the locale related code to the initialization part
+// MinGW C++ std::locale accepts "C" and "POSIX" it does not support other locales.
+	ChangeLocale changeLocale(LC_NUMERIC, "");
+#ifdef __MINGW32__
+	struct lconv *locale_info = localeconv();
+	const char* decimal_point = locale_info->decimal_point;
+#else
 	std::locale l(setlocale(LC_NUMERIC, nullptr));
 	const char decimal_point = std::use_facet< std::numpunct<char> >(l).decimal_point();
+#endif
 	const size_t decimal_point_pos = text.find(decimal_point);
 
-	if (decimal_point_pos == text.npos) {
+	if (decimal_point_pos == std::string::npos) {
 		if (force_decimal_point)
 			result += decimal_point + std::string("0");
 	} else if (decimal_point_pos == text.length()-1) {


synfig-core/src/synfig/string_helper.cpp
```diff
@@ -39,8 +39,16 @@ std::string
 synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point)
 {
 	std::string result(text);
+// TODO(ice0): move all the locale related code to the initialization part
+// MinGW C++ std::locale accepts "C" and "POSIX" it does not support other locales.
+#ifdef __MINGW32__
+	struct lconv *locale_info = localeconv();
+	const char decimal_point = *locale_info->decimal_point;
```
rodolforg, line 46: I think it's better to let it as `char *` as defined in locale_info. `string::find()` works with this parameter type too.

synfig-core/src/synfig/string_helper.cpp
```diff
@@ -39,8 +39,16 @@ std::string
 synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point)
 {
 	std::string result(text);
+// TODO(ice0): move all the locale related code to the initialization part
+// MinGW C++ std::locale accepts "C" and "POSIX" it does not support other locales.
+#ifdef __MINGW32__
+	struct lconv *locale_info = localeconv();
```
rodolforg, line 45: Maybe we should work with `ChangeLocale` guard here? To ensure "system" locale instead of "C" locale? We don't know what caller did before calling `synfig::remove_trailing_zeroes()`

synfig-core/src/synfig/string_helper.cpp
ice0: Agree.

synfig-core/src/synfig/string_helper.cpp
ice0: I don't think this is a good idea to change the locale on every function call. But as a temporary solution, I think it's okay.

rodolforg: ice0, I agree it isn't good, but it is only necessary for `__MINGW32__` case.
The "else" case doesn't need it: read line 50 below.

So, please move the `ChangeLocale` for `#if __MINGW32__` case

rodolforg (changes requested):

synfig-core/src/synfig/string_helper.cpp
ice0: `setlocale(LC_NUMERIC, nullptr)` 
> If locale is NULL, the current locale is only queried, not modified.

Can you explain your idea why `ChangeLocale` is not required for the `#else` branch?

rodolforg: Oh well. Forget about ChangeLocale lol I'm confused.

rodolforg (approved):
Ok. I  finally tested it. It's ok!

Should we keep ChangeLocale?

ice0:
> Ok. I finally tested it. It's ok!
> 
> Should we keep ChangeLocale?

I think yes. This is more correct way. Other improvements will be made later.

PR #2163, opened 2021-05-19
[ETL/ETL] `boxblur` class moved from ETL to Core

DhairyaBahl:
@ice0 @ankit-kumar-dwivedi @AYESDIE 

### Overall Changes I've Made

Boxblur is moved from ETL/ETL to Synfig-Core. Kindly tell me if I have to make any changes in this PR.

### Changes I've made for the ETL/Test

 I haven't moved the tests yet ! Although I've changed the ETL/Test/pen.cpp for the Boxblur with respect to its new location. Either I can move the Tests one by one or I can do it in the end for all of them ! Kindly confirm this once and I will make changes accordingly.


rodolforg (commented):
Why are you renaming it to .cpp?

DhairyaBahl:
> Why are you renaming it to .cpp?

I thought that every file over there is in .cpp format. So, I thought that I have to do that ! but if I don't have to change the extension, then things will also become bit easier.

rodolforg:
Hm… As far as I can see, you're not much familiarized with C++ programming :)

Please take a look at these explanations about header x implementation files for C/C++:
https://en.wikipedia.org/wiki/Class_implementation_file
https://en.wikipedia.org/wiki/Include_directive

http://www.cppforschool.com/tutorial/separate-header-and-implementation-files.html

https://www.learncpp.com/cpp-tutorial/class-code-and-header-files/

PR #2165, opened 2021-05-25
Fixed issue with saving `loop` flag in some dynamic lists (Such as `average` and `weighted_average`)

rodolforg:
Such as 'average' and 'weighted_average'.

Every Dynamic List has loop parameter.


rodolforg:
Maybe backport?

rodolforg:
One difference here: it doesn't write `loop="false"` as current code does for some dynamic lists like `bline`. I chose to do it because `loop` is false by default.

ice0:
> Maybe backport?

Agree.

ice0 (approved):

ice0:
Merged. Thank you!

Issue #600, opened 2018-08-03
Make onionskin work in the same way as in OpenToonz


morevnaproject:
Currently the onionskin is attached to keyframes. This is not comfortable and we need an other way to set onionskin frames.

I am proposing to copy approach from timeline of OpenToonz:
![5](https://user-images.githubusercontent.com/332868/43631181-aa663ab6-972c-11e8-80fd-441d51cbb513.png)

1. Time cursor.
2. Red and green dots are onionskin marks "attached" to time cursor.
3. Blue marks (a bit above of green and red ones) are not "attached" to time cursor. THey do not move when time cursor is moving.

This also implies rework the look of timecursor - rendering it as big yellow-green circle will  make it easier to grab with mouse.

morevnaproject:
Note: Fix for #377 is required first.

zozorg:
0.25 ETH bounty posted ; ~$100 at current price

PR #2166, opened 2021-05-27
[Synfig Studio] Add Onion Skin ability to choose between Keyframes and Frames

### LINKED ISSUE #1562: Synfig's Onion Skin Works Differently
1.3.15 on Win10 (64bit)

**Issue description**:
Instead of showing onion skins for **Only Keyframes**, Synfig's onion skin setting shows onion skins of **Every Frame**,

I'm not sure if this is a bug (i'm pretty sure it is), but i'll post it anyway...

Here's an example on version stable 1.2.2, with onion skin turned on and showing just 1 past keyframe, and the time slider was positioned on 2nd keyframe:

and here's the same onion skin setting but on synfig dev version 1.3.15. Notice how it **shows onion skin for Previous Frame instead of Previous Keyframe**

Found first by tawatoons recently over in this synfig forum thread :
https://forums.synfig.org/t/broken-onion-skin-in-synfig-1-3-14/11192
(they also included the test file for this issue by the way)
I'm just sending the ticket for this issue, that's all ._.v

But i've confirmed that this also happens on my part as well

### LINKED ISSUE #1817: Feature Request: Toggle between frame and keyframe onion skin

Development version
Archlinux 64

**Description**:

I think it is very important to have the frames and also the keyframes as a reference in the work area to create movements because this helps us to achieve greater precision. I suggest an icon or icons where we can enable onion skin for frames and keyframes. Or if you want to have both activated so that they can be seen simultaneously.

Attached is a recreation of how it could be.

Thank you very much in advance for your attention.

Thanks for the excellent work.

1Osk
2OSk
3OSk

FirasH:
This pull request is aimed at introducing the ability for users to use `Onion Skin` with:
- Keyframes (as with previous versions of Synfig)
- Frames

Related issues: #1562, #1817

(Linking to related issues)
Fixes #1562
Fixes #1817

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -1256,6 +1256,7 @@ CanvasView::create_top_toolbar()
 
 	{ // Past onion skin spin button
 		past_onion_spin=Gtk::manage(new class Gtk::SpinButton(past_onion_adjustment_));
+		past_onion_spin->set_value(work_area->get_onion_skins()[0]);
 		past_onion_spin->signal_value_changed().connect(
 			sigc::mem_fun(*this, &CanvasView::set_onion_skins));
 		past_onion_spin->set_tooltip_text(_("Past Onion Skins"));
@@ -1271,6 +1272,7 @@ CanvasView::create_top_toolbar()
 
 	{ // Future onion skin spin button
 		future_onion_spin=Gtk::manage(new class Gtk::SpinButton(future_onion_adjustment_));
+		future_onion_spin->set_value(work_area->get_onion_skins()[1]);
 		future_onion_spin->signal_value_changed().connect(
 			sigc::mem_fun(*this, &CanvasView::set_onion_skins));
 		future_onion_spin->set_tooltip_text(_("Future Onion Skins"));
@@ -1290,7 +1292,7 @@ CanvasView::create_top_toolbar()
 
 		onion_skin_keyframes = Gtk::manage(new class Gtk::ToggleToolButton());
 		// Enable Onion Skin on Keyframes by default
-		onion_skin_keyframes->set_active(true);
+		onion_skin_keyframes->set_active(work_area->get_onion_skin_keyframes());
 		onion_skin_keyframes->set_icon_widget(*icon);
 		onion_skin_keyframes->signal_toggled().connect(
 			sigc::mem_fun(*this, &CanvasView::toggle_onion_skin_keyframes));
@@ -3479,6 +3481,7 @@ CanvasView::on_meta_data_changed()
 	toggling_show_guides=true;
 	toggling_snap_guides=true;
 	toggling_onion_skin=true;
+	toggling_onion_skin_keyframes=true;
 	toggling_background_rendering=true;
 	try
 	{
@@ -3495,6 +3498,7 @@ CanvasView::on_meta_data_changed()
 		action = Glib::RefPtr<Gtk::ToggleAction>::cast_dynamic(action_group->get_action("toggle-guide-snap"));
 		action->set_active((bool)(work_area->get_guide_snap()));
 		// Update the toggle buttons
+		background_rendering_button->set_active(work_area->get_background_rendering());
 		onion_skin->set_active(work_area->get_onion_skin());
 		snap_grid->set_active(work_area->get_grid_snap());
 		show_grid->set_active(work_area->grid_status());
@@ -3503,6 +3507,7 @@ CanvasView::on_meta_data_changed()
 		// Update the onion skin spins
 		past_onion_spin->set_value(work_area->get_onion_skins()[0]);
 		future_onion_spin->set_value(work_area->get_onion_skins()[1]);
+		onion_skin_keyframes->set_active(work_area->get_onion_skin_keyframes());
 	}
 	catch(...)
 	{
@@ -3511,6 +3516,7 @@ CanvasView::on_meta_data_changed()
 		toggling_show_guides=false;
 		toggling_snap_guides=false;
 		toggling_onion_skin=false;
+		toggling_onion_skin_keyframes=false;
 		toggling_background_rendering=false;
 	}
 	toggling_show_grid=false;
@@ -3518,6 +3524,7 @@ CanvasView::on_meta_data_changed()
 	toggling_show_guides=false;
 	toggling_snap_guides=false;
 	toggling_onion_skin=false;
+	toggling_onion_skin_keyframes=false;
 	toggling_background_rendering=false;
 }
 
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -561,6 +561,7 @@ CanvasView::CanvasView(etl::loose_handle<Instance> instance,etl::handle<CanvasIn
 	toggling_show_guides=false;
 	toggling_snap_guides=false;
 	toggling_onion_skin=false;
+	toggling_onion_skin_keyframes=false;
 	toggling_background_rendering=false;
 
 	set_use_scrolled(false);
@@ -1281,6 +1282,23 @@ CanvasView::create_top_toolbar()
 		toolitem->show();
 
 		displaybar->append(*toolitem);
+	}
+
+	{ // Onion skin on Keyframes/Frames toggle button
+		Gtk::Image *icon = manage(new Gtk::Image(Gtk::StockID("synfig-keyframes"), iconsize));
+		icon->show();
+
+		onion_skin_keyframes = Gtk::manage(new class Gtk::ToggleToolButton());
+		// Enable Onion Skin on Keyframes by default
+		onion_skin_keyframes->set_active(true);
+		onion_skin_keyframes->set_icon_widget(*icon);
+		onion_skin_keyframes->signal_toggled().connect(
+			sigc::mem_fun(*this, &CanvasView::toggle_onion_skin_keyframes));
+		onion_skin_keyframes->set_label(_("Keyframes"));
+		onion_skin_keyframes->set_tooltip_text(_("Show Onion Skin on Keyframes when enabled, on Frames when disabled"));
+		onion_skin_keyframes->show();
+
+		displaybar->append(*onion_skin_keyframes);
 	}
 
 	if(App::enable_mainwin_toolbar)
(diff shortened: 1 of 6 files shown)


FirasH:
As of now it's using only keyframes and there's some glitches with the first keyframe IIRC.

FirasH:
Quick demo:
https://user-images.githubusercontent.com/6705690/120122350-90075c00-c1a8-11eb-9242-61b7314cb917.mp4

Toggling is not immediate (slower when going from Frames to Keyframes).

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -2728,6 +2746,17 @@ CanvasView::toggle_onion_skin()
 	toggling_onion_skin=false;
 }
 
+void
+CanvasView::toggle_onion_skin_keyframes()
+{
+	if(toggling_onion_skin_keyframes)
```
rodolforg, line 2752: Due to the current implementation strategy, you have to change `CanvasView::on_meta_data_changed()` too.

synfig-studio/src/gui/workarea.cpp
```diff
@@ -767,6 +768,14 @@ void WorkArea::set_onion_skins(int *onions)
 	save_meta_data();
 }
 
+void
+WorkArea::set_onion_skin_keyframes(bool x)
+{
+	if (onion_skin_keyframes == x)
+		return;
+	onion_skin_keyframes = x;
```
rodolforg, line 776: You have to add here `save_meta_data()` and update this method as well. Otherwise when user reopens this synfig document, it will always be set as Onion Skin per frame, not keyframe.

rodolforg: You have to call `queue_draw()`, don't you?

synfig-studio/src/gui/workarearenderer/renderer_canvas.cpp
```diff
@@ -371,19 +372,91 @@ Renderer_Canvas::build_onion_frames()
...
+			keyframe_time = base_time;
+			for(int i = past; i > 0; --i) {
+				KeyframeList::iterator iter;
+				if (canvas->keyframe_list().find_prev(keyframe_time, iter)) {
+					n_past_keyframes++;
+					keyframe_time = iter->get_time();
+				}
+				else
+					break;
+			}
```
rodolforg, line 394: I suggest to change it a bit:
you store in a std::vector<Time> the keyframe valid times (including if it is in the RendDesc time range). One vector for past and other for future skins.

You can do the same for non-keyframe onion skin.

Then later you just iterate over these vector more easily and clean to read.

rodolforg (changes requested):

synfig-studio/src/gui/canvasview.cpp
FirasH: As of now it's not saving any metadata, is it better to do so?

synfig-studio/src/gui/workarearenderer/renderer_canvas.cpp
FirasH: Followed you suggestion, applied in the last commit

synfig-studio/src/gui/workarea.cpp
FirasH: Yup :+1:

FirasH:
Noticed that on document loading the value of past and future frames for Onion Skin is somehow not in sync... checking this issue as well

FirasH:
Found another issue (not PR related): if the value of `background_rendering` is changed from `MetaData` dock this is not immediately applied to the status of Synfig Studio as for other keys but needs another event to catch that change... Fixing this as well :wink:

FirasH:
As of now this merge request implements the following improvements:

- A toggle button in the toolbar:
ok-button-toolbar

- An action in the `View` menu:
ok-menu-entry

- Fixes loading of Past/Future Onion Skin values from canvas metadata on document opening
- Fixes broken sync of metadata/toolbar/menu entries of `Background` rendering

synfig-studio/src/gui/workarearenderer/renderer_canvas.cpp
```diff
@@ -371,19 +376,75 @@ Renderer_Canvas::build_onion_frames()
...
+					onion_frames.push_back(FrameDesc(time, w, h, alpha));
+			}
+
+			// Future Frames
+			for(int i = future; i > 0; --i) {
+				Time time = base_time + frame_duration*i;
+				ColorReal alpha = base_alpha + (ColorReal)(future - i + 1)/(ColorReal)(future + 1);
+				if (time >= rend_desc.get_time_start() && time <= rend_desc.get_time_end())
+					onion_frames.push_back(FrameDesc(time, w, h, alpha));
+			}
```
rodolforg, line 444: One last thing about this part:

You can see the two codes here are too similar here, right?
It's good to avoid such 'repetition' to prevent someone to change something in one code (fixing it somehow) and forget to check if second code should be changed too.

I'd suggest to populate the vectors for both past and future onion skin times no matter if it is set to keyframes or frames. Next, you loop like you did on lines 415-425.

Something like:
```c++
std::vector<Time> past_times;
std::vector<Time> future_times;

if (is_keyframes) {
   // loop for filling past_times and future_times according to valid keyframes
} else {
   // loop for filling past_times and future_times according to fps and frames
}

for(int i = past_frames; i > 0; --i) {
	ColorReal alpha = base_alpha + (ColorReal)(past_frames - i + 1)/(ColorReal)(past_frames + 1);
	onion_frames.push_back(FrameDesc(past_times[past_frames - i], w, h, alpha));
}

for(int i = future_frames; i > 0; --i) {
	ColorReal alpha = base_alpha + (ColorReal)(future_frames - i + 1)/(ColorReal)(future_frames + 1);
	onion_frames.push_back(FrameDesc(future_times[future_frames - i], w, h, alpha));
}
```

rodolforg (changes requested):

synfig-studio/src/gui/workarearenderer/renderer_canvas.cpp
FirasH: Done :+1:

synfig-studio/src/gui/workarearenderer/renderer_canvas.cpp
```diff
@@ -361,7 +361,12 @@ Renderer_Canvas::build_onion_frames()
 	current_thumb = FrameId(base_time, thumb_w, thumb_h);
 	frame_duration = Time(approximate_greater_lp(fps, 0.f) ? 1.0/(double)fps : 0.0);
 
-	// set onion_frames
+	// Store vector of Past and Future Frames to use
+	std::vector<Time> past_times, future_times;
+	// Total count of actual Past and Future Frames to use
+	int past_frames, future_frames;
```
rodolforg, line 367: Now you can reduce these variable scope. They are only need inside if-statement down below.

synfig-studio/src/gui/workarearenderer/renderer_canvas.cpp
```diff
@@ -371,19 +376,70 @@ Renderer_Canvas::build_onion_frames()
...
 		}
-		for(int i = future; i > 0; --i) {
-			Time time = base_time + frame_duration*i;
-			ColorReal alpha = base_alpha + (ColorReal)(future - i + 1)/(ColorReal)(future + 1);
-			if (time >= rend_desc.get_time_start() && time <= rend_desc.get_time_end())
-				onion_frames.push_back(FrameDesc(time, w, h, alpha));
+
+		// Check number of available Past and Future Frames
+		past_frames = (past > past_times.size()) ? past_times.size() : past;
+		future_frames = (future > future_times.size()) ? future_times.size() : future;
```
rodolforg, line 428: They are only used since this point. So declare them here, not up there.

And, actually, there is no need to check the values anymore, they can be initialized directly the vector sizes. They are already limited by `past`/`future` before in their loops.

rodolforg (changes requested):
final considerations :D

FirasH:
(Linking to related issues)
Fixes #1562
Fixes #1817

synfig-studio/src/gui/workarearenderer/renderer_canvas.cpp
FirasH: Completed!

FirasH:
> After that, please squash all your commits.

Squashed commits in the most significant parts

rodolforg (dismissed):
Tested and it works great :)

(Although I don't like how little alpha changes between skins, but that's the current onion skin behavior :P)

Gilraiser:
Hello, it would also be very useful to be able to have both activated and of different color.

rodolforg:
> Hello, it would also be very useful to be able to have both activated and of different color.

Hm… In this PR, the number of past-future onion-skins is used for either Keyframe-based or Frame-based version.
If user could activate both simultaneously… it would be somehow confusing?

yoSachinkr:
When can we expect this feature to be in development build of Synfig? 🙂

FirasH:
@morevnaproject what about this one for next development release?

morevnaproject:
>> Hello, it would also be very useful to be able to have both activated and of different color.

> Hm… In this PR, the number of past-future onion-skins is used for either Keyframe-based or Frame-based version.
> If user could activate both simultaneously… it would be somehow confusing?

I think this could be useful to display both types of onion skin activated, but I consider this as new feature/improvement and it is better to do this in separate PR. ^__^

ice0:
@morevnaproject @Gilraiser 

> I think this could be useful to display both types of onion skin activated

Can you show example how this can be useful? As for me, these are different approaches to drawing, which is the reason why so many users reacted negatively to the last change.

Actually I was planning to move the mode switch to the Settings window, to avoid overloading interface.

morevnaproject:
> Can you show example how this can be useful? As for me, these are different approaches to drawing, which is the reason why so many users reacted negatively to the last change.

Sometimes animator need to see previous/next keyframe(s) to see the key points of movement.
**And** (in some situations) he needs to see onionskin for regular frames at the same time - to see the trajectory, produced by interpolation.

> Actually I was planning to move the mode switch to the Settings window, to avoid overloading interface.

I think it would be not good to hide the mode far away in Settings Window. Animator often switching configuration of onionskin in the process of work. This is why in OpenToonz the onionskin settings are integrated directly into timeline(see also #600). And this is one of the top features of OpenToonz. For example in Krita theonionskin's settings are in separate panel ant this is less comfortable for animator.

Again, I suggest to discuss further improvements in separate PR or issue. ^__^

ice0:
@FirasH Can you rebase please?

FirasH:
> @FirasH Can you rebase please?

@ice0 Done!

ice0 (approved):

ice0:
Merged. Thank you!

PR #2170, opened 2021-05-30
[Synfig Studio] Make sure paddings are consistent in TreeView headers on all docks

### LINKED ISSUE #2167: [Synfig Studio] Spacings in Window headers is not consistent
master 7ba303a3

**Issue description**:
Looking at this screenshot of the `Library` window, the header labels (ID, Type, ValueBase, Time Track) are evenly spaced from the borders:
window-library

While in other windows like `Keyframes` the spacing is not consistent with the previous one:
window-keyframes

This inconistency can be found in the following windows:
- `Canvas MetaData` (no left spacing)
- `Layers` (no left spacing)
- `Sets` (no left spacing)
- `Parameters` (slightly higher spacing)
- `Keyframes` (no left spacing)
- `History` (no left spacing)

FirasH:
Fix #2167

### DIFF
--- a/synfig-studio/src/gui/docks/dock_children.cpp
+++ b/synfig-studio/src/gui/docks/dock_children.cpp
@@ -32,6 +32,8 @@
 
 #include "docks/dock_children.h"
 
+#include <gtkmm/stylecontext.h>
+
 #include <gui/canvasview.h>
 #include <gui/localization.h>
 #include <gui/trees/childrentreestore.h>
@@ -55,6 +57,9 @@ using namespace studio;
 Dock_Children::Dock_Children():
 	Dock_CanvasSpecific("children",_("Library"),Gtk::StockID("synfig-children"))
 {
+	// Make sure the used style is consistent with other docks
+	get_style_context()->add_class("synfigstudio-efficient-workspace");
+
     //! \todo the combination of set_use_scrolled(false) and
 	//!       add(*tree_view) causes crashes when
 	//!       Dockable::prev_widget_ is left referencing a
--- a/synfig-studio/src/gui/resources/css/synfig.css
+++ b/synfig-studio/src/gui/resources/css/synfig.css
@@ -29,6 +29,13 @@
 }
 
 /*
+ * TreeView header buttons of docks are set with a consistent padding
+ */
+.synfigstudio-efficient-workspace treeview.view header button {
+	padding: 5px;
+}
+
+/*
  * Image padding for buttons are set
  * Now set padding for text inside buttons
  */


synfig-studio/src/gui/resources/css/synfig.css
```diff
@@ -28,6 +28,13 @@
 	padding: 5px;
 }
 
+/*
+ * TreeView header buttons of docks are set with a consistent padding
+ */
+.synfigstudio-efficient-workspace treeview.view header button {
+	padding: 5px;
+}
```
rodolforg, line 36: Are there any TreeView that sets it via code?

FirasH: Currently the style of header is assigned from `synfig.css` using the styling set under `.synfigstudio-efficient-workspace button` which sets padding to 0px.

This part is specific for buttons in `Gtk::TreeView` as in the case of our dock header labels.

From what I remember I didn't find places in the code where this has been explicitly defined and thus I assume we always relied on default Gtk styles.

ice0: I did not notice any changes. Can you make a screenshot showing what exactly has changed?

FirasH: @ice0 it seems this issue is highly dependent on the OS you're using.

Here on my openSUSE this is how the MetaData dock looks with current master:
metadata-dock-before
and this is with the pull request:
metadata-dock-after

The same is applied to remaining docks listed in the issue report.

FirasH:
Tested on Windows and the `Parameters` docks has a slightly different space.
Need to check if the style is applied also there

rodolforg:
Let me ask you: do you know why some TreeView has this class set? `synfigstudio-efficient-workspace`
@Keyikedalube

Keyikedalube:
> Let me ask you: do you know why some TreeView has this class set? `synfigstudio-efficient-workspace`
> @Keyikedalube

The offending code is this
https://github.com/synfig/synfig/blob/b75ef5ed7a8eaa5409aba84cdb3e8f2afdcf9f97/synfig-studio/src/gui/resources/css/synfig.css#L18

It's intended for toolbar buttons and since the column headers in treeviews are also buttons, they are affected too.

Keyikedalube:
@FirasH There's a way to exclude treeview buttons from being styled

https://stackoverflow.com/questions/16201948/how-to-exclude-particular-class-name-in-css-selector

I tried this:
```css
.synfigstudio-efficient-workspace button:not(treeview > button) {
    ...
```

And it seems to work on my XFCE desktop

synfig-studio/src/gui/docks/dock_children.cpp
```diff
@@ -55,6 +57,9 @@ using namespace studio;
 Dock_Children::Dock_Children():
 	Dock_CanvasSpecific("children",_("Library"),Gtk::StockID("synfig-children"))
 {
+	// Make sure the used style is consistent with other docks
+	get_style_context()->add_class("synfigstudio-efficient-workspace");
+
```
Keyikedalube, line 62: Since we already added that class to our parent widget ie., notebook, you don't need to set it here again

Only the CSS code is needed for this PR :)

Screenshot_2021-06-04_08-49-27

synfig-studio/src/gui/resources/css/synfig.css
Keyikedalube: @FirasH if you use the default padding values instead of setting all sides to 5px you wouldn't need to touch Parameters dock file. That class wasn't styled because it has no toolbar buttons.

You can get the default values of treeview buttons padding using gtk inspector. Here's a screenshot:

Screenshot_2021-06-04_09-05-44

Keyikedalube: To learn more about GTK inspector 👉🏽 https://blog.gtk.org/2017/04/05/the-gtk-inspector/

ice0:
> > Let me ask you: do you know why some TreeView has this class set? `synfigstudio-efficient-workspace`
> > @Keyikedalube
> 
> The offending code is this
> https://github.com/synfig/synfig/blob/b75ef5ed7a8eaa5409aba84cdb3e8f2afdcf9f97/synfig-studio/src/gui/resources/css/synfig.css#L18
> 
> It's intended for toolbar buttons and since the column headers in treeviews are also buttons, they are affected too.


In that case I think we need to style `toolbutton` instead of `button`.

Keyikedalube:
> In that case I think we need to style `toolbutton` instead of `button`.

`toolbutton > button` works :)

Screenshot_2021-06-04_10-31-32

We have to take care of the canvas workspace toolbar though. And that's not a toolbar but series of buttons packed in GtkBox I think

ice0:
> > In that case I think we need to style `toolbutton` instead of `button`.
> 
> `toolbutton > button` works :)
> 
> Screenshot_2021-06-04_10-31-32
> 
> We have to take care of the canvas workspace toolbar though. And that's not a toolbar but series of buttons packed in GtkBox I think

Agree. I don't know why these buttons are not made using toolbar, maybe there is some reason for this. 
Anyway, we can add `toolbutton` class to them.

Keyikedalube:
Or we could track down where those buttons are and apply that CSS class to the button's container instead of the entire dock. That way treeview will be left alone with default GTK theme padding values provided by the user's system (or third party) theme. This sounds like a cleaner solution to me than using `toolbutton` selector

Keyikedalube:
For instance, the layers dock toolbar is set from here
https://github.com/synfig/synfig/blob/eafeda4828fea262646abba906b6650fb16591f4/synfig-studio/src/gui/docks/dock_layers.cpp#L181

We could first store that toolbar widget in a variable and set the class like this:
```c++
auto toolbar = static_cast<Gtk::Toolbar*>(App::ui_manager()->get_widget("/toolbar-layer"));
set_toolbar(*toolbar);
toolbar->get_style_context()->add_class("synfigstudio-efficient-workspace");
```

I updated and tested the code and it works, `GtkTreeView` is left intact :)
Same goes for other docks with toolbar

synfig-studio/src/gui/docks/dock_children.cpp
Keyikedalube: Instead of styling the overall dock, style only the toolbar container/widget: https://github.com/synfig/synfig/pull/2170#issuecomment-854363273
Here's an example of how I did it https://github.com/synfig/synfig/pull/2170#issuecomment-854375880

ice0:
> I don't know why these buttons are not made using toolbar, maybe there is some reason for this.

Ok. Just asked @morevnaproject about this. There is no reason why this is not implemented using the toolbar. This is just legacy code, not converted since GTK2. 

> toolbar->get_style_context()->add_class("synfigstudio-efficient-workspace");

GTK recommends to add `toolbar` class to widget.
https://developer.gnome.org/gtk4/stable/ch41s02.html#id-1.7.4.4.75

ice0:
@FirasH Do you have any plans to make any changes here? I mean @Keyikedalube comment.

FirasH:
@ice0 pushed the changes as suggested by @Keyikedalube 

The style is now only applied to the toolbars, so treeviews aren't affected from this issue anymore.

rodolforg:
> Ok. Just asked @morevnaproject about this. There is no reason why this is not implemented using the toolbar. This is just legacy code, not converted since GTK2.

Well, Gtk 4 removed Toolbar widget: https://docs.gtk.org/gtk4/migrating-3to4.html#gtktoolbar-has-been-removed

ice0:
Related issue: https://github.com/synfig/synfig/issues/2188

PR #2171, opened 2021-05-30
[Synfig Studio] Make sure sequence separator string is always in sync with Preferences

### LINKED ISSUE #1460: Image Sequence Separator String setting requires restart to be used on target output
**Synfig version**:
master c51cef4

**Issue description**:
1. Create a new animation
2. Go to: Edit > Preferences... > Render and change `Image Sequence Separator String` to any arbitrary string
3. Go to: File > Render... and export a sequence of image (e.g. use .png to get a sequence using the string set in step 2)

The sequence will use the previously set `Image Sequence Separator String`.
Restarting Synfig is required to apply step 2.

FirasH:
Fix #1460.

The culprit of the issue was that the sequence separator string was updated only on the first initialization of `Render Settings` dialog.

This pull requests makes sure the sequence separator string is always in sync with the current value set in `Preferences` when `Render` button is pressed.

### DIFF
--- a/synfig-studio/src/gui/render.cpp
+++ b/synfig-studio/src/gui/render.cpp
@@ -330,6 +330,7 @@ void
 RenderSettings::on_render_pressed()
 {
 	String filename=entry_filename.get_text();
+	tparam.sequence_separator = App::sequence_separator;
 	
 	if(!check_target_destination())
 	{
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/render.cpp
```diff
@@ -451,7 +452,7 @@ RenderSettings::check_target_destination()
 					n_frame++)
 			{
 				if(Glib::file_test(filename_sans_extension(filename) +
-					tparam.sequence_separator +
```
ice0, line 454: It looks like this change is redundant because you have already updated the `tparam.sequence_separator`.

FirasH: That's right, not necessary :+1:

ice0 (approved):

ice0:
Merged. Thank you!

PR #2172, opened 2021-05-31
Fix some shortcuts/accels - for item deletion and animation stop

rodolforg:
- Add Numpad Delete key to some widgets as an alternative to regular Delete key
- Remove Numpad Delete key from CanvasView accels, because deprecated Gtk::UIManager only allows one accel per action
- Restore old 'Stop' ('Pause' on GUI) : it was renamed by mistake and missed this action


FirasH (approved):

ice0 (approved):

ice0:
Merged. Thank you!

PR #2173, opened 2021-06-01
[Studio] Shortcut editor more organized and asks for user confirmation

rodolforg:
- It visually separates actions in sections (TreeView)
- Actions and sections sorted alphabetically
- It warns about user trying to bind an already-in-use shortcut
- It features a button to Restore Default Accels only (instead of resetting all Synfig Studio settings) and asks for user confirmation before restoration

Captura de tela de 2021-05-31 09-12-35

### DIFF
--- a/synfig-studio/src/gui/dialogs/dialog_setup.cpp
+++ b/synfig-studio/src/gui/dialogs/dialog_setup.cpp
@@ -68,6 +68,8 @@ enum ShortcutsColumns{
 	SHORTCUT_COLUMN_ID_ACTION_NAME = 0,
 	SHORTCUT_COLUMN_ID_ACTION_KEY = 1,
 	SHORTCUT_COLUMN_ID_ACTION_MODS = 2,
+	SHORTCUT_COLUMN_ID_ACTION_SHORT_NAME = 3,
+	SHORTCUT_COLUMN_ID_IS_ACTION = 4,
 };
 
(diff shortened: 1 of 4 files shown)


synfig-studio/src/gui/dialogs/dialog_setup.cpp
```diff
@@ -584,13 +588,46 @@ Dialog_Setup::on_accel_edited(const Glib::ustring& path_string, guint accel_key,
...
+	auto accel_rows = treeview_accels->get_model()->children();
+	auto default_accel_map = App::get_default_accel_map();
+	for (const auto& section_row : accel_rows) {
+		for (auto& row : section_row.children()) {
+			Gtk::AccelKey accel;
+
+			std::string accel_path;
+			row.get_value(SHORTCUT_COLUMN_ID_ACTION_NAME, accel_path);
+
+			bool found = false;
```
ice0, line 616: `found` is set but not used?

rodolforg: Ops. It is useless indeed. `Gtk::AccelKey accel` is properly initialized for 'empty' accel. So no problem using it if nothing found in AccelMap

rodolforg: Fixed

ice0 (approved):

ice0:
Merged. Thank you!

Issue #1243, opened 2020-03-14
[Lottie Exporter] Implement use-def's


AnishGG:
use-def is needed when you import Synfig files to another file.

AnishGG:
Where can I find all the possibilities of how `use-def` is used?
For instance, I saw all these:
```
<param name="canvas" use="circle.sifz#"/>
<param name="segment_list" use="Shape"/>
<param name="canvas" use=":"/>
<param name="gradient" use=":RedGreenBlue"/>
```
@ice0

ice0:
I hink we need @morevnaproject here :)

AnishGG:
@morevnaproject any insights here. ^__^

Issue #1242, opened 2020-03-13
[Lottie Exporter] Lottie Export Fails after importing sifz file into project


RandomGHUser:
Synfig 1.3.12 (Portable) / Windows 10

Problem Recreation: 
1) Create and save an sifz file with a circle in it.
2) Import sifz file into a new project.
3) Export to Lottie format

I receive these errors:

![error](https://user-images.githubusercontent.com/40616218/76664112-75fc4200-6540-11ea-8d6e-5b7978df059b.PNG)

AnishGG:
Yes, This is expected behavior. `use-def`s are yet to be implemented.
I have created an issue regarding the same.
#1243

PR #2182, opened 2021-06-09
[Lottie exporter] Implement use-defs

AnishGG:
Fixes: #1243
Fixes: #1242

I have tested some sample files, and this fix works for them. @rodolforg can you please review this. ^__^

### DIFF
--- a/synfig-studio/plugins/lottie-exporter/common/Param.py
+++ b/synfig-studio/plugins/lottie-exporter/common/Param.py
@@ -50,11 +50,12 @@ class Param:
         layer = self.get_layer()
         canvas = layer.getparent()
         key = self.param.attrib["use"]
-        if ":" in key:  # Start searching from root canvas
-            canvas = settings.ROOT_CANVAS
         keys = key.split("#")
 
         if len(keys) == 2:  # Meaning file-path is not empty
+            if ":" in keys[1]:  # Start searching from root canvas
+                canvas = settings.ROOT_CANVAS
+
             root = etree.parse(keys[0]).getroot()
             # This is a hack which changes the actual root canvas, so that we
             # can store the root canvas of another file in our settings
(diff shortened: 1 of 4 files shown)


AnishGG:
So the lottie-build fails for this. Let me check it out.

AnishGG:
> So the lottie-build fails for this. Let me check it out.

It is fixed now.

synfig-studio/plugins/lottie-exporter/common/Param.py
```diff
@@ -44,10 +44,34 @@ def get_exported_valuenode(self):
...
             return
+
+        # Full ID string format is like: [[file-path]#][canvas-id:][child-canvas-id:][child-canvas-id:]…value-node-id
+        # Refer here: https://github.com/synfig/synfig/issues/993
         layer = self.get_layer()
         canvas = layer.getparent()
         key = self.param.attrib["use"]
-        anim = canvas.get_def(key)
+        if ":" in key:  # Start searching from root canvas
+            canvas = settings.ROOT_CANVAS
```
rodolforg, line 54: Wouldn't it be safer to check (later) whether first token keys[0 or 1].split(":") is empty?

There shouldn't be "something::other" in id, I guess.

AnishGG: I am not sure what you are trying to say here. Do you mean that the file name can have a `:` inside it? If yes, then I will need to put this check later.
Is this what you mean here?

AnishGG: Made the necessary changes.

rodolforg:
That error seems to be gone, but I got another one while testing with that `walk.sif` example file:
Captura de tela de 2021-06-11 06-24-59


Are they related?

AnishGG:
> Are they related?

Nope, this a different but known error.

`079-ATF-skeleton-animated-spline.sif` test file fails with the same error. You can have a look at the same here: https://travis-ci.com/github/AnishGG/synfig/jobs/512991945#L1677

rodolforg (approved):

AnishGG:
@ice0 I think this PR is ready for merging. 😀

ice0:
Merged. Thank you!

PR #2185, opened 2021-06-11
[Synfig Studio] Fixed closed docks bug

### LINKED ISSUE #1803: Don't dock dockable widgets again on next launch if user undocks and closes them
Building from source
Debian sid

**Issue description**:
I'm trying to keep my dock toolbar minimal with fewer but frequently used docks so it's easier to browse around.
Most docks can be closed or removed from the toolbar however some (or just one, haven't tried all) can't,

One such dock that I don't even use a lot/at all is the canvas browser.
I can close it but quit & run synfigstudio again, I'm always presented with this window.

Is this dock that important to always be docked?

Screenshot from 2020-10-24 19-15-43

Keyikedalube:
* Removed explicit dock_toolbox->present()

If user explicitly undocks and closes this dock don't hesitate to keep
it hidden and let user load the dock manually from the menu bar instead
of forcing the dock as dialog

* Removed explicit present() in Dock_Canvases new_instance()

Explicitly calling present() was causing a bug where every new canvas
instance would force this dock (Canvases) to be shown as a dialog even 
though it was undocked and closed/hidden by the user.

Fixes #1803

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -1702,7 +1702,6 @@ App::App(const synfig::String& basepath, int *argc, char ***argv):
 		dock_manager->show_all_dock_dialogs();
 
 		main_window->present();
-		dock_toolbox->present();
 
 		splash_screen.hide();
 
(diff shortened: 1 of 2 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

PR #2193, opened 2021-06-18
Removed excessive use of std namespace

ice0:


synfig-studio/src/synfigapp/actions/valuedesclink.cpp
```diff
@@ -44,11 +44,10 @@
 #include <synfig/blinepoint.h>
 
 #include <synfigapp/localization.h>
-#include <zconf.h>
+//#include <zconf.h>
```
rodolforg, line 47: What is this header doing here anyway?

rodolforg (approved):

synfig-studio/src/synfigapp/actions/valuedesclink.cpp
ice0: This header was added by this commit (https://github.com/synfig/synfig/commit/264825e7c5e730e4611ccfa916ca6188811777ee), apparently by mistake

PR #2194, opened 2021-06-19
Removed excessive use of std namespace (part 2)

ice0:


Keyikedalube (dismissed):

synfig-core/src/modules/mod_yuv420p/trgt_yuv.cpp
```diff
@@ -32,11 +32,10 @@
 
 #include <glib/gstdio.h>
 #include "trgt_yuv.h"
-#include <cstdio>
+//#include <cstdio>
```
rodolforg, line 35: Why not delete it?

ice0: Done

rodolforg (approved):

PR #2206, opened 2021-07-04
Clean synfigapp::ValueDesc a bit

rodolforg:
Remove useless variable member `ValueDesc::scalar`, some NULL/0 usages and old casts.

Basic explanation:
- `scalar`could have only two values: 0.0 (if invalid/useless) and 1.0 (if ValueDesc was a LinkableValueNode, the only situation when scalar had a meaning);
-  This variable was only used in action ValueDescLink, where it would be multiplied by `scalar` of another ValueDesc and checked if it would yell a negative value;
- In any case, the product of two ValueDesc::scalar could only give 0.0 or 1.0, so it would never be negative. The comparison is always false then;
- In a cascade effect, everything that depended on this comparison would never be affected and could be safely deleted.

### DIFF
--- a/synfig-studio/src/synfigapp/actions/valuedesclink.cpp
+++ b/synfig-studio/src/synfigapp/actions/valuedesclink.cpp
@@ -347,13 +347,9 @@ Action::ValueDescLink::prepare()
 					link_is_scaled=true;
 		}
 
-	std::list<ValueDesc>::iterator iter;
-
 	//Now let's loop through all the value desc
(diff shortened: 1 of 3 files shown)


rodolforg:
I split in several commits to better understanding of each step (commit messages try to explain it better).
However, I think we could squash all commits into one if PR is approved.

ice0:
@rodolforg please rebase. I think is better to merge this PR as is.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2209, opened 2021-07-07
[Studio] Fixed crash in rare cases caused by race condition when rebuilding LayerTreeStore

rodolforg:
There was a mistake in 'queued' assignment in queue_rebuild().

Besides, internal class methods directly call rebuild() and it
could crash due to race condition.


ice0 (approved):

rodolforg:
@ice0 Is it OK to backport it to 1.4.2?

ice0:
> @ice0 Is it OK to backport it to 1.4.2?

Sure. It improves stability.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2213, opened 2021-07-10
[Synfig Studio] The end time in the render settings is now correctly updated when changed from the toolbar

### LINKED ISSUE #2197: Can't undo end time set from toolbar
**Describe the bug**
Canvas workspace doesn't indicate one of its properties (end time) has changed when set from toolbar.

**To Reproduce**
1. Go to the toolbar
2. Change the end time
3. Press enter
4. Document is updated but changes cannot be undone

**Expected behavior**
Undo operation should recover our previous canvas document state.

**Screenshots**
Screenshot_2021-06-22_16-38-32

**System information:**
 - OS: Fedora Linux
 - OS version: 34
 - Synfig version: 1.4.1

**Additional comments**
-nil-

### LINKED ISSUE #2198: End time not updated in render settings
> > No need to open Canvas properties. You can edit time duration in place:
> 
> Doesn't work as expected for render settings. End time will be set for canvas but not the render settings. I remember setting both twice on my animation project some time ago.
> This can be rectified in another issue.

Figured this is actually a bug.
1. Change the end time from the toolbar
2. Save the document
3. Close Synfig Studio
4. Open Synfig Studio
5. Re-open the save file
6. Open Render dialog
7. End time is updated

_Originally posted by @Keyikedalube in https://github.com/synfig/synfig/issues/2025#issuecomment-861475064_

FirasH:
Fixes #2197 
Fixes #2198

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -1431,12 +1431,12 @@ CanvasView::on_current_time_widget_changed()
 void
 CanvasView::on_set_end_time_widget_changed()
 {
-	get_canvas()->rend_desc().set_time_end(framedial->get_end_time());
-	//refresh the renddesc
-	refresh_rend_desc();
-	refresh_time_window();
-	//refresh canvas_properties
-	canvas_properties.refresh();
+	if (framedial->get_end_time() == get_canvas()->rend_desc().get_time_end())
+		return;
+
+	RendDesc rend_desc = get_canvas()->rend_desc();
+	rend_desc.set_time_end(framedial->get_end_time());
+	canvas_interface()->set_rend_desc(rend_desc);
 }
 
 void


Keyikedalube:
@FirasH in the linked issue @rodolforg mentioned about not needing explicit refresh calls after `canvas_interface()->set_rend_desc(rend_desc);`

That means we only need to set the new `rend_desc` and the dialogs will respond accordingly.

FirasH:
> 
> 
> @FirasH in the linked issue @rodolforg mentioned about not needing explicit refresh calls after `canvas_interface()->set_rend_desc(rend_desc);`
> 
> That means we only need to set the new `rend_desc` and the dialogs will respond accordingly.

Ouch, missed that. Thanks for checking 👍

Keyikedalube (dismissed):

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1431,12 +1431,9 @@ CanvasView::on_current_time_widget_changed()
...
 {
-	get_canvas()->rend_desc().set_time_end(framedial->get_end_time());
-	//refresh the renddesc
-	refresh_rend_desc();
-	refresh_time_window();
-	//refresh canvas_properties
-	canvas_properties.refresh();
+	RendDesc rend_desc = get_canvas()->rend_desc();
+	rend_desc.set_time_end(framedial->get_end_time());
+	canvas_interface()->set_rend_desc(rend_desc);
```
ice0, line 1436: @rodolforg @FirasH 
do we need to create a copy of `rend_desc` here?
I mean, maybe it's more correct to set `time_end` to canvas as well?

```
    RendDesc **&** rend_desc = get_canvas()->rend_desc();
    rend_desc.set_time_end(framedial->get_end_time());
    canvas_interface()->set_rend_desc(rend_desc);
```

rodolforg: I don't know if it is good to change it twice:
```c++
    RendDesc **&** rend_desc = get_canvas()->rend_desc();
    rend_desc.set_time_end(framedial->get_end_time()); // HERE
    canvas_interface()->set_rend_desc(rend_desc); // AND HERE
```
What if we improve `canvas_interface()->set_rend_desc()` to not react to an argument equals to current value? Something like:
```c++
if (new_rend_desc == rend_desc)
   return;
. . .
```

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1431,12 +1431,11 @@ CanvasView::on_current_time_widget_changed()
...
-	get_canvas()->rend_desc().set_time_end(framedial->get_end_time());
-	//refresh the renddesc
-	refresh_rend_desc();
-	refresh_time_window();
-	//refresh canvas_properties
-	canvas_properties.refresh();
+	RendDesc rend_desc = get_canvas()->rend_desc();
+	rend_desc.set_time_end(framedial->get_end_time());
+
+	if(rend_desc.get_time_end() != get_canvas()->rend_desc().get_time_end())
```
rodolforg, line 1437: I think this `if` could be on the method beginning. You should then compare with framedial, as rend_desc wouldn't exist yet.

rodolforg (approved):

ice0:
Merged. Thank you!

rodolforg:
I suggest to backport it @morevnaproject

PR #2215, opened 2021-07-11
[Studio] Layers in sets can now be selected with a single click, rather than a double click

### LINKED ISSUE #2085: Make Sets select layer with a single click instead of double-clicking
I propose to make behavior of selecting layers in 'Sets' panel similar to 'Layers' panel where active layer changes as soon as user clicks on it. In 'Sets' you need to double-click on a layer in order to make it active, if you just click on it'll simply become highlighted which is basically useless.
SetsHL

Since 'Sets' supposed to ease access to important layers I think it makes sense to select layer as soon as user clicks on it rather than just make it highlighted. It especially becomes an inconvenience when you need to select several layers from the set, because you can't distinguish between the layer being selected or just highlighted.

rodolforg:
fix #2085

### DIFF
--- a/synfig-studio/src/gui/trees/layergrouptree.cpp
+++ b/synfig-studio/src/gui/trees/layergrouptree.cpp
@@ -54,7 +54,7 @@ using namespace studio;
 
 /* === M E T H O D S ======================================================= */
 
-LayerGroupTree::LayerGroupTree() : editable_(false)
+LayerGroupTree::LayerGroupTree()
 {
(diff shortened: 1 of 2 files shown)


ice0:
@rodolforg please rebase

rodolforg:
> @rodolforg please rebase

@ice0  done

ice0 (approved):

ice0:
Merged. Thank you!

Issue #2219, opened 2021-07-12
Incorrect framerate when importing image sequence


goodstuffhappyshare:
**Synfig version & platform**:
- Synfig 1.4.1
- Windows 10 (64 Bit)

**Issue description**:
When I import an image sequence (marked "FPS 30" in the .lst file), it becomes 15 FPS as viewed inside Synfig as well as in the output video. Here is an example:
[image sequence test.zip](https://github.com/synfig/synfig/files/6798705/image.sequence.test.zip)

**Note**:
This does not happen in Synfig 1.4.0.

rodolforg:
Developer note:
For some reason, `FileSystemNative::ReadStream` doesn't perform `seekg`.
Any clue, @ice0 ?

rodolforg:
@morevnaproject I know how to make a quick fix here. But we should fix the `seekg` behavior later.
Should we fix this regression for 1.4.2?

ice0:
> For some reason, `FileSystemNative::ReadStream` doesn't perform `seekg`.
> Any clue, @ice0 ?

@rodolforg 
This is because `ReadStream` should implement `seekpos` method. Without implementation it just returns -1.

> Its default behavior in streambuf is to do nothing and return -1, but derived classes shall override this behavior to alter the internal pointers appropriately: both filebuf and stringbuf override this virtual member function to set the internal pointer specified by which to the absolute position pos.
http://www.cplusplus.com/reference/streambuf/streambuf/seekpos/

Related PR: https://github.com/synfig/synfig/pull/2224

morevnaproject:
> I know how to make a quick fix here. But we should fix the `seekg` behavior later.
> Should we fix this regression for 1.4.2?

Lets' do this for 1.4.3 ^__^

morevnaproject:
Resolved in Synfig version 1.5.1 - https://www.synfig.org/2021/10/30/synfig-studio-1-5-1-released/

PR #2224, opened 2021-07-13
Added `seekpos` method for FileSystemNative::ReadStream

ice0:
This fixes non-working `seekg` method.

Related issue: #2219


rodolforg (approved):
it works! \o/

(`tellg()` keeps stating the same position, but now `seekg()` works indeed!)

rodolforg:
This fix #2219

rodolforg:
hmm… would it be good to implement `seekoff()` too, right?
http://www.cplusplus.com/reference/streambuf/streambuf/seekpos/

PR #2234, opened 2021-07-18
[Studio] Fixed crash in debug builds when right-clicking on some waypoints

rodolforg:
It only makes sense in this case.
Debug builds crash otherwise due to an assert in find_param_desc().

Reported by GregorysonofCarl here:
https://forums.synfig.org/t/crash-when-right-clicking-waypoints/12391/6

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_waypoint.cpp
+++ b/synfig-studio/src/gui/widgets/widget_waypoint.cpp
@@ -254,8 +254,9 @@ void Widget_Waypoint::set_valuedesc(synfigapp::ValueDesc& value_desc)
 
 	value_widget->set_value_desc(value_desc);
 	synfig::ParamDesc param_desc;
-	if (value_desc.find_param_desc(param_desc))
-		value_widget->set_param_desc(param_desc);
+	if (value_desc.parent_is_layer())
+		value_desc.find_param_desc(param_desc);
+	value_widget->set_param_desc(param_desc);
 }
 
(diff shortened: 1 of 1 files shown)


rodolforg:
The crash affects 1.4.x branch too, but only for debug builds.

rodolforg:
Force pushed to improve the commit message only.

synfig-studio/src/gui/widgets/widget_waypoint.cpp
```diff
@@ -254,8 +254,9 @@ void Widget_Waypoint::set_valuedesc(synfigapp::ValueDesc& value_desc)
 
 	value_widget->set_value_desc(value_desc);
 	synfig::ParamDesc param_desc;
-	if (value_desc.find_param_desc(param_desc))
-		value_widget->set_param_desc(param_desc);
+	if (value_desc.parent_is_layer())
```
ice0, line 257: It looks like it should be:
```
	if (value_desc.parent_is_layer()) {
		value_desc.find_param_desc(param_desc);
		value_widget->set_param_desc(param_desc);
        }
```
?

rodolforg: I thought so, but we use a lot of dialogs that are kept hidden/shown during the whole  application lifetime. So I thought it would be safer to reset param_desc.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2240, opened 2021-07-25
[Core] ValueNode_AnimatedFile now supports more data types

### LINKED ISSUE #2211: Add the ability to import layer values from a TSV file
I recall somebody asking for "Animation File" converter to any parameter.
https://forums.synfig.org/t/conversion-to-animation-from-file/11364

The idea is not related to "Import" action, but for the already available converter that Import indirectly uses.
User could write a TSV like:
```
0.00	1.0
0.30	1.5
0.80	1.2
```
Then he could convert the Radius parameter of a Circle layer, for example, to "Animation from File", and then referring to the above TSV file. Thus, the circle radius will be animated by the file contents.

_Originally posted by @rodolforg in https://github.com/synfig/synfig/issues/2207#issuecomment-875261013_

rodolforg:
string, Angle (degrees), bool, integer, real, time and Vector

Useful for letting animation data (like object coordinates along
time) generated by another software to be used on Synfig.

closes #2211

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_animatedfile.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_animatedfile.cpp
@@ -75,7 +75,7 @@ public:
 		auto it = book.find(file_extension);
 		if (it == book.end())
 			return false;
-		return value_type == type_string;
+		return value_type == type_string || !it->second.supports_string_only;
 	}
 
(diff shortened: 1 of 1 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

PR #2249, opened 2021-07-30
Updated snap to 1.4.2

ice0:


rodolforg:
This PR is for master branch or for 1.4.x branch?
Should this required change be done by those "release-bump" scripts?

ice0:
> This PR is for master branch or for 1.4.x branch?

For `master` branch.

> Should this required change be done by those "release-bump" scripts?

Yes, this is possible, but you also need to check if you need to update the dependencies, so I prefer to do it manually.

rodolforg (approved):

ice0:
> This PR is for master branch or for 1.4.x branch?

Addition: The reason why we cannot add this in 1.4.x is because we need to know the hash sums of the source archives, but they will only be available after release. So if you checkout to 1.4.2 you can only see changes for 1.4.1 etc.

For example, the flatpak sources are kept in a different repository (https://github.com/flathub/org.synfig.SynfigStudio), but since there was no dedicated repository for snaps, I think it's better to keep it on the master branch.

PR #2252, opened 2021-07-31
[Studio] fix image size when importing onto canvas with non-standard size

rodolforg:
When importing a raster image without "Resize to Scale" set,
its size could be wrong, if the target canvas is not 60px/unit.

Reported-by: Arturo
https://forums.synfig.org/t/scaling-circle-with-outline/12425/4


ice0 (approved):

rodolforg:
Another PR fix an issue introduced by this one: #2328

PR #2261, opened 2021-08-10
GSOC: Replacing the rhandle class

luboshnikov:
current status of the code


ice0:
@luboshnikov 
Hi!
Can you briefly describe the work done?

luboshnikov:
I replaced the handle class and changed the rhandle class.

ice0:
> I replaced the handle class and changed the rhandle class.

As I can see here you just make Find and Replace `etl::loose_handle -> std::shared_ptr` and `etl::handle -> std::shared_ptr` which is not a correct way, as we discussed before.

1. Why did you do this if I told you not to do it?
2. Can you describe the difference between `etl::loose_handle` and `etl::handle`? Why do you think you can replace both of them with the same class?

morevnaproject:
@ice0 can you please  write here the original plan, which you provided to @luboshnikov at the start?

luboshnikov:
The loose_handle class changes the reference count only when ref() and unref() methods are called, while handle changes the reference count only when it possesses or stops possessing an object.

ice0:
@morevnaproject 

Sure. Here is the work plan that we discussed with Matvey:
1. Documenting the `rhandle` class. Pay special attention to methods that are absent in standard pointer classes.
2. Checking already written tests for `rhandle` and adding new tests to get 100% coverage of `rhandle` methods.
3. Choosing the best smart pointer class from the std library that is suitable for replacement and implementing the rhandle methods using a stub object of this class.
4. Verifying that all tests are working. Fixing problems. 
5. Renaming non-standard method names to standard ones.
6. Removing stub object from `rhandle` class.
7. Inheriting `rhandle` class from smart pointer class.

With this approach, we can still continue to use the `rhandle` class, but at the same time it becomes possible to use the standard smart pointer class instead.

ice0:
Please reopen if you continue development here.

Issue #2259, opened 2021-08-07
[Void Linux] synfig-studio 1.4.2: test suite failure


newbluemoon:
Hi,

I’m trying to update synfig-studio to 1.4.2 for Void Linux and ran into the follwing test suite failure:

**Describe the bug**
Test suite fails on x86_64 with musl libc.
It succeeds, however, on i686 with both musl and glibc and on x86_64 glibc.

**To Reproduce**
Run `make check` after compiling synfig-studio

**Expected behavior**
The tests should pass as they do with 1.4.1.

**System information:**
 - OS: Void Linux
 - OS version: (rolling release)
 - Synfig-Studio version: 1.4.2

**Additional comments**
Console output:
```
...
make[3]: Entering directory '/builddir/synfigstudio-1.4.2/test'
../config/test-driver: line 107:  5693 Segmentation fault      "$@" > $log_file 2>&1
FAIL: app_layerduplicate
============================================================================
Testsuite summary for Synfig Studio 1.4.2
============================================================================
# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See test/test-suite.log
Please report to https://github.com/synfig/synfig/issues
============================================================================
```
test-suite.log:
```
==============================================
   Synfig Studio 1.4.2: test/test-suite.log
==============================================

# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: app_layerduplicate
========================

synfig(5693) [20:52:32] info: rendering threads 16
synfig(5693) [20:52:32] warning: Cannot open ./synfig_modules.cfg
synfig(5693) [20:52:32] warning: Cannot open /tmp/.local/share/synfig/synfig_modules.cfg
synfig(5693) [20:52:32] info: Loading modules from /etc/synfig_modules.cfg
synfig(5693) [20:52:32] info: Loading lyr_std..
synfig(5693) [20:52:32] info: Loading lyr_freetype..
synfig(5693) [20:52:32] info: Loading mod_geometry..
synfig(5693) [20:52:32] info: Loading mod_gradient..
synfig(5693) [20:52:32] info: Loading mod_particle..
synfig(5693) [20:52:32] info: Loading mod_example..
synfig(5693) [20:52:32] info: Loading mod_gif..
[...]

PR #2265, opened 2021-08-13
Fixed crash of test on Void Linux

ice0:
Crash caused by incorrect work with iterators

Fixes issue: #2259

### DIFF
--- a/synfig-studio/src/synfigapp/actions/layerduplicate.cpp
+++ b/synfig-studio/src/synfigapp/actions/layerduplicate.cpp
@@ -416,7 +416,11 @@ LayerDuplicate::replace_valuenodes(const std::map<synfig::Layer::Handle,synfig::
 
 		auto cloned_layer = layer_pair.second;
 
-		for(auto iter=cloned_layer->dynamic_param_list().cbegin();iter!=cloned_layer->dynamic_param_list().cend();++iter)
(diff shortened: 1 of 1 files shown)


rodolforg:
Hm… By reading the `connect_dynamic_param` internal code now, it seems it didn't need the `disconnect` call…
What do you think?

newbluemoon:
Just for confirmation: this fixes the crash and the test passes. Thank you!

ice0:
@rodolforg 

> Hm… By reading the `connect_dynamic_param` internal code now, it seems it didn't need the `disconnect` call…
> What do you think?

Agree. But without it, we lose `static_param_changed()` signal which is called from `disconnect_dynamic_param`.
I don't know why this signal is used, but it looks like a typo in the code.

ice0:
@newbluemoon 
Thank you for testing!

rodolforg (approved):

ice0:
> > Hm… By reading the `connect_dynamic_param` internal code now, it seems it didn't need the `disconnect` call…
> > What do you think?
> 
> Agree. But without it, we lose `static_param_changed()` signal which is called from `disconnect_dynamic_param`. I don't know why this signal is used, but it looks like a typo in the code.

I thought it was a typo and there must be `disconnect_static_param` somewhere, but this is not a typo and there is no such method.

PR #2266, opened 2021-08-13
changed the rhandle class

luboshnikov:
@morevnaproject
@ice0
only the rhandle class has been changed in this code

### DIFF
--- a/ETL/ETL/_handle.h
+++ b/ETL/ETL/_handle.h
@@ -50,7 +50,11 @@ namespace etl {
 // Forward Declarations
 template <class T> class handle;
 template <class T> class loose_handle;
-template <class T> class rhandle;
+template <class T> class rhandle_new;
+template <class T> class rhandle_old;
+
+template <class T>
+using rhandle =  rhandle_old<T>;
 
 
 // ========================================================================
@@ -358,7 +362,7 @@ public:
 **	\writeme
 */
 template <class T>
-class rhandle : public handle<T>
+class rhandle_new : public handle<T>
 {
 	friend class rshared_object;
 public:
@@ -380,8 +384,8 @@ public:
     std::shared_ptr<value_type> obj;
 
 private:
-	rhandle<value_type> *prev_;
-	rhandle<value_type> *next_;
+	rhandle_new<value_type> *prev_;
+	rhandle_new<value_type> *next_;
 
 	void add_to_rlist()
 	{
@@ -398,7 +402,7 @@ private:
 			return;
 		}
 
-		prev_=reinterpret_cast<rhandle<value_type>*>(obj->back_);
+		prev_=reinterpret_cast<rhandle_new<value_type>*>(obj->back_);
 		next_= nullptr;
 		prev_->next_=this;
 		obj->back_=this;
@@ -432,10 +436,10 @@ private:
 public:
 
 	//! Default constructor - empty handle
-	rhandle():obj(nullptr) {}
+	rhandle_new():obj(nullptr) {}
 
 	//! Constructor that constructs from a pointer to new object
-	rhandle(pointer x):obj(x)
+	rhandle_new(pointer x):obj(x)
 	{
 //		value_type*& obj(handle<T>::obj); // Required to keep gcc 3.4.2 from barfing
         if (obj){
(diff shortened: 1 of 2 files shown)


ETL/ETL/_handle.h
```diff
@@ -371,21 +372,14 @@ class rhandle : public handle<T>
...
-	using handle<value_type>::operator*;
-	using handle<value_type>::operator->;
 
-	/*
-	operator const handle<value_type>&()const
-	{ return *this; }
-	*/
+protected:
+#ifdef _DEBUG
+        public:
```
ice0, line 378: Why did you make it public?

ice0:
@luboshnikov 
1. Have you checked why the build is failing? 
2. Have you documented the methods of this class as I recommended you to do to understand how it works?

ETL/ETL/_handle.h
luboshnikov: I just left it the way it was in the handle class.

ETL/test/handle.cpp
```diff
@@ -327,6 +327,24 @@ int rhandle_general_use_test(void)
 		printf(__FILE__":%d: On clear, instance count=%d, should be zero.\n",__LINE__,my_test_obj::instance_count);
 		return 1;
 	}
+    obj_handle obj1;
```
ice0, line 330: Please add a description for the tests. 
It is also better to add how many references should be in `etl::rhandle` and `etl::handle` after copy and move operations.

ETL/test/handle.cpp
```diff
@@ -327,6 +327,24 @@ int rhandle_general_use_test(void)
...
+    {
+        robj_handle robj(new my_test_obj(rand()));
+        obj1 = robj;
+    }
+    if(obj1.count()!=1){
+        printf("FAILED!\n");
+        printf(__FILE__":%d: On copy, handle count=%d, should be %d.\n",__LINE__,obj.count(),1);
```
ice0, line 337: > printf(__FILE__":%d: On copy, handle count=%d, should be %d.\n",__LINE__,obj.count(),1);

I think this variant will be more readable:
`printf(__FILE__":%d: On copy, handle count=%d, should be 1.\n",__LINE__,obj.count());`

ice0 (changes requested):

ice0:
Also I suggest to do something like this:

1) Rename the modified `etl::rhandle` to `etl::rhandle_new`
2) Return back original `etl::rhandle` class (just copy it from master branch) and rename it to `etl::rhandle_old`
3) Add the alias `etl::rhandle` -> `etl::rhandle_old`
4) Add the same tests to the `etl::rhandle_old` and `etl::rhandle_new` classes.

This way you can see if there are any differences in behavior and can easily switch between the old and the new classes.

ETL/test/handle.cpp
```diff
@@ -327,6 +338,208 @@ int rhandle_general_use_test(void)
 		printf(__FILE__":%d: On clear, instance count=%d, should be zero.\n",__LINE__,my_test_obj::instance_count);
 		return 1;
 	}
+    obj_handle obj1;
+	if(!obj1.empty()){ //! default handle constructor should return an empty pointer
+        printf("FAILED!\n");
+        printf(__FILE__":%d: on create an empty handle, the number of instances=%d must be zero.\n",__LINE__,obj.count());
```
ice0, line 344: `obj1`?

ETL/test/handle.cpp
```diff
@@ -327,6 +338,208 @@ int rhandle_general_use_test(void)
...
+        printf("FAILED!\n");
+        printf(__FILE__":%d: On clear, instance count=%d, should be zero.\n",__LINE__,my_test_obj::instance_count);
+    }
+
+	printf("PASSED\n");
+
+	return 0;
+}
+
+int rhandle_new_general_use_test(void)
```
ice0, line 381: Yes, it looks much better that way.

Next step. Since these functions (`rhandle_new_general_use_test`/`rhandle_old_general_use_test`) should be identical (except for the type of pointer used), I suggest rewriting them using a template.

In this case, you can be completely sure that the tests are identical.

ice0 (changes requested):

ETL/ETL/_handle.h
```diff
@@ -627,7 +664,274 @@ class rhandle : public handle<T>
 	*/
 }; // END of template class rhandle
 
+template <class T>
```
ice0, line 667: Please move `rhandle_old` before `rhandle_new`. In this case `git` will show the changes better.

ice0:
Please reopen if you continue development here.

PR #2270, opened 2021-08-16
File menu migration

DhairyaBahl:


synfig-studio/src/gui/app.cpp
```diff
@@ -844,14 +845,115 @@ class Preferences : public synfigapp::Settings
...
+	simple_actions_action_group->add_action("revert");
+	simple_actions_action_group->add_action("import");
+	simple_actions_action_group->add_action("import-sequence");
+	simple_actions_action_group->add_action("render");
+	simple_actions_action_group->add_action("preview");
+	simple_actions_action_group->add_action("close-document");
+	simple_actions_action_group->add_action("quit");
+
+	App::main_window->insert_action_group("menus",simple_menus_action_group);
+	App::main_window->insert_action_group("actions",simple_actions_action_group);
```
DhairyaBahl, line 872: @rodolforg @ice0 BUT the code is giving segmentation fault ! The moment I comment these two lines and revoke the link of action_groups with main window, code compiles perfectly but the file menu is missing coz I commented the Gtk::Action part !

DhairyaBahl: did my best with removing the segmentation fault but couldn't help it ! it would be great if you can rectify my mistake ! and rest of the code is ok according to me ! but we will see if it works or not on seeing the new file menu :)

DhairyaBahl: code was not giving seg fault till this commit 

image

Kindly rectify my mistake, then I will pace up my work :)

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1464,7 +1485,7 @@ CanvasView::init_menus()
 	action_group->add( Gtk::Action::create("export", Gtk::StockID("synfig-export"), _("Export..."), _("Export")),
 		sigc::hide_return(sigc::mem_fun(*get_instance().get(), &Instance::dialog_export))
 	);
-	action_group->add( Gtk::Action::create("save-all", Gtk::StockID("synfig-save_all"), _("Save All"), _("Save all opened documents")),
+	action_group->add( Gtk::Action::create("", Gtk::StockID("synfig-save_all"), _("Save All"), _("Save all opened documents")),
```
rodolforg, line 1488: Why is this action unnamed?

DhairyaBahl: My bad ! I removed that mistakenly :( Will correct this in a minute

synfig-studio/src/gui/app.cpp
rodolforg: I didn't test your code yet. But here are some stuff I saw:

**First**, the try-catch you (correctly) surrounded the builder loading: you only wrote one specific "catch": `std::exception`. Documentation says `create_from_string()` can throw `Gtk::BuilderError` and `Glib::MarkupError` : and none is derived from `std::exception`. Add those `catch`.

**Second**, simplify your test: try a menu with a single item first. "Quit" or "New". "Quit" is preferable or you'll have a bit trouble to quit synfig ;) 

**Third**: You create actions three times. Look how docs/examples do:
https://developer-old.gnome.org/gtkmm-tutorial/3.24/sec-buildapp-app-menu.html.en
```c++
  // Add actions and keyboard accelerators for the application menu.
  add_action("preferences", sigc::mem_fun(*this, &ExampleApplication::on_action_preferences));
  add_action("quit", sigc::mem_fun(*this, &ExampleApplication::on_action_quit));
  set_accel_for_action("app.quit", "<Ctrl>Q");

  auto refBuilder = Gtk::Builder::create();
  try
  {
    refBuilder->add_from_resource("/org/gtkmm/exampleapp/app_menu.ui");
  }
  catch (const Glib::Error& ex)
  {
    std::cerr << "ExampleApplication::on_startup(): " << ex.what() << std::endl;
    return;
  }

  auto object = refBuilder->get_object("appmenu");
  auto app_menu = Glib::RefPtr<Gio::MenuModel>::cast_dynamic(object);
  if (app_menu)
```
(On)/After application startup, they create/add (simple) actions to Application (by naming and binding the callback). Only once.
After that, they instance `Builder` and then read menu model from resource (string, in our case).

You didn't bind any callback method. It's okay by now, for initial testing. But once you are able to make a simple menu with a single item working, fix it :)

**Fourth**: you created action groups. Do you ever add these groups to anything? They can be inserted in Gtk::Widget (and Gtk::Window is a Gtk::Widget). But maybe it's better to do stuff without separate action groups first.

**Fifth**: Segmentation fault is, in large majority (or always?), due to null pointer deference.
Search on your newly added code where do you deference a pointer (operator `->`). Are you sure  that object is not null there?
Anyway, you can always run synfig via `gdb` and find where it crashes.

DhairyaBahl: Understood ! Will try to make the requested changes as soon as possible :)

DhairyaBahl:
@rodolforg Removed the segmentation fault but still, menu is not rendering !

DhairyaBahl:
@rodolforg @ice0 @morevnaproject the current state of application is : I tried to implement everything in main_window only but it is showing empty window ! Kindly tell me why is that ? 

Coz I was having the same condition when dealing with different files ! So, I simplified it ( just for testing purposes) and still it broke ! Kindly tell me why so ?

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -98,26 +99,74 @@ MainWindow::MainWindow() :
...
-	auto visible_vbox = manage(new Gtk::Box(Gtk::ORIENTATION_VERTICAL));
-	auto hidden_box   = manage(new Gtk::Box());
-
-	auto visible_menubar = App::ui_manager()->get_widget("/menubar-main");
-	auto hidden_menubar  = App::ui_manager()->get_widget("/menubar-hidden");
-	if (visible_menubar != NULL)
+	Glib::RefPtr<Gio::SimpleActionGroup> simple_action_group = Gio::SimpleActionGroup::create();
+	simple_action_group->add_action("menu-file");
+	simple_action_group->add_action("quit");
+
```
DhairyaBahl, line 105: Kindly start looking from this part ! It is just showing an empty screen without any menus.

DhairyaBahl: @rodolforg 

1. I corrected the try catch 
2. I am currently working only with quit action
3. Everything is at one place for now ( will later implemented the logic in multiple files )
4. builder is reading from string. I am aware of fact that I am not using any callback ! But will definitely implement those once a basic layout is implemented using builder. nd I am adding those actions to window using ``` insert_action_group("actions",simple_action_group);``` Is it right ?
5. segmentation fault is resolved for now ! ( null pointer exception )

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -98,26 +99,74 @@ MainWindow::MainWindow() :
...
+	catch(Glib::Error& error)
 	{
-		hidden_box->add(*hidden_menubar);
-		hidden_box->hide();
+		std::cerr << "Error while building the menus : " << error.what() << std::endl;
+	}
 
-		visible_vbox->add(*hidden_box);
-		visible_vbox->pack_start(*visible_menubar, false, false, 0);
+	auto builder_box = manage(new Gtk::Box(Gtk::ORIENTATION_VERTICAL));
```
rodolforg, line 135: Why are you creating a Box?

synfig-studio/src/gui/mainwindow.cpp
rodolforg: `Window` is a `Bin` container, that can only have one widget as direct child.
You are adding a (virtually) empty Box as its child; and that should be reason Window showing empty.

Gtk::Application::set_menubar() may be the way to do it in a easier way than yours.

synfig-studio/src/gui/mainwindow.cpp
DhairyaBahl: I added it in a extra box so that I can deal with that separate box only and have to deal with outside box as less as possible.

rodolforg: I didn't understand

DhairyaBahl: I was wrong i guess. I removed that line now :)

synfig-studio/src/gui/mainwindow.cpp
DhairyaBahl: ``` set_menubar``` only works with ``` Gtk::ApplicationWindow``` but normally we use ``` set_app_menu``` ! I tried using that but still no change ui is just the same ! File menu is still missing ! Old ui restored as expected coz I uncommented the commented normal ui code ! 

Whatever I am doing was working perfectly in my own apps !

doing everything according to [this](https://developer-old.gnome.org/gtkmm-tutorial/3.18/sec-menus-examples.html.en)

DhairyaBahl:
@rodolforg @ice0 I will test the same code on some sample apps to check why is it not working ? and why the menubar is not showing up ! Feel free to help me out and give some suggestions on the code ! 

Thanks :)

synfig-studio/src/gui/mainwindow.cpp
ice0: > set_menubar only works with Gtk::ApplicationWindow

I think there should be no problem switching from `Gtk::Window` to `Gtk::ApplicationWindow`. Can you try this?

DhairyaBahl: @ice0 No issues ! I can definitely try this right away ! but still I am curious what was wrong with my code and why the g menu is not showing with the traditional method ! It should show since it was showing in my normal apps ! I am unable to find any bug in that code from past two days ! 

Feel free to tell me what was wrong with my code ? :) 

Thanks :)

ice0: @DhairyaBahl sorry for a delay, I will take a look.

DhairyaBahl: @ice0 Ping. Kindly tell me what is wrong with my PR so I can move forward. Thanks :)

ice0: @DhairyaBahl 
There is no menu, because you doing things in wrong order.
Please check this example: 
https://python-gtk-3-tutorial.readthedocs.io/en/latest/application.html

I think that after you test this example, it should be more clear for you what's wrong with your code. 

Also I recommend you to make a test application and post its code here so that I can show you which lines can be reordered in order to break the menu.

DhairyaBahl: @ice0 sorry for delayed response ! 

Kindly checkout this repository https://github.com/DhairyaBahl/Minus . I store all of my codes over here that I make for practice.

Secondly, Kindly look into this folder https://github.com/DhairyaBahl/Minus/tree/master/quitOrWhat ! I've stored the practice Gtk App that I've created ! 

Its was working pretty well. Kindly checkout the code. Kindly ignore the alignment issues that it may have coz I was trying and experimenting few things with the same.

Thanks :)

PR #2278, opened 2021-08-22
Fix "files left after uninstall" error for "distuninstallcheck" target

morevnaproject:
Currently we are building source tarballs using Autotools - this operation is done using `autobuild/synfigstudio-release.sh` script.

This procedure was broken at some point in master branch and this PR fixes the issue.


ice0 (approved):

ice0:
@rodolforg 
I don't know why, but using quotes breaks uninstall step.

PR #2282, opened 2021-08-23
Extracted layer size update as a separate function

ice0:
Refactoring: duplicate code is extracted into a separate function.

### DIFF
--- a/synfig-studio/src/synfigapp/canvasinterface.cpp
+++ b/synfig-studio/src/synfigapp/canvasinterface.cpp
@@ -699,6 +699,49 @@ CanvasInterface::jump_to_prev_keyframe()
 	//catch(...) { synfig::warning("Unable to find prev keyframe"); }
 }
 
+static void update_layer_size(const RendDesc& rend_desc, Layer::Handle& layer, bool resize_image) {
+	int w = layer->get_param("_width").get(int());
+	int h = layer->get_param("_height").get(int());
+
+	if (w && h) {
+		Vector x, size = rend_desc.get_br() - rend_desc.get_tl();
+
+		// vector from top left of canvas to bottom right
+		if (resize_image) {
+			if (abs(size[0]) < abs(size[1])) { // if canvas is tall and thin (portrait)
+				x[0]=size[0];	// use full width
+				x[1]=size[0]/w*h; // and scale for height
+
+				if ((size[0]<0) ^ (size[1]<0))
+					x[1] = -x[1];
+			} else { // else canvas is short and fat (or maybe square) (landscape)
+				x[1]=size[1];	// use full height
+				x[0]=size[1]/h*w; // and scale for width
+
+				if ((size[0]<0) ^ (size[1]<0))
+					x[0] = -x[0];
+			}
+		} else {
+			x[0] = w*rend_desc.get_pw();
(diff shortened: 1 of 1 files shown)


synfig-studio/src/synfigapp/canvasinterface.cpp
```diff
@@ -699,6 +699,45 @@ CanvasInterface::jump_to_prev_keyframe()
 	//catch(...) { synfig::warning("Unable to find prev keyframe"); }
 }
 
+static void update_layer_size(CanvasInterface& canvas_interface, Layer::Handle& layer, bool resize_image) {
```
ice0, line 702: It looks like better to pass `RendDesc` isntead of `CanvasInterface` here.

Keyikedalube: Agree
Because that function is internal and used only within synfigapp.

[C++ best practices](https://lefticus.gitbooks.io/cpp-best-practices/content/08-Considering_Performance.html) **Reduce Temporary Objects** section may also be helpful to validate why this is better :)

synfig-studio/src/synfigapp/canvasinterface.cpp
```diff
@@ -699,6 +699,45 @@ CanvasInterface::jump_to_prev_keyframe()
...
+	int w, h;
+
+	w = layer->get_param("_width").get(int());
+	h = layer->get_param("_height").get(int());
```
rodolforg, line 706: If you'll make another commit/push (for changing the method parameter list), what about declaring and initializing in the same line?
`int w = layer-> ...`

synfig-studio/src/synfigapp/canvasinterface.cpp
```diff
@@ -699,6 +699,45 @@ CanvasInterface::jump_to_prev_keyframe()
...
+		// vector from top left of canvas to bottom right
+		if (resize_image) {
+			if(abs(size[0])<abs(size[1])) {// if canvas is tall and thin
+				x[0]=size[0];	// use full width
+				x[1]=size[0]/w*h; // and scale for height
+				if ((size[0]<0) ^ (size[1]<0)) x[1] = -x[1];
+			} else { // else canvas is short and fat (or maybe square)
+				x[1]=size[1];	// use full height
+				x[0]=size[1]/h*w; // and scale for width
+				if ((size[0]<0) ^ (size[1]<0)) x[0] = -x[0];
```
rodolforg, line 720: I particulary don't like these single-line `if`, but it's a matter of taste anyway :)

rodolforg (approved):
Approving right away in case of you won't change the method parameter list.
It does clean up code without behavior change. :)

synfig-studio/src/synfigapp/canvasinterface.cpp
ice0: Done

synfig-studio/src/synfigapp/canvasinterface.cpp
ice0: Done

synfig-studio/src/synfigapp/canvasinterface.cpp
ice0: Done

rodolforg (approved):

PR #2291, opened 2021-09-02
Made shared_objects non-copyable

ice0:
Added explicit copy constructors to Transformations (fixes build errors)

### DIFF
--- a/synfig-core/src/modules/lyr_std/perspective.cpp
+++ b/synfig-core/src/modules/lyr_std/perspective.cpp
@@ -352,6 +352,7 @@ namespace {
 		
 		TransformationPerspective() { }
 		explicit TransformationPerspective(const Matrix &matrix): matrix(matrix) { }
+		TransformationPerspective& operator= (const TransformationPerspective& other) { matrix = other.matrix; return *this; }
 		
 	protected:
 		virtual Transformation* clone_vfunc() const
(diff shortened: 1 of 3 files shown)


synfig-core/src/modules/lyr_std/perspective.cpp
```diff
@@ -352,6 +352,7 @@ namespace {
 		
 		TransformationPerspective() { }
 		explicit TransformationPerspective(const Matrix &matrix): matrix(matrix) { }
+		TransformationPerspective& operator= (const TransformationPerspective& other) { matrix = other.matrix; return *this; }
```
rodolforg, line 355: what is the shared-object concept here?

rodolforg (approved):
Now I got it

ice0:
@rodolforg 

Hi, sorry for the late reply.
The concept of ETL smart pointers is different from std smart pointers. 
Std smart pointers store the reference count inside the smart pointer itself. 
Unlike std, ETL stores a reference count within an object.

For this reason, you cannot put <int> object to the etl::handle. Only objects inherited from etl::shared_object.
The same concept is used in Glib::RefPtr (they switched to std::shared_ptr in gtkmm 4)
https://www.murrayc.com/permalink/2017/04/20/gtkmm-4-progress/


I removed the copy constructor just to make sure it doesn't break anything. 
But as you can see there is a code which copying `Transformations` directly, so I added explicit copy constructor to them.

rodolforg:
@ice0 OK :) Approved then

PR #2294, opened 2021-09-05
Studio: avoid lambda functions for signal callbacks

rodolforg:
Specially important for WorkArea queuing rendering.

Reasoning by Kjell Ahlstedt (gtkmm maintainer) follows below:

From the documentation of sigc::mem_fun():

```
 * @note If the object type inherits from sigc::trackable, and the
 * functor returned from mem_fun() is assigned to a sigc::slot, the functor
 * will be automatically cleared when the object goes out of scope. Invoking
 * that slot will then have no effect and will not try to use the destroyed
 * instance.
```

In most examples in the gtkmm tutorial the signal handlers are members of widgets.
All widgets inherit from sigc::trackable. The automatic disconnection when a
widget is destroyed would be lost if a sigc::mem_fun() is replaced by a C++ lambda
expression. Unless it's combined with sigc::track_obj():

```c++
m_button.signal_clicked().connect(sigc::track_obj([this]{ on_button_clicked(); }, *this));
```
but that's no better than sigc::mem_fun(), is it?

This kind of automatic disconnection is not always necessary. Some instances of
mem_fun() could be replaced by lambda expressions, to show that it's possible.
Preferably together with a description of the pros and cons of mem_fun() and
lambda expressions. But I don't recommend replacing the majority of the mem_fun()s.

https://gitlab.gnome.org/GNOME/gtkmm-documentation/-/issues/10

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -2197,13 +2197,13 @@ void App::save_custom_workspace()
 	Gtk::Entry * name_entry = Gtk::manage(new Gtk::Entry());
 	name_entry->set_margin_start(16);
 	name_entry->set_margin_end(16);
-	name_entry->signal_changed().connect([&](){
+	name_entry->signal_changed().connect(sigc::track_obj([&](){
 		std::string name = synfig::trim(name_entry->get_text());
 		bool has_equal_sign = name.find("=") != std::string::npos;
 		ok_button->set_sensitive(!name.empty() && !has_equal_sign);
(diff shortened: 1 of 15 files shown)


rodolforg:
The changes in WorkArea can fix https://forums.synfig.org/t/synfig-dev-v1-5-0-is-crashing/12480 .

A while ago I had some crashes when changing the selected duck (or when unselecting them all) and I tracked down to this problem. I forgot to make a PR.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2295, opened 2021-09-05
Layer_Freetype: fix user font directory on MacOS

rodolforg:


rodolforg:
Some MacOS user, please test it by building from source :(

rodolforg:
It may fix https://forums.synfig.org/t/text-not-showing-at-all-mac/10865/16

PR #2296, opened 2021-09-06
move layer and valuenode iteration methods to synfig-core

rodolforg:
A code cleanup that will help reuse code for GUI layer copying


synfig-studio/src/synfigapp/actions/layerduplicate.cpp
```diff
@@ -418,21 +357,13 @@ LayerDuplicate::replace_valuenodes(const std::map<synfig::Layer::Handle,synfig::
...
-			for (const auto& vn_pair : cloned_valuenode_map) {
-				if (iter->second == vn_pair.first) {
-					cloned_layer->disconnect_dynamic_param(iter->first);
-					cloned_layer->connect_dynamic_param(iter->first, vn_pair.second);
-				} else {
-					do_replace_valuenodes(iter->second, vn_pair);
-				}
+		replace_value_nodes(cloned_layer,
+							[cloned_valuenode_map](ValueNode::LooseHandle vn) -> ValueNode::LooseHandle {
+			try {
```
ice0, line 362: I think better to not use exceptions here:
```
    auto found = cloned_valuenode_map.find(vn);
    if (found == cloned_valuenode_map.end()) {
        return nullptr;
    }
    return found.second;
```

synfig-studio/src/synfigapp/actions/layerduplicate.cpp
```diff
@@ -418,21 +357,13 @@ LayerDuplicate::replace_valuenodes(const std::map<synfig::Layer::Handle,synfig::
...
-		{
-			for (const auto& vn_pair : cloned_valuenode_map) {
-				if (iter->second == vn_pair.first) {
-					cloned_layer->disconnect_dynamic_param(iter->first);
-					cloned_layer->connect_dynamic_param(iter->first, vn_pair.second);
-				} else {
-					do_replace_valuenodes(iter->second, vn_pair);
-				}
+		replace_value_nodes(cloned_layer,
+							[cloned_valuenode_map](ValueNode::LooseHandle vn) -> ValueNode::LooseHandle {
```
ice0, line 361: Why not use `ValueNode::LooseHandle& vn` here?

synfig-core/src/synfig/synfig_iterations.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+SimpleValueNodeReplaceFunctor::SimpleValueNodeReplaceFunctor(const std::map<ValueNode::LooseHandle, ValueNode::LooseHandle> &replacer_map)
+	: replacer_map(replacer_map)
+{
+}
+
+ValueNode::LooseHandle
+SimpleValueNodeReplaceFunctor::operator()(ValueNode::LooseHandle vn)
+{
+	try {
```
ice0, line 215: Same here.

synfig-studio/src/synfigapp/actions/layerduplicate.cpp
rodolforg: Well, I'm not sure :D 

1. To explicitly know if some replacement is available (we could return bool though)
2. We don't necessarily have to replace the item itself
See synfig_iterations.cpp line 150.

I think  it's safer this way.

synfig-studio/src/synfigapp/actions/layerduplicate.cpp
rodolforg: Done

rodolforg:
Rebased :)

synfig-core/src/synfig/synfig_iterations.cpp
rodolforg: Done

rodolforg:
(fixed the new file headers to new syntax)

synfig-studio/src/synfigapp/actions/layerduplicate.cpp
ice0: I mean, `vn` is only used for map lookup here. So you can replace it with `[cloned_valuenode_map] (const ValueNode::LooseHandle& vn)` without losing any functionality. The only thing that will change is it will not copy `vn` on every iteration. Or I miss something?

rodolforg: As `LooseHandle` is just a "pointer", I thought it wouldn't be a problem.

rodolforg: Done

ice0 (approved):

ice0:
Merged. Thank you!

PR #2299, opened 2021-09-07
[Core] fix access modifiers for valuenode classes

rodolforg:
For unknown reason, some *_vfunc() methods were made public
instead of protected, and something like that.

This PR does some cleanup too, as it uniforms class member order in headers:
- first, private variables
- public typedefs for Handle and ConstHandle
- public creators and destructors
- public (and essential) operator()
- public metadada methods (get_name, get_local_name, check_type)
- protected methods: create_new, *_vefunc, etc.

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_atan2.h
+++ b/synfig-core/src/synfig/valuenodes/valuenode_atan2.h
@@ -51,7 +51,7 @@ public:
 	typedef etl::handle<ValueNode_Atan2> Handle;
 	typedef etl::handle<const ValueNode_Atan2> ConstHandle;
 
-	static ValueNode_Atan2* create(const ValueBase &value=ValueBase());
+	static ValueNode_Atan2* create(const ValueBase& value=ValueBase(), etl::loose_handle<Canvas> canvas=nullptr);
 	virtual ~ValueNode_Atan2();
 
 	virtual String get_name() const override;
--- a/synfig-core/src/synfig/valuenodes/valuenode_atan2.h
+++ b/synfig-core/src/synfig/valuenodes/valuenode_atan2.h
@@ -48,31 +48,25 @@ class ValueNode_Atan2 : public LinkableValueNode
 	ValueNode_Atan2(const ValueBase &value);
 
 public:
-
 	typedef etl::handle<ValueNode_Atan2> Handle;
 	typedef etl::handle<const ValueNode_Atan2> ConstHandle;
 
-
-	virtual ValueBase operator()(Time t)const;
-
+	static ValueNode_Atan2* create(const ValueBase &value=ValueBase());
 	virtual ~ValueNode_Atan2();
 
-	virtual String get_name()const;
-	virtual String get_local_name()const;
+	virtual String get_name() const override;
+	virtual String get_local_name() const override;
+	static bool check_type(Type &type);
 
-	virtual ValueNode::LooseHandle get_link_vfunc(int i)const;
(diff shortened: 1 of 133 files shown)


rodolforg:
@ice0  what do you think?
If you agree, I'll continue with all other valuenodes.

rodolforg:
Maybe it would be a good time to change `operator()(Time t)` to `operator()(const Time& t)` but I'm unsure if we should do it (thread concurrence and stuff like that).

synfig-core/src/synfig/valuenodes/valuenode_atan2.h
```diff
@@ -42,34 +42,27 @@ class ValueNode_Atan2 : public LinkableValueNode
...
 public:
-
 	typedef etl::handle<ValueNode_Atan2> Handle;
 	typedef etl::handle<const ValueNode_Atan2> ConstHandle;
 
-
-	virtual ValueBase operator()(Time t)const;
-
+	static ValueNode_Atan2* create(const ValueBase &value=ValueBase());
+	ValueNode_Atan2(const ValueBase &value);
```
ice0, line 50: I don't know the reason why this constructor was private, maybe it is better to leave it as it was?

ice0:
> If you agree, I'll continue with all other valuenodes.

I agree.

ice0:
> Maybe it would be a good time to change `operator()(Time t)` to `operator()(const Time& t)` but I'm unsure if we should do it (thread concurrence and stuff like that).

```
	synfig::Time time;
	std::cout << sizeof(time);   // outputs 8
```

It looks like we can't get any noticeable improvements, so I think it is better not to change it for now.

synfig-core/src/synfig/valuenodes/valuenode_atan2.h
rodolforg: I don't know either.
Maybe I should label them as `protected` instead of `private`? In case of anyone deriving one of those valuenodes classes?

rodolforg: @ice0 ping

ice0: > Maybe I should label them as protected instead of private? In case of anyone deriving one of those valuenodes classes?

I prefer to keep them private until someone really wants to use them, but ok.

rodolforg: Done!

rodolforg:
@ice0 do you think I should make all valuenodes in a single PR?
For now I'm splitting in commits based on their initial letter XD

rodolforg:
Finished and rebased

rodolforg:
Hm... Should I fix all the `create()` methods that miss the second parameter `canvas`?
Like this one:
```
 warning: cast between incompatible function types from ‘synfig::ValueNode_BoneLink* (*)(const synfig::ValueBase&)’ to ‘synfig::ValueNodeRegistry::Factory’ {aka ‘synfig::LinkableValueNode* (*)(const synfig::ValueBase&, etl::loose_handle<synfig::Canvas>)’} [-Wcast-function-type]
  118 |     reinterpret_cast<ValueNodeRegistry::Factory>(&NodeT::create),
```

ice0 (approved):

ice0:
Merged. Thank you!

ice0:
> Hm... Should I fix all the `create()` methods that miss the second parameter `canvas`?

I think yes, better to do this.

PR #2303, opened 2021-09-11
SVG parser: refactor style parser

rodolforg:


synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -2113,3 +2063,173 @@ get_inkscape_version(const xmlpp::Element* svgNodeElement)
...
+std::string
+Style::get(const std::string &property, std::string default_value) const
+{
+	auto item = data.find(property);
+	if (item == data.end())
+		return default_value;
+	return item->second;
+}
+
+double
```
ice0, line 2118: Could you please add a short description of how this method should work?

synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -2113,3 +2063,173 @@ get_inkscape_version(const xmlpp::Element* svgNodeElement)
...
+Style::merge_presentation_attributes(const xmlpp::Element *elem)
+{
+	// presentation attributes: https://www.w3.org/TR/SVG/styling.html#ElementSpecificStyling
+	// this list won't increase in new versions. phew!
+
+	static const std::vector<std::string> circle_ellipse_props {"cx","cy"};
+	// geometry_props is for 'foreignObject', 'image', 'rect', 'svg', 'symbol', and 'use'
+	static const std::vector<std::string> geometry_props {"height", "width", "x", "y"};
+	//static const std::vector<std::string> circle_props {"r"};
+	static const std::vector<std::string> circle_rect_props {"rx", "ry"};
```
ice0, line 2152: > rx, ry | ‘ellipse’ and ‘rect’

Looks like ellipse should be here?

synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -2113,3 +2063,173 @@ get_inkscape_version(const xmlpp::Element* svgNodeElement)
...
+		}
+		previous_pos = pos;
+		pos++;
+	}
+}
+
+void
+Style::push(const std::string &property, const std::string &value)
+{
+	if (!value.empty() && value != "inherit" && !property.empty())
```
ice0, line 2233: Am I correct and `inherit` is just ignored now?

ice0 (changes requested):

synfig-core/src/modules/mod_svg/svg_parser.cpp
rodolforg: Yes.
Currently it is too.

In most cases, it is the natural way of CSS style processing. And the current code and my proposal use it by default if not overridden.

My proposal to a Style object have all explicitly defined properties. When you go to a child element, you take the parent style and merge the child properties. 'inherit' would mean 'keep that value from my parent'.

Some properties, however, are non-inherited by default, and `inherit` would force it. https://developer.mozilla.org/en-US/docs/Web/CSS/inheritance#non-inherited_properties I'll inspect later (I hope) if any of SVG needs it.

synfig-core/src/modules/mod_svg/svg_parser.cpp
rodolforg: Ouch.

synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -2113,3 +2063,173 @@ get_inkscape_version(const xmlpp::Element* svgNodeElement)
...
+	if (item == data.end())
+		return default_value;
+	return item->second;
+}
+
+double
+Style::compute(const std::string &property, std::string default_value, double reference_value) const
+{
+	std::string value = get(property, default_value);
+	if (value.back() != '%')
```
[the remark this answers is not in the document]
rodolforg: Theoretically, `default_value` would not be an empty string, but yes, we should guarantee anyway.

synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -496,22 +482,23 @@ Svg_parser::parser_layer(const xmlpp::Node* node, xmlpp::Element* root, String p
...
 void
-Svg_parser::parser_rect(const xmlpp::Element* nodeElement,xmlpp::Element* root, const String& fill, const String& fill_opacity, const String& opacity)
+Svg_parser::parser_rect(const xmlpp::Element* nodeElement,xmlpp::Element* root, const Style& style)
 {
 	Glib::ustring rect_id		=nodeElement->get_attribute_value("id");
 	Glib::ustring rect_x		=nodeElement->get_attribute_value("x");
 	Glib::ustring rect_y		=nodeElement->get_attribute_value("y");
 	Glib::ustring rect_width	=nodeElement->get_attribute_value("width");
 	Glib::ustring rect_height	=nodeElement->get_attribute_value("height");
+	Glib::ustring fill       	= style.get("fill", "#000");
```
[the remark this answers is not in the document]
rodolforg: We allow 3 hex digits version too. (See `getRed()`)

synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -2113,3 +2063,173 @@ get_inkscape_version(const xmlpp::Element* svgNodeElement)
...
+
+	// merge properties with current style
+	// Normally, we just replace them with new values
+	// However, there are some special cases (SVG 1.1 6.14):
+	//   - relative values will have the same units as the value to which it is relative
+	//      and percentage lengths are relative to current viewport or bounding box dimensions
+	//   - if a 'clip-path' property is specified on an ancestor element, and the current element
+	//       has a 'clip-path' of none, the ancestor's clipping path still applies to current element
+
+	for (auto prop : style.data) {
```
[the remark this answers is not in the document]
rodolforg: Done

synfig-core/src/modules/mod_svg/svg_parser.cpp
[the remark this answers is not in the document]
rodolforg: Done.

synfig-core/src/modules/mod_svg/svg_parser.cpp
rodolforg: Done

synfig-core/src/modules/mod_svg/svg_parser.cpp
rodolforg: Done

rodolforg:
Done your review points and rebased to master :)

synfig-core/src/modules/mod_svg/svg_parser.cpp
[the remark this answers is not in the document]
ice0: Ok

synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -2113,3 +2063,186 @@ get_inkscape_version(const xmlpp::Element* svgNodeElement)
...
+/// This method retrieve a property value, like get() does, but it returns
+/// its numeric value (and it should be used only for numeric properties.
+///
+/// If property is a numerical number, its value should be parsed and computed.
+/// \param property Property name.
+/// \param default_value The value to be used, if this Style does not have this property explicitly set
+/// \param reference_value If property may be percentual, what is the percentage reference
+/// \return the property value as a real number
+double
+Style::compute(const std::string &property, std::string default_value, double reference_value) const
```
ice0, line 2125: Another question. What value it should return for value = `30%`? 
As it seems to me, there should be 0.3, but as I can see from code there will be 30?

rodolforg: Ouch. I didn't create a percentage value test yet here XD

I rewrote this method.

synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -2113,3 +2063,185 @@ get_inkscape_version(const xmlpp::Element* svgNodeElement)
...
+
+	try {
+		if (value.back() != '%')
+			return std::stod(value);
+		else
+			return reference_value * std::stod(value) / 100.;
+	} catch(...) {
+	}
+
+	if (value != default_value) {
```
ice0, line 2137: There might still be a problem with the empty default value here. For example, if you push a wrong value into a property and pass an empty default. 

Also if you add a test for empty value here, when code begins to repeat yourself.
Here my proposed change:

https://github.com/ice0/synfig/commit/2e105a38ccd09bcf0014621e12e6bc46543cf545

rodolforg: done! Thank you :D

ice0:
Merged. Thank you!

PR #2304, opened 2021-09-11
[core] New blending mode: Alpha Intersection

rodolforg:
Overlapping areas are transparent. Otherwise, act as composite.


rodolforg:
I'd propose to move blending stuff out of color.h. Adding a new blending mode makes almost all synfig-core files to rebuild, including value nodes... They are all static methods anyway.

rodolforg:
Discussion about this new blending mode: https://forums.synfig.org/t/proposal-a-new-blend-method-exclusive-or/12518

ice0:
> I'd propose to move blending stuff out of color.h. Adding a new blending mode makes almost all synfig-core files to rebuild, including value nodes... They are all static methods anyway.

Agree. Any idea where it should be located?

ice0 (approved):

rodolforg:
> Agree. Any idea where it should be located?

Maybe in a new `blend.h` in `synfig-core/src/synfig/color` dir?

morevnaproject:
I have noticed there is no antialiasing in the intersection area. Can this be avoided somehow? 
screenshot_001

rodolforg:
Ugh. I did my best here.
All other ways I tried, it was keeping a thin 'line' of the top layer color.

Does anyone has a suggestion about how to fix it?

rodolforg:
> Any more progress here ?

Not by me. I just updated this PR (rebased it to current master and pushed again :)

Svarov-RZM:
I've tested this PR and it produces the proper result.
The reason why there's no Anti-Aliasing on intersection is because it's covered by a solid surface (opacity = 1). It's a technical limitation, just like with sewing shapes.

To workaround it, you throw a Super Sample layer on top of it and here you go:
AI_AA

Test-case file:
[AlphaIntersection.zip](https://github.com/user-attachments/files/23686899/AlphaIntersection.zip)

@rodolforg 
Please rebase, I'll check it against master and then approve.

rodolforg:
Done @Svarov-RZM

rodolforg:
This CI failure is weird (action: [Synfig CI / Ubuntu 22.04 Jammy (Autotools) (pull_request)](https://github.com/synfig/synfig/actions/runs/19594544029/job/56117423470?pr=2304))
.
image

This 'variable' is a static const:
https://github.com/synfig/synfig/blob/5c2b248d9325f6899782c8eeefdae02f3214f939/synfig-core/src/synfig/color/color.h#L42-L53

And its counterpart, `ceil`, is declared the same way and used in some blending methods.

Svarov-RZM:
> This CI failure is weird

I don't understand what's wrong either... 
It compiles and works fine on my MSYS2 environment.

 @ice0 @BobSynfig Do you guys know what is the problem with failing Autotools CI? Should we just ignore it? As I said, it does work on my PC using 2-build-production.sh script.

rodolforg:
Maybe it is a cache issue.
It seems `Color::floor` isn't used anywhere before this PR, so maybe this symbol was discarded by compiler.

Svarov-RZM:
> Maybe it is a cache issue.

Good point! I removed the cache using `gh delete cache` command and it didn't solve the issue, sadly.
Well, I don't know what is happening, will try to investigate more closely on weekends.

ice0:
@rodolforg @Svarov-RZM 

TLDR: The build fails because `approximate_equal(const T &a, const T &b)` requires variable address, but `constexpr` value only exist at compile-time and don't have address.

As a quick workaround we can create temporary variable to provide it address to function:

```cpp
constexpr auto floor = C::floor;
const bool fg_transparent = approximate_equal(fg.get_a(), floor);
const bool bg_transparent = approximate_equal(bg.get_a(), floor);
```

Now the compiler has temporary variable with address, so linking succeeds.

Alternative approach:
Modify the function signatures to avoid requiring address:

```cpp
inline bool approximate_equal(const T a, const T b);
inline constexpr bool approximate_equal_custom(const T a, const T b, const T precision)
```

P.S.
Just for history.
Alternative approach 2. Change constexpr definition:

```cpp
static constexpr value_type floor() { return 0.0f; }

// usage
const bool fg_transparent = approximate_equal(fg.get_a(), C::floor());
const bool bg_transparent = approximate_equal(bg.get_a(), C::floor());
```

This also will work, but need to be tested with MSVC compiler.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2305, opened 2021-09-11
[build] use MSYS2 Mlt package

rodolforg:


rodolforg:
I thought it would be built faster... I guess I was wrong.

rodolforg:
I think I read it wrong: this PR was built in 15min. I think we reduced build time on Windows indeed :)

ice0:
> I think I read it wrong: this PR was built in 15min. I think we reduced build time on Windows indeed :)

Unfortunately no. As you can see here, for example: 
https://ci.appveyor.com/project/Synfig/synfig/builds/40743903

this PR is built in 9 min. Build time mostly depends on the cpp files that have changed. 
We use `ccache` to speed up the build on Appveyor, so if the files don't change (as in this PR), build times should be as fast as possible. 

There are two reasons why we still not switched to MSYS MLT package.
1. When I added the MLT package to MSYS2, I missed some build flags (`--enable-avformat`), so the MLT functions used by Synfig just don't work. Also some changes from this PR have been rolled back: https://github.com/mltframework/mlt/pull/609. 
The problem is that they rolled back changes to the MLT installation paths which are different on Linux and MSYS2 systems. 
So I need time to rebuild the package and check if everything works ok. 
2. As you can see, the size of the artifact has increased from 50 MB to 100 MB. This is because MLT is generating lumas in `pgm` format and they are huge (255 Mb) and this is the reason why I disabled them in Appveyor build (`--disable-lumas`). 
You can check mlt build flags here: https://github.com/synfig/synfig/blob/master/autobuild/msys2/build_mlt.sh
Later I found what `lumas` can be generated in `png` format which is greatly reduces they size, but to fix that I need to rebuild package. 
The main problem here: mlt didn't use autotools - they manually wrote `configure` script, so some standard options just not work with it. CMake also cannot be used because some options were only available with the `configure` script. I didn't check it after `6.22` version, so maybe all that problems is already fixed.

rodolforg:
Wow.

So I misread this night, actually :)

Closing this PR then

PR #2309, opened 2021-09-12
[Studio] Fix dialog responses when user pressing Esc key

rodolforg:
When user presses Esc key or closes the dialog by clicking on
the X (close) button in window decoration, the dialog does not
return "RESPONSE_CANCEL", but "RESPONSE_DELETE_EVENT".

When programmatically closed, the dialog returns with
"RESPONSE_NONE".

In such cases, the dialog is supposed to be canceled, but
that was not correctly interpreted everywhere.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -359,7 +359,10 @@ public:
 		dialog.set_default_response(dflt);
 
 		dialog.show_all();
-		return (Response) dialog.run();
+		int response = dialog.run();
+		if (response != Gtk::RESPONSE_OK)
+			return RESPONSE_CANCEL;
+		return RESPONSE_OK;
 	}
 
 
@@ -387,7 +390,10 @@ public:
 
 		dialog.set_default_response(dflt);
 		dialog.show();
-		return (Response)dialog.run();
+		int response = dialog.run();
+		if (response != Gtk::RESPONSE_YES && response != Gtk::RESPONSE_NO)
+			return RESPONSE_CANCEL;
+		return Response(response);
 	}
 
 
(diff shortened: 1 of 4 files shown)


synfig-studio/src/gui/app.cpp
```diff
@@ -359,7 +359,10 @@ class GlobalUIInterface : public synfigapp::UIInterface
 		dialog.set_default_response(dflt);
 
 		dialog.show_all();
-		return (Response) dialog.run();
+		int response = dialog.run();
+		if (response != Gtk::RESPONSE_OK)
```
ice0, line 363: I think this will make it clearer what happens: 

```
		if (response == Gtk::RESPONSE_OK)
			return RESPONSE_OK;
		return RESPONSE_CANCEL;
```

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -178,7 +178,10 @@ class studio::CanvasViewUIInterface : public UIInterface
 		dialog.set_default_response(dflt);
 
 		dialog.show_all();
-		return (Response) dialog.run();
+		int response = dialog.run();
+		if (response != Gtk::RESPONSE_OK)
```
ice0, line 182: Same here.

synfig-studio/src/gui/app.cpp
rodolforg: I'd prefer to keep consistency with all other cases where I check if it is not favorable (not OK) first. :)
And, indeed, I don't know why I wrote `return Response(response)` as `response` is obviously `RESPONSE_OK`.

Is this ok?
```c++
		if (response != Gtk::RESPONSE_OK)
			return RESPONSE_CANCEL;
		return RESPONSE_OK;
```
or should I write as you proposed?

ice0: > Is this ok?
Sure.

rodolforg: Done

synfig-studio/src/gui/canvasview.cpp
rodolforg: Done

ice0 (approved):

ice0:
Merged. Thank you!

PR #2311, opened 2021-09-13
Replaced `abs` with `std::fabs` for float values and `std::abs` for templates and int values

### LINKED ISSUE #2307: Synfig hangs on attempt to use Halftone 2 Layer
**Describe the bug**
Any attempt to use Halftone 2 Layer results in Synfig hanging.

Synfig is built from source, master branch. The issue not happens on version 1.4.2.

**To Reproduce**
1. From menu select "Layer" -> "New Layer" -> "Filters" -> "Halftone 2"
2. New  "Halftone 2" layer created -> Synfig hangs.

**Expected behavior**
 "Halftone 2" layer should be correctly displayed don Workarea

**System information:**
 - OS: Linux
 - OS version: Mint 19
 - Commit ID: 96f9856dda598a3d6a5baeac557613ffe4708d04

ice0:
Fix #2307

### DIFF
--- a/ETL/ETL/_surface.h
+++ b/ETL/ETL/_surface.h
@@ -306,7 +306,7 @@ public:
 		if(s.data_)
 		{
 			assert(data_);
-			memcpy(data_,s.data_,abs(pitch_)*h_);
+			memcpy(data_,s.data_,std::abs(pitch_)*h_);
 		}
 	}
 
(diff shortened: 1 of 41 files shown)


morevnaproject:
Awesome!

rodolforg (approved):
Why not `std::abs` everywhere? :)

ice0:
> Why not `std::abs` everywhere? :)

It just helps to better understand where you want to use floating point values or integers.

PR #2312, opened 2021-09-14
Added the ability to add a new vertex on a Spline by double-click

### LINKED ISSUE #643: Shortcut Keys + Click to Add New Vertex Point in Vector Shapes (Outlines and Regions)
**Feature Requests for Future Versions**

**Feature description**:
Simple, just some shortcut keys for "Insert Item" and "Insert Item and keep shape"
So it's like holding certain keys and clicking on area's outlines (those faint dotted lines when selecting shapes) then the vertex point can be created

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/64439891-shortcut-keys-click-to-add-new-vertex-point-in-vector-shapes-outlines-and-regions?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

rodolforg:
Fix #643

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -2127,6 +2127,17 @@ CanvasView::popup_param_menu(ValueDesc value_desc, float location, bool bezier)
 }
 
 void
+CanvasView::create_new_vertex_on_bline(float location, synfigapp::ValueDesc value_desc)
+{
+	synfigapp::Action::Handle action = synfigapp::Action::create("ValueNodeDynamicListInsertSmartKeepShape");
+	auto param_list = canvas_interface()->generate_param_list(value_desc);
+	param_list.add("origin", location);
+	action->set_param_list(param_list);
+	if (action->is_ready())
+		canvas_interface()->get_instance()->perform_action(action);
+}
+
+void
 CanvasView::add_actions_to_menu(Gtk::Menu *menu, const Action::ParamList &param_list,Action::Category category)const
 {
 	get_instance()->add_actions_to_menu(menu, param_list, category);
--- a/synfig-studio/src/gui/canvasview.h
+++ b/synfig-studio/src/gui/canvasview.h
@@ -451,6 +451,8 @@ private:
 	//! Pop up menu for the tools but not the bezier ones.
 	void popup_param_menu(synfigapp::ValueDesc value_desc, float location=0, bool bezier=false);
 
+	void create_new_vertex_on_bline(float location, synfigapp::ValueDesc value_desc);
+
 	void workarea_layer_selected(synfig::Layer::Handle layer);
 
 	void selected_layer_color_set(synfig::Color color);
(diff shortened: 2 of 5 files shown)


yoSachinkr:
I've found that with double click, you can't add a new vertex between the last vertex and the first vertex in a path loop.

rodolforg:
> I've found that with double click, you can't add a new vertex between the last vertex and the first vertex in a path loop.

Fixed

ice0 (approved):

ice0:
Merged. Thank you!

yoSachinkr:
Since you've added the ability to add a new vertex with a double-click, I'd like to suggest (ctrl/shift/alt)+double click on a vertex to remove the vertex as well.

rodolforg:
Possible and probably not difficult.
But I wonder how user may know these new ways to do things :( Double-click I saw in other apps.

yoSachinkr:
There's already a dedicated windows that lists all the "keyboard shortcuts" which can be accessed from the help menu.

ice0:
I suggest using `Ctrl + Click` to add a new vertex and `Ctrl + Shift + Click` to delete.
By using `Ctrl/Cmd` as the action key, we can change the mouse cursor to give the user a hint about the planned action.

yoSachinkr:
but ctrl + click is generally used to select vertices one at a time. I don't think that's a good idea. Double click is working just fine.

> I suggest using `Ctrl + Click` to add a new vertex and `Ctrl + Shift + Click` to delete.
> By using `Ctrl/Cmd` as the action key, we can change the mouse cursor to give the user a hint about the planned action.

ice0:
> but ctrl + click is generally used to select vertices one at a time. I don't think that's a good idea.

What do you mean by "select vertices one at a time"? I think left click selects one vertex at a time.

> Double click is working just fine.

The problem with double-click is what you can't use hint using cursor change, and this becomes unintuitive.

https://user-images.githubusercontent.com/5604544/134125158-c1388b05-869a-4e17-aa55-099db3b809f7.mp4

ice0:
We can also change the current behavior. 
At the moment, clicking on the spline allows you to move it. 
We can move this behavior to `Сtrl + Click`.

Then we can use:
- `Click` on Spline to add vertex (this will work without any modifier keys, so most used action should use it).
- `Shift+Click` - on vertex to remove it
- `Ctrl+Click` on vertex - add it to selection
- `Ctrl+Click` on Spline to move it

I didn't suggest `Alt` as it reserved in XFCE for window movement, but we can use `Ctrl+Alt` or `Shift+Alt` combinations.

yoSachinkr:
> What do you mean by "select vertices one at a time"? I think left click selects one vertex at a time

It's basically when you select multiple vertices, while holding ctrl and and clicking the required vertices you want to select.

morevnaproject:
> * `Click` on Spline to add vertex (this will work without any modifier keys, so most used action should use it).

Adding a vertex is not most used action in my opinion. ^__^
In my experience changing the shape of spline is used more often during animation.

morevnaproject:
I think it would be better to continue this discussion in separate issue. Otherwise we might lost it at some point. ^__^

PR #2314, opened 2021-09-16
[core] Improve hints for Dash item parameters

rodolforg:
Dash Item length and offset are measured as distances,
differently of Width Point position, that is relative to
Spline length.

The parameter hint "dash" was not used/parsed anywhere, so
I removed it.

Captura de tela de 2021-09-16 04-07-35


ice0 (approved):

ice0:
Merged. Thank you!

PR #2325, opened 2021-09-21
[Studio] Bone Tool cleanup

rodolforg:


synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -729,148 +742,19 @@ StateBone_Context::event_mouse_release_handler(const Smach::event& x)
...
-					bool is_currently_on(get_work_area()->get_type_mask()&Duck::TYPE_WIDTH);
-					if(!is_currently_on){
-						get_canvas_view()->toggle_duck_mask(Duck::TYPE_WIDTH);
-					}
+							ValueNode::Handle new_active_bone_node;
+							if (skel_layer) {
+								new_active_bone_node = ValueNode_Bone::create(bone,get_canvas());
+								action->set_param("item",new_active_bone_node);
+							} else {
+								Bone bone2 = bone;
```
ice0, line 745: It looks like you not need `bone2` anymore. 
So you can just use:
```
ValueNode_Composite::create(std::pair<Bone,Bone>(bone, bone)
```

ice0:
@rodolforg Code looks good to me.

@morevnaproject Can you test this PR?

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -960,22 +948,27 @@ StateBone_Context::find_bone(Point point,Layer::Handle layer) const
...
+		|| (type == SKELETON_DEFORMATION_TYPE && !is_width_duck_currently_on)) {
+		get_canvas_view()->toggle_duck_mask(Duck::TYPE_WIDTH);
+	}
+}
+
 void
 StateBone_Context::make_layer(){
 	egress_on_selection_change=false;
 
 	bool is_currently_on(get_work_area()->get_type_mask()&Duck::TYPE_WIDTH);
```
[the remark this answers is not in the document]
rodolforg: Feito

synfig-studio/src/gui/states/state_bone.cpp
rodolforg: Done

morevnaproject:
> @morevnaproject Can you test this PR?

Added to my list now. Thank you! ^__^

rodolforg:
@morevnaproject  I suggest you to write/record your tests here, because I'll submit another PR related to Bone Tool following this PR approval XD
It will fix a crash with active bone selection changes (after undoes)

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -626,57 +623,67 @@ StateBone_Context::event_mouse_release_handler(const Smach::event& x)
...
 							ValueNode_Bone::Handle bone_node;
+							if (deform_layer) {
+								ValueNode_Composite::Handle comp = ValueNode_Composite::Handle::cast_dynamic(value_desc.get_value_node());
+								value_desc = ValueDesc(comp,comp->get_link_index_from_name("first"),value_desc);
+							}
+
 							if (!(bone_node = ValueNode_Bone::Handle::cast_dynamic(value_desc.get_value_node())))
 							{
-								error("expected a ValueNode_Bone");
+								error("2. expected a ValueNode_Bone");
```
morevnaproject, line 686: Why "2. " at the start of the string? ^__^

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -731,148 +737,17 @@ StateBone_Context::event_mouse_release_handler(const Smach::event& x)
...
-							bone2.set_angle(Angle::rad(acos(0.0)));
-							if((clickOrigin-releaseOrigin).mag()>=0.01) {
-								bone2.set_scalelx((releaseOrigin - clickOrigin).mag());
-								bone2.set_angle((releaseOrigin - clickOrigin).angle());
+							ValueNode::Handle new_active_bone_node;
+							if (skel_layer) {
+								new_active_bone_node = ValueNode_Bone::create(bone,get_canvas());
+								action->set_param("item",new_active_bone_node);
+							} else {
+								ValueNode_Composite::Handle bone_pair = ValueNode_Composite::create(std::pair<Bone,Bone>(bone,bone),get_canvas());
```
morevnaproject, line 745:
```
std::pair<Bone,Bone>(bone,bone)
```
I am not sure - is it safe to use a same single `bone` to create a bone pair here?
Maybe use two bone objects (`bone1` and `bone2`)?

synfig-studio/src/gui/states/state_bone.cpp
rodolforg: Oh, it wasn't supposed to be committed.

I was investigating a crash and numbering all those strings. From the crash fix, I did this partial PR. Next it'll be the actual crash fix.

synfig-studio/src/gui/states/state_bone.cpp
rodolforg: I rechecked now and it's safe.
It'll actually store a value copy.

For this case, `ValueNode_Composite::create` calls `ValueNode_Const::create` with a `Pair<Bone,Bone>` value, and that calls `ValueNode_Bone::create` that _finally_ calls `ValueNode_Bone::ValueNode_Bone(const ValueBase &value, etl::loose_handle<Canvas> canvas)` for each Pair<Bone,Bone> item component (first and second).

In this last method, it uses the pair item component (first or second) here (named as `value`): `Bone bone(value.get(Bone()));`

morevnaproject: Got it! Thank you for the explanation!

synfig-studio/src/gui/states/state_bone.cpp
```diff
@@ -1036,108 +904,71 @@ StateBone_Context::event_layer_selection_changed_handler(const Smach::event& /*x
...
-			return -1;
-		}
+	std::vector<Bone> bone_list;
+	ValueDesc list_desc(layer,"bones");
+	if(Layer_Skeleton::Handle::cast_dynamic(layer)){
+		bone_list = (*list_desc.get_value_node())(get_canvas()->get_time()).get_list_of(Bone());
+	} else if (Layer_SkeletonDeformation::Handle::cast_dynamic(layer)) {
+		std::vector<std::pair<Bone,Bone>> bone_pair_list = (*list_desc.get_value_node())(get_canvas()->get_time()).get_list_of(std::pair<Bone,Bone>());
+		for (const auto& item : bone_pair_list)
+			bone_list.push_back(item.second);
```
morevnaproject, line 916: After reviewing this part I have realized that there is a bug in original code: it should use "first" bone instead of "second". Because of this bug I was experiencing an issue when it was not possible to select bone as active at some when second bone had different position than second (this usually happens when user deforms image).

So, we should change this part of code from "item.second" to "item.first', but I suggest to do that in separate PR, because it's better to have this PR for cleanup&refactoring only. ^__^

morevnaproject (dismissed):

morevnaproject:
>  I suggest you to write/record your tests here, because I'll submit another PR related to Bone Tool following this PR approval XD

Woops, I already did my tests before I saw your message! This is a great idea and I will do that next time (I believe we have many more fixes ahead XD).

morevnaproject:
The PR is looks good to me! Thank you!

synfig-studio/src/gui/states/state_bone.cpp
rodolforg: First item is for resting and second is for posing state, right?

rodolforg: Maybe it depends on layer activate status?

synfig-studio/src/gui/states/state_bone.cpp
rodolforg: cleaned up now and rebased! :)

synfig-studio/src/gui/states/state_bone.cpp
morevnaproject: > First item is for resting state and the second one is for posing state, right?

Exactly!

morevnaproject: > Maybe it depends on layer activate status?

In general case - yes. But Bone Tool always sets the Skeleton Deformation layer to “inactive” state. So, in our case we always deal with “first”.

morevnaproject (dismissed):

morevnaproject (approved):

synfig-studio/src/gui/states/state_bone.cpp
rodolforg: Fixed via https://github.com/synfig/synfig/pull/2340

morevnaproject: Thank you! ^__^

PR #2326, opened 2021-09-22
Autotools: Remove @MAGICKPP_LIBS@ variable from LDFLAGS

morevnaproject:
...because in 391f8a1540b4c9339bb1141f the @MAGICKPP_LIBS@ variable is placed in correct location now - at 'libmod_magickpp_la_LIBADD' list.

See here - 
https://github.com/synfig/synfig/blob/391f8a1540b4c9339bb1141f4067221f6cc87b8b/synfig-core/src/modules/mod_magickpp/Makefile.am#L36-L39


ice0 (approved):

ice0:
Merged. Thank you!

PR #2327, opened 2021-09-23
SVG: fix parser for <path> element except A command

rodolforg:
Here I try to fix all inconsistencies I found in implementation for each `<path>` command, except for `A` and `a`.

Specs:
SVG 1.1: https://www.w3.org/TR/SVG11/paths.html
SVG 2: https://www.w3.org/TR/SVG/paths.html


rodolforg:
Last push was to fix LGTM warning (an error actually).

PR #2334, opened 2021-09-27
add explicit namespace synfigapp for Action macros

rodolforg:
don't require a `using namespace Action` in a file that use
such macros.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2341, opened 2021-10-02
Bone Tool: fix crash when undoing Child Bone until first one

rodolforg:
If someone uses Bone Tool and create 3 bones in a row and
undoes 2 times, creating a new (second) bone with Bone Tool
crashes.

The problem was active-bone was based on list index, that
changes when we add/remove items.


morevnaproject:
I am testing this one now.

PR #2343, opened 2021-10-02
Bone Tool: fix default bone width preference loading

rodolforg:
If saved with pixel unit system, the value would be loaded wrong, as
if it was the Synfig unit system


ice0 (approved):

ice0:
Merged. Thank you!

PR #2350, opened 2021-10-10
Bevel parameter of Rectangle layer is a relative value, not distance

rodolforg:
Range: 0% - 100%, being 100% half of rectangle width/height


ice0:
> @ice0 no. It is just a visual tip for which "cell renderer" to use (WidgetDistance or normal one). It is already stored as a real number.

Ok. Merged. Thank you!

PR #2364, opened 2021-10-14
Canvas class: hide std::list methods and do some cleanup

rodolforg:
Although `Canvas` itself is (argh) a `std::list` of layers, it must always have a forced last list element: a null layer.
As a consequence, it must `override` some `std::list` methods to assure a programmer doesn't access or remove this special layer: like `size()` and `end()`.

However, std containers are not supposed to be derived as they don't have virtual methods. A misplaced cast (like passing a Canvas as argument to a function that accepts `std::list` as parameter) may broken the list.
As original code was written way before C++11, some methods were added to `std::list` (and more may will be added). A distracted programmer could use those new methods and break the special behavior of this derived `std::list`. Example: `push_back()` with different signatures of the original - and they do exist now on C++11: current implementation only overrides one.

Therefore, it is best to just hide this ancestral class by making this inheritance `private` and make it available to programmer only via wrapped or customized methods.

### DIFF
--- a/synfig-core/src/synfig/canvas.cpp
+++ b/synfig-core/src/synfig/canvas.cpp
@@ -938,26 +938,7 @@ Canvas::insert(iterator iter,etl::handle<Layer> x)
 
 	add_child(x.get());
 
-	LooseHandle correct_canvas(this);
-	//while(correct_canvas->is_inline())correct_canvas=correct_canvas->parent();
-	Layer::LooseHandle loose_layer(x);
-
-	add_connection(loose_layer,
-				   sigc::connection(
-					   x->signal_added_to_group().connect(
-						   sigc::bind(
-							   sigc::mem_fun(
-								   *correct_canvas,
-								   &Canvas::add_group_pair),
-							   loose_layer))));
-	add_connection(loose_layer,
-				   sigc::connection(
-					   x->signal_removed_from_group().connect(
-						   sigc::bind(
-							   sigc::mem_fun(
-								   *correct_canvas,
-								   &Canvas::remove_group_pair),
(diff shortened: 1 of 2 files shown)


rodolforg:
OK. Autotools in my machine build.
Errors were in CMake. Any idea?

ice0:
> OK. Autotools in my machine build. Errors were in CMake. Any idea?

That's just a luck )
Same error with autotools build:
```
make[2]: *** [Makefile:1317: action_export_icon.png] Segmentation fault (core dumped)
make[2]: *** Deleting file 'action_export_icon.png'
make[2]: *** Waiting for unfinished jobs....
make[2]: *** [Makefile:1317: action_unexport_icon.png] Segmentation fault (core dumped)
make[2]: *** Deleting file 'action_unexport_icon.png'
```

it faults on different files, so you need to restart it several times to reproduce. 
But for me it just hangs.

You can test it using this command:
```
./synfig ~/projects/synfig/synfig-studio/images/action_interpolate_interpolation_icon.sif -o ./test.png --time 0f
```

I noticed that it loops in `remove_group_pair` method, so I added debug output to `canvas.cpp`:
```
Canvas::remove_group_pair(String group, etl::handle<Layer> layer)
{
	static int counter = 0;
	std::cout << "Removed from: " << group << ++counter << std::endl;
```

and this is result:
```
Removed from: Shading13832
Removed from: Shading13833
Removed from: Shading13834
Removed from: Shading13835
Removed from: Shading13836
Removed from: Shading13837
Removed from: Shading13838
```

It hangs because layer inside `group_db_` is NULL, so it never exits (I guess because of the recursive call).

PR #2366, opened 2021-10-15
Circle Tool: remove deprecated falloff option

rodolforg:
The Circle layer does not have this parameter anymore


ice0:
Merged. Thank you!

PR #2367, opened 2021-10-16
Fix inverted contour rendering for consecutive blank pixel rows between shapes

rodolforg:
Before:
screenshot_014

After:
screenshot_018

Though not used by current GUI tools, a Contour may be composed
of more than one polygon or bline-shape.

Let's assume we are rendering a shape like the character 'i':
it is a dot above a straight line segment (or a bit more elaborated
shape), right?
There is at least a pixel row separating them.

If we set 'invert' parameter of the contour that contains both
polygons (dot + straight line segment),
the row(s) between those shapes are not filled, but kept as if
'invert' was not set.

### DIFF
--- a/synfig-core/src/synfig/rendering/software/function/contour.cpp
+++ b/synfig-core/src/synfig/rendering/software/function/contour.cpp
@@ -185,6 +185,21 @@ software::Contour::render_polyspan(
 					p.put_hline(window.maxx - x);
 				}
 
+				// fill any empty line until next mark
+				if (++y != cur_mark->y)
+				{
+					if (simple_fill)
+					{
+						sp.move_to(window.minx, y);
+						sp.put_block(cur_mark->y - y, window.maxx - window.minx);
+					}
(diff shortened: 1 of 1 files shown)


ice0:
@rodolforg please rebase

ice0:
> Alternative is to use "Squash and merge" green button, but in this case all commits will be squashed into one. Maybe this is acceptable for this PR? ^__^

Ok, I will try it next time!

rodolforg:
> Strange enough `generic_pen` does not provide virtual methods, even though another class - in the same file - `alpha_pen` inherits it, and 'overrides' some of its methods.

So I removed the trouble commit.
Here it is: https://github.com/rodolforg/synfig/commit/48edcddbcbafc03e2c1aee1afc5da94868c5ac9a
I tried to make the `generic_pen` methods virtual, but it caused some building issue related to a `const synfig::Color&` thing. So I gave up.

PR #2368, opened 2021-10-17
[Synfig Studio] Refresh Widget_RendDesc on width/height link toggling

FirasH:
This fix is necessary to avoid out of sync checkbuttons when toggling width/height link from their actual status.
See `Other > Pixel Aspect` status checkbutton for example.

Screenshot_84

Screenshot_92


ice0:
Merged. Thank you!

Issue #2297, opened 2021-09-06
Feature request: Change in onion skin behaviour.


yoSachinkr:
In current implementation, the onion skin is drawn basically by changing the opacity and showing the overlapping images of the past, present and future frames, which, in my opinion, is not the right way to implement the onion skin, as it causes confusion as well as distraction.
I propose, 
1. Opacity of only past and future keyframes should be changed. Present keyframe should remain fully opaque.
2. Onion skins should appear above the present frame.
3. Color of past and future keyframes should be different, like red for past keyframe and blue for future keyframes.

FirasH:
I can take a look at this one, not sure how long this will take.
Maybe we can work on the first 2 points first and then proceed with the others.

yoSachinkr:
Onion skin is one of the most helpful and important tool in animation. Any improvement will be a welcome change. 😀

FirasH:
I don't usually use `Onion Skin`, but I feel this is a key feature for advanced users.
For this reason I have some observations about the suggested points that we can discuss about and gather ideas for next improvements.

Point 1 and 2 are straightforward and in the mentioned pull request you can find a modified version that includes those changes.

Point 3 and 4 are more detailed and require more information.

_Point 3_
Looking at the screenshot I can infer that there's and addition of a red/blue plane for past/future onion frames.
For example, the yellow shape is added to the blue "future" color and the resulting shape in the onion frame is green, right?
Maybe we can add an option in the settings where the user can select those two colors?

_Point 4_
Do we let the user define the opacity of the onion skins, maybe with a slider in the UI?
Do all onion skins keep the same opacity? Or we need to adjust their opacity depending on their distance from current frame?
I don't think allowing too much customization is great as it would overcomplicate its use.

@morevnaproject, @yoSachinkr or anyone else feel free to comment about it.

yoSachinkr:
For point 3, I'll suggest a little bit of change. Instead of adding (as using add blend method in synfig) red/ blue color, if the blend method would be "color" with 0.7 or 0.5 opacity. it would give a perfect red/blue colored past and future keyframe look.
For point 4, slider would be perfect for opacity change for the onion skins.
 And about having variable opacity as per distance, i don't think it would be that useful. I think it's a matter of discussion.

yoSachinkr:
> Maybe we can add an option in the settings where the user can select those two colors?

It's a good idea to provide an option to use either colored onion skins or the normal onion skins, as many would want to switch between them time to time.

[...]

PR #2369, opened 2021-10-17
[Synfig Studio] Improvements to Onion Skin behaviour (opacity and z-order of frames)

FirasH:
The current goals are to deal with:
- [x] Only the opacity of past and future Keyframes should be changed, current Keyframe should remain fully opaque
- [ ] Onion Skins should appear behind the present frame

See #2297 for reference.

### DIFF
--- a/synfig-studio/src/gui/workarearenderer/renderer_canvas.cpp
+++ b/synfig-studio/src/gui/workarearenderer/renderer_canvas.cpp
@@ -421,6 +421,9 @@ Renderer_Canvas::build_onion_frames()
 			}
 		}
 
+		// Add current time Frame keeping its full opacity
+		onion_frames.push_back(FrameDesc(current_frame, 1.f));
+
 		// Cycle through past_times and retrieve their Time values
 		for(int i = past_times.size(); i > 0; --i) {
 			ColorReal alpha = base_alpha + (ColorReal)(past_times.size() - i + 1)/(ColorReal)(past_times.size() + 1);
@@ -437,9 +440,10 @@ Renderer_Canvas::build_onion_frames()
 		ColorReal summary = 0.f;
 
 		// If there's more than one onion frame, normalize using their cumulative alpha value
-		if (onion_frames.size() > 1) {
-			for(FrameList::const_iterator i = onion_frames.begin(); i != onion_frames.end(); ++i)
-				summary += i->alpha;
+		if (onion_frames.size() > 2) {
+			for(FrameList::const_iterator i = onion_frames.begin(); i != onion_frames.end(); i++)
+				if (i->id.time != current_frame.time)
+					summary += i->alpha;
 		}
 		// Otherwise normalize by the maximum possible alpha value
 		else {
@@ -447,11 +451,10 @@ Renderer_Canvas::build_onion_frames()
 		}
 
 		ColorReal k = approximate_greater(summary, ColorReal(1.f)) ? 1.f/summary : 1.f;
-		for(FrameList::iterator i = onion_frames.begin(); i != onion_frames.end(); ++i)
-			i->alpha *= k;
-
-		// Add current time Frame at last keeping its full opacity
-		onion_frames.push_back(FrameDesc(current_frame, 1.f));
+		for(FrameList::iterator i = onion_frames.begin(); i != onion_frames.end(); i++)
+			// Don't modify current frame opacity
+			if (i->id.time != current_frame.time)
+				i->alpha *= k;
 	} else {
 		onion_frames.push_back(FrameDesc(current_frame, 1.f));
 	}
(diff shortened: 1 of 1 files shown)


rodolforg:
>  https://user-images.githubusercontent.com/6705690/137630667-73371858-ec26-471e-8ec5-02667788e6ca.mp4

I can't see this video. It says the file is corrupted.

FirasH:
@rodolforg I have a question about the implementation of this feature, not sure if I'm going in the right direction...

Here's what I want to do.

Currently the onion skin frames are added one on top of each other, which is not enough if we want to color the onion frames.
I tried to color the frame and that can be done with cairomm, but as of now if we simply add frames on top of each other we'll end up with all the onion frame being colored (changed and unchanged parts of the animation) and stacked one on top the other.

Of course that's not a good solution and for this reason I was thinking to use a [mask](https://cairographics.org/documentation/cairomm/reference/classCairo_1_1Context.html#a2f76e3b13027efc4c8745de5e04e00ec) when adding the onion frames to the current one, adding and coloring only the changed parts of the animation. This requires to compare somehow the current frame data with the onion frames data and exclude unchanged areas.

The problem is that I have a very basic idea on how to edit a `Cairo::ImageSurface`. From the documentation I can get `get_data()` ([cairomm documentation](https://cairographics.org/documentation/cairomm/reference/classCairo_1_1ImageSurface.html#a94ba52fe4a201579c8a5541717822bdb)), but I have no clue how to use it. The best possible solution would be to use something like an `and` logic operator, create a mask and then proceed with the next operations.

Does it makes any sense? And if yes do you have any idea how to do that?

FirasH:
Here's a screencast of the issue:
https://user-images.githubusercontent.com/6705690/139593811-d9c96187-99ed-4a13-b29a-bc023d930d16.mp4

(the code is not pushed to the pull request)

FirasH:
@rodolforg I think I have a solution, hope it works :wink:

ice0:
@FirasH 
Can you check for a case where past and future keyframes overlap? How will it look like?

yoSachinkr:
I would like to suggest a solution, if I may. How about having a toggle button available next to layers toggle button in the layers panel with default set to "show onion skin". So, when the onion skin feature is set to "on", only those layers will show whose "show onion skin" toggle is "on".

FirasH:
After playing a bit with the code I could better understand the logic behind the current implementation.
I'm reporting here my observation so we can discuss a bit more in depth about the possible improvements.

As of now each frame is rendered independently and the frame that has been computed for current frame can be used exactly the same as an onion frame. Each frame is an image that results from rendering all the active layers in a single plane, so we cannot simply "remove a layer" from a rendered frame and color only the changed parts, or at least without rendering again another frame (and add all the necessary code changes). This approach for sure has the advantage of reducing the number of rendered frames.

When onion skin is enabled, the frames at different time positions are merged together tuning their transparency (alpha). The tuning is quite interesting as it normalizes the used alpha values.
This means that the unchanged parts of canvas will look unchanged (sums the same value at different alphas resulting in the same single frame value).

If I have the following 2 frames:



The results with one past onion frame is:
onion

The background and the red intersection of the circles are unchanged thus with full opacity.


Of course the changes I proposed here are breaking this logic and need revision for sure.
I'm afraid if we want advanced onion skin coloring some deeper changes need to be made.

What do you think?

rodolforg:
Hm... The ideal would be rendering only the layer outlines of previous and future frames, not the whole frames themselves.
Like wireframes.

yoSachinkr:
> The big issue is that as of now the whole past and future onion frames are colored and if we have a large area that doesn't change (like a large white rectangle or solid) all of the onion frame get colored and used:

It is a very special case. And If you think about it. A constant colored background can't have an onion frame at all. So it is supposed to get colored completely when the onion frames are on even if it doesn't move.

yoSachinkr:
> here is how it looks when a past and a future keyframe overlap:
> keyframes-overlap

This looks perfect.

yoSachinkr:
Now only the manual opacity controls are missing for the onion skin.

yoSachinkr:
I've tested the changes made in this pull request. 
The observations that I made are as follows:
!. The opacity of past and future frames add up when they are overlapping each other. ( If the opacity of overlapping areas can be made constant with the non-overlapping areas of the onion skin, then it'll look the best)
2. I've noticed as the number of onion skin increases, the onion skin close to the present frame as the least opacity. ( I propose, opacity controls should be made manual.)

yoSachinkr:
> The big issue is that as of now the whole past and future onion frames are colored and if we have a large area that doesn't change (like a large white rectangle or solid) all of the onion frame get colored and used:
> keyframe-overlap-solid
> (in this case a white background becomes all colored with the color that results from past and future onion frame colors overlapping).

@FirasH @rodolforg The solution to this problem is to allow users to set onion skin per layer instead of showing aggregated onion skin. (just like in krita and other animation software)
image

PR #2376, opened 2021-10-19
Remove dead/dummy translations

### LINKED ISSUE #2221: Remove "Hindi (India)" translation [hi_IN]
In Transifex we have "Hindi" [hi] language, but not "Hindi (India)" [hi_IN].

The "Hindi (India)" [hi_IN] translation files in source code are empty. 

So I suggest to remove this language to avoid confusions.

morevnaproject:
We have empty translations for some locales. Those locales has no translators on Transifex, so they will remain empty. Some of those empty translations creating problems for displaying other translations (as with #2221). So I have removed all them.

If someone interested to get specific locale back - please request that on our Transifex page (https://www.transifex.com/morevnaproject/synfig/languages/). ^__^

Closes #2221.


ice0:
Merged. Thank you!

PR #2395, opened 2021-10-31
Clear Redo History on new action by default; new user preference

rodolforg:
It is dangerous to keep Redo stack when user undoes some stuff and
performs a new action: not very action reacts well to changes of
previous status: some Actions are not stateless and stores some
info/values/pointers when their methods `set_param()` or `prepare()`
are called before perform.

App crash can happen!

So let's prevent this redo-stack-keeper by disabling it by default.

A new user preference is created (in System section) for those who
really want it enabled.

RedoFeature

### DIFF
--- a/synfig-studio/src/gui/instance.cpp
+++ b/synfig-studio/src/gui/instance.cpp
@@ -73,6 +73,8 @@
 #include <synfig/widthpoint.h>
 #include <synfig/zstreambuf.h>
 
+#include <synfigapp/main.h>
+
 #include <sys/stat.h>
 
 #endif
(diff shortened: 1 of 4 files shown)


rodolforg:
Waiting #2372 be merged to clean up this PR. :)

rodolforg:
@morevnaproject I know you like this feature, but it is dangerous! XD

synfig-studio/src/gui/instance.cpp
```diff
@@ -102,6 +104,8 @@ Instance::Instance(synfig::Canvas::Handle canvas, synfig::FileSystem::Handle con
 {
 	id_=instance_count_++;
 
+	set_clear_redo_stack_on_new_action(true);
```
ice0, line 107: Is it required here? You will set it at `Instance::create` anyway.

rodolforg: I never know if somebody will write a `new Instance()` in internal code XD

ice0 (approved):

ice0:
Merged. Thank you!

morevnaproject:
> @morevnaproject I know you like this feature, but it is dangerous! XD

@rodolforg I agree that it's better to save users from shooting into his own foot. Thank you for keeping this as an option! ^__^

PR #2402, opened 2021-11-04
Fix distance values on settings

rodolforg:
What of the two commits is best?

Many tools ignore the unit system a preference was saved - that may be different of current App preference when loaded...
So 1 pixel can become the gigantic 1u (60px).

Test it:
1. In `Preferences`, assure the Unit `System` is '`Pixels`'
2. Select `Bline Tool`
3. In `Tool Options Panel`, set `Brush Size` to `1.0 px`
4. Draw any Bline in Work Area
4. Change to `Normal Tool`
5. In `Preferences`, assure the Unit `System` is '`Units`'
6. (optional: close Synfig and reopen it)
6. Select `Bline Tool` again
7. Draw another Bline
8. It's way thicker than before. If you check `Tool Options Panel`, you'll see `Brush Size` as `1.0u`

Mentioned first here:
https://github.com/synfig/synfig/pull/2372#pullrequestreview-783987051

### DIFF
--- a/synfig-studio/src/gui/states/state_bline.cpp
+++ b/synfig-studio/src/gui/states/state_bline.cpp
@@ -328,10 +328,7 @@ StateBLine_Context::load_settings()
 
 		set_opacity(settings.get_value("bline.opacity", 1.0));
 
-		set_bline_width(Distance(
-							settings.get_value("bline.bline_width", 1.0),
-							App::distance_system)
-						);
+		set_bline_width(settings.get_value("bline.bline_width", Distance("1.0px")));
 
 		set_layer_region_flag(settings.get_value("bline.layer_region", true));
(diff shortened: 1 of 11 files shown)


ice0:
> What of the two commits is best?

The second option looks better to me, because it hides additional actions from the user, so the code looks cleaner and does not require to remember to do something important. 

```
bool
Settings::set_value(const synfig::String &key, const synfig::Distance &value)
{
	ChangeLocale(LC_NUMERIC, "C");
	return set_value(key, value.get_string());
}
```

rodolforg:
Squashed and rebased!

ice0 (approved):

ice0:
Merged. Thank you!

PR #2404, opened 2021-11-06
Tab/Shift-Tab support for Widget_Vector

### LINKED ISSUE #1208: Widget_Vector should reacts to TAB key press
When editing value of vector parameters, eg layer `Origin`, user types `x` value first, press `Enter` (to activate/confirm edition) and then focus moves to `Y` value text field.

On the other hand, `TAB` (and `shift`+`TAB`)  is a common navigation key among form fields and this editor widget ignores it.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/89773521-widget_vector-should-reacts-to-tab-key-press?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

veermetri05:
Added support for Widget_Vector to respond to Tab and Shift-Tab as described in https://github.com/synfig/synfig/issues/1208

Close #1208

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_vector.cpp
+++ b/synfig-studio/src/gui/widgets/widget_vector.cpp
@@ -35,12 +35,10 @@
 #include <gui/widgets/widget_vector.h>
 
 #include <gtkmm/spinbutton.h>
-
 #include <gui/app.h>
 #include <gui/widgets/widget_distance.h>
 #include <synfig/general.h>
 #include <synfig/string_helper.h>
-
 #endif
 
 /* === U S I N G =========================================================== */
@@ -118,6 +116,9 @@ void Widget_Vector::init() {
 	distance_y->signal_value_changed().connect(sigc::mem_fun(*this,&studio::Widget_Vector::on_value_changed));
 	//distance_y->signal_activate().connect(sigc::mem_fun(*this,&studio::Widget_Vector::activate));
 	pack_start(*distance_y, Gtk::PACK_EXPAND_WIDGET);
+
+
+	add_events(Gdk::KEY_PRESS_MASK);
 
 	//spinbutton_x->show();
 	//spinbutton_y->show();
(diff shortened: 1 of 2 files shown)


synfig-studio/src/gui/widgets/widget_vector.cpp
```diff
@@ -315,6 +316,29 @@ Widget_Vector::show_all_vfunc()
...
+  if((key_event->keyval == GDK_KEY_ISO_Left_Tab)) {
+	  if(entry_y->is_focus()){
+		  entry_x->grab_focus();
+	  }
+	  else {
+		  Widget_Vector::activate();
+	  }
+  }
+  return Gtk::Box::on_key_press_event(key_event);
+}
```
rodolforg, line 339: You are mixing indentation here.
Please use `Tab` only for alignment in this case.

synfig-studio/src/gui/widgets/widget_vector.cpp
```diff
@@ -315,6 +316,29 @@ Widget_Vector::show_all_vfunc()
...
+{
+  if(key_event->keyval == GDK_KEY_Tab)
+  { 
+    if(entry_x->is_focus()) { 
+		entry_y->grab_focus() ;
+	} 
+	else { 
+		Widget_Vector::activate() ;
+	}
+  }
```
rodolforg, line 329: Add `return true` right before finishing the `if` (same for the `if` below).

We don't want/need to call `Gtk::Box::on_key_press_event(key_event);` if we already handled the key event.

synfig-studio/src/gui/widgets/widget_vector.cpp
```diff
@@ -315,6 +316,29 @@ Widget_Vector::show_all_vfunc()
...
+  if((key_event->keyval == GDK_KEY_ISO_Left_Tab)) {
```
rodolforg, line 330: As it was not so obvious for us, what about add an inline comment after `{`: `// Shift+Tab`?

rodolforg (changes requested):
Thank you and congrats @veermetri05 :)
Please  address my comments here

veermetri05:
I have done the requested changes, is there anything else ?

ice0:
@veermetri05 Looks good, awaiting @rodolforg review.

rodolforg:
@ice0 it's okay for me too.

Do you think he should change to `switch` instead of `if`?

Besides, we can squash and merge :)

veermetri05:
@rodolforg Should I change to `switch` ? 

I don't think it would be good to use `switch` case, for now we are just checking `keyval`, **maybe** we need to use `state` sometimes then `switch` wouldn't be a good choice.

rodolforg:
> **maybe** we need to use `state` sometimes then `switch` wouldn't be a good choice.

You could check `state` inside (or not) the `switch`. But it isn't essential/critical for now.

rodolforg (approved):

ice0:
Merged. Thank you!

PR #2407, opened 2021-11-07
Svg_parser: support more outline/stroke styles

rodolforg:
- linejoin "bevel"
- linecap "square"
- dashes ("dash_array" and "dash_offset")
   * Something is off here, but better than nothing

### DIFF
--- a/synfig-core/src/modules/mod_svg/svg_parser.cpp
+++ b/synfig-core/src/modules/mod_svg/svg_parser.cpp
@@ -40,6 +40,7 @@
 
 #include <cstring>
 
+#include <synfig/valuenodes/valuenode_bline.h>
 #include <synfig/general.h>
 #include <synfig/loadcanvas.h>
 #include <synfig/localization.h>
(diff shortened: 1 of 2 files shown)


synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -505,16 +512,134 @@ Svg_parser::build_outline(xmlpp::Node* root, Style style, const std::list<BLine>
...
+				v.back().set_split_tangent_radius(point.split_radius);
+			}
+
+			path_length = scale_factor * bline_length(v, k.front().loop, nullptr);
+		}
+
+		std::vector<float> dashes;
+
+		float dash_sum = 0.f;
+		for (String token : tokenize(stroke_dasharray, " ,\0x09\0x0a\0x0d"))
```
ice0, line 577: Is that correct? Looks like you want to escape `<tab>, <cr>, <lf>`.
In that case it should be `\x09` without zero.
https://en.cppreference.com/w/cpp/language/escape

rodolforg: Fixed!

ice0 (approved):

ice0:
Merged. Thank you!

PR #2419, opened 2021-11-15
fix dragging area of a waypoint in CellRenderer_Timetrack

rodolforg:
After PR #2384, you could click near the waypoint and drag it,
even though the mouse was not exactly over it.

### DIFF
--- a/synfig-studio/src/gui/cellrenderer/cellrenderer_timetrack.cpp
+++ b/synfig-studio/src/gui/cellrenderer/cellrenderer_timetrack.cpp
@@ -488,9 +488,6 @@ CellRenderer_TimeTrack::activate_vfunc(
 	if (time_plot_data.is_invalid())
 		return false;
 
-	const int waypoint_width = cell_area.get_height();
-	const Time waypoint_width_time = time_plot_data.get_delta_t_from_delta_pixel_coord(waypoint_width);
-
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/cellrenderer/cellrenderer_timetrack.cpp
```diff
@@ -529,9 +529,9 @@ CellRenderer_TimeTrack::activate_vfunc(
 		//Deal with time point selection, but only if they aren't involved in the insanity...
 		Time stime;
 		const Node::time_set *tset = get_times_from_vdesc(value_desc);
-		bool clickfound = tset && get_closest_time(*tset, actual_time*time_dilation + time_offset, waypoint_width_time, stime);
```
ice0, line 532: I suggest moving the variable declaration here and do division in-place: 

```
	const int waypoint_width = cell_area.get_height();
	const Time waypoint_width_time = time_plot_data.get_delta_t_from_delta_pixel_coord(waypoint_width) / 2;
```

rodolforg: Done :)

ice0 (approved):

ice0:
Merged. Thank you!

PR #2425, opened 2021-11-18
Fixed build warnings

ice0:
- warning: 'internal_read' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
- warning: variable 'mag' is uninitialized when passed as a const reference argument here [-Wuninitialized-const-reference]
- warning: loop variable 'layer' creates a copy from type 'const Layer::Handle' (aka 'const handle<synfig::Layer>') [-Wrange-loop-construct]
- warning: equality comparison with extraneous parentheses [-Wparentheses-equality]
- commented unused `blur_iir_coefficients.cpp`

### DIFF
--- a/synfig-core/src/synfig/filesystemnative.h
+++ b/synfig-core/src/synfig/filesystemnative.h
@@ -54,7 +54,7 @@ namespace synfig
 			friend class FileSystemNative;
 			FILE *file_;
 			ReadStream(FileSystem::Handle file_system, FILE *file);
-			virtual size_t internal_read(void *buffer, size_t size);
+			size_t internal_read(void *buffer, size_t size) override;
 			std::istream::pos_type seekpos(std::istream::pos_type pos, std::ios_base::openmode which = std::ios_base::in | std::ios_base::out) override;
 		public:
 			virtual ~ReadStream();
--- a/synfig-core/src/synfig/rendering/software/function/blur.cpp
+++ b/synfig-core/src/synfig/rendering/software/function/blur.cpp
@@ -41,7 +41,7 @@
 
 #include "blurtemplates.h"
 #include "fft.h"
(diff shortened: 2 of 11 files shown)


synfig-core/src/synfig/filesystemnative.h
```diff
@@ -54,7 +54,7 @@ namespace synfig
 			friend class FileSystemNative;
 			FILE *file_;
 			ReadStream(FileSystem::Handle file_system, FILE *file);
-			virtual size_t internal_read(void *buffer, size_t size);
```
rodolforg, line 57: `virtual` and `override` are not mutually exclusive.
Should we keep `virtual`?

ice0: If I understand correctly it's still virtual:
https://en.cppreference.com/w/cpp/language/virtual
```
If some member function vf is declared as virtual in a class Base, 
and some class Derived, which is derived, directly or indirectly, from Base
...
Then this function in the class Derived is also virtual 
(whether or not the keyword virtual is used in its declaration) 
and overrides Base::vf (whether or not the word override is used in its declaration).
```

This code prints `C` (https://godbolt.org/z/Mnf3vPGaa):
```
#include <iostream>

class A {
    virtual void do_f() = 0; // private member
 public:
    void f() { do_f(); } // public interface
};
struct B : public A {
    void do_f() override { std::cout << "B" << std::endl; }; // overrides A::do_f
};
struct C : public B {
    void do_f() override { std::cout << "C" << std::endl; }; // overrides B::do_f
};

 
int main() {
    C c;
    A* a = &c;
    a->f(); // internally calls C::do_f();
}
```
So I think that we can remove this without any consequences.

rodolforg (approved):

PR #2428, opened 2021-11-20
[CMake] Improved search for libraries

ice0:
The main goal here was to get rid of the `include_directories` because they add include directories to all sub targets.


synfig-core/src/modules/mod_noise/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ add_library(mod_noise MODULE
         "${CMAKE_CURRENT_LIST_DIR}/valuenode_random.cpp"
 )
 
-target_link_libraries(mod_noise synfig ${SIGCPP_LIBRARIES})
+target_link_libraries(mod_noise PRIVATE synfig)
```
rodolforg, line 9: Why `PRIVATE` here, but not above with `mod_jpeg`, `mod_magickpp`?

ice0: At first I forgot to fix it, but since I plan to return to this task, I left it as is.

ice0: The main goal here was to get rid of the `include_directories` because they add include directories to all sub targets.

rodolforg: > At first I forgot to fix it, but since I plan to return to this task, I left it as is.

So should I approve?

ice0: Yes.

rodolforg (approved):

synfig-core/src/modules/mod_noise/CMakeLists.txt
rodolforg: Done

PR #2438, opened 2021-11-22
Wider time track space at start and end

### LINKED ISSUE #2417: Wider time track space at start and end
This issue continues #1145 and PR #2400. ^__^

It would be nice to have wider margins at the start at at the end of timetrack.

The PR #2400 sets extra margin = 3px.
And it is easy to make it wider, just by changing one line of code.
I have set it to 8px and it looks nice.

But after changing margin to 8px there is a problem appears: now it allows to drag waypoint from frame 0 to frame -1 or -2.
This can be confusing for user.

screenshot_003

Video - https://youtu.be/lwuFQxbO-LE

Is there a possibility to add check like this? - "if user grabs at frame 0, then forbid dragging in negative direction".
Or maybe other solutions?

morevnaproject:
Closes #2417.

### DIFF
--- a/synfig-studio/src/gui/timeplotdata.cpp
+++ b/synfig-studio/src/gui/timeplotdata.cpp
@@ -53,7 +53,7 @@ namespace studio {
 TimePlotData::TimePlotData(Gtk::Widget* widget, Glib::RefPtr<Gtk::Adjustment> vertical_adjustment) :
 	invalid(true),
 	k(0),
-	extra_margin(3),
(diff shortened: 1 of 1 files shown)


FirasH:
Maybe this should be applied also to `Graphs`?

ice0:
Merged. Thank you!

rodolforg:
> Maybe this should be applied also to `Graphs`?

It affects old/new TimeTrack, Graph & Sound :)

Issue #2403, opened 2021-11-06
Add ability to multiple selection when importing files


Bala621:
i can't import more than one image or file. it is very hard to import multiple files.

if add ability to ctrl+ and shift+ selection. it will be very easy
![importb select](https://user-images.githubusercontent.com/68197434/140599634-a9bc59a2-48b7-4454-8bea-76993a3f430e.jpg)

b-aaz:
There is a workaround (atleast in linux)
Just select your files and drag and drop!

nickleus27:
I am interested in this issue.  Is there any contributing documentation I should read? Or can anyone point me in the right direction as far as files I should be looking at?

ice0:
@nickleus27 Hi!

You can start from here: https://synfig-docs-dev.readthedocs.io/en/latest/

GUI files located in `synfig/synfig-studio` folder.

nickleus27:
@ice0 cool thanks.

nickleus27:
@ice0 Hi how are you today? I think I have found the gtk method for selecting multiple files.  I am new to C++ GUI's and I would need to add it.  After looking through the codebase I am thinking somewhere in synfig/synfig-studio/src/gui/docks/, synfig/synfig-studio/src/gui/trees/, or in the synfig-studio/src/gui/app.cpp file itself?  Would appreciate any advice.

rodolforg:
You can `git grep` to search for `import` :) I think it is on app.cpp. A lot of open dialogs are there.

nickleus27:
@ice0 thank you that is really helpful. right now I am having trouble building from source.  When I run ./2-build-debug.sh after installing all the dependencies I run into problem.  When the build is configuring it exits with error:  You need to install Boost >= 1.53.0.  However, I thought that was installed when I ran the previous ./1-setup-osx-brew.sh?  I am on macos, is there any path that I need to export in my .zshrc file?  I have tried with the path in opt/homebrew/cellar/boost/1.76.0

nickleus27:
@ice0 I fixed my last problem with installing boost from the tar file.  But now the build is exiting with error:

  CXXLD    libmod_openexr.la
ld: warning: ignoring file /usr/local/opt/gettext/lib/libintl.dylib, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
Undefined symbols for architecture arm64:
  "half::_eLut", referenced from:
      half::half(float) in libmod_openexr_la-trgt_openexr.o
[...]

PR #2439, opened 2021-11-22
Added ability to import multiple files

nickleus27:
#2403
@ice0 This pull request allows users of the gui to select and import multiple files at once from the import dialog.  I overloaded the dialog_open_file with this:
`App::dialog_open_file(const std::string &title, std::vector<std::string> &filenames, std::string preference)`
which adds the the parameters that is a string vector.  This PR adds a lot of lines of code since I overloaded the original function.  I tried making it work a bool parameter, but to be honest I was a little confused how to make it work.  I know I can figure out how to make it work if need be to make maintaining code easier.

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -3575,7 +3575,7 @@ CanvasView::import_file()
 	String errors, warnings;
 	if(App::dialog_open_file(_("Please select files"), filenames, IMAGE_DIR_PREFERENCE))
 	{
-		for(std::string filename : filenames){
+		for(const std::string filename : filenames){
 		// Don't let user import a file to itself
 		// Check if it's the same file of this canvas
 		{
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -3570,10 +3570,12 @@ void
 CanvasView::import_file()
 {
 	// String filename(dirname(get_canvas()->get_file_name()));
-	String filename("*.*");
+	//String filename("*.*");
+	std::vector<std::string> filenames;
 	String errors, warnings;
-	if(App::dialog_open_file(_("Please select files"), filename, IMAGE_DIR_PREFERENCE))
+	if(App::dialog_open_file(_("Please select files"), filenames, IMAGE_DIR_PREFERENCE))
 	{
+		for(std::string filename : filenames){
 		// Don't let user import a file to itself
 		// Check if it's the same file of this canvas
 		{
@@ -3624,6 +3626,7 @@ CanvasView::import_file()
 			get_selection_manager()->clear_selected_layers();
 			get_selection_manager()->set_selected_layer(layer);
 		}
+	}
 	}
 }
 
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -3570,8 +3570,8 @@ void
 CanvasView::import_file()
 {
 	// String filename(dirname(get_canvas()->get_file_name()));
-	//String filename("*.*");
 	std::vector<std::string> filenames;
+	filenames.push_back("*.*");
 	String errors, warnings;
 	if(App::dialog_open_file(_("Please select files"), filenames, IMAGE_DIR_PREFERENCE))
 	{
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -3571,6 +3571,7 @@ CanvasView::import_file()
 {
 	// String filename(dirname(get_canvas()->get_file_name()));
 	std::vector<std::string> filenames;
+	LayerTree::LayerList layers;
 	filenames.push_back("*.*");
 	String errors, warnings;
 	if(App::dialog_open_file(_("Please select files"), filenames, IMAGE_DIR_PREFERENCE))
@@ -3623,10 +3624,11 @@ CanvasView::import_file()
 				_("Close"));
 
 		if (layer) {
-			get_selection_manager()->clear_selected_layers();
-			get_selection_manager()->set_selected_layer(layer);
+			layers.push_back(layer);
 		}
 	}
+	get_selection_manager()->clear_selected_layers();
+	get_selection_manager()->set_selected_layers(layers);
 	}
 }
 
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -3575,7 +3575,7 @@ CanvasView::import_file()
 	String errors, warnings;
 	if(App::dialog_open_file(_("Please select files"), filenames, IMAGE_DIR_PREFERENCE))
 	{
-		for(const std::string filename : filenames){
+		for(const std::string &filename : filenames){
 		// Don't let user import a file to itself
 		// Check if it's the same file of this canvas
 		{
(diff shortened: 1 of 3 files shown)


rodolforg:
> #2403 @ice0 This pull request allows users of the gui to select and import multiple files at once from the import dialog. I overloaded the dialog_open_file with this: `App::dialog_open_file(const std::string &title, std::vector<std::string> &filenames, std::string preference)` which adds the the parameters that is a string vector. This PR adds a lot of lines of code since I overloaded the original function. I tried making it work a bool parameter, but to be honest I was a little confused how to make it work. I know I can figure out how to make it work if need be to make maintaining code easier.

Hi, nickleus27! Please (re?)read the second approach proposed by ice0 here: https://github.com/synfig/synfig/issues/2403#issuecomment-975169490
That overload you did that copy'n paste almost everything is really bad for maintaining as ice0 and you already stated. His proposal is very good IMO.

nickleus27:
@ice0 and @rodolforg I didn't understand ice0 example at first reading, but going through it again I see that it is a really great idea.  I appreciate you showing me how to do that.  I think it helped me learn. 💯 I added a commit with ice0's suggestion.

ice0:
Hi, Nick!

> I tried making it work a bool parameter, but to be honest I was a little confused how to make it work.

I made a commit to see changes in this method.
https://github.com/ice0/synfig/commit/14a025317a86ea1ab027cd4f98ff986713480ee8

So let's check them.

1.
```
dialog->set_select_multiple(true);
```

Here you can use the passed parameter (`bool select_multiple`) to enable or disable multiple selection. Something like that:

```
dialog->set_select_multiple(select_multiple);
```

2. 

```
if (filename.empty())
    dialog->set_filename(prev_path);
else if (is_absolute_path(filename))
    dialog->set_filename(filename);
else
    dialog->set_filename(prev_path + ETL_DIRECTORY_SEPARATOR + filename);
```

This part sets the currently selected filename in the dialog.
You can check the documentation here: 
https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1FileChooser.html#a35bbea274ddb0a7b948e6f23b70b9bed

This brings up an additional question. If we allow multiple selection, do we need to open a window with multiple selected files? 
Or just the last selected folder? 


Now this part:
```
for(std::string filename : filenames){
	// info("Saving preference %s = '%s' in App::dialog_open_file()", preference.c_str(), dirname(filename).c_str());
	_preferences.set_value(preference, dirname(filename));
}
```

Take a look at this line:
```
  _preferences.set_value(preference, dirname(filename));
```

This code saves the last opened folder in the preferences so that you can open it next time when you open the dialog box. 
In your case, there is no need to iterate over the files. The directory where the files are located must always be the same. 
Quote from [docs](https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1FileChooser.html#a71bbd79a36b8fb445b6d735345233a27): 
> Lists all the selected files and subfolders in the current folder of chooser.

So you can get that path from the first file.
```
_preferences.set_value(preference, dirname(filenames.front()));
```

P.S. I'm not sure if filenames can be empty after selection, so it's better to check this. 
```
if (!filenames.empty()) {
    _preferences.set_value(preference, dirname(filenames.front()));
}
```

nickleus27:
@ice0 
I made a commit that fixed steps 1 and 3.  However after reading the docs you provided I am still uncertain how to use this code:
> ```
> if (filename.empty())
>     dialog->set_filename(prev_path);
> else if (is_absolute_path(filename))
>     dialog->set_filename(filename);
> else
>     dialog->set_filename(prev_path + ETL_DIRECTORY_SEPARATOR + filename);
> ```

Originally filename was passed in from CanvasView::import_file() as "*.*" so I am uncertain how to go about this. I could replace filename with "*.*" in these lines of code but that wouldn't work for 
> if (filename.empty())

ice0:
> Originally filename was passed in from CanvasView::import_file() as "*.*" so I am uncertain how to go about this. I could replace filename with "*.*" in these lines of code but that wouldn't work for

It is also used in `Widget_Filename::on_button_choose_pressed`. 
In order not to break compatibility, we can use the following logic: 
If the user wants to open one file, the logic is the same as now (we just put the filename as the first element in the vector)
```
App::dialog_open_file(const std::string &title, std::string &filename, std::string preference)
{
	std::vector<std::string> filenames;
        if (!filename.empty()) {
               filenames.push_back()
        }
	if(dialog_open_file_ext(title, filenames, preference, false)) {
		filename = filenames.front();
		return true;
	}
	return false;
}
```

If the user wants to open multiple files, then we we open last opened directory, without selecting any files.
In this case, we can rewrite code like this:
```
if (filenames.empty()) // No filename was passed
    dialog->set_filename(prev_path);
else if (is_absolute_path(filenames.front()))
    dialog->set_filename(filenames.front());
else
    dialog->set_filename(prev_path + ETL_DIRECTORY_SEPARATOR + filenames.front());
```


> From how the code was previously it seems that filename is always passed in as "*.*" from CanvasView::import_file(). Unless I am mis-understanding something?

Yes, but it can be called with different filename mask from `Widget_Filename::on_button_choose_pressed`.

nickleus27:
@ice0 Thank you, That makes sense now.  I failed to recognize ```Widget_Filename::on_button_choose_pressed``` .  I should poke around more next time before asking the question.

nickleus27:
Hi @ice0 and @rodolforg! I think these changes are looking a lot better.  Thank you for all the help along the way.

nickleus27:
@ice0 Hey ice how are you?  As you can probably tell I am a super newb at github.  I tried to git merge and then git push to keep this branch up to date with master.  But after doing so it added other peoples commits it seems.  Did i do this correctly? Thanks for being patient with answering my questions.

ice0:
Hi, Nick!

Please try `git pull --rebase origin master` command to move your changes on top of master branch.

After that you will need to force push your changes (since this rewrites git history).
`git push <remote> --force`

Good documentation on `rebase` here:
https://git-scm.com/docs/git-rebase

nickleus27:
@ice0 thank you ice! Much appreciated!

nickleus27:
@ice0 Thank you that worked great, and this is looking a lot better now.  So git >>>pull --rebase origin master

nickleus27:
I am excited to say I figured out what I did wrong with the previous push. :)

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -3570,10 +3570,12 @@ void
...
 {
 	// String filename(dirname(get_canvas()->get_file_name()));
-	String filename("*.*");
+	std::vector<std::string> filenames;
+	filenames.push_back("*.*");
 	String errors, warnings;
-	if(App::dialog_open_file(_("Please select files"), filename, IMAGE_DIR_PREFERENCE))
+	if(App::dialog_open_file(_("Please select files"), filenames, IMAGE_DIR_PREFERENCE))
 	{
+		for(std::string filename : filenames){
```
ice0, line 3578: Use 
```
for (const std::string& filename : filenames) {
```

this will allow to avoid copying the string.

ice0:
@nickleus27 
Looks good, but I want to suggest one more improvement. 
After importing several files, only one of them is selected in the layer list. I suggest to make all of them selected.

Screenshot_1

nickleus27:
@ice0 can you look over my ``` imported files selected in layers list``` commit to see if looks okay. Thanks.

ice0:
@nickleus27 Looks good to me!

ice0 (approved):

ice0:
Merged. Thank you!

PR #2440, opened 2021-11-22
[Synfig Studio] Extract accels methods and call them when needed

### LINKED ISSUE #2437: [Synfig Studio] Preferences > Restore Defaults doesn't restore default Keyboard Shortcuts
master d312076

**Issue description**:
1) Go to: `Edit` > `Preferences` > `Shortcuts`
2) Edit a shortcut e.g. `redo` to `Ctrl+Shift+G`
3) Restore the default preferences with `Restore Defaults`

The shortcut set in step 2 is not restored.

FirasH:
Fixes #2437


rodolforg (commented):
The two code slices you replaced with the newly created method `restore_default_accels()` are not exactly equal. One changes the treeview, another change the App accel map directly.
I can't recall now why I made them different and I'm too sleepy to think about it.

FirasH:
> The two code slices you replaced with the newly created method `restore_default_accels()` are not exactly equal. One changes the treeview, another change the App accel map directly. I can't recall now why I made them different and I'm too sleepy to think about it.

Yes, that's right.

With the changes proposed the behavior is the following:
When `Restore Defaults` (all preferences) or `Restore default shortcuts` (only shortcuts) are pressed, the settings are restored to default values but not applied using `restore_default_accels()`.

Then:
- if `OK` is pressed, those new settings are applied using `apply_current_accels()`.
- if `Cancel` is pressed, those new settings are discarded.

PR #2424, opened 2021-11-18
[MacOS M1] Fixed autotools build

ice0:
- Autodetect number of threads for compilation on MacOS
- Fixed CPP/LDFLAGS to correctly work with M1 Homebrew
- Fixed boost detection

PR #2442, opened 2021-11-23
 Change #include to not... include... harfbuzz or fribidi folder

rodolforg:
From ice0 comment on PR #2424:

 For example, currently we are using #include <harfbuzz/hb.h> instead of <hb.h>
 but pkg-config returns `harfbuzz/include/harfbuzz`, not just `harfbuzz/include`.
 This works fine on Linux because all system libraries is already included by compiler.
 This need to be fixed. (TODO)

--

From ice0 comment on PR #2442:

 Please remove prefix from `fribidi` too.

```
pkg-config --cflags fribidi
-I/usr/include/fribidi
```

### DIFF
--- a/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
+++ b/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
@@ -46,8 +46,8 @@
 #include <glibmm.h>
 
 #if HAVE_HARFBUZZ
-#include <fribidi/fribidi.h>
-#include <harfbuzz/hb-ft.h>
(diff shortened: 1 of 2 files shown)


ice0:
Please remove prefix from `fribidi` too.

```
pkg-config --cflags fribidi            
-I/usr/include/fribidi
```

rodolforg:
> Please remove prefix from `fribidi` too.

Done

ice0 (approved):

ice0:
Merged. Thank you!

PR #2445, opened 2021-11-24
Fixed crash when using Draw tool (Fedora/Flatpak)

ice0:
This issue is reproduced with Flatpak build and on Fedora.
The reason is that Fedora enables _GLIBCXX_ASSERTIONS  compile flag by default.
https://fedoraproject.org/wiki/Changes/HardeningFlags28

And Synfig crashes when reads off the end of the vector.

So I enabled this flag for CMake debug build too.
This will help us find such problems faster.


rodolforg (approved):

rodolforg (changes requested):
Don't you think you should split into two commits or PR?

ice0:
> Don't you think you should split into two commits or PR? 

Done.

> By the way, is `_GLIBCXX_ASSERTIONS` a GCC extension?

This is `libstc++` define, so it is not compiler dependent. 
https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_macros.html

rodolforg (approved):

PR #2451, opened 2021-12-04
Refactor import file: is_same_file function

nickleus27:
#2403
Made new function in ```canvasview.cpp``` called ```is_same_file()``` This puts the code from ```import_file()``` that checks if file selected is same file into a separate bool function to make ```import_file()``` cleaner.  I am unsure if this new helper function should be a private or public function.  At the moment I have it as public.


ice0 (approved):

ice0:
Merged. Thank you!

Issue #2411, opened 2021-11-09
`Esc` should cancel the in-going edition of a layer parameter value


rodolforg:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I'm typing a new value for a parameter and I gave up on this edition, but I can't cancel it by keyboard.

**Describe the solution you'd like**
By pressing `Esc` I should be able to cancel my in-going edition on the Value column of the Parameters Panel as I'm able to when editing a layer name in the Layers Panel.

nickleus27:
Hi @rodolforg ! Is anyone working on this issue? If not I am interested.

ice0:
Hi @nickleus27!
Assigned it to you.

nickleus27:
@ice0 I am looking in ```layertree.cpp``` with no luck.  Also, the only ```GDK_KEY_Escape``` case is in ```state_bline.cpp``` in the ```event_key_release_handler``` method, and in ```selectdraghelper.h``` in the ```process_key_release_event``` method.  I cannot see how these methods connect with the layer_tree?

ice0:
I think Rodolfo wrote about this panel:

![Screenshot_2](https://user-images.githubusercontent.com/5604544/144701312-f056f046-73b3-46b5-bd93-9382508f903a.png)

it should be located in the `dock_params.cpp` file. 

> Also, the only GDK_KEY_Escape case is in state_bline.cpp in the event_key_release_handler method, and in selectdraghelper.h in the process_key_release_event method. I cannot see how these methods connect with the layer_tree?

I think the `Esc` key is currently not connected to this panel and that is the reason why you cannot find it. 
As for me it's best to start with a simple `gtkmm` app here, just to test how it works out of the box. 
Also you can check the `gtk-demo` application, it should be located in `/opt/homebrew/bin/gtk3-demo`.

nickleus27:
@ice0 awesome, will do.  This will help get me going 👍

nickleus27:
Hi @rodolforg !  ice0 recommended i look in ```dock_params.cpp``` in regards to this issue.  After looking there for a while I ended up in ```cellrenderer_value.cpp``` .  Any advice if I am looking in the right area on this issue?  I see that ```on_editing_done()``` in ```cellrenderer_value.cpp```  is called when the you press enter in the "value" cell of the param table.  But, it does not respond when you press esc.

nickleus27:
[...]

PR #2455, opened 2021-12-09
Added ability for escape key to close  editing box in Value column in the Parameters panel

nickleus27:
#2411

I added changes in ```cellrenderer_value.cpp```.  I added the ```on_key_press_event``` method to check for escape key and call ```on_editing_done()```. I also had to add the ```signal_key_press_event()``` for the value widget.

@ice0 @rodolforg Hey guys can you check out these changes? Thanks again for all the advice to get me started.

### DIFF
--- a/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
+++ b/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
@@ -180,15 +180,8 @@ public:
 		{
 			hide();
 			if (parent) parent->grab_focus();
-			if(!property_editing_canceled())
-			{
-				property_editing_canceled() = true;
-				Gtk::CellEditable::editing_done();
-			}
-			else
-			{
-				synfig::error("property_editing_canceled(): Called twice!");
-			}
+			property_editing_canceled() = true;
+			Gtk::CellEditable::editing_done();
 			return true;
 		}
 		return Gtk::EventBox::on_key_press_event(key_event);
@@ -695,22 +688,22 @@ CellRenderer_ValueBase::start_editing_vfunc(
 void
 CellRenderer_ValueBase::on_value_editing_done()
 {
+	if (value_entry && value_entry->property_editing_canceled())
+		return;
+		
 	if (edit_value_done_called)
 	{
 		synfig::error("on_value_editing_done(): Called twice!");
-		//return;
+//		return;
 	}
 
 	edit_value_done_called = true;
 
 	if (value_entry)
 	{
-		if (!value_entry->property_editing_canceled())
-		{
-			ValueBase value(value_entry->get_value());
+		ValueBase value(value_entry->get_value());
 
-			if (saved_data != value)
-				signal_edited_(value_entry->get_path(), value);
-		}
+		if (saved_data != value)
+			signal_edited_(value_entry->get_path(), value);
 	}
 }
--- a/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
+++ b/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
@@ -181,7 +181,7 @@ public:
 			on_editing_done();
 			return true;
 		}
-		return Gtk::EventBox::on_key_release_event(key_event);
+		return Gtk::EventBox::on_key_press_event(key_event);
 		SYNFIG_EXCEPTION_GUARD_END_BOOL(true)
 	}
 
--- a/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
+++ b/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
@@ -687,8 +687,6 @@ CellRenderer_ValueBase::start_editing_vfunc(
 void
 CellRenderer_ValueBase::on_value_editing_done()
 {
-	std::cout<<"in on_value_editing_done()"<<std::endl;
-	std::cout<<value_entry->property_editing_canceled()<<std::endl;
 	if(value_entry->property_editing_canceled()){
 		CellRenderer::stop_editing(true);
 	}
--- a/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
+++ b/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
@@ -111,7 +111,7 @@ public:
 			synfig::error("on_editing_done(): Called twice!");
 		}
 	}
-	
+
 	void set_parent(Gtk::Widget* x) { parent = x; }
 
 	void on_remove_widget()
--- a/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
+++ b/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
@@ -75,14 +75,12 @@ class studio::ValueBase_Entry : public Gtk::CellEditable, public Gtk::EventBox
 	Widget_ValueBase *valuewidget;
 	Gtk::Widget      *parent;
 	bool              edit_done_called;
-	bool			  edit_canceled_called;
 public:
 	ValueBase_Entry():
 		Glib::ObjectBase(typeid(ValueBase_Entry))
 	{
 		parent           = nullptr;
 		edit_done_called = false;
-		edit_canceled_called = false;
 
 		valuewidget = manage(new class Widget_ValueBase());
 		valuewidget->inside_cellrenderer();
@@ -114,14 +112,16 @@ public:
 		}
 	}
 
+<<<<<<< HEAD
+=======
 	void on_editing_canceled()
 	{
 		hide();
 		if (parent) parent->grab_focus();
 		if (!edit_done_called)
 		{
-			edit_canceled_called = true;
-			//Gtk::CellEditable::on_editing_canceled();
+			edit_done_called = true;
+			Gtk::CellEditable::on_editing_done();
 		}
 		else
 		{
@@ -129,6 +129,7 @@ public:
 		}
 	}
 
+>>>>>>> parent of 17e6817e4 (changed L123 and L124)
 	void set_parent(Gtk::Widget* x) { parent = x; }
 
 	void on_remove_widget()
@@ -195,7 +196,7 @@ public:
 		SYNFIG_EXCEPTION_GUARD_BEGIN()
 		if(key_event->keyval == GDK_KEY_Escape)
 		{
-			on_editing_canceled();
+			on_editing_done();
 			return true;
 		}
 		return Gtk::EventBox::on_key_press_event(key_event);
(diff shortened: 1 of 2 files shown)


ice0:
@nickleus27 
Works like a charm!
Ready to merge, but I want Rodolfo's review of this PR, maybe he has some ideas.

ice0 (approved):

nickleus27:
@ice0 Thank you. Sounds good.

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
```diff
@@ -172,6 +173,18 @@ class studio::ValueBase_Entry : public Gtk::CellEditable, public Gtk::EventBox
...
 
+	bool on_key_press_event(GdkEventKey* key_event)
+	{
+		SYNFIG_EXCEPTION_GUARD_BEGIN()
+		if(key_event->keyval == GDK_KEY_Escape)
+		{
+			on_editing_done();
+			return true;
+		}
+		return Gtk::EventBox::on_key_release_event(key_event);
```
rodolforg, line 184: Why is `release` and not `press` here?

rodolforg (changes requested):
It was a nice trick to simply prevent to call `activate`.

But somehow I think it should set `property_editing_canceled` to true before finish editing.
As I'm not sure about it, let it be :)

nickleus27:
@rodolforg Yes, I think you are right. ```property_editing_canceled``` should be set to ```true```. I will work on that and fix the change you requested.

ice0:
> It was a nice trick to simply prevent to call `activate`.

Where?

nickleus27:
@rodolforg I am having some trouble getting to work with ```bool property_editing_canceled```. I am trying to connect ```on_editing_canceled()``` methods with ```signal_editing_canceled``` unsuccessfully.  Any advice?

rodolforg:
It's a CellRenderer signal. So, CellRenderer_ValueBase should implement it.
In fact, now I see it even has a on_editing_canceled() as a default handler!

nickleus27:
@rodolforg this is the error i am getting ```cellrenderer_value.cpp:699:16: error: no member named 'signal_editing_canceled' in 'studio::ValueBase_Entry'
                value_entry->signal_editing_canceled().connect(sigc::mem_fun(*this, &CellRenderer_ValueBase::on_value_editing_cancelled));``` Not sure what I am doing wrong?

nickleus27:
@ice0 @rodolforg Hey fellows, can you check these changes out when you have a chance. There is a problem with these changes for sure.  When you escape a value entry more than once it produces this output error:

```synfig(11949) [12:22:36] error: on_value_editing_canceled(): Called twice!```

**Comment by rodolforg** [comment]
Maybe we should leave as you did before XD
Simpler and worked. I'm unsure about this property behavior now :P

**Comment by nickleus27** [comment]
@rodolforg No problem, I reverted back to previous solution.  I enjoyed learning about ```on_editing_canceled()```. I wish I could have gotten it to work properly. Nothing I tried seemed to work though.

**Comment by ice0** [comment]
Hi, guys!
I made a screenshot, just to show what is going on here.

First. Cell in `Value` row is `CellRenderer_ValueBase : public Gtk::CellRendererText` ([docs](https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1CellRendererText.html))

It is not editable and is only used to display the cell value. But when you click on it, it emits a `start_editing` signal in which we can create a temporary editable cell ([docs](https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1CellRenderer.html#aa21fc6826b9a3b0cb8a4dc364df7331e)). This signal is not caught in Synfig code, instead of this overridden `start_editing_vfunc` is used here. 

Screenshot_3

Next we creating editable cell `ValueBase_Entry : public Gtk::CellEditable`
https://github.com/synfig/synfig/blob/ec35bd850a913b0c6e99660850cc4a771decd99a/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp#L655-L659

This class is the place where we can catch keypresses and other events and should signal to the parent cell when editing is done or canceled. If editing is canceled then we should set `property_editing_canceled()` to `true`, so parent can check this property and handle the event correctly (it should call `stop_editing(true)` method). After that we should emit `signal_editing_done()`. 

I think the current implementation works as expected only because it just hides the widget without emitting signals. 
For example if you change `on_editing_done()` to `editing_done()` as it should be, then the value will be changed even then you hit the `Esc` key.

To make it work correctly you need to set `property_editing_canceled()` in `ValueBase_Entry` and check this property in `CellRenderer_ValueBase::on_value_editing_done`.

**Comment by nickleus27** [comment]
@ice0 Am I understanding this correctly? I should keep ```on_key_press_event()``` where it is in class ```ValueBase_Entry```?  I added ```property_editing_canceled() = true;``` to ```on_key_press_event()``` and I am checking for it in ```CellRenderer_ValueBase::on_value_editing_done```.  I am having trouble with the ```stop_editing(true)``` and emitting ```signal_editing_done``` correctly. Every way I have tried so far does not work as expected.

**Comment by ice0** [comment]
> @ice0 Am I understanding this correctly? I should keep `on_key_press_event()` where it is in class `ValueBase_Entry`?...

@nickleus27 
Yes, this is correct. Just push your code, so I can check it.

**Comment by nickleus27** [comment]
@ice0 here it is. Thanks for taking a look.  This is not working correctly. The first time you escape value cell ```on_value_editing_done()``` does not get called.  Then and eventually ```"on_value_editing_done(): Called twice!"``` is output from repeatedly "escaping" or "entering" on value cell.

**Comment by ice0** [comment]
The problem is with this line:
```
CellRenderer::signal_editing_canceled().connect(sigc::mem_fun(*this, &CellRenderer_ValueBase::on_value_editing_done));
```

By doing this you connecting CellRenderer_ValueBase to itself. How this works:

1. You hit `Esc` key. It emits `on_key_press_event(...)`.
2. `on_key_press_event` sets `property_editing_canceled() = true` and emits `signal_editing_done` signal;
3. `signal_editing_done` calls `CellRenderer_ValueBase::on_value_editing_done`
4. `CellRenderer_ValueBase::on_value_editing_done` checks `property_editing_canceled()` flag and emits `signal_editing_canceled()`
5. `signal_editing_canceled` calls `CellRenderer_ValueBase::on_value_editing_done` and here you will get `Called twice!` message.

This is the first problem.
The next problem is that this signal (`signal_editing_canceled`) could be triggered by someone else. For example it can be emitted by Gtk::Tree component when it refreshes.

**Comment by nickleus27** [comment]
@ice0 my code doesn't seem to be calling ```on_value_editing_canceled()``` when you press ```esc```? Not sure what is missing.

**Comment by ice0** [comment]
> @ice0 my code doesn't seem to be calling `on_value_editing_done()` when you press `esc`? Not sure what is missing.

Yes. `on_value_editing_done()` is called when `signal_editing_done()` is emitted.
```
value_entry->signal_editing_done().connect(sigc::mem_fun(*this, &CellRenderer_ValueBase::on_value_editing_done));
```

But Esc key handler doesn't emit signal, it's just call `on_editing_done()` method. 
```
if(key_event->keyval == GDK_KEY_Escape)
{
	...
	on_editing_done();
```

To emit `editing_done()` signal you should use `editing_done()` method here([docs](https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1CellEditable.html#a364374949e443a813185aa53f324e681)).

**Comment by nickleus27** [comment]
@ice0 alright I think this is it.  Thank you for being so patient.  I had made this assumption the on_editing_done() should be emitting the signal_editing_done(), but that assumption was wrong.  You had mentioned this a while ago, but I was reluctant to make sense of it.

**Comment by ice0** [comment]
> @ice0 alright I think this is it. Thank you for being so patient. 

You are welcome!

File: synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp [hunk 2]
```diff
@@ -172,6 +173,28 @@ class studio::ValueBase_Entry : public Gtk::CellEditable, public Gtk::EventBox
...
 	}
 
+	bool on_key_press_event(GdkEventKey* key_event)
+	{
+		SYNFIG_EXCEPTION_GUARD_BEGIN()
+		if(key_event->keyval == GDK_KEY_Escape)
+		{
+			hide();
+			if (parent) parent->grab_focus();
+			if(!property_editing_canceled())
```
ice0, line 183: I think better to move this condition back to `CellRenderer_ValueBase::on_value_editing_done`.
In this case, everything will be very simple: 
```
bool on_key_press_event(GdkEventKey* key_event)
{
  if(key_event->keyval == GDK_KEY_Escape)
  {
	  hide();
	  if (parent) parent->grab_focus();
	  property_editing_canceled() = true;
	  Gtk::CellEditable::editing_done();
	  return true;
  }
  return Gtk::EventBox::on_key_press_event(key_event);
}

void CellRenderer_ValueBase::on_value_editing_done() 
{
	if (value_entry && value_entry->property_editing_canceled()) {
		return;
	}
	...
```

Also, there is no need for a second check of `property_editing_canceled()`, because in that case you are already exited.

nickleus27:
@ice0 Yes, good idea! I like the idea of returning from ```on_value_editing_done``` if ```property_editing_canceled``` is true. Much cleaner and simpler.

ice0 (approved):

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
```diff
@@ -172,6 +173,21 @@ class studio::ValueBase_Entry : public Gtk::CellEditable, public Gtk::EventBox
...
 
+	bool on_key_press_event(GdkEventKey* key_event)
+	{
+		SYNFIG_EXCEPTION_GUARD_BEGIN()
+		if(key_event->keyval == GDK_KEY_Escape)
+		{
+			hide();
+			if (parent) parent->grab_focus();
+			property_editing_canceled() = true;
+			Gtk::CellEditable::editing_done();
```
rodolforg, line 184: The only thing I'd propose is to delete this namespace.
In case of an 'overrider' implementation of `editing_done()`, it won't be called :)

rodolforg (dismissed):
In case of @ice0 doesn't think my comment is relevant, I let here my approval :)

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
nickleus27: @rodolforg take out ```Gtk::CellEditable``` namespace?

nickleus27: @rodolforg Alright I removed ```Gtk::CellEditable``` namespace. Let me know if that is what you were thinking.

rodolforg (approved):

rodolforg:
Very nice and clean :)

ice0 (approved):

ice0:
Merged. Thank you!

nickleus27:
@ice0 thank you

PR #2457, opened 2021-12-12
Bone Tool: Correctly apply offset to Second bone when First bone changed

### LINKED ISSUE #1626: [Skeleton Tool] Edit Second bone too when Skeleton Deformation Layer edited

morevnaproject:
Bone Tool: Correctly apply offset to Second bone when First bone of Skeleton Deformation Layer changed.

This issue is described in this comment - https://github.com/synfig/synfig/issues/1626#issuecomment-927001289

Closes #1626.

### DIFF
--- a/synfig-studio/src/synfigapp/actions/valuedescset.cpp
+++ b/synfig-studio/src/synfigapp/actions/valuedescset.cpp
@@ -239,33 +239,10 @@ Action::ValueDescSet::prepare()
 
 					if(index==origin_index) {
 						Point p = value.get(Point());
-						ValueBase parentBone_vb = (*bone->get_link("parent"))(time);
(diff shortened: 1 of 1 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

Issue #2410, opened 2021-11-09
[Windows] Can't import file into a saved SIF animation (if both are on the same disk)


rodolforg:
**Describe the bug**

> When I want to import a png to a sfz file, I get “you cannot import a file onto itself” IF I have already saved the sfz. I can only import them at the start.

(Reported by velizet on forums: https://forums.synfig.org/t/cannot-import-file-onto-itself/12721)

**To Reproduce**
1. Open Synfig
2. Save as a file
3. Menu File > Import
4. Select a PNG file
5. It doesn't accept the import with the error message "You cannot import a file to itself"

**Expected behavior**

**Screencast**
https://youtu.be/WRoxhZccthc
(By Bala)

**System information:**
 - OS: Windows
 - OS version: 11
 - Synfig version: 1.5.1

**Additional comments**
Bug introduced in #2239  and #2243 
Can't reproduce on Linux.

ice0:
@rodolforg
This is a packaging problem. 
First I checked and the problem doesn't reproduced on latest master. 
Then I downloaded the archive from the latest build  https://ci.appveyor.com/api/buildjobs/60kig9u0ohdes0s5/artifacts/SynfigStudio-1.5.0-testing-2021-11-17-win64-27ea6.zip. Problem is not reproduced here too.

After that I replaced all third-party libs (GTK/GIO, etc) in the official release and the problem is gone.
It looks like a bug in `Glib/glibmm` which has already been fixed. 

@morevnaproject Can you take a look?

Issue #1701, opened 2020-09-12
Removing last Waypoint of an Exported Value crashes Synfig Studio


FirasH:
**Synfig version & platform**:
master 88e56d2
openSUSE 15.2

**Issue description**:
1) Create a new animation
2) Draw a `Rectangle` on Canvas
3) Go to the `Parameters` Panel and select `Point 1`, right-click `Export Value`
4) Enable animation editing mode
5) Create a couple of Waypoints for `Point 1` exported value

At this point the exported value has some waypoints

6) Go to the `Library` Panel, under `ValueBase Nodes` remove the exported value Waypoints

After deleting the last Waypoint, Synfig Studio crashes consistently.

The expected behavior should be that the Value becomes constant (the same state when added without any Waypoints).

rodolforg:
I reproduce it with Rectangle (shape) layer, but not with region-type layer.
Backtrace:

```
#6  0x0000555555df62b9 in Gtk::TreeRow::get_value<synfigapp::ValueDesc>(Gtk::TreeModelColumn<synfigapp::ValueDesc> const&) const (this=0x7fffffff9530, column=...) at /usr/include/gtkmm-3.0/gtkmm/treeiter.h:604
#7  0x0000555555df5d50 in Gtk::TreeValueProxy<synfigapp::ValueDesc>::operator synfigapp::ValueDesc() const (this=0x7fffffff9480) at /usr/include/gtkmm-3.0/gtkmm/treeiter.h:558
#8  0x0000555555fe3377 in studio::CanvasTreeStore::get_value_vfunc(Gtk::TreeIter const&, int, Glib::ValueBase&) const (this=0x555558633c00, iter=..., column=3, value=...)
    at /home/rodolfo/projetos/synfig-puro/synfig-studio/src/gui/trees/canvastreestore.cpp:334
#9  0x00007ffff6687b1e in Gtk::TreeModel_Class::get_value_vfunc_callback(_GtkTreeModel*, _GtkTreeIter*, int, _GValue*) () at /usr/lib/x86_64-linux-gnu/libgtkmm-3.0.so.1
```

```
#127 0x000055555618ad98 in studio::CanvasViewUIInterface::error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    (this=0x555558c20e50, err="Remover Ponto de passagem: Empty ID") at /home/rodolfo/projetos/synfig-puro/synfig-studio/src/gui/canvasview.cpp:289
[...]

PR #2462, opened 2021-12-13
Version 1.4.3

morevnaproject:
This is an update to stable branch (critical fixes).


morevnaproject:
Hi! I have made a test Windows build from this branch - [SynfigStudio-1.4.3-testing-2022.04.05-win64-71775.exe](https://cloud.morevnaproject.org/s/NrH2Dx8DFL4dtNT)

...and I found an important issue:
It is not possible to import any file from the same drive.

I.e., if user has synfig's file saved on drive D, then it is not possible to import any file from drive D. 
It shows error: "You cannot import a file to itself".

Here is a demonstration video - https://www.youtube.com/watch?v=0Q2mShLWle4

ice0:
> ...and I found an important issue: It is not possible to import any file from the same drive.
> 
> I.e., if user has synfig's file saved on drive D, then it is not possible to import any file from drive D. It shows error: "You cannot import a file to itself".

This is packaging issue.
https://github.com/synfig/synfig/issues/2410

ice0:
@morevnaproject 

OK. Just checked, and it still looks like a packaging issue.
What I've done:
1) Tried to build `1.4.x` locally. Failed because CMake not working in `1.4`, autotools build fails because MSYS2 removed the `libdl` shared library (only static is available), but autotools depends on the shared library.
2) Downloaded version `1.5.1`. The issues are reproduced here too.
So I build `v1.5.1` locally using CMake. The problem is not reproduced.

So I think the problem is in the outdated `glib/glibmm` libraries used for the build. I only see `Gio::...` methods when checking a file, and they are part of `glib` libs.

morevnaproject:
I have updated glib 2.50.0 -> 2.69.3 and glibmm 2.50.0 -> 2.58.1.
The issue has gone.
Though, I've got other issues with Windows version. Will test them more and report.

Also, I made clean rebuild of everything again and got error in Linux build - #2612

morevnaproject:
> Unfortunately, the fix with the replacement of libraries does not work anymore

@ice0 I wonder why library replacement doesn't work anymore. 🤔 Could it be an SJLJ vs SEH issue? - https://microeducate.tech/what-is-the-difference-between-mingw-seh-and-mingw-sjlj/.

I would prefer to have this possibility to replace libraries from MSYS2. So, if this is an SJLJ vs SEH issue, then I can change compiler in my build environment.

ice0:
> > Unfortunately, the fix with the replacement of libraries does not work anymore
> 
> @ice0 I wonder why library replacement doesn't work anymore. 🤔 Could it be an SJLJ vs SEH issue? - https://microeducate.tech/what-is-the-difference-between-mingw-seh-and-mingw-sjlj/.


Just checked again, and after replacing everything works fine.
I think maybe I forgot to replace the 'lib' and 'share' folders last time.

morevnaproject:
> Just checked again, and after replacing everything works fine.
> I think maybe I forgot to replace the 'lib' and 'share' folders last time.

Good to know! Thanks!

morevnaproject:
Here is the Windows build with new Glib - [synfigstudio-1.4.3-20220419-new-glib.exe](https://cloud.morevnaproject.org/s/a3fbrMGqan3xyFF).

About the bug I have mentioned earlier - looks like it happens in master branch only. Will investigate more soon.

morevnaproject:
So far, the release is blocked only by this issue - https://github.com/synfig/synfig/issues/2608

Also, I still have to review this one - https://github.com/synfig/synfig/pull/2495

ice0:
@morevnaproject 
Can you check if 1.4.3 has this issue #1701? (it is also fixed)

rodolforg:
May I push the changes to be backported (https://github.com/synfig/synfig/projects/12) directly here in this PR?

rodolforg:
On July 29, it's been a year since last stable release :) Will we have a new one for celebration? XD

morevnaproject:
> Can you check if 1.4.3 has this issue #1701? (this has also been fixed)

Yes, 1.4.3 is affected. Fix backported now.

morevnaproject:
> On July 29, it will be a year since last stable release :) Will we have a new one for celebration? XD

Sorry for taking so long... Releasing now. ^__^

morevnaproject:
> May I push the changes to be backported (https://github.com/synfig/synfig/projects/12) directly here in this PR?

I am not sure what would be better approach with 1.4.x branch...
In general, it is okay if other devs will push into it... But when we have release time (like no) this is not desirable.
We should somehow identify when it is "release time" and when it isn't. ^__^

morevnaproject:
> > May I push the changes to be backported (https://github.com/synfig/synfig/projects/12) directly here in this PR?
> 
> I am not sure what would be better approach with 1.4.x branch... In general, it is okay if other devs will push into it... But when we have release time (like no) this is not desirable. We should somehow identify when it is "release time" and when it isn't. ^__^

Ah, I got it. It's better to submit PRs to 1.4.x branch. instead of pushing directly.

This way we have one person, who is responsible for maintaining 1.4.x release (currently that's me) and others can help by preparing backports for the branch (for the cases when commits can't be cherry-picked without conflicts).

PR #2467, opened 2021-12-19
Show bone shape when Skeleton Deformation layer is disabled

### LINKED ISSUE #1625: [Skeleton Tool] Show bone shape when Skeleton Deformation layer is disabled
When Skeleton Deformation layer is in "disabled" mode, then it is hard to see which area is actually covered by bones:

screenshot_003

To make this more intuitive I suggest to improve drawing of bone controls on workarea like this:

screenshot_004

I guess this can be done by adding modifications to `duckmatic.cpp` and/or `src/gui/workarearenderer`.

rodolforg:
fix #1625

### DIFF
--- a/synfig-studio/po/POTFILES.in
+++ b/synfig-studio/po/POTFILES.in
@@ -228,6 +228,8 @@ src/gui/workarearenderer/renderer_background.cpp
 src/gui/workarearenderer/renderer_background.h
 src/gui/workarearenderer/renderer_bbox.cpp
 src/gui/workarearenderer/renderer_bbox.h
+src/gui/workarearenderer/renderer_bonedeformarea.cpp
+src/gui/workarearenderer/renderer_bonedeformarea.h
 src/gui/workarearenderer/renderer_bonesetup.cpp
 src/gui/workarearenderer/renderer_canvas.cpp
 src/gui/workarearenderer/renderer_canvas.h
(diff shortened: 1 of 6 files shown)


morevnaproject:
Woooow!!!

rodolforg:
I force-pushed because I didn't update POFILES.in before :(

morevnaproject:
The feature looks fantastic!

Though, I have found one issue while testing.
Here is a sample file -
[scene014-bug.sif.zip](https://github.com/synfig/synfig/files/7801385/scene014-bug.sif.zip)

Open it in Synfig, you will see one Skeleton Deformation Layer inside. Turn it off and Synfig will crash with an error:
```
synfigstudio: ../../../../src/cairo-arc.c:189: _cairo_arc_in_direction: Assertion `angle_max >= angle_min' failed.
```

Thank you! ^__^

rodolforg:
> Though, I have found one issue while testing.

Fixed! It was the case when a bone root influence area was large enough to circle the whole influence area of its own tip (or vice-versa)

morevnaproject:
Works perfectly now! Thank you so much!

We use Deformation bones more and more recently ^__^ - https://youtu.be/EfUFBN-5U4w?t=7640

morevnaproject (approved):

PR #2458, opened 2021-12-12
ValueDescSet Skeleton Deformation bone: minor refactoring

rodolforg:
Improve code readability.

PR #2470, opened 2021-12-21
fix my mistake in #2458: it should not return at that point

rodolforg:
otherwise rest bone won't be set. Sorry.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2473, opened 2021-12-24
Svg_parser: fix not creating last vertex of cubic path segment

rodolforg:
It happened when it coincides with first vertex - and it didn't
make sense not creating it. It isn't the same thing of closing
path, for example.


ice0:
@rodolforg Can you add before/after image? Just for a better understanding.

rodolforg:
Sure. It fix the blue wall clock here:
v 1.4.2 (and 1.5.1)
Captura de tela de 2021-12-24 13-00-32

And this PR fix the clock (and yellow smile sticker and a plant leaf):
Captura de tela de 2021-12-24 13-02-12

ice0 (approved):

PR #2477, opened 2021-12-25
Fix exported file with audio being smaller than it should

### LINKED ISSUE #2475: Rendering video+audio can be shorter than expected video length
**Describe the bug**
When audio track is shorter than video track, the movie file lasts only as long as audio track.

**To Reproduce**
1. Open Synfig
2. Even in the empty default animation, add a new layer sound with an audio 2 seconds (less than the default 5s/192f)
3. Render with default settings (or explicitly with ffmpeg target)
4. Check output file info or open it on a movie player and check how long it lasts.

If you set a delay, it lasts audio length + delay.

**Expected behavior**
Renderer should render video that lasts as long as set in Canvas Properties.

**System information:**
 - OS: any
 - OS version: any
 - Synfig version: 1.4.2

**Additional comments**
Reported-by: ollieclotheshoffe in https://forums.synfig.org/t/avi-rendering-question/12824
The problem seems to be the `-shortest` passed to ffmpeg.

rodolforg:
Bug: When audio track is shorter than video track, the movie file lasts
only as long as audio track.

fix #2475

### DIFF
--- a/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
+++ b/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
@@ -267,7 +267,8 @@ ffmpeg_trgt::init(ProgressCallback *cb=NULL)
 	// MPEG-1 cannot work with 'le' audio, it requires 'be'
 	vargs.emplace_back(video_codec == "mpeg1video" ? "pcm_s16be" : "pcm_s16le");
 	vargs.emplace_back("-y");
-	vargs.emplace_back("-shortest");
+	vargs.emplace_back("-t");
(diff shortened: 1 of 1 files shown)


rodolforg:
https://www.ffmpeg.org/ffmpeg.html#Main-options

`-t` or `-to` ? o.O

ice0:
> https://www.ffmpeg.org/ffmpeg.html#Main-options
> 
> `-t` or `-to` ? o.O

`-t` specifies the duration, `-to` the end time position.

ice0 (approved):

ice0:
Merged. Thank you!

morevnaproject:
Backported to 1.4.3 ^__^

PR #2487, opened 2022-01-02
make UIInterface::confirmation() implementation consistent to signature

rodolforg:
Base class synfigapp::UIInterface has this method signature:
```
virtual Response confirmation(
			const std::string &message,
			const std::string &details,
			const std::string &confirm,
			const std::string &cancel,
			Response dflt = RESPONSE_OK
) = 0;
```

Somehow the order of button label text parameters was changed for:
- `GlobalUIInterface` (app.cpp)
- `CanvasViewUIInterface` (canvasview.cpp)

And, for that reason, calls from `synfigapp::Action` may have wrong
default responses.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -341,8 +341,8 @@ public:
 	virtual Response confirmation(
 			const std::string &message,
 			const std::string &details,
-			const std::string &cancel,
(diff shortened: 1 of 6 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

PR #2491, opened 2022-01-03
refactor: move non-portable popen code to one place

### LINKED ISSUE #2108: Move non-portable `popen` code to one place (discussion)
We have a non-portable pipe code that starts a process and reads its output using `popen`, `pclose`.
It is used in plugins (to run Python binaries), `mod_ffmpeg` (to run ffmpeg),` mod_dv` (encodedv), and `mod_imagemagick` (convert). 

I suggest moving this code to one place. Maybe to `synfig::FileSystem` as it work with files?

rodolforg:
fix #2108

Known issues (they exist even before this PR)

- ImageMagick importer still doesn't use wchar (argument starting with "png32:) (it already didn't work correctly before)
- Windows package does not provide ImageMagick converter.exe (packaging issue; https://github.com/synfig/synfig/pull/2491#issuecomment-1248258741)
- I believe Windows RunPipe version didn't and doesn't provide exit status :( (I tested once by providing an intentionally malformed ffmpeg call) (it already didn't work correctly before)

### DIFF
--- a/synfig-core/src/modules/mod_ffmpeg/mptr_ffmpeg.cpp
+++ b/synfig-core/src/modules/mod_ffmpeg/mptr_ffmpeg.cpp
@@ -34,15 +34,12 @@
 #endif
 
 #include "mptr_ffmpeg.h"
-#include <cstdio>
 
 #include <ETL/stringf>
 
 #include <synfig/general.h>
 #include <synfig/localization.h>
-#if HAVE_SYS_WAIT_H
- #include <sys/wait.h>
-#endif
+
 #if HAVE_IO_H
  #include <io.h>
 #endif
@@ -59,13 +56,6 @@
 
 using namespace synfig;
 using namespace etl;
-
-#if defined(HAVE_FORK) && defined(HAVE_PIPE) && defined(HAVE_WAITPID)
- #include <unistd.h> // for popen
- #define UNIX_PIPE_TO_PROCESSES
-#else
- #define WIN32_PIPE_TO_PROCESSES
-#endif
 
 /* === G L O B A L S ======================================================= */
 
@@ -87,87 +77,35 @@ ffmpeg_mptr::seek_to(const Time& time)
 {
 	//if(frame<cur_frame || !file)
 	//{
-		if(file)
-		{
-#if defined(WIN32_PIPE_TO_PROCESSES)
-			_pclose(file);
-#elif defined(UNIX_PIPE_TO_PROCESSES)
-			fclose(file);
-			int status;
-			waitpid(pid,&status,0);
-#endif
-		}
+		pipe = nullptr;
 		
 		// FIXME: 24 fps is hardcoded now, but in fact we have to get it from canvas
 		//float position = (frame+1)/24; // ffmpeg didn't work with 0 frame
 		//float position = 1000/24; // ffmpeg didn't work with 0 frame
 		const std::string position = time.get_string(Time::FORMAT_NORMAL);
 
-#if defined(WIN32_PIPE_TO_PROCESSES)
+		OS::RunArgs args;
+		args.push_pair("-ss", position);
+		args.push_pair_filename("-i", identifier.filename);
+		args.push_pair("-vframes", "1");
+		args.push("-an");
+		args.push_pair("-f", "image2pipe");
+		args.push_pair("-vcodec", "ppm");
+		args.push("-");
 
-		std::string command;
-		
+#ifdef _WIN32
 		String binary_path = synfig::get_binary_path("");
 		if (binary_path != "")
 			binary_path = etl::dirname(binary_path)+ETL_DIRECTORY_SEPARATOR;
 		binary_path += "ffmpeg.exe";
-
-		command=strprintf("\"%s\" -ss %s -i \"%s\" -vframes 1 -an -f image2pipe -vcodec ppm -\n", binary_path.c_str(), position.c_str(), identifier.filename.c_str());
-		
-		// This covers the dumb cmd.exe behavior.
-		// See: http://eli.thegreenplace.net/2011/01/28/on-spaces-in-the-paths-of-programs-and-files-on-windows/
-		command = "\"" + command + "\"";
-
-		file=_popen(command.c_str(),POPEN_BINARY_READ_TYPE);
-
-#elif defined(UNIX_PIPE_TO_PROCESSES)
-
-		int p[2];
-
-		if (pipe(p)) {
-			std::cerr<<"Unable to open pipe to ffmpeg (no pipe)"<<std::endl;
-			return false;
-		};
-
-		pid = fork();
-
-		if (pid == -1) {
-			std::cerr<<"Unable to open pipe to ffmpeg (pid == -1)"<<std::endl;
-			return false;
-		}
-
-		if (pid == 0){
-			// Child process
-			// Close pipein, not needed
-			close(p[0]);
-			// Dup pipein to stdout
-			if( dup2( p[1], STDOUT_FILENO ) == -1 ){
-				std::cerr<<"Unable to open pipe to ffmpeg (dup2( p[1], STDOUT_FILENO ) == -1)"<<std::endl;
-				return false;
-			}
-			// Close the unneeded pipein
-			close(p[1]);
-			/*std::string command = strprintf("\"%s\" -ss '%s' -i \"%s\" -vframes 1 -an -f image2pipe -vcodec ppm -\n", "ffmpeg", position2, identifier.filename.c_str());
-			synfig::warning("ffmpeg command: '%s'", command.c_str());*/
-			execlp("ffmpeg", "ffmpeg", "-ss", position.c_str(), "-i", identifier.filename.c_str(), "-vframes", "1","-an", "-f", "image2pipe", "-vcodec", "ppm", "-", (const char*)nullptr);
-			// We should never reach here unless the exec failed
-			std::cerr<<"Unable to open pipe to ffmpeg (exec failed)"<<std::endl;
-			_exit(1);
-		} else {
-			// Parent process
-			// Close pipeout, not needed
-			close(p[1]);
-			// Save pipein to file handle, will read from it later
-			file = fdopen(p[0], "rb");
-		}
-
 #else
-	#error There are no known APIs for creating child processes
+		String binary_path = "ffmpeg";
 #endif
+		pipe = OS::run_async(binary_path, args, OS::RUN_MODE_READ);
 
-		if(!file)
+		if(!pipe)
 		{
-			std::cerr<<"Unable to open pipe to ffmpeg"<<std::endl;
+			synfig::error(_("Unable to open pipe to ffmpeg"));
 			return false;
 		}
 		cur_frame=-1;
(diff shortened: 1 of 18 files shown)


ice0:
@rodolforg 
In my opinion, it's better not to add the `RunArgs` class because it mostly copies `std::vector`.
Instead of that you can extract `push_filename(...)` method and rename it to `escape_filename(...)`. 
As for me, in this case it will be much clearer what is going on.

synfig-core/src/synfig/os.h
```diff
@@ -0,0 +1,101 @@
...
+	RUN_MODE_NONE,
+	RUN_MODE_READ,
+	RUN_MODE_WRITE,
+	RUN_MODE_READWRITE,
+};
+
+struct RunPipe {
+	typedef std::unique_ptr<RunPipe> Handle;
+	virtual ~RunPipe() = default;
+	virtual bool init(std::string binary_path, const RunArgs& binary_args, RunMode mode, const std::string& stdout_redir_file = "", const std::string& stdin_redir_file = "") = 0;
```
ice0, line 70: I think it's better to rename it to `open(...)` to match the `close()` method.

rodolforg: Indeed.

rodolforg:
> Instead of that you can extract `push_filename(...)` method and rename it to `escape_filename(...)`.

Would it be a function directly in `synfig::OS` namespace? 
By the way, what is better: `synfig::OS` or `synfig::os`?

rodolforg:
> > Instead of that you can extract `push_filename(...)` method and rename it to `escape_filename(...)`.
> 
> Would it be a function directly in `synfig::OS` namespace? By the way, what is better: `synfig::OS` or `synfig::os`?

@ice0

rodolforg:
> Instead of that you can extract `push_filename(...)` method and rename it to `escape_filename(...)`.

Not all filenames must be escaped; it depends on what OS we are using, for example. :/

synfig-core/src/synfig/os.cpp
```diff
@@ -0,0 +1,421 @@
...
+	run_args.clear();
+}
+
+std::string OS::RunArgs::get_string() const
+{
+	std::string ret;
+	for (const auto& arg : run_args) {
+		ret.append(arg + " ");
+	}
+	ret.pop_back();
```
ice0, line 143: The behavior is undefined if the string is empty. (https://en.cppreference.com/w/cpp/string/basic_string/pop_back)

synfig-core/src/modules/mod_ffmpeg/mptr_ffmpeg.cpp
```diff
@@ -85,87 +74,35 @@ ffmpeg_mptr::seek_to(const Time& time)
...
-		}
+		pipe = nullptr;
 		
 		// FIXME: 24 fps is hardcoded now, but in fact we have to get it from canvas
 		//float position = (frame+1)/24; // ffmpeg didn't work with 0 frame
 		//float position = 1000/24; // ffmpeg didn't work with 0 frame
 		const std::string position = time.get_string(Time::FORMAT_NORMAL);
 
-#if defined(WIN32_PIPE_TO_PROCESSES)
+		OS::RunArgs args;
```
ice0, line 84: In case if replace it with `std::vector<std::string> args;`:
Before:
```
OS::RunArgs args;
args.push_pair("-ss", position);
args.push_pair_filename("-i", identifier.filename);
args.push_pair("-vframes", "1");
args.push("-an");
args.push_pair("-f", "image2pipe");
args.push_pair("-vcodec", "ppm");
args.push("-");
```

After (variant 1):
```
std::vector<std::string> args;
args.push_back({"-ss", position, "-i", identifier.filename.escape(), "-vframes", "1", "-an", "-f", "image2pipe", "-vcodec", "ppm", "-"});
```

After (variant 2):
```
std::vector<std::string> args;
args.push_back({"-ss", position});
args.push_back({"-i", identifier.filename.escape()});
args.push_back({"-vframes", "1"});
args.push_back("-an");
args.push_back({"-f", "image2pipe"});
args.push_back({"-vcodec", "ppm"});
args.push_back("-");
```

The main difference is that the contributor knows what is going on without checking the `OS::RunArgs` class.

rodolforg: The problem is that Windows needs filename quotes if it has spaces (because it uses a single string for the whole command) but Unix pipes must not have quotes (because it uses command string + arg array):

https://github.com/synfig/synfig/blob/cc59c1247fcdaa1c293bc1fc704922bd9348f7a1/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp#L249-L253

https://github.com/synfig/synfig/blob/cc59c1247fcdaa1c293bc1fc704922bd9348f7a1/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp#L294-L298

https://github.com/synfig/synfig/blob/cc59c1247fcdaa1c293bc1fc704922bd9348f7a1/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp#L300-L306

What about create a method `push(const synfig::filesystem::Path& filename)` instead of `push_filename()` then?

rodolforg:
I can't test on Windows for now. :(

synfig-core/src/modules/mod_ffmpeg/mptr_ffmpeg.cpp
ice0: Agree

ice0:
> I can't test on Windows for now. :(

Sorry for the late answer.
I can test it, but I want to merge #2725 first, in that case I can also test the MSVC build.

synfig-core/src/modules/mod_ffmpeg/mptr_ffmpeg.cpp
ice0: Maybe name it `push_back` (because it adds data to the end of the list)?

rodolforg: OK
I'll fix it after the current code is working.

rodolforg:
Argh, I don't know why the pipe suddenly closes on Windows when exporting via FFmpeg

ice0:
I think better to move step by step, then. This will help us to find where it breaks.
I suggest moving the popen code from ffmpeg module only to the os module first (without changing current behavior).

rodolforg:
I believe now I got it right :)

rodolforg:
> I suggest moving the popen code from ffmpeg module only to the os module first (without changing current behavior).

It indeed works, but not allow stdout/stderr redirection to a file, and it is needed for target_dv.

Interesting enough , using Win32 API is way slower than _popen...

ice0:
> Interesting enough , using Win32 API is way slower than _popen...

O_o

rodolforg:
You can test it XD
Ignore last commit: it's broken. You''ll use Win32 API.
Four commits back ([refactor: ImageMagick importer now uses new synfig::OS::run_sync()](https://github.com/synfig/synfig/pull/2491/commits/c527e9364ed479a49b4d6ad3e6f50941017d8e7c) ) is (still) the _wpopen implementation.

Exporting ffmpeg 1920x1080 is terrible.

rodolforg:
Note:
- with `_wpopen` on Windows: no stdout/stderr/stdin redirection to/from a file. It means `trgt_dv` keeps broken.
- with Win32 API: pretty slow exporting emprty file with 1920x1080 pixels via FFmpeg.

synfig-core/src/modules/mod_ffmpeg/mptr_ffmpeg.cpp
rodolforg: Done

ice0:
> Exporting ffmpeg 1920x1080 is terrible.

How much time does it take?
I tested this PR (Release build) and it takes about 6 seconds, which is basically the same as on Linux.

Windows:
```
ptime.exe synfig.exe c:\projects\synfig\empty.sifz -o out.avi -w 1920 -h 1080 -t ffmpeg
Execution time: 5.980 s
```

Linux:
```
./synfig ~/Downloads/empty.sifz -o out.avi -w 1920 -h 1080 -t ffmpeg  8,15s user 0,98s system 117% cpu 7,744 total
```

rodolforg:
> I tested this PR (Release build) and it takes about 6 seconds, which is basically the same as on Linux.

Are you saying it is properly working? :o  I can make it ready for review?

rodolforg:
Yes, it's working indeed. What did happen here before? :O 

Anyway, I'm making this PR Ready to review then (and finally).

rodolforg:
Hmm
Not solved problems:
- last commit drops support to `scanf()` support. I'm out of ideas for fixing it.
- ImageMagick importer still doesn't use wchar  (argument starting with `"png32:`)
- Windows package doesnot provide ImageMagick `converter.exe`.

synfig-core/src/synfig/os.cpp
```diff
@@ -0,0 +1,632 @@
...
+						exit(-1000);
+					}
+				}
+				// Map pipeout as stdout
+				if (dup2(p[1], STDOUT_FILENO) == -1) {
+					synfig::error(_("Unable to open pipe to %s (dup2( p[1], STDOUT_FILENO ) == -1)"), binary_path.c_str());
+					exit(-2000);
+				}
+				// Map stderr to pipeout as well
+				if (dup2(p[1], STDERR_FILENO) == -1) {
```
rodolforg, line 457: I realized it's not correct to merge stdout and stderr into one single pipe (or redirection file).
FFmpeg always output its "about message" via stderr and it prevents to import a file o.O

Should I change the `open()` signature to add a new parameter?
`bool open(std::string binary_path, const OS::RunArgs& binary_args, OS::RunMode run_mode, const std::string& stdout_redir_file="", const std::string& stdin_redir_file="") override`

it's so loooong

ice0: I think better to do this.
You can check `Glib::spawn_sync` for ideas.

ice0:
> Hmm Not solved problems:
> 
> * last commit drops support to `scanf()` support. I'm out of ideas for fixing it.

I'm not sure if this is really necessary. In the case of `Glib::spawn_sync` user can get the output string from the pipe and use `sscanf` on it if necessary.

> * ImageMagick importer still doesn't use wchar  (argument starting with `"png32:`)

Can be fixed later.

> * Windows package does not provide ImageMagick `converter.exe`

Ok, I'll add it.

> * I believe Windows RunPipe version didn't and doesn't provide exit status :( (I tested once by providing an intentionally malformed ffmpeg call)
> 
> PS: Is this a refactor, right?

Yes.

rodolforg:
> I'm not sure if this is really necessary. In the case of `Glib::spawn_sync` user can get the output string from the pipe and use `sscanf` on it if necessary.

`scanf` is used by `mptr_ffmpeg`. The problem is MS Windows doesn't provide sscanf for Win32 pipes :( At least I couldn't find.

> > * Windows package does not provide ImageMagick `converter.exe`
> 
> Ok, I'll add it.

Thank you. It will provide #2795 feature to MS Windows users.


----

So, the only main issue here is `scanf()` missing for MS Windows (blocks mptr_ffmpeg: **regression**).


1. last commit drops support to scanf() support. I'm out of ideas for fixing it. **(regression)**
2. ImageMagick importer still doesn't use wchar (argument starting with "png32:) (it already didn't work correctly before)
3. Windows package does not provide ImageMagick converter.exe (packaging issue; [ice0 told he'll fix it](https://github.com/synfig/synfig/pull/2491#issuecomment-1248258741))
4. I believe Windows RunPipe version didn't and doesn't provide exit status :( (I tested once by providing an intentionally malformed ffmpeg call)  (it already didn't work correctly before)

rodolforg:
@ice0 do you suggest any API changes here?

ice0:
> @ice0 do you suggest any API changes here?

Hi! I need a time to think about it.

rodolforg:
OK.
Fixed missed scanf() on Windows

ice0 (approved):

ice0:
Merged. Thank you!

PR #2493, opened 2022-01-03
Bone Tool: Fixed issue where skeleton "name" wasn't updated as new Layers are created

### LINKED ISSUE #2082: [Synfig Studio] Skeleton "Name" value is not updated as new Layers are created
**Synfig version**:
master 1ffe5c3

**Issue description**:
1) Create a new animation
2) Select the `Skeleton Tool`
Now in the `Toolbox` window a `Name:` value is assigned automatically like "NewSkeleton014".
3) Create more Layers using the `Create Layer` button

Only the first Layer created uses the value assigned then it reuses the same value for all the next ones.
The tool should update the name value on each time a new Layer is created and keep consistent the used and the displayed value.

rodolforg:
fix #2082


ice0 (approved):

ice0 (approved):

ice0:
Merged. Thank you!

Issue #2212, opened 2021-07-09
[Windows] Synfig doesn't open file in a file server via double-click


rodolforg:
**Describe the bug**
Originally reported by sam.forsey here: https://forums.synfig.org/t/synfig-crashing-randomly-windows-10/12348/16

When a synfig document is in a network shared filesystem, user can't open a synfig file directly from Windows Explorer, but he can do it from open dialog of Synfig.

**To Reproduce**
1. Double-click on file on 'user area'
2. Synfig opens, but reports something like 'unable to load “file location”
Cant find linked file #project.sifz'

**Expected behavior**
Launching Synfig with a filepath argument should open the file, as it is currently possible via open dialog .

**System information:**
 - OS: Windows
 - OS version: 10
 - Synfig version: 1.4.1

PR #2497, opened 2022-01-05
Porting to Gtk::Application - part 1

rodolforg:
Need to test:
- [x] Opening of files named with non-latin characters on MS Windows
- [x] Splash Screen is working
- [x] `--console` is working on MS Windows
- [x] Synfig Studio plugins are working on MS Windows
- [x] `synfig` tool still works


synfig-core/src/synfig/main.cpp
```diff
@@ -202,7 +202,7 @@ synfig::Main::Main(const synfig::String& basepath,ProgressCallback *cb):
 
 	// Paths
 
-	root_path       = etl::dirname(basepath);
```
ice0, line 205: In this case, you also need to fix `synfig-core/src/tool/main.cpp`.
As for now it can't find files:
```
debug/output/Debug/bin/synfig: warning: Unable to find module "lyr_freetype" (file not found)
Unable to load module 'lyr_freetype'
```

ice0 (changes requested):

synfig-studio/src/gui/app.cpp
```diff
@@ -1362,15 +1361,68 @@ App::get_default_accel_map()
...
+		return -1;
+#ifdef _WIN32
+	if (options->contains("console")) {
+		redirectIOToConsole();
+	} else {
+		// QuickHack: to avoid strange bug with stderr
+		freopen("NUL", "w", stdout);
+		freopen("NUL", "w", stderr);
+		freopen("NUL", "r", stdin);
+		ios::sync_with_stdio();
```
ice0, line 1389: `std::ios::sync_with_stdio`

synfig-studio/src/gui/app.cpp
```diff
@@ -1433,14 +1483,15 @@ App::App(const synfig::String& basepath, int *argc, char ***argv):
 	Glib::set_application_name(_("Synfig Studio"));
 
 	Splash splash_screen;
-	splash_screen.show();
+	// Is it safe? I don't know, but that's how I got Splash Screen back
+	splash_screen.show_now();
```
ice0, line 1487: `splash_screen.show();` works fine for me. Both on Linux and Windows. 
Did you have any issues here?

synfig-studio/src/gui/main.cpp
```diff
@@ -68,75 +67,43 @@ using namespace studio;
...
+
+	app->signal_startup().connect([app, rootpath, &is_remote]() {
+		app->init(rootpath);
+		is_remote = false;
+	});
+
+	int exit_code = app->run(argc, argv);
+
+	// app->is_remote() give us a nasty warning
+	if (is_remote) {
```
ice0, line 98: > app->is_remote() give us a nasty warning

Which one? I see only standard message after changing this to `if (app->is_remote())`.

synfig-studio/src/gui/app.cpp
```diff
@@ -1362,15 +1361,68 @@ App::get_default_accel_map()
...
+App::App()
+	: Gtk::Application("org.synfig.SynfigStudio", Gio::APPLICATION_HANDLES_OPEN)
+{
+	add_main_option_entry(Gio::Application::OPTION_TYPE_BOOL, "console", 'c', N_("Opens a console that shows some Synfig Studio output"));
+	signal_handle_local_options().connect(sigc::mem_fun(*this, &App::on_handle_local_options));
+}
+
+int App::on_handle_local_options(const Glib::RefPtr<Glib::VariantDict> &options)
+{
+	if (!options)
```
ice0, line 1379: It's not intuitive why we are returning -1 here, so I think maybe add a comment here? 
```
To continue, return -1 to let the default option processing continue.
https://developer-old.gnome.org/glibmm/unstable/classGio_1_1Application.html#a46b21ab9629b123b7524fe906290d32b
```

synfig-studio/src/gui/app.cpp
```diff
@@ -4355,9 +4382,9 @@ void
 studio::App::process_all_events(long unsigned int us)
 {
 	Glib::usleep(us);
```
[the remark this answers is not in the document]
rodolforg: Thank you. I don't understand anything at this code XD

synfig-core/src/synfig/main.cpp
rodolforg: "Whoops I did it again" :(

synfig-studio/src/gui/app.cpp
rodolforg: OK. But I don't even know why such code exists: the comment left by blackwarthog does not mention what bug it prevents: `stderr`, `stdout`, `stdin` does not even exist on Windows without the `--console` approach.

synfig-studio/src/gui/app.cpp
rodolforg: Yes, no Splash Screen appears.
But maybe it was something memory cached and opened it fast enough to "hide" it

rodolforg: Yes, it's working now o.O And I fought against it for 30min ¬¬

synfig-studio/src/gui/main.cpp
rodolforg: I suppose that `app` is already closed at that line, so `app` is not registered anymore.
It makes sense: once app shuts down, it should de-register itself.

ice0: Can you describe how you are testing this? 
In my case I launch Synfig and try to launch it again from another console.

rodolforg: Same way.

1. Now I edited the code to be like this (for my previous post I did it in another way):
Captura de tela de 2022-01-05 13-56-05
2. Launched it normally.
3. While it is open, I launch it again from another terminal and I got this:
Captura de tela de 2022-01-05 13-57-16
4. Close primary instance and I got this:
Captura de tela de 2022-01-05 13-58-55

synfig-studio/src/gui/app.cpp
```diff
@@ -4354,11 +4381,10 @@ studio::App::setup_changed()
...
 studio::App::process_all_events(long unsigned int us)
 {
-	Glib::usleep(us);
-	while(studio::App::events_pending()) {
-		while(studio::App::events_pending())
-			studio::App::iteration(false);
-		Glib::usleep(us);
+	const auto& ctx = Glib::MainContext::get_default();
+	while(ctx->pending()) {
+		while(ctx->pending())
```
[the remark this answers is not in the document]
rodolforg: Ouch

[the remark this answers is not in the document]
rodolforg: fixed

rodolforg:
@ice0  please read and review last commit

synfig-studio/src/gui/main.cpp
ice0: Ok. This is what I found:
1. `Gio::Application::is_remote()` [docs](https://developer-old.gnome.org/glibmm/unstable/classGio_1_1Application.html#afb0429cbaaaf8bc48f461a90507fc157)
> **bool Gio::Application::is_remote() const**
> If application is remote then it means that another instance of application already exists (the 'primary' instance). Calls to perform actions on application will result in the actions being performed by the primary instance.

> The value of this property cannot be accessed before g_application_register() has been called. See g_application_get_is_registered().

2. `g_application_register()` [docs](https://docs.gtk.org/gio/method.Application.register.html)
> The GApplication::startup signal is emitted if registration succeeds and application is the primary instance (including the non-unique case).

3. https://gitlab.gnome.org/GNOME/glib/-/blob/HEAD/gio/tests/gapplication-example-actions.c

Conclusions:
1. `Gio.Application` registers itself after startup, and a good point for check is the `startup` signal.
2. We can't use `startup` signal, because it is only emitted for the primary instance, so we have to use a different signal or a different approach. For example, the current one.
3. After checking `gapplication-example-actions.c` I made a change that work without error.

```
	Glib::RefPtr<studio::App> app = studio::App::instance();

	app->signal_startup().connect([app, rootpath]() {
		app->init(rootpath);
	});

	app->register_application();
	if (app->is_remote()) {
		std::cout << endl;
		std::cout << "   " << _("synfig studio is already running") << std::endl << std::endl;
		std::cout << "   " << _("the existing process will be used") << std::endl << std::endl;
	}
```

synfig-studio/src/gui/app.cpp
ice0: I suggest removing this code. In any case, it is better to return the "strange" error and fix it.

synfig-studio/src/gui/app.cpp
```diff
@@ -133,6 +133,10 @@
 #include <gui/widgets/widget_enum.h>
 #include <gui/workspacehandler.h>
 
+#ifdef _WIN32
+#include <gui/main_win32.h>
```
ice0, line 137: Move it to line 163

synfig-studio/src/gui/main.cpp
rodolforg: It seems to launch the app once more on quitting (Splash Screen appears and all) before do quit.

ice0: I didn’t notice it. What should I do to reproduce this?

rodolforg: I launched it.
In other terminal, I tried to relaunch it - everything worked fine: it said it is already running, popped up the already open Synfig Studio window and exits.
I close Synfig Studio: splash screen appeared, and said loading all that stuff as if it was starting Studio again, but closes at end.

ice0: What's strange. Is it reproduced if you not trying to relaunch it?
Can you push these changes to a new branch so I can check them?

ice0: Hm, only reproduced on Debian 12. I will take a look.

ice0: I think this is happening because your code looks like this: 
```
	app->signal_startup().connect([app, rootpath]() {
		app->init(rootpath);
	});

	int exit_code = app->run(argc, argv);

	app->register_application();
	if (app->is_remote()) {
		std::cout << endl;
		std::cout << "   " << _("synfig studio is already running") << std::endl << std::endl;
		std::cout << "   " << _("the existing process will be used") << std::endl << std::endl;
	}
```

but it should be:
```
	app->signal_startup().connect([app, rootpath]() {
		app->init(rootpath);
	});

	app->register_application();
	if (app->is_remote()) {
		std::cout << endl;
		std::cout << "   " << _("synfig studio is already running") << std::endl << std::endl;
		std::cout << "   " << _("the existing process will be used") << std::endl << std::endl;
	}

	int exit_code = app->run(argc, argv);
```

rodolforg: Ouch.

Well these are the patches Debian team apply themselves to gtk3 in debian 12 (testing/Bookworm): https://sources.debian.org/patches/gtk+3.0/3.24.31-1/

rodolforg: gtkmm has no patches: https://sources.debian.org/patches/gtkmm3.0/3.24.5-1/

ice0: > Hm, only reproduced on Debian 12. I will take a look.

I mean this is not a Debian issue, it is related to the newer GTK. 
Can you check my assumption about `app->run()` location?

rodolforg: I answered you before Github updates your proposal XD
It works! I made it wrong, sorry :(

rodolforg: Done :)

ice0:
@rodolforg
I think it's ready to be merged.
#2212 is not fixed by this PR, I made another PR to fix this issue: https://github.com/synfig/synfig/pull/2505

rodolforg:
OK :)
I hope Synfig gets more integrated into MacOS too - although I don't use it XD
I read Gtk3 (with Application) has more to do about it.

ice0:
@morevnaproject @rodolforg 

I need to make one more PR (fix windows plugins) before this will be merged.

ice0:
@rodolforg 
I found another way to fix plugins error, so please rebase, and I merge this PR.

rodolforg:
Rebased!

ice0 (approved):

ice0:
Merged. Thank you!

PR #2498, opened 2022-01-05
Add Color "destructive-action" and "suggested-action" to dialog when trying to quit before saving changes

### LINKED ISSUE #2208: Colored action buttons for `Gtk::FileChooserDialog`
Synfig file chooser dialog utilizes header bar while some dialogs utilize standard `Gtk::Dialog` with the title bar. It'll be more useful if action buttons placed in the header bar are colored like normal GTK/GNOME apps to visually guide/assist the user.

GTK provides CSS classes `suggested-action` and `destructive-action` for action buttons **OK/Open/Save** and **Cancel/Close** respectively

-nil-

**Additional comments**
- _**Current dialog: action buttons not colored**_
Screenshot_2021-07-07_08-49-16

- _**Suggested feature: colored action buttons**_
Screenshot_2021-07-07_07-26-30

nickleus27:
fix #2208

@rodolforg Hi Rodolfo, is this the change you are looking for?
This only creates "destructive-action" ```cancel``` button when the user tries to exit without saving.

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -193,6 +193,7 @@ public:
 				const std::string &button1,
 				const std::string &button2,
 				const std::string &button3,
+				bool hasDestructiveAction,
 				Response dflt=RESPONSE_YES )
 	{
 		view->present();
@@ -207,10 +208,12 @@ public:
 		);
 
 		dialog.set_secondary_text(details);
-		dialog.add_button(button1, RESPONSE_NO);
+		Gtk::Button* no_button = dialog.add_button(button1, RESPONSE_NO);
 		dialog.add_button(button2, RESPONSE_CANCEL);
 		dialog.add_button(button3, RESPONSE_YES);
-
+		//add destructive-action colored button if closed without saving
+		if (hasDestructiveAction)
+			no_button->get_style_context()->add_class("destructive-action");
 		dialog.set_default_response(dflt);
 		dialog.show();
 		int response = dialog.run();
@@ -3638,6 +3641,7 @@ CanvasView::import_sequence()
 				_("No"),
 				_("Cancel"),
 				_("Yes"),
+				false,
 				UIInterface::RESPONSE_NO);
 		if(answer!=UIInterface::RESPONSE_CANCEL){
 			canvas_interface()->import_sequence(filenames, errors, warnings, App::resize_imported_images,answer);
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -377,6 +377,7 @@ public:
 				const std::string &button1,
 				const std::string &button2,
 				const std::string &button3,
+				bool hasDestructiveAction,
 				Response dflt=RESPONSE_YES
 	)
 	{
--- a/synfig-studio/src/gui/instance.cpp
+++ b/synfig-studio/src/gui/instance.cpp
@@ -813,6 +813,7 @@ Instance::safe_close()
 						_("Close without Saving"),
 						_("Cancel"),
 						_("Save"),
+						true,
 						synfigapp::UIInterface::RESPONSE_YES
 			);
 
--- a/synfig-studio/src/synfigapp/uimanager.cpp
+++ b/synfig-studio/src/synfigapp/uimanager.cpp
@@ -91,6 +91,7 @@ ConsoleUIInterface::yes_no_cancel(
 			const std::string &/*button1*/,
 			const std::string &/*button2*/,
 			const std::string &/*button3*/,
+			bool hasDestructiveAction,
 			Response dflt
 )
 {
--- a/synfig-studio/src/synfigapp/uimanager.h
+++ b/synfig-studio/src/synfigapp/uimanager.h
@@ -70,6 +70,7 @@ public:
 				const std::string &button1,
 				const std::string &button2,
 				const std::string &button3,
+				bool hasDestructiveAction,
 				Response dflt=RESPONSE_YES
 	) = 0;
 };
@@ -93,6 +94,7 @@ public:
 			const std::string &/*button1*/,
 			const std::string &/*button2*/,
 			const std::string &/*button3*/,
+			bool hasDestructiveAction,
 			Response dflt
 	)
 	{ return dflt; }
@@ -127,6 +129,7 @@ public:
 			const std::string &/*button1*/,
 			const std::string &/*button2*/,
 			const std::string &/*button3*/,
+			bool hasDestructiveAction,
 			Response /*dflt*/
 	)
 	{ return RESPONSE_YES; }
@@ -160,6 +163,7 @@ public:
 			const std::string &button1,
 			const std::string &button2,
 			const std::string &button3,
+			bool hasDestructiveAction,
 			Response dflt
 	);
 


rodolforg:
Another PR?

nickleus27:
@rodolforg Yes. I can close the previous PR if this is what you are looking for instead?

nickleus27:
This change only colors the ```cancel``` button and the ```save``` button when the user tries to exit before saving.

rodolforg:
Hmm. Aren't you coloring any dialog this way?

nickleus27:
Yes.  The code change is in ```canvasview.cpp``` in ```virtual Response yes_no_cancel```.  I only found two places this is called and that is when a user closes without saving, and when a user tries to import duplicate files from ```CanvasView::import_sequence()```.  These both seem like destructive situations, should I find a way to only happen when the user closes without saving?

nickleus27:
Screen Shot 2022-01-06 at 11 40 18 AM
Screen Shot 2022-01-06 at 11 41 05 AM
I think these are the two situations.

nickleus27:
@rodolforg I see what you mean.  It doesn't make sense to color the buttons when importing a sequence.  I pushed changes.  I am comparing strings to check for "Close without Saving".

rodolforg:
Sorry my long delay for a feddback, @nickleus27 
Hmm... I'm starting to realize maybe this issue is not so simple to solve.
1. How can the create-dialog method know what buttons led to a destructive action?
2. Colorize both red and blue buttons in a single dialog make the dialog a circus XD Or suggest an action button or warn the dangerous one.
3. Your solution is good for the dramatic case of close without saving: but it only works for English :) you should surround the string with `_()` macro (defined in localization.h) :)

nickleus27:
@rodolforg thank you for the suggestions.  I have never used ```gettext``` that is defined in ```localization.h```.  I tried reading a little about it, and I am not sure if I used it correctly. Please let me know.

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -207,9 +208,14 @@ class studio::CanvasViewUIInterface : public UIInterface
 		);
 
 		dialog.set_secondary_text(details);
-		dialog.add_button(button1, RESPONSE_NO);
+		Gtk::Button* close_button = dialog.add_button(button1, RESPONSE_NO);
 		dialog.add_button(button2, RESPONSE_CANCEL);
 		dialog.add_button(button3, RESPONSE_YES);
+		//add action buttons if closed without saving
+		if (button1.compare(_("Close without Saving")) == 0)
```
rodolforg, line 215: Any particular reason for using `compare()` instead of `==` directly?

nickleus27: No reason. It seems the == is calling .compare under the hood, but it is easier to read so I changed it to use == directly.

ice0:
@nickleus27 
I think it's better to add the `bool isColored = false` flag to the method arguments and use it.
In this case, we can just pass this flag from anywhere to switch to colored buttons.

nickleus27:
@ice0 Hi Ice, I made the changes you suggested.  The only downside is I had to make more changes in the code because ```yes_no_cancel``` is a virtual function.  Would like to hear any suggestions. Thank you.

rodolforg:
> I think it's better to add the `bool isColored = false` flag

What button should be colored in what way? :/

nickleus27:
@rodolforg I changed '''isColored''' to '''noBtn_hasDestActnColor''' . Let me know if that name is too long.  But it is more descriptive to what you asked.  Also, do you think the button is the right color or should I make the '''save''' button blue instead
Screen Shot 2022-02-05 at 8 21 42 PM
?

nickleus27:
@rodolforg @ice0 Hey guys, I just made a build from the synfig master branch, and all I do is open synfig-studio app and close it and I get a segfault.  I am on a mac with m1 chip. Should I open an issue about this?  Maybe you guys could try for yourself and see if this is true.  Maybe I just made a rookie mistake some how..
Screen Shot 2022-02-05 at 8 42 11 PM
.

rodolforg:
> I changed `isColored` to `noBtn_hasDestActnColor`

What about `noButton_isDestructiveAction` ? XD



> do you think the button is the right color or should I make the '''save''' button blue instead

It's fine that way for me.



> Should I open an issue about this?

Make sure that happens on a clean repo clone. If so, then open it. My guess: the problem came from #2497

nickleus27:
@rodolforg Thanks for the suggestion noButton_isDestructiveAction. I changed it per your suggestion XD

ice0:
Hi guys! Sorry for the late reply.

I was thinking about the name `isDestructiveAction` too. 
The reason I suggested the name `isColored` is just because that's exactly what we do with a dialog - colorize its buttons.
In this case button action is not important, we color the buttons to attract the user's attention, maybe someone decides to color the buttons not for a destructive action, and another contributor looking at the code decides that he is missing some important detail (since the action is destructive).
Anyway that is not critical and this parameter can be renamed at any moment.

nickleus27:
I like having the bool variable describe what button is being colored in what way. I do agree that the idea of color is left out with ```noButton_isDestructiveAction```. That is why originally I named it ```noBtn_hasDestActnColor```

rodolforg:
> maybe someone decides to color the buttons not for a destructive action

Well, we're setting a class to the button: `destructive-action` - and maybe Gtk theme doesn't even implement it (Adwaita does).
The focus itself is not colorize a button with red, but warn it is an important/dangerous action.

It could be colored as blue (in Adwaita theme) as a suggested action.

Oh I lost myself in my arguments... lol English is a barrier for me, but my mind is now confused too XD

ice0:
> Well, we're setting a class to the button: `destructive-action` - and maybe Gtk theme doesn't even implement it (Adwaita does). The focus itself is not colorize a button with red, but warn it is an important/dangerous action.

Hmm. I thought `destructive-action` is our own class.
In this case, my suggestion is to name the parameter `hasDestructiveAction`. 

https://developer.apple.com/documentation/appkit/nsbutton/3622469-hasdestructiveaction?language=objc
https://docs.gtk.org/gtk3/const.STYLE_CLASS_DESTRUCTIVE_ACTION.html

nickleus27:
What about ```noButton_hasDestructiveAction``` since the bool flag sets only the ```no button``` to ```destructive-action``` not any the other buttons.

ice0:
> What about `noButton_hasDestructiveAction` since the bool flag sets only the `no button` to `destructive-action` not any the other buttons.

I think from the user's point of view `hasDestructiveAction` is more clear, especially since we are talking about the whole dialog.

nickleus27:
@rodolforg What do you think? Is ```hasDestructiveAction``` okay with you?

rodolforg:
Well, just "hasDestructiveAction" .... Some future dialog may have "Yes" as the destructive action.
But let it to the possible (not probable) future case ;)

rodolforg:
@nickleus27 please squash all commits into a single one and force-push here!
Thank you for your contribution and sorry for the long "chat" here :)

nickleus27:
@rodolforg and @ice0 thank you. Here is one commit. :D
by the way, is there any issues that could introduce me to another part of the codebase? I have only been working on UI features so far. Thank you for any suggestions.

rodolforg (approved):

rodolforg:
> is there any issues you could recommend that could introduce me to another part of the codebase? I have only been working on UI features so far.

Don't you like to work on UI? :) What part would you like to play with?

nickleus27:
Actually, I do like working on the UI. XD I would like to play with the code that makes the animation work? The src code behind the gui?  Not really sure actually.

ice0:
There is an old idea to make the MacOS menu native, I have not touched it yet, but you can check it out if you want. 
https://www.gtk.org/docs/installations/macos

Also here is a long list of feature requests to improve the GUI (https://github.com/synfig/synfig/issues?q=is%3Aissue+is%3Aopen+label%3A"Feature+request"+gui)

nickleus27:
Thank you : )

ice0:
Merged. Thank you!

rodolforg:
@nickleus27 found any issue to work on? :)

nickleus27:
@rodolforg I am looking forward to my next issue, but I have not found one yet. I have been busy, but I hope to find one soon : )

nickleus27:
> There is an old idea to make the MacOS menu native, I have not touched it yet, but you can check it out if >you want.
>https://www.gtk.org/docs/installations/macos
@ice0 I was looking into the gtk macos integration, and I am unsure if it works with gtkmm?  What I found seemed to be examples for gtk.  If this is the case, I should probably not continue looking into this.

nickleus27:
@rodolforg @ice0 if I were to work on integrating native MacOS menubar, do you guys have an idea where the menubar is setup in synfig code?

ice0:
@nickleus27 

Yes. Here it prepares the menu description and loads it with this code: 

```
	App::ui_manager()->add_ui_from_string(ui_info);
	App::ui_manager()->add_ui_from_string(hidden_ui_info);
```

P.S.
I'm not sure, but maybe we need to update this code first (because `add_ui_from_string` is already deprecated).

nickleus27:
@ice0 
>P.S.
I'm not sure, but maybe we need to update this code first (because add_ui_from_string is already deprecated).

Would the current method be ```Gtk::Builder::add_from_string()``` ?

ice0:
> Would the current method be Gtk::Builder::add_from_string() ?

Yes.

> Converting to Gtk::Builder has been giving me some challenges. For example methods in UIManager with no corresponding methods in Gtk::Builder such as remove_ui.

I'm not sure, but as far as I know, `remove_ui` is not used for the main menu.
It should be used for the popup menu inside the main window. So in your case you can just skip it.
I propose to split this task into two steps:
1) Extract the main menu creation into a new `CreateMainMenubar` method or something like that. (This will allow to see all required dependencies)
2) Switch from UIManager to Gtk::Builder in this method.

nickleus27:
@ice0 Great, that makes sense.  I will approach the problem in the way you described. Thank you.

rodolforg:
> if I were to work on integrating native MacOS menubar, do you guys have an idea where the menubar is setup in synfig code

You could look into `Gtk::Application::set_menubar()`.
https://developer-old.gnome.org/gtkmm/stable/classGtk_1_1Application.html#aef25d09a3867f74fc48c36c65b3b2e7f
https://docs.gtk.org/gtk3/method.Application.set_menubar.html

nickleus27:
@ice0 Before I go too far down this path I wanted to checkin to see if this was the correct way of updating this method.  It looks like I will need to use ```Gtk::Builder::get_widget()``` for the ```insert_action_group``` method.  However, this method takes ```Gio::ActionGroup``` as a parameter.  So do I need to change the ```Gtk::ActionGroup``` variable ```menu_action_groups``` that is declared at the top of the method to a ```Gio::ActionGroup``` type?

rodolforg:
The Gtk made Actions be independent of GUI stuff. They can even be called by DBus, for example. That's why they move them from "Gtk" (the GUI part) to Gio (their I/O base).

The menu is done via Gio::MenuModel. Take a look here:
(C)
https://developer.gnome.org/documentation/tutorials/menus.html
https://developer.gnome.org/documentation/tutorials/actions.html

https://wiki.gnome.org/HowDoI/GMenu
https://wiki.gnome.org/HowDoI/GAction

https://docs.gtk.org/gtk3/getting_started.html#an-application-menu

(C++)
https://developer-old.gnome.org/gtkmm-tutorial/stable/sec-menubar-and-toolbar.html.en

nickleus27:
Sweet thanks @rodolforg ! This is great material. I will spend sometime studying this.

nickleus27:
@ice0 @rodolforg  Hey guys, I am starting to get an understanding of the menu with ```Gtk::Builder``` I am working on the example app found here https://docs.huihoo.com/gtkmm/programming-with-gtkmm-3/3.17.50/sl/sec-menus-examples.html .  
I have kind of a rookie problem I have ran into:  The example says you can use ```add_action``` without any object because it inherits from ```Gtk::Application```.  My problem is I need to set up the actions with a callback but I am running into an error that I have not been able to solve. Here is an example of the code producing error ```add_action("quit", sigc::mem_fun(*this, &App::shutdown_request));``` and the error I get is ```error: invalid use of 'this' outside of a non-static member function```.  This line of code is within my new function replacing ```init_ui_manager()``` The signature is is a static void public member function in the header.

nickleus27:
The menu looks good with the ```Gtk::Builder``` though.  It runs natively in the ```MacOs``` menubar.

nickleus27:
@ice0 From you previous comment I realized about the nature of this being static so I added this line:
```Glib::RefPtr<Gio::SimpleActionGroup> refActionGroup = Gio::SimpleActionGroup::create();```
however when I try the ```sigc::mem_fun``` I am still getting a bunch of errors.  My code looks like this: ```refActionGroup->add_action("quit", sigc::mem_fun(*App::instance().get(), &App::shutdown_request));``` and the error I am getting is ``` error: no matching function for call to 'mem_fun'```. Do you see what I am doing wrong?

**Comment by rodolforg** [comment]
The problem is the method signature: `bool App::shutdown_request(GdkEventAny*)` vs expected signal slot for `Glib::RefPtr<SimpleAction> Gio::ActionMap::add_action(const Glib::ustring & name, const ActivateSlot & slot)`.

`ActivateSlot` is :
```c++
using Gio::ActionMap::ActivateSlot = sigc::slot<void>
```
i.e.: it expects something like `void App::shutdown_request()`. You have to use sigc tricks to bind return and parameter values. Or call `App::quit()` directly.

nickleus27:
@ice0 Thank you for explaining the slot needs to be void.  That makes sense now. However, using ```App::quit``` is still producing the same error for me. ````error`: no matching function for call to 'mem_fun'``` ````note:` candidate template ignored: could not match 'T_obj *' against 'studio::App'``` and ````note:` candidate template ignored: could not match 'T_return (T_obj2::*)()' against 'void (*)()'``` My apologize having to ask so many questions about this.

rodolforg:
Oops. Sorry.
Both `quit()` and `shutdown_request()` are static methods on a static object. So they are actually mere functions. Instead of using `sigc::mem_fun()` (from "**mem**ber_**fun**ction), use `sigc::ptr_fun(&App::quit)` - without the object.

nickleus27:
> Oops. Sorry.
Both quit() and shutdown_request() are static methods on a static object. So they are actually mere functions. Instead of using sigc::mem_fun() (from "member_function), use sigc::ptr_fun(&App::quit) - without the object.

@rodolforg oh geez, that makes so much sense! That worked, and taught me a lot. Thank you!

nickleus27:
Hi @ice0 @rodolforg ,
Here is a pic for reference:
4AD8AAFF-7850-4C5B-B930-01682C964613_4_5005_c

So at the moment I have been hardcoding the icons and accelerators into the ```ui_string```. 
C78D5BD5-8F9D-42A5-819F-87C93C7ABB4D_4_5005_c

I have not found any methods for adding them in ```Gtk::Builder``` I have been getting the stock icon strings from this reference: https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html . Are these the icons I should be using, or are there some custom synfig icons that I should be using? Also, I have been hardcoding the accelerators into the ```ui_string```.  Is there another way I should be doing this?

nickleus27:
I am struggling to add the ```save``` action.  It seems the save method is ```Instance::Status
studio::Instance::save()``` in ```synfig-studio/src/gui/instance.cpp``` this is a member function and I cannot seem to find the object in ```app.cpp```. I have tried using ```studio::App::get_selected_instance().get()``` but that hasn't worked.  Also, how do you use sigc::bind to alter functions without void return signature? I have figured out how to use bind for adding arguments...

ice0:
> Hi @ice0 @rodolforg , Here is a pic for reference: 4AD8AAFF-7850-4C5B-B930-01682C964613_4_5005_c

Looks awesome!

> So at the moment I have been hardcoding the icons and accelerators into the `ui_string`. 
> I have not found any methods for adding them in `Gtk::Builder` I have been getting the stock icon strings from this reference: https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html . Are these the icons I should be using, or are there some custom synfig icons that I should be using? I see from the link that Rodolfo gave me, https://wiki.gnome.org/HowDoI/GMenu , that you can add a path to a .png in the xml string. 

Some icons in the menu are standard icons, some are custom icons (for example, the icons near `Preview...` or `Render...` menu items). So you need to use both. 

> Also, I have been hardcoding the accelerators into the `ui_string`. Is there another way I should be doing this?

I'm not sure if these accelerators can be changed, so it's ok to hard-code them until the problem is found.

ice0:
> I am struggling to add the `save` action. It seems the save method is `Instance::Status studio::Instance::save()` in `synfig-studio/src/gui/instance.cpp` this is a member function and I cannot seem to find the object in `app.cpp`. I have tried using `studio::App::get_selected_instance().get()` but that hasn't worked.

`studio::App::get_selected_instance()->save()` should do the job.

Also I wanted to note that I suggested using a static method to ease porting.
If it makes the task more difficult, then just remove the static modifier from the method. 

> Also, how do you use sigc::bind to alter functions without void return signature? I have figured out how to use bind for adding arguments...

If you want to alter return type, you can use [sigc::bind_return](https://developer-old.gnome.org/libsigc%2B%2B/2.2/group__bind.html).

If not, then can you describe your task in more detail?

nickleus27:
>Also I wanted to note that I suggested using a static method to ease porting.
If it makes the task more difficult, then just remove the static modifier from the method.

@ice0 sorry, suggested what making what method static?

ice0:
> @ice0 sorry, I am not sure what method you suggested making static? 

I meant this:
> Extract the main menu creation into a new CreateMainMenubar method or something like that. (This will allow to see all required dependencies)

> If you mean the method I am creating I am not adding function to `app.h` just like `void init_ui_manager()` since it only is called within `app.cpp`

Ah, ok.

nickleus27:
@iceo Sorry for the bombardment of questions at this moment.  Here is a screenshot of where I am having trouble adding the ```save action```. The line of code is under the comment ```//NEED HELP...```
E11F2BB6-976D-45AD-BB58-C96A708FFC9F

ice0:
1. A small improvement to your code: 
```
App::instance()->add_action("open", sigc::ptr_fun(&App::dialog_open));
```
This is the same as with `get()`, but shorter. 

2. You want to call `Instance::save()`. It's not a static method, so you need to pass the owner of the object first. This will be:
```
sigc::mem_fun(App::get_selected_instance().get(), &Instance::save);
```
`.get()` is used here because `get_selected_instance()` returns `etl::loose_handle` and sigc doesn't know how to deal with it.

Next you need to discard the return value because `add_action` expects a `void` to be returned. You can use `sigc::hide_return` for this. 
```
sigc::hide_return(sigc::mem_fun(App::get_selected_instance().get(), &Instance::save));
```

So the final code should looks like:

```
App::instance()->add_action("save", sigc::hide_return(sigc::mem_fun(App::get_selected_instance().get(), &Instance::save)));
```

nickleus27:
@ice0 
>So the final code should looks like:
App::instance()->add_action("save", sigc::hide_return(sigc::mem_fun(App::get_selected_instance().get(), &Instance::save)));

Ah yes, very nice.

rodolforg:
> I have been getting the stock icon strings from this reference: https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html

I suggest you to open a draft PR:
1. to share with us your current code
2. to have this discussion there :)

I hope we can use the icons without full icon path o.O by pointing them to icon themes (that currently Synfig doesn't use yet. Some draft work was done here in #1508. Feel free to implement/correct/improve/discard it :)



> Also, I have been hardcoding the accelerators into the `ui_string`. Is there another way I should be doing this?

I suppose with Gio::ActionMap, similar as we do now. However, we still use the deprecated Gtk::UIManager, and now it should be done via Gtk::Application (that heirs Gio::ActionMap, by the way).



> `sigc::mem_fun(App::get_selected_instance().get(), &Instance::save);`

I think it's missing a deference here ;)

ice0:
Btw, it is recommended to use `sigc::mem_fun` instead of a lambda to be able to automatically disable the signal from the widget. 

> https://developer-old.gnome.org/libsigc++/stable/group__mem__fun.html
> If the object type inherits from [sigc::trackable](https://developer-old.gnome.org/libsigc++/stable/structsigc_1_1trackable.html), and the functor returned from [mem_fun()](https://developer-old.gnome.org/libsigc++/stable/group__mem__fun.html#ga5d28dcf1bb88830a6e50d08d1fe7ec08) is assigned to a sigc::slot, the functor will be automatically cleared when the object goes out of scope. Invoking that slot will then have no effect and will not try to use the destroyed instance.

But in the case of an application (`App`), this is not required, so you can rewrite the code like this: 
```
App::instance()->add_action("save", []() {
	App::get_selected_instance()->save();
});
```

ice0:
> > `sigc::mem_fun(App::get_selected_instance().get(), &Instance::save);`
> 
> I think it's missing a deference here ;)

`sigc::mem_fun` allows both pointer or reference argument.

PR #2500, opened 2022-01-05
fix crash when using Width Tool on Skeleton Layer

### LINKED ISSUE #2381: Crash when using Width Tool on Skeleton Layer
**Describe the bug**
When user the Width tool on a Skeleton layer, Synfig crashes.

**To Reproduce**
1. With Bone Tool, create a skeleton layer with two bones at least
2. Assure the Skeleton Layer is selected
3. Swap to Width Tool
3. Click on a bone on workarea
4. It crashes

**Expected behavior**
If nothing is supposed to happen, at least don't crash.

**System information:**
 - OS: Linux - Debian 12 (testing)
 - OS version: Debian 12
 - Commit ID: 1025c0108b0704cfe0e18420c70dc65b9a71b3bb (via `git bisect`)

**Additional comments**
It seems the cause is the use of bezier to show bone parenting. It is a good idea, but Width Tool (and probably other tools and actions) don't check something about these fake bezier ducks (they doesn't represent a spline) and then raise the problem.

rodolforg:
fix #2381

### DIFF
--- a/synfig-studio/src/gui/states/state_width.cpp
+++ b/synfig-studio/src/gui/states/state_width.cpp
@@ -445,7 +445,7 @@ StateWidth_Context::AdjustWidth(handle<Duckmatic::Bezier> c, float t, Real mult,
 	synfig::ValueNode::Handle p1pvn(p1->get_value_desc().get_parent_value_node());
 	synfig::ValueNode::Handle p2pvn(p2->get_value_desc().get_parent_value_node());
 	// if the bezier position ducks are linkable valuenode children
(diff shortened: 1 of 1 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

rodolforg:
Backport?
https://forums.synfig.org/t/dancing-skeleton-crashes-everytime-i-use-width-tool/12872

morevnaproject:
Backported now! Thank you!

PR #2503, opened 2022-01-08
reuse existent method for traveling layer parameters

nattymbt:


rodolforg:
Nice! But we should avoid lambda functions when it needs to capture stuff.
I'm not an expert on it, so I tried to avoid them all: #2294

Issue #1714, opened 2020-09-29
Create child bone menu option missing for skeleton layer


Keyikedalube:
**Synfig version & platform**:
Building from source
Debian sid

**Issue description**:
1. Add a new skeleton layer
2. Right-click on any points to bring up its context menu

**Create child bone** menu option is missing. Is this done with intention?

The solution I use is add more skeletons and link them to a parent bone from the properties toolbox. But that's making my layers look ugly and I have to track which bone goes to which layer object. I like the old style where a user adds only one skeleton and then add multiple child bones with ease using the right-click style.

ice0:
Can you check this with 1.2 release?

Keyikedalube:
I think 1.2 is ok... U guys made the **How to create animation in Synfig (3rd edition)** video course with 1.2... right?
*I don't think 1.2 is affected*

Plus I can't downgrade to older version now. All synfig files are created from latest master branch. Cutting-edge commits LOL

rodolforg:
You can create skeleton bones in master via Skeleton tool!
I don't know if it was removed intentionally though.

rodolforg:
It was: https://github.com/synfig/synfig/pull/1663

Keyikedalube:
That PR intentionally removed child bone menu option.

The benefits??? I'm confused... Was it done to fix something that was already broken? Because the linked issue only mentioned about the need for removing the said menu after reworking skeleton layer. Looks like users must always use Layers panel for adding new child bones now.

Creating child bone using right-click context menu option was faster :(

Now we have to:
1. Manually add skeleton layer
2. Connect to parent bone
3. Maybe even group the skeleton and layer object together to hierarchically organize our layers to avoid clutter

morevnaproject:
> That PR intentionally removed child bone menu option.

Yes. Now you can use Skeleton tool to add new bone - https://youtu.be/8qawMmSN55o

Documentation - https://synfig.readthedocs.io/en/latest/tools/skeleton.html

Keyikedalube:
New UX :)

I didn't realize I can click around to directly add more child bones!!!

Closing this issue. Sorry for the misunderstanding

PR #2507, opened 2022-01-15
Re-enable "Create Child Bone" action

morevnaproject:
Got a feedback from several Synfig users, who are on latest development snapshot of Synfig:
Many of them got stuck on tutorials because of missing "Create Child Bones" action.
The issue https://github.com/synfig/synfig/issues/1714 confirms that too.

So, it looks like I was wrong when [suggested to disable this](https://github.com/synfig/synfig/issues/1558). ^__^

Let's re-enable this and wait while most tutorials will be updated and people will get used to Skeleton Tool.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2512, opened 2022-01-16
Warn user about linking layer to a skeleton bone from a different canvas

rodolforg:
They must be on the same group/hierarchy level.


rodolforg:
@morevnaproject I realized I made one commit with 4 different effects.
I now split them in 4 different commits.
(rebased it to master branch too)

rodolforg:
Maybe PR title should change XD

morevnaproject:
I have made a quick test of the PR and found two issues.

1. It shows warning even if Skeleton Layer is in the same group with linked layer. See this video from 0:18 -
https://www.youtube.com/watch?v=JhDRO-UThlU

2. I think the warning shouldn't be displayed IF (linked layer is inside of group(s) AND all those groups have default transformation).
By "default transformation" I mean **position=(0;0)** & **angle=0** & **scale=(1;1)**.
In this situation the warning isn't necessary, I think.

rodolforg:
> 1. It shows warning even if Skeleton Layer is in the same group with linked layer. See this video from 0:18 -

Thank you. I only tested with bone in the root canvas :P I'll fix it later.



> 2\. I think the warning shouldn't be displayed IF (linked layer is inside of group(s) AND all those groups have default transformation).

Wouldn't the problem happen if user decides to manually move the parent group(s) later after linking?

morevnaproject:
> Wouldn't the problem happen if user decides to manually move the parent group(s) later after linking?

This is a good question! Let me think...

Well, let's imagine user have made a skeleton outside of untransformed group.
He has attached all parts (from untransformed group) to the skeleton.

After that he transformed the group.

Now he wishes to change skeleton's pose.

So, he selects the skeleton.
And he see that the bones of skeleton are not aligned to the parts in the transformed group.
So, I assume the flaw is visible to user in that case and user can figure out why that happened.

So, I still suggest to apply point 2 from my previous comment.
Thank you!

PR #2519, opened 2022-01-23
fix "origin" for Layer_PasteCanvas - it was not intuitive (reversed!)

nattymbt:


nattymbt:
It does not fix previously saved files, though. I don't have any idea about it.

ice0:
@nattymbt 
Hi! Can you describe the original issue?

nattymbt:
I believe the actual position of an element/layer in a `Group Layer` should be `Group origin` + `Group offset` + `Element origin`.

Currently, we should displace the `Group origin` in 'negative direction' to actually do it.

1. Create a circle (e.g. Origin: (0,0) Radius: 60px)
2. Encapsulate the circle in a Group Layer
3. Set the Group origin to (10px, 0)

Instead of the actual circle center changes to (10px, 0), it goes to (-10px, 0).

If, instead of set 3, I had directly set the Circle origin to (10px, 0), the Circle center would go to (10px, 0) as expected.

ice0:
@morevnaproject @rodolforg any ideas here?

rodolforg:
Hm... I got the problem: indeed it is a bit weird.

We need to check first if this change does not cause any artifact in Studio (layer hit-check, layer rectangle boundaries, ...).



> It does not fix previously saved files, though. I don't have any idea about it.

Almost every compatibility stuff is in `synfig/loadcanvas.cpp`. It could be done in Layer_Group itself if you check its version when setting parameters (initially loadcanvas.cpp sets the .sfz-file layer version and , at the end, it resets to current version).

PR #2522, opened 2022-01-26
Fixed PVS-Studio warning v730

ice0:
V730. Not all members of a class are initialized inside the constructor.
https://pvs-studio.com/en/docs/warnings/v730/

### DIFF
--- a/ETL/ETL/_bezier.h
+++ b/ETL/ETL/_bezier.h
@@ -243,7 +243,7 @@ private:
 	value_type _coeff[4];
 	time_type drs; // reciprocal of (s-r)
 public:
-	bezier_base():r(0.0),s(1.0),drs(1.0) { }
+	bezier_base(): a{0.f}, b{0.f}, c{0.f}, d{0.f}, r(0.0),s(1.0),drs(1.0) { }
 	bezier_base(
 		const value_type &a, const value_type &b, const value_type &c, const value_type &d,
 		const time_type &r=0.0, const time_type &s=1.0):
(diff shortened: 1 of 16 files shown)


synfig-core/src/synfig/layers/layer_shape.h
```diff
@@ -73,7 +73,7 @@ class Layer_Shape : public Layer_Composite, public Layer_NoDeform
 	Vector feather;
 
 	mutable Time last_sync_time;
-	mutable Real last_sync_outline_grow;
+	mutable Real last_sync_outline_grow = 0.l;
```
rodolforg, line 76: Is this not zero to force the first `Layer::sync()`?

ETL/ETL/_bezier.h
```diff
@@ -323,7 +323,7 @@ class bezier_base<double,float> : public std::unary_function<float,double>
 	value_type _coeff[4];
 	time_type drs; // reciprocal of (s-r)
 public:
-	bezier_base():r(0.0),s(1.0),drs(1.0) { }
+	bezier_base(): a{0.l}, b{0.l}, c{0.l}, d{0.l}, r(0.0),s(1.0),drs(1.0) { }
```
rodolforg, line 326: Why aren't these `0.f` as previous case?

ice0: Just to show they are double. Your question confirms that this is not obvious.

Previous class:
> typedef float value_type;

Current class:
> typedef double value_type;

synfig-core/src/synfig/layers/layer_shape.h
ice0: > Is this not zero to force the first `Layer::sync()`?

```
Layer_Shape::sync(bool force) const
{
	if ( force
	  || !last_sync_time.is_equal(get_time_mark())
	  || fabs(last_sync_outline_grow - get_outline_grow_mark()) > 1e-8 )
```

I don't know. My idea is to initialize it with zero and fix if it breaks something.
At least this will be consistent result.

ETL/ETL/_bezier.h
rodolforg: I supposed 1.0f that means is float :)

PR #2524, opened 2022-01-27
Fixed PVS-Studio warning v656

ice0:
V656. Variables are initialized through the call to the same function.
It's probably an error or un-optimized code.

https://pvs-studio.com/en/docs/warnings/v656/


rodolforg (approved):
WOW!

PR #2555, opened 2022-03-01
only show widget for Brush settings if Brush Tool is enabled

rodolforg:


ice0 (approved):

rodolforg:
Maybe backport it for better UX

ice0:
> Maybe backport it for better UX

@morevnaproject said it needs to check if this affects other tools.

rodolforg:
@ice0 @morevnaproject Originally yes: BrushWidget sets `synfigapp::Main::set_bline_width()`.
https://github.com/synfig/synfig/blob/cc59c1247fcdaa1c293bc1fc704922bd9348f7a1/synfig-studio/src/gui/widgets/widget_defaults.cpp#L149-L167

However... "All" tools does not read this value anymore, each one has its own "Line Width" in its Tool Options, saved to and restored from settings.

https://github.com/synfig/synfig/blob/cc59c1247fcdaa1c293bc1fc704922bd9348f7a1/synfig-studio/src/gui/states/state_bline.cpp#L317-L327

"All" tools except Brush (AFAIK).

morevnaproject:
Originally, the width defined in brush widget should affect width of all draw tools. In older versions it behaves like this.
I guess this got broken at some point.
I suggest to restore this functionality. ^__^

rodolforg:
> I suggest to restore this functionality. ^__^

How it is supposed to work? Now every draw tool has it Brush Size saved and restored from Preferences file.
image

PR #2559, opened 2022-03-03
add more synfig::Node tests

rodolforg:
`deleting_node_removes_it_as_parent_from_its_children()` keeps failing

I don't think if it is easy to solve without storing not only parents
nodes, but children also.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2562, opened 2022-03-07
refactor: port more layers to Cobra renderer: some Gradient layers

rodolforg:


rodolforg:
Could you guys test them? I make the (temporary) last commit to be able to compare the original version vs. Cobra version.
@FirasH @morevnaproject

Keyikedalube:
test gradients

So the port is successful I guess?

FirasH:
I've tried to export to `.png` format the 4 gradient types and compare them to current render engine, didn't notice any difference. If there are some they're imperceptible to me.

morevnaproject:
I also tested this PR and it seems everything works good. ^__^

But, we have regression testing suite, which is invoked by calling `./autobuild/synfigrenderer-regression-test.sh` script.
The set of tests is stored here - https://gitlab.com/synfig/synfig-tests/-/tree/master/rendering/sources/layers (there you can find tests for gradients).
It would be really nice to run those tests and see if the tests are passing.

Also, I have noticed one more thing: previously we had a Travis CI job, which was running tests for regression automatically. Now Travis has gone and tests are not automatically started anymore. It would be nice to configure CI job again (using GitHub Actions). ^__^

P.S. Sorry for bringing this topic here, but I see this PR as a great opportunity to resurrect our regression testing, which will make easier merging any future changes for rendering engine. ^__^

ice0:
Just tested with regression testing package.
We have regressions in the current master.
I'm not sure should we fix all the regressions in master first? 

**gradient-linear-black-white-loop test**

**Original**

gradient-linear-black-white-loop-original

**Master branch**

gradient-linear-black-white-loop-master

**Animated diff**

animated

rodolforg:
> We have regressions in the current master.
> I'm not sure should we fix all the regressions in master first?

With this PR we go back to original rendering or get some different render result?

rodolforg:
> With this PR we go back to original rendering or get some different render result?

If I didn't test incorrectly, this port makes your example back to original rendering.

ice0:
> > With this PR we go back to original rendering or get some different render result?
> 
> If I didn't test incorrectly, this port makes your example back to original rendering.

The animated image shows a slight change in the coordinates of the gradient layer, but I need time to check with the 1.4.3 AppImage.
Another approach is to merge it as is and fix the tests after that.

rodolforg:
The image I rendered with this PR (with Cobra enabled in the layers) is exactly equals to the reference image the Gitlab repository provides. :)
 Unless I did something wrong :oP

rodolforg:
(rebased)

rodolforg:
(rebased)

morevnaproject:
> The animated image shows a slight change in the coordinates of the gradient layer

This slight change is 1 px offset. This offset is produced by every layer, which isn't ported to Cobra (and yes, this is a bug with engine, not related with this PR).

morevnaproject:
I have tested this PR and render results are the same as in Synfig 1.0.2.
Let's remove last commit and it is ready for merge!

rodolforg:
rebased and removed the temporary last commit :)

ice0:
@rodolforg 
Please rebase :)

ice0 (approved):

rodolforg:
> Please rebase :)

@ice0 rebased :)

ice0:
Merged! Thank you!

Issue #2036, opened 2021-02-19
[Feature request] Change mouse cursor while repositioning canvas


Keyikedalube:
**Synfig version & platform**:
Building from source
Fedora 33

**Issue description**:

https://user-images.githubusercontent.com/24214509/108483373-aad05780-72c0-11eb-8f08-c02b9b830832.mov

Just like on Inkscape, I'd love Synfig to change its mouse cursor while dragging the canvas around the screen using the middle mouse button.

![Untitled](https://user-images.githubusercontent.com/24214509/108482983-2e3d7900-72c0-11eb-8c29-4b0ccae373fb.png)

PR #2564, opened 2022-03-07
Change mouse cursor when repositioning canvas

hesham-essam:
Hi all,
I'm Hesham from Egypt, and I'm interested in applying with you for GSoC this year.

This is my attempt at fixing #2036. As far as I understand, the working area has a finite state machine that handles each tool. However, repositioning the canvas is handled by the WorkingArea class itself, so we can set the cursor there when dragging. The problem then is restoring the cursor afterwards.
My solution is to add a refresh cursor event  to each state which should do that job.
If this solution is ok, I can go ahead and handle this event for the rest of the states.


ice0:
> Hi all, I'm Hesham from Egypt, and I'm interested in applying with you for GSoC this year.

Hi, Hesham!

> This is my attempt at fixing #2036. As far as I understand, the working area has a finite state machine that handles each tool. 

Yes.

> However, repositioning the canvas is handled by the WorkingArea class itself, so we can set the cursor there when dragging. 

Yes.

> The problem then is restoring the cursor afterwards. 

Why do you think this is a problem? Can you explain it?

hesham-essam:
> Why do you think this is a problem? Can you explain it?
Each tool sets its own cursor. So while dragging the window I will change the cursor, and afterwards I will need to restore it to the one the tool uses.
> As you have already noted: "changing the position of the canvas is handled by the WorkArea class itself", so why would you care about states?
I care about them since tool states changes the cursor, so I will need to restore it.

I thought about storing the cursor style before changing it, but I thought this way was less error prone maybe.

ice0:
> I care about them since tool states changes the cursor, so I will need to restore it.
> 
> I thought about storing the cursor style before changing it, but I thought this way was less error prone maybe.

Agree. But current behavior does not use a state machine for dragging/zooming, so my suggestion is to fix this issue using straightforward way (save previous cursor before drag and restore it after drag).

If you want implement this behavior using states, then it seems to me that it is easier to add a new state, rather than a new event. 
In this case, you need to enter this state when the drag starts and exit the state when it ends. 
Anyway, you will still need to save the previous cursor when you enter this state, and restore it when you exit. Also you don't need to handle this event in all possible states.

hesham-essam:
What you say makes sense. I opted for the simpler solution of simply storing the previous cursor.

synfig-studio/src/gui/workarea.cpp
```diff
@@ -2185,15 +2194,22 @@ studio::WorkArea::set_cursor(const Glib::RefPtr<Gdk::Cursor> &x)
 	//!Check if the window we want draw is ready
 	Glib::RefPtr<Gdk::Window> draw_area_window = drawing_area->get_window();
 	if (!draw_area_window) return;
-	draw_area_window->set_cursor(x);
+
+	// do not change cursor in drag window mode
+	if (drag_mode == DRAG_WINDOW) {
+		previous_cursor = x;
```
hesham-essam, line 2200: this is necessary to prevent changing the cursor while dragging the window (for example by pressing alt to scale).  So we simply store the request cursor and restore it when dragging is finished

synfig-studio/src/gui/workarea.cpp
```diff
@@ -2185,15 +2194,22 @@ studio::WorkArea::set_cursor(const Glib::RefPtr<Gdk::Cursor> &x)
 	//!Check if the window we want draw is ready
 	Glib::RefPtr<Gdk::Window> draw_area_window = drawing_area->get_window();
 	if (!draw_area_window) return;
-	draw_area_window->set_cursor(x);
+
+	// do not change cursor in drag window mode
```
rodolforg, line 2198: Please change to
`	// do not change cursor while in drag window mode; postpone this change until after stopping window dragging`

Or something like that :)

synfig-studio/src/gui/workarea.cpp
```diff
@@ -343,8 +343,17 @@ WorkArea::set_drag_mode(DragMode mode)
...
+	// change cursor in drag window mode
+	if (mode == DRAG_WINDOW)
+		set_cursor(Gdk::FLEUR);
+
+	DragMode old_mode = drag_mode;
 	drag_mode = mode;
 
+	// reset cursor to previous cursor after finishing drag window mode
+	if (old_mode == DRAG_WINDOW)
+		set_cursor(previous_cursor);
```
rodolforg, line 355: What about do not create this variable `old_mode`, and rename the parameter name to `new_drag_mode`?
```c++
if (new_drag_mode == DRAG_WINDOW)
	set_cursor(...);
else // reset cursor [...]
	set_cursor(...);
drag_mode = new_drag_mode;
```

synfig-studio/src/gui/workarea.h
```diff
@@ -227,6 +227,7 @@ class WorkArea : public Gtk::Grid, public Duckmatic
 	bool curr_guide_is_x;
 
 	etl::handle<LockDucks> lock_ducks;
+	Glib::RefPtr<Gdk::Cursor> previous_cursor;
```
rodolforg, line 230: Please add a comment describing it.

rodolforg (changes requested):

synfig-studio/src/gui/workarea.cpp
hesham-essam: The problem here is that if `drag_mode == DRAG_WINDOW`, then `set_cursor` would have no effect. So I have first to change `drag_mode` then change the cursor.
Similarly, if `new_drag_mode == DRAG_WINDOW`, then I have to change the cursor first, then I have to change `drag_mode`. Which makes this awkward code order necessary.
Maybe I can define `set_cursor_forced` private function that would just change the cursor to make the code flow better.

ice0:
Hi guys! I have to go for today, so I will answer tomorrow.

ice0:
@hesham-essam 

Hi, Hesham!

I just want to notice that the time for writing proposals is open.
Full timeline is here: https://summerofcode.withgoogle.com/programs/2022
I recommend you to introduce yourself in [this thread](https://forums.synfig.org/t/gsoc-2022-google-summer-of-code-2022/12986) on the forum.
Have you already decided on the idea you would like to work on?

synfig-studio/src/gui/workarea.h
```diff
@@ -472,6 +475,9 @@ class WorkArea : public Gtk::Grid, public Duckmatic
 	bool on_hruler_event(GdkEvent* event);
 	bool on_vruler_event(GdkEvent* event);
 	void on_duck_selection_single(const etl::handle<Duck>& duck_guid);
+
+    // Sets the current cursor regardless of drag mode
+    void set_cursor_forced(const Glib::RefPtr<Gdk::Cursor> &x);
```
rodolforg, line 480: fix indentation (we use tab, not spaces)

rodolforg (changes requested):

synfig-studio/src/gui/workarea.h
hesham-essam: done

rodolforg (approved):

PR #2568, opened 2022-03-08
add new ValueNode "Modulo" for modulo operation

rodolforg:
Suggested by woody in
https://forums.synfig.org/t/if-else-expression/12528/9


synfig-core/src/synfig/valuenodes/valuenode_modulo.cpp
```diff
@@ -0,0 +1,257 @@
...
+ValueNode_Modulo::ValueNode_Modulo(const ValueBase &value):
+	LinkableValueNode(value.get_type())
+{
+	Vocab ret(get_children_vocab());
+	set_children_vocab(ret);
+	set_link("divisor",ValueNode_Const::create(int(2)));
+	set_link("scalar",ValueNode_Const::create(Real(1.0)));
+	Type& type(value.get_type());
+
+	if (type == type_angle)
```
ice0, line 77: Why don't do something like:
```
if (!check_type(type))
{
	assert(0);
	throw std::runtime_error(get_local_name()+_(":Bad type ")+type.description.local_name);
}
set_link("link",ValueNode_Const::create(value));
```

rodolforg: Nice.
I just based my code in another valuenode implementation (I think it was `add`) and didn't think about it :)
I'll fix it.

rodolforg: Done :)

synfig-core/src/synfig/valuenodes/valuenode_modulo.cpp
```diff
@@ -0,0 +1,239 @@
...
+/* === U S I N G =========================================================== */
+
+using namespace etl;
+using namespace synfig;
+
+/* === M A C R O S ========================================================= */
+
+/* === G L O B A L S ======================================================= */
+
+REGISTER_VALUENODE(ValueNode_Modulo, RELEASE_VERSION_1_4_0, "modulo", N_("Modulo"))
```
rodolforg, line 62: I'm hesitant with this version detail: should we create RELEASE_VERSION_1_4_0?

ice0: Do you mean `1_4_3`?

rodolforg: hmmm Maybe 1_6_0?

ice0: > hmmm Maybe 1_6_0?

Agree, it makes sense. 1.4.x is mostly bug fixes, so 1.6.0 is much better for a new feature.

rodolforg:
@ice0 what do you think about the commit I just added?

ice0 (approved):

ice0:
> @ice0 what do you think about the commit I just added?

Agree.

ice0:
Merged. Thank you!

Issue #1251, opened 2020-03-16
Strings in REGISTER_VALUENODE cannot be translated


FirasH:
**Synfig version & platform**:
Master 15ad054

**Issue description**:
The strings used in `REGISTER_VALUENODE` cannot be translated.
Adding `_()` does not work, see #1162 

Those strings are the used in the Parameters window: right-click menu > Convert > `Node_type`

rodolforg:
@FirasH Is it fixed now on master due to #2569 ?

FirasH:
> @FirasH Is it fixed now on master due to #2569 ?

I’ll check and report as soon as possible

FirasH:
Tried the lastest build, but the translation files are not updated so strings are always displayed in English.
I guess everything is ok, but the final veridict will be once we update the translation files.

PR #2569, opened 2022-03-08
fix ValueNode local name not being translated

rodolforg:
Now ValueNode Converter menu in Synfig Studio shows translated names
as expected.


FirasH:
Does this fix #1251?

rodolforg:
It seems so! :)

ice0 (approved):

PR #2571, opened 2022-03-11
Slight easing of the build process using c-make.

mohamedAdhamc:
Reasons why I think this would be helpful:
- this would eliminate the need to manually change the script then run it. Although not a problem for experienced fellas, I believe this would prove helpful to new contributors as they wouldn't know what to change and even if there was something to change.

Where the idea originated:
For building using Autotools two scripts are available for Release and Debug ---- "2-build-debug.sh" and "2-build-production.sh".
so I thought it would be helpful to do the same with c-make especially as @ice0 mentioned that it is currently the preferred way of building on windows.

p.s.
If this is to be merged I will make another proposal to describe this in the docs at the part which is about c-make building.


mohamedAdhamc:
Whoops looks like other parts would have to be changed for this to work. Will change this to draft and try to work on this tommorow hopefully (:

ice0:
> Whoops looks like other parts would have to be changed for this to work. Will change this to draft and try to work on this tommorow hopefully (: Meanwhile if anyone can give insight on to what is happening or if this is even a good idea it would be appreciated 😊.

The current CI script runs `2-build-msys-cmake.sh` to perform a test build. After renaming this file, it cannot find it and fails. 

I think it's better to set the build type using an environment variable. In this case, this will not break the current behavior and will allow to change the build type on the fly. 
Currently we using Release build on the Windows as default just because of long linking time for Debug build.

mohamedAdhamc:
> > Whoops looks like other parts would have to be changed for this to work. Will change this to draft and try to work on this tommorow hopefully (: Meanwhile if anyone can give insight on to what is happening or if this is even a good idea it would be appreciated 😊.
> 
> The current CI script runs `2-build-msys-cmake.sh` to perform a test build. After renaming this file, it cannot find it and fails.
> 
> I think it's better to set the build type using an environment variable. In this case, this will not break the current behavior and will allow to change the build type on the fly. Currently we using Release build on the Windows as default just because of long linking time for Debug build.

Hey, @ice0 
Yes I actually thought of doing that as well but thought just changing the name would be more error-prone (turns out not xd) .
Anyways I made an alternative which would just ask the user for the type of build but I think theres a problem with appveyor as it stops on the line where it should enter "Debug" or "Release". However i tried it manually and it worked. what do you think?

mohamedAdhamc:
ok, this is getting too hectic xd. I will close this, and open a new one with the new idea.

mohamedAdhamc:
#2572  is the new idea

PR #2572, opened 2022-03-11
Automating build-type selection for c-make 

mohamedAdhamc:
This would ask the user to type either "Debug" or "Release" for each respective build. therefore, removing the need to manually change it. At first, I thought of doing this for the original main script but it seems to be causing an issue for appveyor as it stops on the line which requires an input. So to get away from all this mess I just made this a separate one and the original one will be left out for the purpose of the test build to run smoothly and this would be the one provided in the dev-docs 

 Ultimately, if this is to be merged I would update the docs for it and instead provide there this script there mentioning this.

### DIFF
--- a/2-build-msys-cmake -auto.sh
+++ b/2-build-msys-cmake -auto.sh
@@ -2,6 +2,8 @@
 
 export PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:/opt/mlt-7.2.0/lib/pkgconfig"
 mkdir -p cmake-build-msys && cd cmake-build-msys
-cmake -GNinja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_INSTALL_PREFIX=./install -DCMAKE_BUILD_TYPE=Release ..
+echo type Debug for debug build, or Release for a release build
+read buildType
+cmake -GNinja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_INSTALL_PREFIX=./install -DCMAKE_BUILD_TYPE=$buildType ..
 cmake --build .
 cmake --install . >/dev/null
--- a/2-build-msys-cmake -auto.sh
+++ b/2-build-msys-cmake -auto.sh
@@ -2,8 +2,18 @@
 
 export PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:/opt/mlt-7.2.0/lib/pkgconfig"
 mkdir -p cmake-build-msys && cd cmake-build-msys
+for ((;;))
+do
 echo type Debug for debug build, or Release for a release build
 read buildType
+if [[ $buildType = Debug ]]
+then 
+break
+elif [[ $buildType = Release ]]
+then
+break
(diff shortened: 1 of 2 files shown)


Keyikedalube:
I think this PR is okay because you are committing changes to a custom build bash script that calls in CMake scripts.
The original file `2-build-msys-cmake.sh` already automates most of the heavy build generation.
So creating another is already redundant and unnecessary and also confusing.
Why not just use arguments? That way instead of waiting for the user to type in "Debug"/"Release" just set "Debug" as default choice in case no argument was passed while calling the script.
```
./2-build-msys-cmake.sh           <--- Debug build
./2-build-msys-cmake.sh release   <--- Release build
```

mohamedAdhamc:
> I think this PR is okay because you are committing changes to a custom build bash script that calls in CMake scripts. The original file `2-build-msys-cmake.sh` already automates most of the heavy build generation. So creating another is already redundant and unnecessary and also confusing. Why not just use arguments? That way instead of waiting for the user to type in "Debug"/"Release" just set "Debug" as default choice in case no argument was passed while calling the script.
> 
> ```
> ./2-build-msys-cmake.sh           <--- Debug build
> ./2-build-msys-cmake.sh release   <--- Release build
> ```



> I think this PR is okay because you are committing changes to a custom build bash script that calls in CMake scripts. The original file `2-build-msys-cmake.sh` already automates most of the heavy build generation. So creating another is already redundant and unnecessary and also confusing. Why not just use arguments? That way instead of waiting for the user to type in "Debug"/"Release" just set "Debug" as default choice in case no argument was passed while calling the script.
> 
> ```
> ./2-build-msys-cmake.sh           <--- Debug build
> ./2-build-msys-cmake.sh release   <--- Release build
> ```
Well the idea  was that ./2-build-msys-cmake.sh automatically runs the Release build and the only way to run the debug build was to open it and manually change 
-DCMAKE_BUILD_TYPE=Release
into
-DCMAKE_BUILD_TYPE=Debug

mohamedAdhamc:
> I think this PR is okay because you are committing changes to a custom build bash script that calls in CMake scripts. The original file `2-build-msys-cmake.sh` already automates most of the heavy build generation. So creating another is already redundant and unnecessary and also confusing. Why not just use arguments? That way instead of waiting for the user to type in "Debug"/"Release" just set "Debug" as default choice in case no argument was passed while calling the script.
> 
> ```
> ./2-build-msys-cmake.sh           <--- Debug build
> ./2-build-msys-cmake.sh release   <--- Release build
> ```

Hey @Keyikedalube I made another verision using arguments should I submit it here and remove this ?

Keyikedalube:
I tagged @ice0 for his approval.

I only gave my opinion on why using arguments would be an improvement over duplicated build scripts. Still, final approval goes through him. Wait for him to respond OK.

mohamedAdhamc:
> I tagged @ice0 for his approval.
> 
> I only gave my opinion on why using arguments would be an improvement over duplicated build scripts. Still, final approval goes through him. Wait for him to respond OK.

Oh ok then I guess I can submit it as a separate proposal and tag it here so he can review them both at the same time. Anyway, thanks for the feedback  :)

mohamedAdhamc:
#2573 for the other approach which maybe the more efficient one.

Keyikedalube:
Closing this in favor of #2573

PR #2573, opened 2022-03-12
Added the ability to set the build type from the command line to the 2-build-msys-cmake.sh script

mohamedAdhamc:
this way 
./2-build-msys-cmake.sh    
runs release build which is the default because as @ice0  mentioned "Currently we use Release build on the Windows as default just because of long linking time for Debug build"
and 
./2-build-msys-cmake.sh debug
runs debug build

### DIFF
--- a/2-build-msys-cmake.sh
+++ b/2-build-msys-cmake.sh
@@ -2,13 +2,22 @@
 
 export PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:/opt/mlt-7.2.0/lib/pkgconfig"
 mkdir -p cmake-build-msys && cd cmake-build-msys
-if [[ $1 = debug ]]
-then 
-cmake -GNinja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_INSTALL_PREFIX=./install -DCMAKE_BUILD_TYPE=Debug ..
+mode="Release"
+if [[ -n $1 ]] 
+then
+ for ((;;)); do
+   if [[ $1 = Debug ]]
+   then
+   mode=$1
+   else 
+   echo type Debug for debug build
+   read mode
+   fi
+   if [[ $mode = Debug ]]
+   then break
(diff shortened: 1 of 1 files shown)


ice0:
@mohamedAdhamc 

You can improve script in this way:
```
mode="Release"
if [ -n $1 ]; then
    mode=$1
fi

...
cmake -GNinja ... -DCMAKE_BUILD_TYPE=${mode} ..
```

mohamedAdhamc:
> @mohamedAdhamc
> 
> You can improve the script in this way:
> 
> ```
> mode="Release"
> if [ -n $1 ]; then
>     mode=$1
> fi
> 
> ...
> cmake -GNinja ... -DCMAKE_BUILD_TYPE=${mode} ..
> ```
> 
> In this way no need for copy/paste code.

@ice0  yes thanks will do

mohamedAdhamc:
@ice0  done

ice0:
I think it makes no sense to check the mode. 
If you specify it incorrectly, the script itself will generate an error. 
For example, now you do not check `RelWithDebInfo` mode. 
So my suggestion is just pass it as is.

mohamedAdhamc:
> I think it makes no sense to check the mode. If you specify it incorrectly, the script itself will generate an error. For example, now you do not check `RelWithDebInfo` mode. So my suggestion is just pass it as is.
Alright then, will remove the check and update.

ice0:
@mohamedAdhamc 
Looks good to me.
Can you also update the documentation showing the ability to set the build type from the command line? 

https://synfig-docs-dev.readthedocs.io/en/latest/building/Building%20using%20CMake.html

ice0 (approved):

mohamedAdhamc:
> @mohamedAdhamc Looks good to me. Can you also update the documentation showing the ability to set the build type from the command line?
> 
> https://synfig-docs-dev.readthedocs.io/en/latest/building/Building%20using%20CMake.html

Yes i will right now actually.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2574, opened 2022-03-13
fix RendDesc::get_pixels_to_world_matrix() contents

rodolforg:


ice0:
This method doesn't seem to be used anywhere. Are you planning to use it? Maybe we can just delete it?

rodolforg:
> Are you planning to use it?

Yes. And it took me one day to find that was the problem :'(

ice0 (approved):

ice0:
Merged. Thank you!

PR #2575, opened 2022-03-13
[MSYS2] Fixed MSYS2 dependencies

ice0:
Ninja is required to build MLT
Packages reordered alphabetically
Removed cairomm and pango packages
Added harfbuzz package


1-setup-windows-msys2.sh
```diff
@@ -28,25 +28,26 @@ libtool \
 make \
 patch \
 tar \
-$MINGW_PACKAGE_PREFIX-gcc \
+$MINGW_PACKAGE_PREFIX-boost \
+$MINGW_PACKAGE_PREFIX-cairomm \
```
rodolforg, line 32: There is no need of explicit `cairomm` anymore :)

1-setup-windows-msys2.sh
```diff
@@ -28,25 +28,26 @@ libtool \
...
+$MINGW_PACKAGE_PREFIX-libltdl \
+$MINGW_PACKAGE_PREFIX-libmng \
 $MINGW_PACKAGE_PREFIX-libxml++2.6 \
-$MINGW_PACKAGE_PREFIX-pango \
-$MINGW_PACKAGE_PREFIX-gtkmm3 \
+$MINGW_PACKAGE_PREFIX-make \
+$MINGW_PACKAGE_PREFIX-ninja \
 $MINGW_PACKAGE_PREFIX-openexr \
-$MINGW_PACKAGE_PREFIX-libmng
+$MINGW_PACKAGE_PREFIX-pango \
```
rodolforg, line 48: There is no need of explicit `pango`

1-setup-windows-msys2.sh
```diff
@@ -28,25 +28,26 @@ libtool \
...
+$MINGW_PACKAGE_PREFIX-pkgconf \
+$MINGW_PACKAGE_PREFIX-SDL2
```
rodolforg, line 50: As far as I know, the only reason we depend on SDL2 is `melt` of `mlt`.
Since we don't use `'melt`, is there a way to avoid building it to avoid SDL2 dependency?

1-setup-windows-msys2.sh
ice0: Done

1-setup-windows-msys2.sh
ice0: Done

1-setup-windows-msys2.sh
ice0: I will check it.

rodolforg (approved):

1-setup-windows-msys2.sh
rodolforg: It may be another PR :)

ice0: > It may be another PR :)

Yes. I just thought about the HarfBuzz dependency. Should it be added?

rodolforg: If we trust on MSYS2 as our "package provider", it makes `gtkmm3` depends on `pangomm` that depends on `pango` that (finally) depends on `harfbuzz`! :)
But I'd make sure `harfbuzz` is listed there anyway. Because `synfig-core` uses `harfbuzz` anyway.

ice0:
Added harfbuzz package

rodolforg (approved):

Issue #2200, opened 2021-06-30
Can't determine if state tools button is inactive or active for Materia GTK theme


Keyikedalube:
**Describe the bug**
State tools button's focusability is making it hard to determine if a button is active or inactive for Materia GTK theme

**To Reproduce**
1. Activate rectangle tool
2. Click on any active button
3. Inactive button is still vaguely highlighted because focus is set to this button

**Expected behavior**
When button is inactive/untoggled it should appear like the rest without grabbing focus.

**Screencast**

https://user-images.githubusercontent.com/24214509/123899279-7e220000-d984-11eb-971f-8d2ab414bdfa.mp4

**System information:**
 - OS: Fedora Linux
 - OS version: 34
 - Commit ID: 4273435309266fa4181269fdb3e9cc454cf14337

**Additional comments**
On Adwaita, a border is shown instead when button state is changed from active to inactive

mohamedAdhamc:
is this still an issue?  if it is i would be willing to work on it.

Keyikedalube:
@mohamedAdhamc Yes this is still an issue maybe only for me and probably other materia gtk theme users.

If you'd like to crack this issue I can assign you :)

mohamedAdhamc:
> @mohamedAdhamc Yes this is still an issue maybe only for me and probably other materia gtk theme users.
> 
> If you'd like to crack this issue I can assign you :)

yes, sure I'd like to have a go at it.

mohamedAdhamc:
> @mohamedAdhamc Yes this is still an issue maybe only for me and probably other materia gtk theme users.
> 
> If you'd like to crack this issue I can assign you :)

Hey sorry was carried away a bit in the past 2 days with all the build and IDE configuration errors however I will be starting to try and  tackle this, but I just have a few questions to clarify things here a little.

1- What is the expected behaviour here exactly ? My understanding is that maybe you can have multiple ones active and the problem is that inactive ones are still vaguely highlighted. Or is it that only one should be active and they should be grouped in a mutually exclusive way for example.

2- Are there any specific places or functions where you would think it'd  be helpful to look at? I'm guessing I should start state_rectangle.cpp and see where that leads me.

Keyikedalube:
> 1- What is the expected behaviour here exactly ? My understanding is that maybe you can have multiple ones active and the problem is that inactive ones are still vaguely highlighted.

Yes, that's correct :)

> I'm guessing I should start at state_rectangle.cpp and see where that leads me.

[...]

PR #2577, opened 2022-03-15
Fixing Rectangle Tool State Indeterminacy

mohamedAdhamc:
this I believe solves the bug #2200  . I am proposing this just for the rectangle tool as a demo and if merged I could easily repeat this for all the others which have the same problem.

### DIFF
--- a/synfig-studio/src/gui/states/state_rectangle.cpp
+++ b/synfig-studio/src/gui/states/state_rectangle.cpp
@@ -1264,4 +1264,13 @@ StateRectangle_Context::toggle_layer_creation()
   layer_advanced_outline_flag = get_layer_advanced_outline_flag();
   layer_curve_gradient_flag = get_layer_curve_gradient_flag();
   layer_plant_flag = get_layer_plant_flag();
+   
+  // disable focus for buttons where focus makes active/inactive states confusing
+  layer_rectangle_togglebutton.set_can_focus(false);
+  layer_region_togglebutton.set_can_focus(false);
+  layer_outline_togglebutton.set_can_focus(false);
+  layer_advanced_outline_togglebutton.set_can_focus(false);
+  layer_curve_gradient_togglebutton.set_can_focus(false);
+  layer_plant_togglebutton.set_can_focus(false);
+
 }


Keyikedalube:
The behavior is working as expected but did you notice that when you place your cursor on the text box above those togglable buttons and press tab key all those toggle buttons are skipped jumping directly to ComboBox and vice versa for shift+tab key press.
That's the minor inconvenience for `*.set_can_focus(false)`

If we wanted to keep the keyboard tab input intact we could use signals to detect when the button is toggled and thereby grabbing/ungrabbing focus intelligently. Code implementation would be longer though.

I like the current solution because I don't think most users use keyboard as input for state buttons rather they prefer quick mouse hover and clicking to select/unselect. And also because the implementation is simple and clear :)

Keyikedalube (approved):

Keyikedalube (approved):
Much better 👍🏽 🎉

ice0:
@mohamedAdhamc 

There is a more elegant way to do this. Please check `LAYER_CREATION`.

mohamedAdhamc:
> @mohamedAdhamc
> 
> There is a more elegant way to do this. Please check `LAYER_CREATION`.

@ice0  
done, not sure if this is exactly what you meant though.

ice0:
> > @mohamedAdhamc
> > There is a more elegant way to do this. Please check `LAYER_CREATION`.
> 
> @ice0 done, not sure if this is exactly what you meant, but I tested it and the functionality is consistent and it is indeed way more efficient.

This is exactly what I meant. You do the same work in nice and clean way.
It also automatically sets the correct properties for new buttons on that toolbar.
Please fix the indentation and it will be ready to be merged.

mohamedAdhamc:
@ice0  done

ice0:
> @ice0 done Also, this is a problem for other tool buttons, so should I make another proposal doing this same change for them so #2200 can finally be closed ?

In total, I counted eight tools that have tool buttons, so I suggest fixing them all in this PR, as there will be few changes.

mohamedAdhamc:
@ice0  done, also for the text tool there is only one tool button, and implementing this there right now doesn't really fix anything so I was wondering if I should edit it or not, but I decided to do so as to keep the consistency and in case new ones are added in the future.

p.s I don't know if you or other people here use qt creator but for some reason, it messes up the indentation, I had to fix it using visual studio.

Keyikedalube:
> p.s I don't know if you or other people here use qt creator but for some reason, it messes up the indentation

This section tells how to setup indentation level and type for qtcreator https://synfig-docs-dev.readthedocs.io/en/latest/ide/QtCreator.html#source-code-indentation

PR #2578, opened 2022-03-16
simplify Widget_Sublayer by using Gtk::ComboBoxText

rodolforg:
also, make some method protected instead of public

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_sublayer.cpp
+++ b/synfig-studio/src/gui/widgets/widget_sublayer.cpp
@@ -56,41 +56,24 @@ using namespace studio;
 
 /* === M E T H O D S ======================================================= */
 
-Widget_Sublayer::Widget_Sublayer():
-	value()
-{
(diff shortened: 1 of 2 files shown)


ice0 (approved):

synfig-studio/src/gui/widgets/widget_sublayer.cpp
```diff
@@ -100,24 +83,13 @@ Widget_Sublayer::set_value_desc(const synfigapp::ValueDesc &x)
...
-		{
-			set_active(iter);
-			return;
-		}
-	}
+	set_active_id(value);
 }
 
 void
 Widget_Sublayer::set_value(std::string data)
```
ice0, line 90: Maybe change it to `const std::string& data`?

rodolforg: Done :)

ice0 (approved):

ice0:
Merged. Thank you!

PR #2582, opened 2022-03-18
Added a mnemonic for the Text layer input dialog.

### LINKED ISSUE #2041: Menu mnemonics for text layer input dialog
Can't use keyboard mnemonics `Alt + O` or `Alt + C` to apply or cancel changes

Implement the mnemonics

Use mouse click to apply or cancel changes

**Additional context**
Screenshot from 2021-02-21 13-29-19

mohamedAdhamc:
Fix #2041 
Alt + O and Alt + C to apply and cancel changes

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -3769,8 +3769,8 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
 
 	dialog.get_content_area()->pack_start(text_view);
 
-	dialog.add_button(_("Cancel"), Gtk::RESPONSE_CANCEL)->set_image_from_icon_name("gtk-cancel", Gtk::ICON_SIZE_BUTTON);
-	dialog.add_button(_("OK"),   Gtk::RESPONSE_OK)->set_image_from_icon_name("gtk-ok", Gtk::ICON_SIZE_BUTTON);
+	dialog.add_button(_("_Cancel"), Gtk::RESPONSE_CANCEL)->set_image_from_icon_name("gtk-cancel", Gtk::ICON_SIZE_BUTTON);
+	dialog.add_button(_("_OK"),   Gtk::RESPONSE_OK)->set_image_from_icon_name("gtk-ok", Gtk::ICON_SIZE_BUTTON);
 	dialog.set_default_response(Gtk::RESPONSE_OK);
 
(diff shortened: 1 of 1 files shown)


rodolforg (approved):

ice0:
@mohamedAdhamc 
Do you plan to implement `Ctrl`+`Enter`/`Esc` keys support?

mohamedAdhamc:
> @mohamedAdhamc Do you plan to implement `Ctrl`+`Enter`/`Esc` keys support?

If they would be preferred over this mnemonics approach I guess I could. Should I ?

rodolforg:
> If they would be preferred over this mnemonics approach I guess I could

They are not mutually exclusive :)

mohamedAdhamc:
> > If they would be preferred over this mnemonics approach I guess I could
> 
> They are not mutually exclusive :)

Ok, sounds good 😂. Will be working on it.

mohamedAdhamc:
Hey guys, I was working on implement the short cut keys and at first I was thinking of accelerators but then realized they are more tailored for menu item things. So now the approach I will be taking is to catch the keypress event using signals, and then implementing something where if the ctr+enter are pressed Gtk::RESPONSE_OK is executed or just that the button is pressed. Is a correct approach for this ?

p.s I found something already in dialog preview which is similar to what I was thinking 
`bool Dialog_Preview::on_key_pressed(GdkEventKey *ev)
{
    if (ev->keyval == gdk_keyval_from_name("Escape") )
    {
        close_window_handler();
        return true;
    }

    return false;
}`

Also sorry for the multiple questions xd, but could someone elaborate on what the synfig exception guard is ? from looking it up in the source code I have a very basic idea that it could be related to catching exceptions which is seen more often accross the source code but I cant really confirm that and why for example it was needed here.

rodolforg:
> could someone elaborate on what the synfig exception guard is ?

It is a pair of macros to prevent any application crash if an exception is thrown and not "caught" in (mainly) gtk signal handlers.
Synfig uses exceptions (like IDNotFound, IDAlreadyUsed, BadTime,...), that's a fact. However, not every code worries about it. Sometimes there are uncovered corner cases, sometimes it is in fact unknown that some bad stuff could indirectly happen there...

So, I wrote it that guard macro and (tried to) place it in every custom Gtk signal handler even though it could not be necessary there - at least at time that I did it. We don't know what future commits would do, and it's currently hard to maintain Synfig and remember these precautions.

ice0:
Merged. Thank you!

PR #2585, opened 2022-03-21
Allow to call regression test script outside of Travis

morevnaproject:
We do not use Travis anymore, so let's allow to call regression test script without TRAVIS_BUILD_DIR variable.

This is useful for running tests locally.

Also, I have added condition to avoid re-downloading whole tests repository every time.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2587, opened 2022-03-21
Fix crash when deleting last waypoint

### LINKED ISSUE #1701: Removing last Waypoint of an Exported Value crashes Synfig Studio

hesham-essam:
Fixes #1701,  As far as I understand, the problem is that when the last waypoint is removed, the node representing the exported value is replaced, as in here
https://github.com/synfig/synfig/blob/f64f99f020f0689fb006969a7c649c101c52d066/synfig-studio/src/synfigapp/actions/waypointremove.cpp#L137-L163
The old `ValueNode` is removed from the canvas, and hence invalidating all the `ValueDesc`s that point to it. A signal is emitted of the change, for which the different tree stores handle. Many of the tree stores don't handle the event instantly, but instead wait for a period of time, then handles it as in here
https://github.com/synfig/synfig/blob/f64f99f020f0689fb006969a7c649c101c52d066/synfig-studio/src/gui/trees/childrentreestore.cpp#L299-L309 
The problem happens when the invalidated `ValueDesc` is accessed (for example before the treestore is updated.)
I found this to happen in `Dialog_Waypoint`:
https://github.com/synfig/synfig/blob/f64f99f020f0689fb006969a7c649c101c52d066/synfig-studio/src/gui/dialogs/dialog_waypoint.cpp#L145-L154
And also when a redraw event is called which accesses the invalid `ValueDesc` through any of the (yet not updated due to signal timeout) treestores.

My fix was to add checks for invalid `ValueDesc`, and eventually when all signals are handled, this invalid object won't be accessed.

### DIFF
--- a/synfig-studio/src/synfigapp/value_desc.cpp
+++ b/synfig-studio/src/synfigapp/value_desc.cpp
@@ -57,7 +57,11 @@ SYNFIGAPP_EXPORT const ValueDesc ValueDesc::blank;
 void ValueDesc::on_id_changed()
 {
 	try {
-		name = get_value_node()->get_id();
+		synfig::ValueNode::Handle value_node = get_value_node();
+		if (value_node)
+				name = value_node->get_id();
+		else
+				name.clear();
 	} catch (Exception::IDNotFound &) {
 		name.clear();
 	}
--- a/synfig-studio/src/synfigapp/value_desc.h
+++ b/synfig-studio/src/synfigapp/value_desc.h
@@ -130,7 +130,7 @@ public:
 
 		if (id_changed_connection.connected())
 			id_changed_connection.disconnect();
-		if (other.id_changed_connection.connected())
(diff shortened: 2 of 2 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

morevnaproject:
Backported to 1.4.3

PR #2588, opened 2022-03-23
Added support for `Ctr+Enter` to complete text input in Text tool dialog

### LINKED ISSUE #2041: Menu mnemonics for text layer input dialog

mohamedAdhamc:
Fix #2041 
Hey so I have been trying to do this for a bit, I started first with trying to implement accelerators but then realized that they are more for "menu" related buttons and such. So I was thinking of catching keypress events and then if the correct key is pressed then the text will be applied. However, in the process of implementing i'm facing multiple isssues.

1- `signal_key_press_event().connect(sigc::ptr_fun(&on_key_pressed)); `    in all other places where this is implemented its always done in the constructor of the class related to it. for example for dialog-preview it is in its constructor. also in gtkmm documentation it mentions it being placed in the constructor. by placing it here the build gives out an error :'signal_key_press_event' was not declared in this scope
the problem here is that dialog-paragraph is a method not a class so I cant really place it in its constructor. to which I have thought of trying to place it instead in the app class constructor (since it is the class which includes this method)... still to no successs

I'm getting puzzled here as to what I could do or where to place it.

2-also for the line 3785
 I'm getting two errors "expected primary-expression before '*' token" and "ev was not declared in this scope"

p.s
the body of "on_key_pressed" is not really made at all I'm focusing now on how to actually apply this method.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -3766,7 +3766,7 @@ App::on_key_pressed(GdkEventKey *ev)
            (ev->keyval == (0xff8d)||(0xfe34)||(0xfd1e)) &&
            (ev->state == GDK_CONTROL_MASK) )  //pressing control and enter
       {
-        ok_button->clicked();
+        //ok_button->clicked();
         std::cout << "hey";
         return true;
       }
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -3754,25 +3754,7 @@ App::dialog_sets_entry(const std::string &action, const std::string &content, st
 	return true;
 }
 
-bool App::response_ok_activate_check = 0 ;
 
-bool
-App::on_key_pressed(GdkEventKey *ev)
-  {
-
-    SYNFIG_EXCEPTION_GUARD_BEGIN()
-      if (  (ev->type == GDK_KEY_PRESS) &&
-           (ev->keyval == (gdk_keyval_from_name("s"))) &&
-           (ev->state == GDK_CONTROL_MASK) )  //pressing control and enter
-      {
-        response_ok_activate_check =1 ;
-       //  std::cout << "hey";
-        return true;
-      }
-
-      return false;
-      SYNFIG_EXCEPTION_GUARD_END_BOOL(true)
- }
 
 bool
 App::dialog_paragraph(const std::string &title, const std::string &message,std::string &text)
@@ -3796,28 +3778,31 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
 
 	dialog.add_button(_("Cancel"), Gtk::RESPONSE_CANCEL)->set_image_from_icon_name("gtk-cancel", Gtk::ICON_SIZE_BUTTON);
 	dialog.add_button(_("OK"),   Gtk::RESPONSE_OK)->set_image_from_icon_name("gtk-ok", Gtk::ICON_SIZE_BUTTON);
-    //Gtk::Button* ok_button = dialog.add_button(_("Ok"), Gtk::RESPONSE_OK);
-	dialog.set_default_response(Gtk::RESPONSE_OK);
+    dialog.set_default_response(Gtk::RESPONSE_OK);
 
-    dialog.signal_key_press_event().connect(sigc::ptr_fun(&on_key_pressed)); //catching key-press event
+
+    text_view.signal_key_press_event().connect( [&](GdkEventKey *ev) {
+        SYNFIG_EXCEPTION_GUARD_BEGIN()
+          if (  (ev->type == GDK_KEY_PRESS) &&
+               (ev->keyval == (GDK_KEY_Return) || ev->keyval == (GDK_KEY_KP_Enter) || ev->keyval == (gdk_keyval_from_name("s")) ) &&
+               (ev->state == GDK_CONTROL_MASK) )  //pressing control and enter
+          {
+            dialog.response(-5);
+            return true;
+          }
+
+            return false;
+          SYNFIG_EXCEPTION_GUARD_END_BOOL(true)
+    } );
 
 	//text_entry.signal_activate().connect(sigc::bind(sigc::mem_fun(dialog,&Gtk::Dialog::response),Gtk::RESPONSE_OK));
 	dialog.show();
-
-    if (response_ok_activate_check == 1)
-             {
-              // ok_button->clicked();
-               dialog.response(Gtk::RESPONSE_OK ) ;
-             //   std::cout<<" response emitted ";
-             }
 
 
 	if(dialog.run()!=Gtk::RESPONSE_OK)
 		return false;
 
-	text=text_buffer->get_text();
-
-    response_ok_activate_check = 0 ;
+    text=text_buffer->get_text();
 
 	return true;
 }
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -53,7 +53,6 @@
 #include <glibmm/main.h>
 #include <glibmm/miscutils.h>
 #include <glibmm/timer.h>
-#include <glibmm/shell.h>
 #include <glibmm/spawn.h>
 
 #include <gtkmm/accelmap.h>
@@ -145,6 +144,8 @@
 #include <synfig/soundprocessor.h>
 #include <synfig/string_helper.h>
 #include <synfig/version.h>
+#include <gui/exception_guard.h>
+
 
 #include <synfigapp/action.h>
 #include <synfigapp/canvasinterface.h>
@@ -152,7 +153,6 @@
 #include <synfigapp/settings.h>
 
 #include <thread>
-#include <gtkmm/icontheme.h>
 
 #ifdef _WIN32
 
@@ -170,7 +170,6 @@
 
 #endif // _WIN32
 
-
 #endif
 
 /* === U S I N G =========================================================== */
@@ -181,7 +180,6 @@ using namespace studio;
 
 /* === M A C R O S ========================================================= */
 
-#define DEFAULT_ICON_THEME_NAME "classic"
 /* === S I G N A L S ======================================================= */
 
 static sigc::signal<void> signal_present_all_;
@@ -283,7 +281,6 @@ bool   studio::App::resize_imported_images       = false;
 bool   studio::App::animation_thumbnail_preview  = true;
 bool   studio::App::enable_experimental_features = false;
 bool   studio::App::use_dark_theme               = false;
-String studio::App::icon_theme_name              = "";
 bool   studio::App::show_file_toolbar            = true;
 String studio::App::custom_filename_prefix       (DEFAULT_FILENAME_PREFIX);
 int    studio::App::preferred_x_size             = 480;
@@ -454,7 +451,7 @@ class SetsModel : public Gtk::TreeModel::ColumnRecord
 
 /* === P R O C E D U R E S ================================================= */
 
-class App::Preferences : public synfigapp::Settings
+class Preferences : public synfigapp::Settings
 {
 public:
 	virtual bool get_raw_value(const synfig::String& key, synfig::String& value)const override
@@ -510,11 +507,6 @@ public:
 			if(key=="use_dark_theme")
 			{
 				value=strprintf("%i",(int)App::use_dark_theme);
-				return true;
-			}
-			if(key=="icon_theme_name")
-			{
-				value=strprintf("%s", App::get_raw_icon_theme_name().c_str());
 				return true;
 			}
 			if(key=="show_file_toolbar")
@@ -703,11 +695,6 @@ public:
 				App::use_dark_theme=i;
 				return true;
 			}
-			if(key=="icon_theme_name")
-			{
-				App::icon_theme_name = value;
-				return true;
-			}
 			if(key=="show_file_toolbar")
 			{
 				int i(atoi(value.c_str()));
@@ -847,7 +834,6 @@ public:
 		ret.push_back("animation_thumbnail_preview");
 		ret.push_back("enable_experimental_features");
 		ret.push_back("use_dark_theme");
-		ret.push_back("icon_theme_name");
 		ret.push_back("show_file_toolbar");
 		ret.push_back("brushes_path");
 		ret.push_back("custom_filename_prefix");
@@ -875,7 +861,7 @@ public:
 	}
 };
 
-App::Preferences App::_preferences;
+static ::Preferences _preferences;
 
 void
 init_ui_manager()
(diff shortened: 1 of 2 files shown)


ice0:
> 1- `signal_key_press_event().connect(sigc::ptr_fun(&on_key_pressed)); ` in all other places where this is implemented its always done in the constructor of the class related to it. for example for dialog-preview it is in its constructor.

You can attach to event in any place.
In your case this will be `<widget>.signal_key_press_event().connect(...);`.
For example `dialog.signal_key_press_event().connect`.

> also in gtkmm documentation it mentions it being placed in the constructor. by placing it here the build gives out an error :'signal_key_press_event' was not declared in this scope the problem here is that dialog-paragraph is a method not a class so I cant really place it in its constructor. to which I have thought of trying to place it instead in the app class constructor (since it is the class which includes this method)... still to no successs

I think it's better to start from simple test application, so you will not be confused with unrelated stuff.

> 
> I'm getting puzzled here as to what I could do or where to place it.
> 
> 2-also for the line 3785 I'm getting two errors "expected primary-expression before '*' token" and "ev was not declared in this scope"
> 
> p.s the body of "on_key_pressed" is not really made at all I'm focusing now on how to actually apply this method.

It looks like you are trying to make a function inside a function. Please check this discussion for better understanding: https://stackoverflow.com/questions/4324763/can-we-have-functions-inside-functions-in-c

mohamedAdhamc:
Hey, @ice0 thanks for the head up. now the key-press signal is being handled correctly I believe as when the dialog is open and i hit ctr+enter it prints out "hey" in the command prompt.

I believe I'm almost done here the last thing would be to implement line 3772 `//ok_button->clicked();`
so then when the shortcut is hit the button will be clicked. However, I'm having quite a problem here trying to access the button which is the dialog_paragraph method from the on_key_pressed method.  any suggestions as to what i might do ?

p.s, a few side questions
1- as you might have noticed I have made an entirely new button xd just so that I can have its name as I couldn't find in the documentation when dialog.add_button is used if "ok" is the name or if it even has one. any suggestions?

2- also I found that there are three key values for enter,  is that related to each machine type, and if so should I keep all three as is ? (GDK_KEY_KP_Enter, GDK_KEY_ISO_Enter, GDK_KEY_3270_Enter)

ice0:
> now the key-press signal is being handled correctly I believe as when the dialog is open and i hit ctr+enter it prints out "hey" in the command prompt.

I'm not sure about this because the `ev->keyval` check is wrong.
Check the LGTM warning first to see what I mean. 

> I believe I'm almost done here the last thing would be to implement line 3772 `//ok_button->clicked();` so then when the shortcut is hit the button will be clicked. However, I'm having quite a problem here trying to access the button which is the dialog_paragraph method from the on_key_pressed method. any suggestions as to what i might do ?

I think there is no need to "click" the button. This the quote from documentation (https://docs.gtk.org/gtk3/class.Dialog.html):

>  When adding buttons using add_button(), clicking the button will emit
 signal_response() with a "response id" you specified. You are encouraged
 to use the Gtk::ResponseType enum.

It's better to call the `Gtk::Dialog::response(...)` method with the required `response_id` to emit required signal. 


> p.s, a few side questions 1- as you might have noticed I have made an entirely new button xd just so that I can have its name as I couldn't find in the documentation when dialog.add_button is used if "ok" is the name or if it even has one. any suggestions?

If you need the button which has `RESPONSE_OK` you can use `get_widget_for_response(...)` method.
But, as I mentioned earlier, there is no reason to do this, because you can just emit required signal. 

> 
> 2- also I found that there are three key values for enter, is that related to each machine type, and if so should I keep all three as is ? (GDK_KEY_KP_Enter, GDK_KEY_ISO_Enter, GDK_KEY_3270_Enter)

You need 2 values: `GDK_KEY_Return` and `GDK_KEY_KP_Enter`
Return is "Enter" on the main keyboard and "KP_Enter" on the keypad/numpad.

mohamedAdhamc:
> 
> I'm not sure about this because the `ev->keyval` check is wrong. Check the LGTM warning first to see what I mean.
> 
> > I believe I'm almost done here the last thing would be to implement line 3772 `//ok_button->clicked();` so then when the shortcut is hit the button will be clicked. However, I'm having quite a problem here trying to access the button which is the dialog_paragraph method from the on_key_pressed method. any suggestions as to what i might do ?
> 


@ice0 yes, I can how using a logical operator here in that manner could possibly cause an error. however, it suggests using bitwise operators which would just output a whole new value which isn't needed I believe. However, I don't believe that's much of an issue, worst case scenario would be having to split it into two if-conditions.

Now, the bigger problem and which is causing more of an issue is that enter isn't being detected. I put a `std::cout<<(ev->keyval);`  at the very beginning of the method called when a key is pressed. I tried different combinations and the only ones  that seem to be picked are (control+letter/numbers), letters by themselves aren't, also enter, and basically anything which actually manipulates the text box it seems. could this actually be the reason ?

ice0:
> Now, the bigger problem and which is causing more of an issue is that enter isn't being detected. I put a `std::cout<<(ev->keyval);` at the very beginning of the method called when a key is pressed. 

There is a possibility that the keystroke is intercepted somewhere else. For example, in the main Synfig window.
And that's exactly the reason why I recommended that you start with a simple application to make sure that there are no other factors influencing the behavior. 

> I think it's better to start from simple test application, so you will not be confused with unrelated stuff.

rodolforg:
The signal_key_press_event must be attached to the text area, not to the OK/Cancel buttons.

mohamedAdhamc:
> The signal_key_press_event must be attached to the text area, not to the OK/Cancel buttons.

yes, I had it attached at the beginning to the dialog not the buttons, but for some reason I thought the shortcut applying the button would look better to the eye if it was to be noticed, but yea as ice said using the dialog response right away is better.

 The problem I seem to be facing now though is that I cant access the dialog which is in the paragraph layer function from the on_key_press function. Also for some reason ctr+ enter/up/down/left or right isn't captured and I'm suspecting it has to do with the text buffer maybe picking the signal first or something as also letters by themselves aren't picked up they're only picked up when ctrl is pressed and the text cursor kind of pauses. I was looking through the possible solutions but was quite busy in the past couple days I will hopefully be back at it tommorow. Any suggestions would be appreciated though  😊 .

mohamedAdhamc:
quick description of what I have done so far:

1- `dialog.signal_key_press_event().connect(sigc::ptr_fun(&on_key_pressed));`   for catching the key-press event

2- then the on_key_press method which sets a variable to one 

3- supposedly when this variable is 1 a gtk response ok signal is emitted which the run() handles by default.

4- at the end of the dialog execution the variable value is set to zero again.

what I'm having trouble with though I believe is step 3. From what I read the run function executes a recursive main loop until a button is pressed or a response signal is emitted then when either of them happens it returns the response id and the program flow continues normally. I even tried directly harcoding an ok response right before the .run() and still it didnt return anything ! any ideas on what I can do ?

p.s I was using Github desktop and it ruined the files I will be adding the correct commits later on today as I have to go for now.

ice0:
> what I'm having trouble with though I believe is step 3.

I'm not sure about this. I made a simple application as I already recommended to you and found the reason why the keypress is not being handled correctly. After that, I switched to Synfig and implemented the necessary logic.
I used a lambda instead of the explicit `on_key_pressed` function just to keep all the logic in one place.
The whole solution consists of only 9 lines of code.

mohamedAdhamc:
> > what I'm having trouble with though I believe is step 3.
> 
> I'm not sure about this. I made a simple application as I already recommended to you and found the reason why the keypress is not being handled correctly. After that, I switched to Synfig and implemented the necessary logic. I used a lambda instead of the explicit `on_key_pressed` function just to keep all the logic in one place. The whole solution consists of only 9 lines of code.

I tried the simple application method, but I found that when I mimic the exact behaviour of this dialog I get the same problems, so I though then maybe doing it in the same place would save time ().

ice0:
> However, now since you have already done it should I close this PR ?

No. I just want to say that this problem has a simple solution and there is a good chance to learn something new here.

mohamedAdhamc:
> > However, now since you have already done it should I close this PR ?
> 
> No. I just want to say that this problem has a simple solution and there is a good chance to learn something new here.

Oh alright then should I continue working on it ?

ice0:
> Oh alright, then I should keep working on it correct ?

Sure.

mohamedAdhamc:
Unfortunately due to my exams, which will be everyday in the upcoming week starting saturday, I will have to take a break for this week. it's been a good couple weeks now with you guys. Looking forward to being back after a week with even more enthusiasm to continue this and other new things 😊.

mohamedAdhamc:
Hey all, I'm finally back 😊
so the lambda seems to be implemented correctly. Also ctrl+s succesfully applies a gtk response ok. However, still working on making ctrl+enter work.

 when I made a simple dialog without the text view ctrl+ enter worked just fine. Also when I did the same for synfig by commenting text_view.show(); it worked correctly (the signal was attached to dialog not text_view while doing so). I then also tried using  text_view.set_editable(false); on the condition that control is pressed which also makes ctrl+enter works (but obviously messses up other things). So basically I'm 99% sure the problem here is from the text view/ text buffer as whenever a key is pressed which manipulates them it's handled right away by the text_view/buffer before being able to be handled by the lambda.

I have been going through the docs trying to find out a good way of overriding this behaviour only for ctr+enter. I found multiple things like `im_context_filter_keypress( )` , and I tried to implement them but I'm also not sure if this is the "correct" way there might be an even easier way for overriding this default behaviour. I guess thats all for today tommrow i will be continuing hopefully.

mohamedAdhamc:
@ice0  done. now ctr+enter/return works correctly to apply change in the dialog paragraph. This alongside #2582  should finally fix #2041.

2-build-msys-cmake.sh
```diff
@@ -3,7 +3,7 @@
 export PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:/opt/mlt-7.2.0/lib/pkgconfig"
 mkdir -p cmake-build-msys && cd cmake-build-msys
 mode="Release"
-if [[ -n $1 ]]
```
ice0, line 6: You need to remove this change.

synfig-studio/src/gui/app.cpp
```diff
@@ -3777,6 +3778,22 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
...
 	dialog.set_default_response(Gtk::RESPONSE_OK);
 
+	text_view.signal_key_press_event().connect(( [&](GdkEventKey *ev) {
+
+		SYNFIG_EXCEPTION_GUARD_BEGIN()
+		if (	(ev->type == GDK_KEY_PRESS) &&
+				(ev->keyval == (GDK_KEY_Return) || ev->keyval == (GDK_KEY_KP_Enter) ) &&
+				(ev->state == GDK_CONTROL_MASK) )
+			{
+			dialog.response(-5);
```
ice0, line 3788: Please use a named constant here (instead of -5).

ice0:
I suggest to copy changes from https://github.com/synfig/synfig/pull/2582 here

synfig-studio/src/gui/app.cpp
mohamedAdhamc: oh yes definetly xd, was planning on doing so at the end but forgot.

mohamedAdhamc:
@ice0 done.

synfig-studio/src/gui/app.cpp
```diff
@@ -3773,10 +3774,26 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
...
+
+		SYNFIG_EXCEPTION_GUARD_BEGIN()
+		if (	(ev->type == GDK_KEY_PRESS) &&
+				(ev->keyval == (GDK_KEY_Return) || ev->keyval == (GDK_KEY_KP_Enter) ) &&
+				(ev->state == GDK_CONTROL_MASK) )
+			{
+			dialog.response(Gtk::RESPONSE_OK);
+			return true;
+			}
+			return false;
```
rodolforg, line 3791: Indentation is a bit weird here :)

mohamedAdhamc: yea, I admit it did look a bit weirx xd. I fixed it up abit and indented the  "if statement" in the way you guys proposed in the forums. I am guilty of using a space or two though to fix the allignment of the multi-lined condition but I think its fine. if not please let me know 😊.

synfig-studio/src/gui/app.cpp
```diff
@@ -3773,10 +3774,25 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
...
+		    (ev->state == GDK_CONTROL_MASK)){
+				dialog.response(Gtk::RESPONSE_OK);
+				return true;
+		}
+		return false;
+		SYNFIG_EXCEPTION_GUARD_END_BOOL(true)
+
+	}), false );
+
+
```
rodolforg, line 3795: No need for those 2 extra blank lines

synfig-studio/src/gui/app.cpp
```diff
@@ -3773,10 +3774,25 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
 
 	dialog.get_content_area()->pack_start(text_view);
 
-	dialog.add_button(_("Cancel"), Gtk::RESPONSE_CANCEL)->set_image_from_icon_name("gtk-cancel", Gtk::ICON_SIZE_BUTTON);
-	dialog.add_button(_("OK"),   Gtk::RESPONSE_OK)->set_image_from_icon_name("gtk-ok", Gtk::ICON_SIZE_BUTTON);
+	dialog.add_button(_("_Cancel"), Gtk::RESPONSE_CANCEL)->set_image_from_icon_name("gtk-cancel", Gtk::ICON_SIZE_BUTTON);
+	dialog.add_button(_("_OK"),   Gtk::RESPONSE_OK)->set_image_from_icon_name("gtk-ok", Gtk::ICON_SIZE_BUTTON);
 	dialog.set_default_response(Gtk::RESPONSE_OK);
 
+	text_view.signal_key_press_event().connect(( [&](GdkEventKey *ev) {
```
rodolforg, line 3781: You don't have to place all App object in lambda context (And we should avoid some lambda context usages).
Use only the needed variable: `[&dialog]`.

mohamedAdhamc:
damn, I tried rebasing using github desktop and now lots of thing got messede up (* _ *)

synfig-studio/src/gui/app.cpp
mohamedAdhamc: yes makes much more sense to not include everything in the laambda. Thanks !

ice0:
Merged via #2689

Issue #1975, opened 2021-01-11
[Synfig Studio] Main window incorrectly positioned at startup


ice0:
**Synfig version & platform**:
PR https://github.com/synfig/synfig/pull/1973

**Issue description**:
When window position is set to 0,0 it moves out of screen.

Bug is reproducible when you try to `move()` after present(). If you call `move()` before showing window - then position is correct.

If you call move() again after window is already visible it positions incorrectly. 
Reproduced only on multi-monitor configuration, if monitors have different DPI and primary monitor have low DPI" and secondary HiDPI.

pixelr0:
,

rodolforg:
Is it specifically to Windows?

ice0:
> Is it specifically on Windows?

Unfortunately, I can only test this on Windows for now.

PR #2591, opened 2022-03-24
Fix crash when switching workspaces

### LINKED ISSUE #2584: Switching workspace crashes Synfig
**Describe the bug**
Can no longer switch between workspaces: Animating, Compositing, Default, and/or user custom ones.
 The program just crashes.

**To Reproduce**
1. Go to 'Window > Workspace'
2. Click on 'Compositing'
3. Crash

**Expected behavior**
Workspace switches from Default to Compositing successfully.

**Screenshots**
Screenshot from 2022-03-20 14-12-37

**System information:**
 - OS: Fedora
 - OS version: 35
 - Commit ID: f64f99f020f0689fb006969a7c649c101c52d066

**Additional comments**
Crash reproducible both on Debug and Release build.
I got this console output at the end:
```
Application appears to have terminated successfully
pure virtual method called
terminate called without an active exception
Aborted (core dumped)
```

hesham-essam:
Fixes #2584. Hiding a window causes it to be removed from Gtk::Application. When all windows are removed, the main event loop stops.

### DIFF
--- a/synfig-studio/src/gui/docks/dockmanager.cpp
+++ b/synfig-studio/src/gui/docks/dockmanager.cpp
@@ -447,12 +447,24 @@ static void check_and_fix_window_position(int &x, int &y, int &width, int &heigh
 	const Gdk::Rectangle window_rect(x,y,width,height);
 	Gdk::Rectangle window_visible_rect;
 
+	Gdk::Rectangle monitor_rect;
+
 #if GTK_CHECK_VERSION(3,22,0)
 	int n_monitors = display->get_n_monitors();
+	auto monitor_at_topleft = display->get_monitor_at_point(x, y);
+	monitor_at_topleft->get_geometry(monitor_rect);
 #else
 	auto screen = display->get_default_screen();
-	int n_monitors = display->get_default_screen()->get_n_monitors();
+	int n_monitors = screen->get_n_monitors();
+	auto monitor_at_topleft_idx = screen->get_monitor_at_point(x, y);
+	screen->get_monitor_geometry(monitor_at_topleft_idx, monitor_rect);
 #endif
+
+	// adjust topleft point to lie in the monitor
+	if (x < monitor_rect.get_x() || x > monitor_rect.get_x() + monitor_rect.get_width())
+		x = monitor_rect.get_x();
+	if (y < monitor_rect.get_y() || y > monitor_rect.get_y() + monitor_rect.get_height())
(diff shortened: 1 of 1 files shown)


rodolforg:
Yes, this was the source of the problem when we switched to Gtk::Application :)

ice0:
Hi!
A little explanation about this hack (with hiding the window). 

https://github.com/synfig/synfig/blob/042c7378917c32af4354bac5d35283a63a7b828c/synfig-studio/src/gui/docks/dockmanager.cpp#L582-L584

The initial problem is that if you close the minimized Synfig window, it would open offscreen at startup and not be visible. (https://github.com/synfig/synfig/issues/1871)
Probably this is only a Windows OS issue just in case.

Rodolfo fixed this problem, but another issue appeared: on a multi-monitor configuration, the window opened a little higher than it should be, and the minimize/maximize buttons were out of view. 
https://github.com/synfig/synfig/pull/1973
https://github.com/synfig/synfig/issues/1975


I noticed that this happens if the window is at coordinates (0, 0). (In fact, this is not entirely true, more on that later). But if you hide the window and move it to (0, 0) and then show it, then its position is correct and it opens where it should, in the top left corner.

This was done out of a misunderstanding of what was really going on. 

The fact is that on a multi-monitor configuration, the monitors form a virtual rectangle that combines them into one large display. However, if the monitors are not the same geometry (as in my case), then the top left corner of one of the monitors will not be at the (0, 0) coordinates.
In my case, this is (0, 173) pixels, and this is exactly the information we can get using the `display->get_monitor(i)->get_geometry(rect)` method. 
This is done here:
https://github.com/synfig/synfig/blob/042c7378917c32af4354bac5d35283a63a7b828c/synfig-studio/src/gui/docks/dockmanager.cpp#L429

I.e. the solution is quite simple, we need to move the window not to (0, 0) but to the top left corner of the monitor on which window is located. The code that hides the window then becomes unnecessary and everything will work correctly.

So I can merge the current PR as is and fix the hack in another PR (this will remove all current changes and hack) or maybe you want to continue working in this PR?

hesham-essam:
I will fix the hack

hesham-essam:
I don't have multiple monitors to test this, but this should ensure that the top left point is visible.

ice0:
Looks good to me.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2595, opened 2022-03-27
Palette editor invert option

Kemosalamy:
Added the ability to right click a color in the palette editor and invert it's RGB values (basically 255-current value). Does not affect alpha.

### DIFF
--- a/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
+++ b/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
@@ -329,6 +329,14 @@ Dock_PalEdit::show_menu(int i)
 	item->show_all();
 	menu->append(*item);
 
+	item = manage(new Gtk::ImageMenuItem(Gtk::StockID("synfig-invert")));
+	item->signal_activate().connect(
+		sigc::bind(
+			sigc::mem_fun(*this,&studio::Dock_PalEdit::invert_color),
+			i ));
+	item->show_all();
+	menu->append(*item);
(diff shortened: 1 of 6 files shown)


synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -371,6 +379,16 @@ Dock_PalEdit::erase_color(int i)
 	refresh();
 }
 
+void
+Dock_PalEdit::invert_color(int i)
+{
+	palette_[i].color.set_r(1.00f-palette_[i].color.get_r());
+	palette_[i].color.set_g(1.00f-palette_[i].color.get_g());
+	palette_[i].color.set_b(1.00f-palette_[i].color.get_b());
```
rodolforg, line 387: There is an `operator ~` in `synfig::Color` that invert color :)

I don't know if this (interesting) PR will be accepted, but I think maybe it would be good to use this operator.

Kemosalamy: Thank you very much for the suggestion! I will implement the inverting using that ~ operator and commit it to the branch.

morevnaproject:
Hi @Kemosalamy! The PR is written well, but I have doubts about function itself. Can you please provide any real use-case for such function?

Kemosalamy:
closing this pull request as I have initiated another one which implements more practical features: [#2597](https://github.com/synfig/synfig/pull/2597)

PR #2597, opened 2022-04-01
feat: palette clipboard functionality

Kemosalamy:
Added the ability to right click a color in the palette and copy it's hexadecimal value to clipboard. 

image

Also added the ability to add a color to the palette from the clipboard.

image

### DIFF
--- a/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
+++ b/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
@@ -385,20 +385,20 @@ Dock_PalEdit::add_color(const synfig::Color& x)
 bool
 Dock_PalEdit::check_hex_format(const std::string& hexcolor)
 {
-	int strsize = hexcolor.size();
+	const int strsize = hexcolor.size();
 
-	//Check size is correct for the format: #ffffff or #fff or #f
-	//these are the sizes supported by the Color::set_hex() function 
-	if(strsize!=7&&strsize!=4&&strsize!=2)
+	// Check size is correct for the format: #ffffff or #fff or #f
+	// these are the sizes supported by Color::set_hex()
+	if (strsize != 7 && strsize != 4 && strsize != 2)
 		return false;
 
-	//Check if first character is a # for the correct format
-	if(hexcolor[0]!='#')
+	// Check if first character is a # for the correct format
+	if (hexcolor[0] != '#')
 		return false;
 
-	//Checking if all characters are between 0-F uppercase or lowercase in ascii
-	for(int i=1;i<strsize;i++)
-		if(!std::isxdigit(hexcolor[i]))
+	// Checking if all characters are between 0-F uppercase or lowercase in ascii
+	for (int i=1; i < strsize; ++i)
+		if (!std::isxdigit(hexcolor[i]))
 			return false;
 
 	return true;
@@ -410,9 +410,9 @@ Dock_PalEdit::add_from_clipboard()
 	Glib::RefPtr<Gtk::Clipboard> refClipboard = Gtk::Clipboard::get();
 	std::string hexcolor = refClipboard->wait_for_text();
 
-	//Clipboard is in hexformat. Add the color.
-	if(check_hex_format(hexcolor)){
-		hexcolor = hexcolor.substr(1,6);
+	// Clipboard is in hexformat. Add the color.
+	if (check_hex_format(hexcolor)) {
+		hexcolor = hexcolor.substr(1, 6);
 		
 		synfig::Color col;
 		col.set_hex(hexcolor);
@@ -434,7 +434,7 @@ Dock_PalEdit::copy_color(int i)
 	}
 	// Taking first 7 characters of the color string as they contain the formated #hexadecimal color
 	Glib::RefPtr<Gtk::Clipboard> refClipboard = Gtk::Clipboard::get();
-	refClipboard->set_text(palette_[i].color.get_string().substr(0,7));
+	refClipboard->set_text(palette_[i].color.get_string().substr(0, 7));
 }
 
 void


synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -349,6 +369,68 @@ Dock_PalEdit::add_color(const synfig::Color& x)
 	return size()-1;
 }
 
+void
+Dock_PalEdit::add_from_clipboard()
+{
+	if(clipboard_==NULL)
```
ice0, line 375: Maybe use `Gdk::Clipboard` here?
https://developer-old.gnome.org/gtkmm/stable/classGdk_1_1Clipboard.html#details

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -349,6 +369,68 @@ Dock_PalEdit::add_color(const synfig::Color& x)
...
+	//Bool to check if the clipboard string is formatted in hexadecimal.  
+	bool hexfrmt = false;
+
+	//Check if clipboard is empty
+	if(hexcolor!=NULL){
+		hexfrmt = true;
+
+		//Check size is correct for the format: #ffffff
+		//Check if first character is a # for the correct format
+		if(strlen(hexcolor)!=7||hexcolor[0]!='#')
```
ice0, line 389: There may also be a shortened version of the color representation.  (#rgb)
https://www.w3schools.com/css/css_colors_hex.asp

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -349,6 +369,68 @@ Dock_PalEdit::add_color(const synfig::Color& x)
...
+	if(clipboard_==NULL)
+		clipboard_ = gtk_clipboard_get(GDK_SELECTION_CLIPBOARD);
+	
+	gchar* hexcolor = gtk_clipboard_wait_for_text(clipboard_);
+
+	//Bool to check if the clipboard string is formatted in hexadecimal.  
+	bool hexfrmt = false;
+
+	//Check if clipboard is empty
+	if(hexcolor!=NULL){
```
ice0, line 384: I suggest to extract format check to a separate function. This will make code cleaner.

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
Kemosalamy: I have added support for those additional hex representations 👍.

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
Kemosalamy: The hex format checking has been moved to a helper function.

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
Kemosalamy: According to the link that was posted I need to have the following library:
```#include <gdkmm/clipboard.h>```
however when I try to compile with adding that include it gives me the following error: 
``` fatal error: gdkmm/clipboard.h: No such file or directory```
  ``` 34 | #include <gdkmm/clipboard.h> ```
   
it seems like that part of gdk at least is not part of the synfig dependencies?

Comment (ice0) [reply synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp]: Sorry I gave you the wrong link. Missed this part: 

> [Since gtkmm 3.94:](https://developer-old.gnome.org/gtkmm/stable/since_3_94.html#_since_3_94000002)

So this link is for GTK4.
Found only example for GTK3: https://developer-old.gnome.org/gtkmm-tutorial/3.24/sec-clipboard-examples.html.en

But also I found this comment:
https://gitlab.gnome.org/GNOME/gtkmm/-/issues/110

> I will not continue forever with maintenance of gtkmm and other *mm modules.
> Unless someone else continues the work, glibmm, gtkmm and other modules will
> be abandoned in a near future.

So now I'm not sure if we should prefer gtkmm to gtk. 

@rodolforg @morevnaproject Any ideas here?

File: synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp [hunk 4]
```diff
@@ -349,6 +369,76 @@ Dock_PalEdit::add_color(const synfig::Color& x)
...
+	//these are the sizes supported by the Color::set_hex() function 
+	if(strsize!=7&&strsize!=4&&strsize!=2)
+		return false;
+
+	//Checking if all characters are between 0-F uppercase or lowercase in ascii
+	for(int i=1;i<strsize;i++){
+		char c = hexcolor[i];
+
+		//    0   -   9  || A   -   F  || a   -    f
+		if(!((47<c&&c<58)||(64<c&&c<71)||(96<c&&c<103)))
```
ice0, line 395: You can use [std::isxdigit](https://en.cppreference.com/w/cpp/string/byte/isxdigit) here.

Kemosalamy: the for loop has been updated to use that function. That's a really cool function I didn't know about 😅, thanks for bringing my attention to it.

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -349,6 +369,70 @@ Dock_PalEdit::add_color(const synfig::Color& x)
...
+
+	signal_changed()();
+	refresh();
+}
+
+void
+Dock_PalEdit::copy_color(int i)
+{
+	//Taking first 7 characters of the color string as they contain the formated #hexadecimal color
+	GtkClipboard* clipboard = gtk_clipboard_get(GDK_SELECTION_CLIPBOARD);
```
rodolforg, line 429: Why do you use Gtk instead of gtkmm? :)
`Glib::RefPtr<Gtk::Clipboard> refClipboard = Gtk::Clipboard::get();`
https://developer-old.gnome.org/gtkmm-tutorial/3.24/sec-clipboard-copy.html.en
https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1Clipboard.html

Kemosalamy: oh thanks for the heads up! I wasn't aware that I should be using specifically gtkmm. I have changed the implementation to use the gtkmm functions from those links :).

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
Kemosalamy: I've ended up sticking with the functions described in this page 
https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1Clipboard.html 
which was posted by rodolforg.

ice0 (approved):

rodolforg:
@ice0 should I squash all commits in a single one?

ice0:
> @ice0 should I squash all commits in a single one?

I think yes, better to do this.
BTW I'm currently checking why the build is taking so long - still in progress.

rodolforg:
@ice0  it looks like it was fixed :)

ice0:
> @ice0 it looks like it was fixed :)

Yes, where was an issue with disk read speed (because it was 90% full). After cleaning, the speed stabilized.

PR #2603, opened 2022-04-03
Switched Ubuntu version on Github Workflows to 18.04 (#2322)

ice0:
Version 16.04 is no longer available.
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners


rodolforg:
@ice0 I can't recall why v1.4.x branch fails.

ice0:
Ubuntu failed because 16.04 is no longer supported on Github.
MacOS build still fails but it's ok for now.

PR #2613, opened 2022-04-17
feat: Reproducible build

rodolforg:
Reasoning is stated here: https://reproducible-builds.org/.

UPD:
Build can be reproduced by setting environment variable `SOURCE_DATE_EPOCH`

### DIFF
--- a/synfig-core/src/synfig/main.cpp
+++ b/synfig-core/src/synfig/main.cpp
@@ -34,6 +34,9 @@
 #	include <config.h>
 #endif
 
+#include <cstring>
+#include <ctime>
+
 #include <synfig/localization.h>
 #include <synfig/general.h>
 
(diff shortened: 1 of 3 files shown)


synfig-core/src/synfig/main.cpp
```diff
@@ -125,7 +128,23 @@ synfig::get_version()
 const char *
 synfig::get_build_date()
 {
-	return __DATE__;
+	const int max_date_length = 50;
+	static char date_str[max_date_length] = {0};
+
+	if (date_str[0] == 0) {
+		if (char* source_date_epoch = getenv("SOURCE_DATE_EPOCH")) {
```
ice0, line 135: I suggest to add a comment here:
```
// https://reproducible-builds.org/specs/source-date-epoch/
```

synfig-core/src/synfig/main.cpp
```diff
@@ -125,7 +128,23 @@ synfig::get_version()
...
+			std::istringstream iss(source_date_epoch);
+			std::time_t t;
+			iss >> t;
+			if (iss.fail()
+			    || !iss.eof()
+			    || !std::strftime(date_str, sizeof(date_str), "%x", std::localtime(&t))) {
+				    std::strncpy(date_str, _("Unknown date"), max_date_length-1);
```
ice0, line 142: What about `"Unknown build date"`? 
I think it will be more clear.

synfig-studio/src/gui/dialogs/about.cpp
```diff
@@ -312,8 +314,24 @@ About::About()
 	#endif
 
 	extra_info += "\n";
-
-	extra_info += etl::strprintf(_("Built on %s" /* at %s */ "\n\n"), __DATE__ /* , __TIME__ */ );
+	{
+		const int max_date_length = 50;
```
ice0, line 318: Maybe extract this part to `synfig/general.h`?

rodolforg: I don't know.
One is for synfig/libsynfig. Other is for synfigstudio.
I don't really know if build date is really necessary to be shown to user. :)

synfig-core/src/synfig/main.cpp
rodolforg: Done

synfig-core/src/synfig/main.cpp
rodolforg: Done

ice0 (approved):

ice0:
Merged. Thank you!

Issue #1267, opened 2020-03-21
[Timetrack] Buttons to change interpolation for selected waypoints


morevnaproject:
It would be nice to add buttons, which allow to change interpolation for selected waypoints:
![zzzz](https://user-images.githubusercontent.com/332868/77225647-6796d180-6ba3-11ea-8d7e-c2ced6724a68.png)

5 buttons, each button has icon of corresponding interpolation type.

Clicking on button changes interpolation on selected waypoints.

Note: Those buttons act differently than move/duplicate/scale buttons in TimeTrack panel. They do not switch to specific "mode" (tool), instead they just apply particular action to selected waypoints. This is why I added separator to toolbar, so interpolation buttons visually stand away from other buttons.

rodolforg:
If those buttons needs more space, should we use a popup/popover menu/bar?

morevnaproject:
Good idea!

mohamedAdhamc:
Hey @morevnaproject @rodolforg, should these five buttons apply the selected interpolation type for the in/out interpolation together?

mohamedAdhamc:
Merged in #2615

PR #2615, opened 2022-04-20
feat(studio): added buttons to change interpolation for selected waypoints

mohamedAdhamc:
Added the five buttons asked for in #1267 .
prototype

Yet to do 
1- add the separator 
2- add the button functionality

update:
p.s. I believe these buttons should be normal gtk buttons not radio buttons but I will leave this for after I'm done with correctly applying the functionality to the buttons.


mohamedAdhamc:
Hey guys @ice0 @Keyikedalube  and anyone who could help xd,
I'm not even sure if the signal is connected properly but I came across something I find quite weird.
the CanvasView member function canvas_interface() is used without an object many times in canvasview.cpp
an example is in the on_waypoint_clicked_canvasview function
`item->signal_activate().connect(/*connecting the signal*/\
                    sigc::bind(sigc::ptr_fun(set_waypoint_model), waypoint_set, model, canvas_interface()));`
and everything works correctly 
and when I do something very similar in another file as I did here 
`tcb_button->signal_toggled().connect(sigc::bind(sigc::ptr_fun(set_waypoint_model), waypoint_set_g, model, CanvasView::canvas_interface()))`
it surely complains and says member function canvas_interface() cant be called without an object. which makes sense, what doesn't is the former case.

p.s
am I getting too carried away here ?

Keyikedalube:
> CanvasView member function canvas_interface() is used without an object many times in canvasview.cpp

Didn't really get what you meant by that but I'm hoping you mean `CanvasInterface` object itself

`CanvasInterface` object gets instantiated in `CanvasView` constructor initializer lists in line 517
https://github.com/synfig/synfig/blob/08202f400bffb75031b907c7eab02f517ad99fef/synfig-studio/src/gui/canvasview.cpp#L511-L517

In your code I believe `CanvasView` object is non-existent excepting these two protected member functions `init_canvas_view_vfunc(...)` and `changed_canvas_view_vfunc(...)` where the said object exists for a short time (locally). Hence, your call to this `CanvasView::canvas_interface()` fails because the object for `CanvasView` is not instantiated at all so you can't really access its member function `canvas_interface()`

I tried adding a new private member `etl::loose_handle<CanvasView> canvas_view`  and assigned that when either init or changed canvas view vfunc was called. That way we actually have a `canvas_view` object in your modified `Dock_Timtrack2` class. Changed your signal call to
```
if ( tcb_button->signal_toggled().connect(sigc::bind(sigc::ptr_fun(set_waypoint_model), waypoint_set_g, model, canvas_view->canvas_interface())) )
...
```
Code compiles successfully this time but the application crashes on launch.

I don't understand why you have a global function `set_waypoint_model(...)` on `CanvasView` class but that's not my concern LOL.

I hope my small contribution helps.

mohamedAdhamc:
update: realized what was done wrong, we could just use get_canvas_interface() and it works fine for getting the canvas_interface as its name implies xd. Im now working on it and will commit the updates soon.

mohamedAdhamc:
so as far as I see here the WaypointItem is a struct which is used along with WaypointSD to assist in dragging multiple waypoints however I couldnt draw a connection between it and the "waypoint_set" which is of a different -synfig::waypoint- type. I guess I could implement the select drag helper somehow and just set everything up for being able to capture the selection of waypoint_sets and maybe put them in a synfig::WaypointList somehow. However, things are a bit messy right now trying to figure out all these things so in the mean time while im working on it. @ice0 and @Keyikedalube  what do you guys think is this the way to go for this or is there something I'm missing here ?

Keyikedalube:
@mohamedAdhamc I'm sorry I can't help you here because I haven't explored these core parts of Synfig. I only contribute to the frontend ie., GUI, not the backend. @ice0 and @rodolforg are your best guides here.

BTW, hooray to you on your successful implementation of single-waypoints interpolation change!

ice0:
> 2- switch over the toggle buttons into regular gtk buttons as it makes more sense considering the functionality.

Maybe it's better to use Gtk::RadioButton for this?
P.S. I haven't checked if they can only be used with icons.

mohamedAdhamc:
> > 2- switch over the toggle buttons into regular gtk buttons as it makes more sense considering the functionality.
> 
> Maybe it's better to use Gtk::RadioButton for this? P.S. I haven't checked if they can be used with icons only.

  Well part of the reason I felt like maybe the regular buttons would be good for this is because one would just have to click on a single waypoint/select waypoints then hit the button which sets the needed interpolation type and thats it. However I with the check button it may get a bit more complicated due to a few points
1- size : Im also not sure of they can somehow be only used with icons, didnt even know they could be used with icons xd, however if not. Radio buttons would probably too big to fit into the tool panel. would probably have to add to a sepearate perhaps popup tool bar
2- there would have to be processing done on selection of waypoints to have the "tiny mark looking thing" in the proper interpolation type button upon selection of waypoints  to display original type before the users sets the new one. Which also poses another question, on selection of multiple one of different type what should the behavior be ? 

 However, I believe this could probably be debated both ways and the button type wouldn't be really a problem to implement either way. What really needs a lot of work now is getting this to work for multiple waypoints at once 😄.

mohamedAdhamc:
> @mohamedAdhamc I'm sorry I can't help you here because I haven't explored these core parts of Synfig. I only contribute to the frontend ie., GUI, not the backend. @ice0 and @rodolforg are your best guides here.

Yes, no problem😄. Actually, this is part of the reason i'm having a bit more trouble here, its probably one of the first time i come into such direct contact with the core xd.

> 
> BTW, hooray to you on your successful implementation of single-waypoints interpolation change!

Thanks! lets hope the whole feature would be finished soon.

mohamedAdhamc:
update: looks like implemennting a struct which inherits from select drag helper in order to handle waypoint_set selection is the way to go. I'm currently starting its implementation, looks like it might take a bit but lets hope for the best, I believe this is a pretty cool feature so its worth it xD.

mohamedAdhamc:
Got delayed a lot unfortuanetly due to college work which keeps on increasing and never decreases lol. however I was able to do something.

ok, so I used a diffrent approach. Just getting the selected waypoint items and storing them in a vector. then iterating through the vector and setting the waypoints. It worked and changes the selected waypoints interpolation.  I put the code in "on_waypoint_clicked" just to test out if the concept would actually work.

so I guess now I would just have to add a new action which is set when the button is pressed and which calls a new function which I would make with a block of code very similar to the one i tried out here.

mohamedAdhamc:
Alright, finally now I tested it and the functionality works fine. What's left would be just to switch over the radio buttons into regular buttons, and add the separator.

p.s. the additions are all over the place so its a bit hectic xD

mohamedAdhamc:
Hey guys,
@ice0 , @Keyikedalube , finally this feature is now 99% done. one can click on the waypoint then the corresponding interpolation button to set the interpolation. or select multiple waypoints and do the same to set their interpolation.

What's left is just to add the separator which for some reason isnt being added correctly. Also the buttons tool tip which I just realized I forgot to add. 

Excited to hear your reviews guys xd.

p.s this is whats happening with the separator. I will paste the question i asked in the gnome forums.

So I know that tool palettes cant have widgets added to them directly, only tool item groups. I have a group with all the buttons i want and its added and everythings good. then I wanted to add a separator, and i did the regular way, here is a snippet of the code.

Gtk::SeparatorToolItem *seperator = Gtk::manage(new Gtk::SeparatorToolItem());
   seperator->show();
   tool_item_group->insert(*seperator);
but the problem here is that the separator ends up looking like this.

p

also it runs with the error.
Gtk-CRITICAL **: 03:20:29.865: gtk_widget_set_focus_on_click: assertion 'GTK_IS_WIDGET (widget)' failed

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -198,6 +198,74 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
 	return ok;
 }
 
+ void Widget_Timetrack::interpolate_selected(std::string type)
```
ice0, line 201: Use `synfig::Interpolation` instead.

ice0:
Also we are using tabs instead of spaces. Please switch.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
mohamedAdhamc: Yes it makes more sense will change

mohamedAdhamc:
> Also we are using tabs instead of spaces. Please switch.

Yes the laptop was restarted somewhere along the way and the settings were reset to spaces only without me noticing, will fix the indentation though.

rodolforg:
> Any thoughts on the separator issue though ?

These tips here work? https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1SeparatorToolItem.html#details https://docs.gtk.org/gtk3/class.SeparatorToolItem.html

rodolforg:
You can play with GtkIntrospector too to find (live) what property combo works :)

mohamedAdhamc:
> These tips here work? https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1SeparatorToolItem.html#details https://docs.gtk.org/gtk3/class.SeparatorToolItem.html

unfortunately, I tried changing these two as described in the documentation but the only difference it made is that one time I would get a shaded square as sent in the picture and the other time it was completely blank.



> You can play with GtkIntrospector too to find (live) what property combo works :)

This sounds like a good idea. will definetly do so when I get to this feature in my project  :-)

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,74 @@ void Dock_Timetrack2::setup_tool_palette()
 		action_button_map[tool_button->get_name()] = tool_button;
 		tool_item_group->add(*tool_button);
 	}
+    Gtk::RadioButtonGroup interpolation_group;
```
rodolforg, line 236: Delete. Not used.

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,74 @@ void Dock_Timetrack2::setup_tool_palette()
 		action_button_map[tool_button->get_name()] = tool_button;
 		tool_item_group->add(*tool_button);
 	}
+    Gtk::RadioButtonGroup interpolation_group;
+    //button1
+	Gtk::ToolButton *clamped_button = manage(new Gtk::ToolButton(
+													*manage(new Gtk::Image(
+																Gtk::StockID("synfig-interpolation_type_clamped"),
```
rodolforg, line 240: Let's avoid deprecated `StockID`.
create default `Image`; and right after call its `set_from_icon_name()` method. Don't use `Image(icon_name)` constructor because it is only available after gtkmm 3.24, and we intend to support 3.18.

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,74 @@ void Dock_Timetrack2::setup_tool_palette()
...
+																Gtk::IconSize::from_name("synfig-small_icon_16x16") )),
+																"_TCB" ));
+
+	tcb_button->signal_clicked().connect( [this](){
+
+		current_widget_timetrack->set_interpolation(synfig::INTERPOLATION_TCB);
+
+	} );
+
+	tool_item_group->add(*tcb_button);
```
rodolforg, line 263: As this code is a bit repetitive, let's make a loop here.
You can make a vector with those data: icon name, translatable tooltip and Interpolation constant.

At each iteration, you create and the set the toolbutton and add it to toolbar.

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,74 @@ void Dock_Timetrack2::setup_tool_palette()
...
+    //button3
+	Gtk::ToolButton *const_button = manage(new Gtk::ToolButton(
+													*manage(new Gtk::Image(
+																Gtk::StockID("synfig-interpolation_type_const"),
+																Gtk::IconSize::from_name("synfig-small_icon_16x16") )),
+																"_Constant" ));
```
rodolforg, line 269: did you mean `_("Constant")`? This will translate "`Constant`".
Look other similar cases here.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -337,6 +378,10 @@ void Widget_Timetrack::set_action_state(Widget_Timetrack::ActionState action_sta
 {
 	waypoint_sd.set_action(action_state);
 }
+void Widget_Timetrack::set_interpolation(synfig::Interpolation type)
+{
+	waypoint_sd.set_widget_interp(type);
```
rodolforg, line 383: Strange call here: you make Dock tool button call `Widget_Timetrack::set_interpolation()` that only calls `Widget_Timetrack::WaypointSD::set_widget_interp()` that, in its turn, only calls `Widget_Timetrack::interpolate_selected()`.

Why  doesn't `Widget_Timetrack::set_interpolation()` contents is directly the one of `Widget_Timetrack::interpolate_selected()`?

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -198,6 +198,47 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
...
 }
 
+ void Widget_Timetrack::interpolate_selected(synfig::Interpolation type)
+{
+	synfig::Waypoint::Model model;
+	model.set_before(type);
+	model.set_after(type);
+	std::vector<WaypointItem*> selection = waypoint_sd.get_selected_items(); //storing selected waypoint items
+	std::vector<std::set<synfig::Waypoint, std::less<synfig::UniqueID> >> accumilated_selection_set;
+	accumilated_selection_set.clear();
```
rodolforg, line 208: no  need to call `clear()` here: the constructor properly initializes it.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -198,6 +198,47 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
 	return ok;
 }
 
+ void Widget_Timetrack::interpolate_selected(synfig::Interpolation type)
+{
+	synfig::Waypoint::Model model;
+	model.set_before(type);
+	model.set_after(type);
+	std::vector<WaypointItem*> selection = waypoint_sd.get_selected_items(); //storing selected waypoint items
+	std::vector<std::set<synfig::Waypoint, std::less<synfig::UniqueID> >> accumilated_selection_set;
```
rodolforg, line 207: 1. typo: accumulated
2. why do you need an accumulated set?

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -198,6 +198,47 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
...
+    {
+		std::set<synfig::Waypoint, std::less<synfig::UniqueID> > waypoint_set_new; //declaration
+		fetch_waypoints(*selection[i], waypoint_set_new); //setting up waypoint_set_new
+		std::set<synfig::Waypoint, std::less<synfig::UniqueID> >::const_iterator iter;
+		accumilated_selection_set.push_back(waypoint_set_new);
+
+		for(iter=accumilated_selection_set[i].begin();iter!=accumilated_selection_set[i].end();++iter)
+        {
+			synfigapp::Action::PassiveGrouper group(get_canvas_interface()->get_instance().get(),_("Change Waypoint Group"));
+
```
rodolforg, line 220: It seems you'll:
1. create multiples Change Waypoint Group groups
2. set the waypoints from previous iterations multiple times

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: done

synfig-studio/src/gui/widgets/widget_timetrack.cpp
mohamedAdhamc: hmm, I dont really know what I was thinking while doing that xD, anyways I removed the extra unnecessary functions and just left `Widget_Timetrack::interpolate_selected()`

synfig-studio/src/gui/widgets/widget_timetrack.cpp
mohamedAdhamc: it seems like I used it to store the fetched waypoint set and iterate through them -cant remember why-, but yes its not necessary we can use `waypoint_set_new` right away, better to remove it.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
mohamedAdhamc: how about now?

p.s. im uploading the commit in 5 mins if you see this rn

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: for some reason doing that the icon gives the image it gives when the icon isn't found, I went to check how the interpolation icons are set. it seems they are not set the regular way I know of, they are all in the same .sif and there seems to be some configurations done to them.

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: oh yes the gettext, I didn't know about it back then. thanks for the notifying :).

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: done

synfig-studio/src/gui/docks/dock_timetrack2.cpp
rodolforg: The problem probably was the icon name you were using.
These are the real icon names:
https://github.com/synfig/synfig/blob/ac8214e47fd71dbec1066f475e98bd9dda5859e7/synfig-studio/src/gui/iconcontroller.cpp#L335-L357

Please remove the StockID dependency.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -198,6 +198,44 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
...
+	for(int i =0 ; i<selection.size() ; i++) //iteratre through selected waypoint items to fetch the set and apply the interpolation type
+    {
+		std::set<synfig::Waypoint, std::less<synfig::UniqueID> > waypoint_set_new;
+		fetch_waypoints(*selection[i], waypoint_set_new);
+		std::set<synfig::Waypoint, std::less<synfig::UniqueID> >::const_iterator iter;
+
+		for(iter=waypoint_set_new.begin();iter!=waypoint_set_new.end();++iter)
+        {
+			synfigapp::Action::PassiveGrouper group(get_canvas_interface()->get_instance().get(),_("Change Waypoint Group"));
```
rodolforg, line 216: 1. I think you don't need the `.get()` of `get_canvas_interface()->get_instance().get()`
2. The intention of PassiveGrouper is to group several actions inside a single one: with a single Undo, all inside actions are undone at once. So it does not make sense to make several PassiveGroups in a row, each one containing a single action. Place it outside the loops.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -198,6 +198,44 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
...
 }
 
+ void Widget_Timetrack::interpolate_selected(synfig::Interpolation type)
+{
+	synfig::Waypoint::Model model;
+	model.set_before(type);
+	model.set_after(type);
+	std::vector<WaypointItem*> selection = waypoint_sd.get_selected_items(); //storing selected waypoint items
+
+	for(int i =0 ; i<selection.size() ; i++) //iteratre through selected waypoint items to fetch the set and apply the interpolation type
```
rodolforg, line 208: 1. You don't need the comments here. Neither the ones on previous line.
2. Prefer [Range-based for loop](https://www.geeksforgeeks.org/range-based-loop-c/).

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -198,6 +198,44 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
...
+    {
+		std::set<synfig::Waypoint, std::less<synfig::UniqueID> > waypoint_set_new;
+		fetch_waypoints(*selection[i], waypoint_set_new);
+		std::set<synfig::Waypoint, std::less<synfig::UniqueID> >::const_iterator iter;
+
+		for(iter=waypoint_set_new.begin();iter!=waypoint_set_new.end();++iter)
```
rodolforg, line 214: Prefer [Range-based for loop](https://www.geeksforgeeks.org/range-based-loop-c/).

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,29 @@ void Dock_Timetrack2::setup_tool_palette()
...
+		Gtk::ToolButton *tool_button = manage(new Gtk::ToolButton(
+														*manage(new Gtk::Image(
+																	Gtk::StockID(stock_id[i]),
+																	Gtk::IconSize::from_name("synfig-small_icon_16x16"))),
+																	_(name[i].c_str())));
+		tool_button->signal_clicked().connect( [this, interp_type, i](){
+
+			current_widget_timetrack->interpolate_selected(interp_type[i]);
+
+		});
```
rodolforg, line 255: Please avoid lambda signal slot to avoid crash.
Use `sigc::memfun()`. You may use `sigc::bind()` to pass a variable as signal slot (class method) argument.

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,29 @@ void Dock_Timetrack2::setup_tool_palette()
 		action_button_map[tool_button->get_name()] = tool_button;
 		tool_item_group->add(*tool_button);
 	}
+
+	std::vector<std::string> stock_id{ "synfig-interpolation_type_clamped", "synfig-interpolation_type_tcb",
+								  "synfig-interpolation_type_const", "synfig-interpolation_type_ease" ,
+								  "synfig-interpolation_type_linear"};
+	std::vector<std::string> name{ "Clamped", "TCB" , "Constant" , "Ease In/Out" , "Linear"};
```
rodolforg, line 240: To mark string as translatable, but without translating in this code point, use `N_()`, like `N_("Clamped")`.
It does nothing at runtime, but it makes the translatable string parser (i.e. gettext) know that string must be included in .pot/.po files.

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,29 @@ void Dock_Timetrack2::setup_tool_palette()
...
+
+	std::vector<std::string> stock_id{ "synfig-interpolation_type_clamped", "synfig-interpolation_type_tcb",
+								  "synfig-interpolation_type_const", "synfig-interpolation_type_ease" ,
+								  "synfig-interpolation_type_linear"};
+	std::vector<std::string> name{ "Clamped", "TCB" , "Constant" , "Ease In/Out" , "Linear"};
+	std::vector<synfig::Interpolation> interp_type{ synfig::INTERPOLATION_CLAMPED, synfig::INTERPOLATION_TCB,
+												  synfig::INTERPOLATION_CONSTANT, synfig::INTERPOLATION_HALT,
+												  synfig::INTERPOLATION_LINEAR};
+	//button1
+	for (int i=0; i<=4; i++) {
```
rodolforg, line 245: Prefer [Range-based for loop](https://www.geeksforgeeks.org/range-based-loop-c/).
So it would be good to make a struct with fields: icon_name, name/label, type. Instead of 3 `std::vector`s, we would have one only.
It could be initialized like this:
```c++
std::vector<InterpolationInfo> interpolation_info = {
	{"interpolation_type_clamped_icon", N_("Clamped"), synfig::INTERPOLATION_CLAMPED},
	...
};
```

rodolforg (changes requested):

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: I tried doing so before committing as I remembered what you said earlier about the complications that come with the lambda but for some reason, it was giving out seg errors.

mohamedAdhamc: hmm ok, I used the "safe" alternative for the current lambda which is using sigc::track_obj with it as mentioned in the [link ](https://gitlab.gnome.org/GNOME/gtkmm-documentation/-/issues/10)you provided earlier, works like a charm :).

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: alright makes sense thanks!

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: done

synfig-studio/src/gui/widgets/widget_timetrack.cpp
mohamedAdhamc: 1 - i wonder why ? 
     removing gives out the error
   ` error: no matching function for call to 'synfigapp::Action::PassiveGrouper::PassiveGrouper(const etl::loose_handle<synfigapp::Instance>&, char*)'
  208 |         synfigapp::Action::PassiveGrouper group(get_canvas_interface()->get_instance(),_("Change Waypoint Group"));`
2- definetly makes sense, thanks for notifying.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
mohamedAdhamc: done

synfig-studio/src/gui/widgets/widget_timetrack.cpp
mohamedAdhamc: done, although I'm not sure if replacing this const_iter by what I did to use a range-based for loop is safe, so I'm gonna leave this un-resolved just to be sure.

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: yes these were the same names I tried it still gives out the image missing icon for some reason :(

mohamedAdhamc:
@rodolforg done, now all the req changes should be addressed (although some I wasnt so sure so I left unresolved). However the icons are still not showing and giving out the image missing icon :(.

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,31 @@ void Dock_Timetrack2::setup_tool_palette()
...
+		{"interpolation_type_tcb_icon", N_("TCB"), synfig::INTERPOLATION_TCB},
+		{"interpolation_type_const_icon", N_("Constant"), synfig::INTERPOLATION_CONSTANT},
+		{"interpolation_type_ease_icon", N_("Ease In/Out"), synfig::INTERPOLATION_HALT},
+		{"interpolation_type_linear_icon", N_("Linear"), synfig::INTERPOLATION_LINEAR}
+	};
+
+	for (const auto & interp_button_info: interp_buttons_info) {
+		Gtk::Image* image= Gtk::manage(new Gtk::Image());
+		image->set_from_icon_name(interp_button_info.icon_name, Gtk::IconSize::from_name("synfig-small_icon_16x16"));
+		Gtk::ToolButton *tool_button = manage(new Gtk::ToolButton(*image, interp_button_info.name));
```
rodolforg, line 254: `_(interp_button_info.name)` to actually translate it :)

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,31 @@ void Dock_Timetrack2::setup_tool_palette()
 		action_button_map[tool_button->get_name()] = tool_button;
 		tool_item_group->add(*tool_button);
 	}
+
+	struct interpolationButtonInfo {
```
rodolforg, line 237: start type names with capital letter: `InterpolationButtonInfo`

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -198,6 +198,40 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
...
+	model.set_after(type);
+	std::vector<WaypointItem*> selection = waypoint_sd.get_selected_items();
+
+	synfigapp::Action::PassiveGrouper group(get_canvas_interface()->get_instance().get(),_("Change Waypoint Group"));
+
+	for (WaypointItem* waypoint_item : selection) {
+		std::set<synfig::Waypoint, std::less<synfig::UniqueID> > waypoint_set_new;
+		fetch_waypoints(*waypoint_item, waypoint_set_new);
+
+		for (synfig::Waypoint waypoint_new : waypoint_set_new) {
```
rodolforg, line 214: To avoid unnecessary copy: `const synfig::Waypoint&`

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -198,6 +198,40 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
...
 }
 
+ void Widget_Timetrack::interpolate_selected(synfig::Interpolation type)
+{
+	synfig::Waypoint::Model model;
+	model.set_before(type);
+	model.set_after(type);
+	std::vector<WaypointItem*> selection = waypoint_sd.get_selected_items();
+
+	synfigapp::Action::PassiveGrouper group(get_canvas_interface()->get_instance().get(),_("Change Waypoint Group"));
```
rodolforg, line 208: check if `selection` is empty before call the PassiveGrouper.
Otherwise the Undo history will have an empty action added.
Or does the PassiveGrouper self deletes if empty?

rodolforg (changes requested):

synfig-studio/src/gui/docks/dock_timetrack2.cpp
rodolforg: Maybe, I can't remember. Maybe I was just experimenting `sigc::track_obj` XD

synfig-studio/src/gui/docks/dock_timetrack2.cpp
rodolforg: I guess it is that way because your PR is based on a commit previous 865f1977d0a3e4129a1031cc39d7ea7514bc7158 :)

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -233,6 +233,31 @@ void Dock_Timetrack2::setup_tool_palette()
...
 	}
+
+	struct interpolationButtonInfo {
+		std::string icon_name;
+		std::string name;
+		synfig::Interpolation interpolation;
+	};
+
+	const std::vector<interpolationButtonInfo> interp_buttons_info{
+		{"interpolation_type_clamped_icon", N_("Clamped"), synfig::INTERPOLATION_CLAMPED},
```
ice0, line 244: Please rebase and use `interpolation_icon_name` from `iconcontroller.cpp` to get icon names.

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: oops for some reason I thought N_("") was enough :"

synfig-studio/src/gui/widgets/widget_timetrack.cpp
mohamedAdhamc: I am not sure if it "self deletes" but clicking the buttons while nothing is selected doesnt make any new action and undo seems to be working as expected so I'm assuming it is taken care of ?

But similar to this I think I should check if the new interpolation is different or not to avoid making a new action and going through the methods if there is no change.

mohamedAdhamc:
ok, I think now everything is good I'm just rebuilding to be sure then I will update here

mohamedAdhamc:
ok so back to the separator issue, so I think it looks kind of off and this is confirmed by the warning I get on running 
`(synfigstudio.exe:8104): Gtk-CRITICAL **: 12:54:56.334: gtk_widget_set_focus_on_click: assertion 'GTK_IS_WIDGET (widget)' failed`. I think it might be a bug in gtk, I had tried asking about it in their forums before but to no response :".

synfig-studio/src/gui/widgets/widget_timetrack.cpp
rodolforg: > But similar to this I think I should check if the new interpolation is different or not to avoid making a new action and going through the methods if there is no change.

Indeed :)

rodolforg:
> ok so back to the separator issue, so I think it looks kind of off

It's a Gtk theme issue. I think we should not use the separator :/

@morevnaproject @ice0

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -229,6 +230,34 @@ void Dock_Timetrack2::setup_tool_palette()
...
+		{N_("TCB"), synfig::INTERPOLATION_TCB},
+		{N_("Constant"), synfig::INTERPOLATION_CONSTANT},
+		{N_("Ease In/Out"), synfig::INTERPOLATION_HALT},
+		{N_("Linear"), synfig::INTERPOLATION_LINEAR}
+	};
+
+	for (const auto & interp_button_info: interp_buttons_info) {
+		Gtk::Image* image= Gtk::manage(new Gtk::Image());
+		image->set_from_icon_name(interpolation_icon_name(interp_button_info.interpolation), Gtk::IconSize::from_name("synfig-small_icon_16x16"));
+		Gtk::ToolButton *tool_button = manage(new Gtk::ToolButton(*image, _("interp_button_info.name")));
```
rodolforg, line 254: `_("interp_button_info.name")` this is not right.
It will translate the string `"interp_button_info.name"` instead of the variable contents.

Use `_(interp_button_info.name)`

rodolforg (changes requested):

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: oops my bad didn't notice.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -199,6 +199,42 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
...
```
rodolforg, line 209: "Change Waypoint Group" -> "Change Waypoint Interpolation"

mohamedAdhamc: ok, makes sense.

mohamedAdhamc:
@rodolforg done, should the separator be removed?

ice0:
> It's a Gtk theme issue. I think we should not use the separator :/

Where did you find this information?

As for me it looks like this feature just not implemented for `Gtk::ToolPalette` widget.

These are images of how the Gtk developers think it should be used. And I don't see any separators here.
Since the `ToolPalette` is a grid, it is not clear what to do if a separator is added. Should you split it horizontally or vertically?
How should you position the buttons?

Moreover, in the gtk tests, the separator is only tested with the toolbar.

https://github.com/GNOME/gtk/blob/51e636be0f1f154a3a007892401fc51104f88ca1/tests/testtoolbar.c#L609-L611

image

image

https://lazka.github.io/pgi-docs/Gtk-3.0/classes/ToolPalette.html
https://developer-old.gnome.org/gtkmm-tutorial/3.24/toolpalette-example.html.en

After switching from `Gtk::ToolPalette` to `Gtk::Toolbar` all works without issues.

Screenshot_110

@mohamedAdhamc 
I suggest making these buttons disabled if there are no waypoints selected and add tooltips to them.

mohamedAdhamc:
> @mohamedAdhamc I suggest making these buttons disabled if there are no waypoints selected and add tooltips to them.

while switching over to `Gtk::ToolPalette' or keeping the current 'Gtk::Toolbar' ? I'm assuming it would be the latter, but I'm just making sure here.

ice0:
> > @mohamedAdhamc I suggest making these buttons disabled if there are no waypoints selected and add tooltips to them.
> 
> while switching over to `Gtk::ToolPalette' or keeping the current 'Gtk::Toolbar' ? I'm assuming it would be the latter, but I'm just making sure here.

I suggest to switch to `Gtk::Toolbar` from `Gtk::ToolPalette`

rodolforg:
> Where did you find this information?

https://docs.gtk.org/gtk3/class.SeparatorToolItem.html
> Depending on the theme, a GtkSeparatorToolItem will often look like a vertical line on horizontally docked toolbars.

mohamedAdhamc:
@ice0  done.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -956,6 +993,13 @@ void Widget_Timetrack::on_waypoint_action_changed()
 	signal_action_state_changed().emit();
 }
 
+void Widget_Timetrack::on_waypoint_selection_changed()
+{
+	std::vector<WaypointItem*> selection = waypoint_sd.get_selected_items();
+	bool show_tool_bar_buttons = (selection.size() != 0);
+	signal_update_interpolation_buttons_visiblity().emit(show_tool_bar_buttons);
```
rodolforg, line 1000: Instead of this signal name, that describe an action other entity should perform, the name should describe what event happened. Example: waypoint_selection_changed

rodolforg (changes requested):

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -119,7 +123,7 @@ void Dock_Timetrack2::changed_canvas_view_vfunc(etl::loose_handle<CanvasView> ca
 
 		hscrollbar.unset_adjustment();
 
-		tool_palette.hide();
+		tool_bar->hide();
```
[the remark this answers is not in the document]
mohamedAdhamc: done

synfig-studio/src/gui/widgets/widget_timetrack.cpp
mohamedAdhamc: oh yes I went a little unorthodox with this one i guess xd, done

mohamedAdhamc:
@rodolforg done :)

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -133,15 +137,15 @@ void Dock_Timetrack2::changed_canvas_view_vfunc(etl::loose_handle<CanvasView> ca
...
+		update_toolbar_action();
+		set_interp_buttons_visibility(current_widget_timetrack->get_num_waypoints_selected());
 
 		grid.attach(widget_kf_list,            0, 0, 1, 1);
 		grid.attach(widget_timeslider,         0, 1, 1, 1);
 		grid.attach(*current_widget_timetrack, 0, 2, 1, 1);
 		grid.attach(hscrollbar,                0, 4, 2, 1);
 		grid.attach(vscrollbar,                1, 0, 1, 4);
-		grid.attach(tool_palette,              2, 0, 1, 4);
+		grid.attach(*toolbar, 		       2, 0, 1, 4);
```
rodolforg, line 148: don't use `tab` to align, only for indenting

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -227,13 +232,42 @@ void Dock_Timetrack2::setup_tool_palette()
...
+
+	for (const auto & interp_button_info: interp_buttons_info) {
+		Gtk::Image* image= Gtk::manage(new Gtk::Image());
+		image->set_from_icon_name(interpolation_icon_name(interp_button_info.interpolation), Gtk::IconSize::from_name("synfig-small_icon_16x16"));
+		Gtk::ToolButton *tool_button = manage(new Gtk::ToolButton(*image, _((interp_button_info.name).c_str())));
+		tool_button->signal_clicked().connect(sigc::track_obj([this, interp_button_info](){
+			current_widget_timetrack->interpolate_selected(interp_button_info.interpolation);
+		}, *this));
+		std::string tool_tip = N_("Change waypoint interpolation to " + interp_button_info.name);
+		tool_button->set_tooltip_text(_(tool_tip.c_str()));
```
rodolforg, line 263: No, you should split/join string you want to translate.
Different languages have different word order in their grammars.

Use string format instead.

```c++
strprintf(_("Change waypoint interpolation to %s"), interp_button_info.name);
```

rodolforg (dismissed):

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: oops, didn't know that, Done

synfig-studio/src/gui/docks/dock_timetrack2.cpp
mohamedAdhamc: alright, makes sense. Done, not exactly sure if it is how it should be thougg, so i will leave this open

mohamedAdhamc: @rodolforg can't I just remove the N_ on line 262 and concactenate the string regularly then use the (gettext) _() on 263 ?

rodolforg: No, you can't. It breaks proper translations.

The problem was I typed/wrote the wrong code. This is right:

```
strprintf(_("Change waypoint interpolation to %s"), interp_button_info.name.c_str());
```

mohamedAdhamc: oh, yes fair enough it now works well.

mohamedAdhamc:
@rodolforg done. Anything else needed here ?

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -249,3 +282,14 @@ void Dock_Timetrack2::update_tool_palette_action()
 	if (button)
 		button->set_active(true);
 }
+
+void Dock_Timetrack2::set_interp_buttons_visibility(bool visible)
+{
+	for (int i = 0; i < toolbar->get_n_items(); i++){
```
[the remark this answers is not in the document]
mohamedAdhamc: I agree as well, for some reason i misread the suggestion as hiding instead of disabling :).  done

synfig-studio/src/gui/docks/dock_timetrack2.cpp
```diff
@@ -227,13 +232,41 @@ void Dock_Timetrack2::setup_tool_palette()
...
+	for (const auto & interp_button_info: interp_buttons_info) {
+		Gtk::Image* image= Gtk::manage(new Gtk::Image());
+		image->set_from_icon_name(interpolation_icon_name(interp_button_info.interpolation), Gtk::IconSize::from_name("synfig-small_icon_16x16"));
+		Gtk::ToolButton *tool_button = manage(new Gtk::ToolButton(*image, _((interp_button_info.name).c_str())));
+		tool_button->signal_clicked().connect(sigc::track_obj([this, interp_button_info](){
+			current_widget_timetrack->interpolate_selected(interp_button_info.interpolation);
+		}, *this));
+		tool_button->set_tooltip_text(synfig::strprintf(_("Change waypoint interpolation to %s"), interp_button_info.name.c_str()));
+		tool_button->set_name(interp_button_info.name);
+		toolbar->append(*tool_button);
```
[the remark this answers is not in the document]
mohamedAdhamc: looks great, done.

synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -199,6 +200,42 @@ bool Widget_Timetrack::move_selected(synfig::Time delta_time)
...
+	synfigapp::Action::PassiveGrouper group(get_canvas_interface()->get_instance().get(),_("Change Waypoint Interpolation"));
+
+	for (WaypointItem* waypoint_item : selection) {
+		std::set<synfig::Waypoint, std::less<synfig::UniqueID> > waypoint_set_new;
+		fetch_waypoints(*waypoint_item, waypoint_set_new);
+
+		for (const synfig::Waypoint& waypoint_new : waypoint_set_new) {
+			synfig::Waypoint waypoint(waypoint_new);
+			if (waypoint.get_before() != type || waypoint.get_after() != type){
+				waypoint.apply_model(model);
```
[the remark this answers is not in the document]
mohamedAdhamc: hmm, I'm not really ssure what you mean here.  I thought maybe you mean to set the waypoint parameter for multiple waypoints then perform the action once, but it gives out errors (also I think setting the parameter "waypoint" multiple time before perform action willl just overwrite it so its probably not what you meant)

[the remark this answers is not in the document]
ice0: > hmm, I'm not really ssure what you mean here. I thought maybe you mean to set the waypoint parameter for multiple waypoints then perform the action once, but it gives out errors 

Yes.

> (also I think setting the parameter "waypoint" multiple time before perform action will just overwrite it so its probably not what you meant)

You can check `WaypointSet` action to be sure it doesn't - it adds it to the vector. 

The only issue I see here, is the `value_node` parameter. I don't know why it is passed, if it can be resolved from waypoint parameter. Also `value_node` overwrites previous value, so you need to check if:
1) it is used
2) it can't be resolved from waypoint parameter.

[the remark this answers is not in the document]
mohamedAdhamc: > You can check `WaypointSet` action to be sure it doesn't - it adds it to the vector.

oh yes I checked it does add them to a vector but what is puzzling me is the value node, it is also what is causing the error when trying to perform the action for all the selected waypoints at once, the value node isn't set for each one of them. In fact I'm confused how it could be set for each one of them when there is only one value node handle in the `WayPointSet` ... maybe we need a value node vector similar to the waypoint vector or maybe there is one i'm not sure i will try to investigate here a bit.

> The only issue I see here, is the `value_node` parameter. I don't know why it is passed, if it can be resolved from waypoint parameter. Also `value_node` overwrites previous value, so you need to check if:
> 
> 1. it is used
> 2. it can't be resolved from waypoint parameter.

1- I believe yes it is used as it seems for the action `WaypointSet` to be performed it needs the waypoint and the valuenode of the waypoint, failure to have the valuenode gives out the error `unable to find waypoint` 

2- i'm not sure what you mean exactly by resolve, however the valuenode of the waypoint can be gotten from the waypoint itself using for example `get_parent_value_node`, im thinking maybe i can add an extra default parameter for doing waypoint related actions like this without having to set the valuenode explicitly for each one, to make it implicit by making the valuenode be set from `WaypointSet` itself and stored in a vector. But I have to first understand a bit more of whats going on, so I will have to investigate further.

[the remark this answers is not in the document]
mohamedAdhamc: @ice0 , @rodolforg would this change be necessary here though, as in this PR. from what I see currently a similar situation which is setting interpolation from right mouse click groups the set waypoint actions into a passive group similar to what is done here, however to instead set the action once then waypointset class would have to be modified somewhat significantly.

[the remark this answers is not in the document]
ice0: > would this change be necessary here though

Ok, this can be fixed later.

ice0 (approved):

ice0:
@morevnaproject Can you test it?

ice0:
Merged. Thank you!

PR #2620, opened 2022-04-27
Transformation submenu now always expanded

### LINKED ISSUE #466: Transformation submenu behavior
S1.2.1 / W10 64bit / 1709

**Issue description**:
The behavior of the submenu  "transformation" is very annoying. Example: I want to change the "Angle" of say 10 objects from 0.00° to 30°.
These is my workflow now:
1. Click on object one
2. Click on the triangle "Transformation" in "Parameters"
3. The Transformation Submenu opens
4. Doubleclick on "Angle"
5. Change the Value
6. Click on object two (-> The Transformation submenue closes, grrrrrr)
7. Click on the triangle "Transformation" in "Parameters"
...

1st:
the Transformation submenu closing is VERY annoying. 
2nd:
There would be a much better option if i could select all 10 objects (with ctrl+click) and then change the values all together. But if i do so, the "Transormation"-submenue is grayed out.

Thank you
Blaggejoe

Keyikedalube:
Fixes #466

### DIFF
--- a/synfig-studio/src/gui/trees/layerparamtreestore.cpp
+++ b/synfig-studio/src/gui/trees/layerparamtreestore.cpp
@@ -477,7 +477,7 @@ LayerParamTreeStore::queue_rebuild()
 			sigc::mem_fun(*this,&LayerParamTreeStore::rebuild),
 			false
 		)
-	,150);
+	,20);
 
 }
 
--- a/synfig-studio/src/gui/trees/layertree.cpp
+++ b/synfig-studio/src/gui/trees/layertree.cpp
@@ -805,6 +805,23 @@ LayerTree::on_layer_tree_event(GdkEvent *event)
 		}
 		break;
 	case GDK_BUTTON_RELEASE:
+		{
+			// Transformation parameter property is in 5th row
+			// Row is expanded only once and not recursively to avoid lengthy tree
+			auto transform_row   = param_tree_view().get_model()->children()[4];
+			auto layer_rows_path = layer_tree_view().get_selection()->get_selected_rows();
+			bool is_group_layer  = true;
+
+			for (auto path : layer_rows_path) {
+				if (!layer_tree_view().get_model()->get_iter(path)->children()) {
+					is_group_layer = false;
+					break;
+				}
+			}
+
+			if (is_group_layer && transform_row.children())
+				param_tree_view().expand_row(Gtk::TreePath("4"), false);
+		}
 		break;
 	default:
 		break;
--- a/synfig-studio/src/gui/trees/layertree.cpp
+++ b/synfig-studio/src/gui/trees/layertree.cpp
@@ -737,8 +737,8 @@ bool
 LayerTree::on_layer_tree_event(GdkEvent *event)
 {
 	SYNFIG_EXCEPTION_GUARD_BEGIN()
-    switch(event->type)
-    {
+	switch(event->type)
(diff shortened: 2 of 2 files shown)


ice0:
What if I want this submenu to be collapsed?

I think it's better to save the expanded/collapsed state and restore it when you switch to another object.

Svarov-RZM:
> What if I want this submenu to be collapsed?

I am pretty sure that no one wants this menu collapsed, there's no reason for that. It should have been expanded from the very beginning because transformational parameters **are essential in animation process** and therefore should be easily accessible.

Personally, I would remove this collapsing group completely and put all parameters in the root. 'Z range' parameters are done this way, so why 'transformation' parameters got different treatment? I don't see a good reason.

morevnaproject:
If we will implement the possibility to save the expanded/collapsed state of any parameter (as @ice0 suggested), then it will solve the problem with transformation parameter. And we can have it expanded by default. On the other hand, if user will want to have it collapsed, then it will be possible. So that will be a win-win for everyone.

Also, we will be able to apply the same approach to the list of spline vertices - user will be able to expand/collapse it and this state will be remembered, which is also very useful.

> Personally, I would remove this collapsing group completely and put all parameters in the root. 'Z range' parameters are done this way, so why 'transformation' parameters got different treatment? I don't see a good reason.

There are several reasons for that.

1. This is related with mechanism how groups are connected to bones. Bone defines whole transformation of group, so the group's Transformation parameter is connected to Bone's Transformation.
2. Transformation parameter displays "merged" waypoints of all sub-parameters and this is useful in some cases when user works with animation.

P.S. I agree that Transformation parameter should be expanded by default.

ice0:
In this case, I propose to split this problem into two smaller ones.
1) Add a flag to ParamDesc (something like `expanded`) which will allow to set default expanded state for any parameter (current PR could be improved to do this).
2) Add the ability to save the expanded/collapsed state for all parameters.

Keyikedalube:
I'll work on this later. Maybe this weekend. Sorry for the delay.

I agree with @ice0 saving expand/collapse states.

But what happens when user selects multiple layers where some transformation group are collapsed and some expanded? Override and expand all? Or collapse all?

Keyikedalube:
I'm a bit lost why setting new updated row like this breaks the program:
```c++
// on_param_row_expanded() function body
auto row = *iter;
auto new_param_desc = cellrenderer_value->get_param_desc();

if (new_param_desc.get_name() == "transformation")
	row[param_model.param_desc] = new_param_desc.set_row_expanded();
```

ice0:
> I'm a bit lost why setting new updated row like this breaks the program:

Can you add more details? I added this code to this PR and I don't know what to expect.

Keyikedalube:
> > I'm a bit lost why setting new updated row like this breaks the program:
> 
> Can you add more details? I added this code to this PR and I don't know what to expect.

Transformation parameter name changes to other random ValueBase names. Keep expanding/collapsing that row and sometimes you'd get "I'M A BUG SO REPORT ME" label displayed.

The doc for gtk treeview mentions one to get the model and modify that in order to update the user's treeview. I feel I'm doing exactly that but I might be missing something here.

Keyikedalube:
Inactive for 3 months, yikes!

Sorry for the delay. Was lost in other stuff + mostly wasted time not being productive.
Be back soon-ish! I'll finish this PR.

PR #2622, opened 2022-05-02
refactor(core): Cleaned up registration function

ice0:
Removed all unnecessary "magic"

### DIFF
--- a/synfig-core/src/synfig/valuenode_registry.cpp
+++ b/synfig-core/src/synfig/valuenode_registry.cpp
@@ -57,10 +57,11 @@ String
 ValueNodeRegistry::localize_name(const String &local_name)
 	{ return _(local_name.c_str()); }
 
-void
+bool
 ValueNodeRegistry::register_node_type(const String &name, const String &local_name, ReleaseVersion version, Factory factory, CheckType check_type)
 {
 	book().insert({name, {local_name, factory, check_type, version}});
+	return true;
 }
 
 bool
--- a/synfig-core/src/synfig/valuenode_registry.h
+++ b/synfig-core/src/synfig/valuenode_registry.h
@@ -35,38 +35,14 @@
 /* === M A C R O S ========================================================= */
 
 /// This macro automatically registers valuenode and defines its
-/// get_name & get_local_name functions.
-/// NOTE: this relies on valuenode being either in synfig namespace
-/// or in 'used' namespace.
-/// TODO: move all nodes into synfig::valuenodes namespace
-#define REGISTER_VALUENODE(klass, _version, _name, _local_name)				\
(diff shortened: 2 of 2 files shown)


synfig-core/src/synfig/valuenode_registry.cpp
```diff
@@ -52,23 +52,13 @@ String
 ValueNodeRegistry::localize_name(const String &local_name)
 	{ return _(local_name.c_str()); }
 
-void
+bool
```
rodolforg, line 55: why `bool` if it always will return true?

synfig-core/src/synfig/valuenode_registry.h
```diff
@@ -35,38 +35,13 @@
...
-			return _local_name;												\
-		}																	\
-																			\
-		const ReleaseVersion Register_##klass::release_version = _version;	\
-		const char* Register_##klass::name = _name;							\
-		const char* Register_##klass::local_name = _local_name;				\
-	}
+/// get_name & get_local_name methods.
+#define REGISTER_VALUENODE(klass, _version, _name, _local_name) \
+static bool result_##klass = synfig::ValueNodeRegistry::register_node_type(_name, \
```
rodolforg, line 40: Don't you think indentation would help realize next lines are part of `#define`?

synfig-core/src/synfig/valuenode_registry.cpp
ice0: To call the auto-registration function "automatically", we create a static variable. In this case, we need some sort of return value.

So all the "magic" of auto-registration is to make something like this:
```
static bool result = register_me("class_name");
```
You can't do this on void function.

synfig-core/src/synfig/valuenode_registry.h
ice0: How do you propose to indent them?

ice0: @rodolforg 
Done.

hesham-essam:
This is much better, and it works with msvc. The old trick is such a head scratcher, I still don't know what does that even mean
```
template<do_register&> struct register_ref {}; // do_register is a variable not a type!
```

ice0:
> This is much better, and it works with msvc. The old trick is such a head scratcher, I still don't know what does that even mean
> 
> ```
> template<do_register&> struct register_ref {}; // do_register is a variable not a type!
> ```

Same for me. It looks like a hack for MSVC compiler, but I am not sure about it.

Anyway, if you check/review the PR and don't see any issues, please approve it, just to show that someone else has checked it too.

hesham-essam (approved):

morevnaproject (approved):

PR #2589, opened 2022-03-23
Use GtkBuilder to create App Menu

nickleus27:
This is a draft to update deprecated ```Gtk::UIManager``` methods to create the applications menu bar.  Instead this pull request uses ```Gtk::Builder``` to creae application menu bar.  This is currently a work in progress.

UPD:
Discussion start here: https://github.com/synfig/synfig/pull/2498#issuecomment-1054922175

PR #2624, opened 2022-05-03
Added basic support for Gtk theme

ice0:
- Icons moved to `128x128` folder
- Added index.theme
- `File -> Save as` icon switched to IconTheme icon (as example)

This PR will allow seamless transition to Gtk::IconTheme.

**Details:**
Almost all GTK methods using icons from the default GTK theme.
Unfortunately, we can't change the IconTheme name if we get it from the default screen with the `Gtk::IconTheme::get_default()` method (https://docs.gtk.org/gtk3/method.IconTheme.set_custom_theme.html)

> Sets the name of the icon theme that the GtkIconTheme object uses overriding system configuration. This function cannot be called on the icon theme objects returned from gtk_icon_theme_get_default() and gtk_icon_theme_get_for_screen().

Also, I didn't find a way to change the app IconTheme to a custom one. 
However, we can change the IconTheme for the default screen using the `Gtk::Settings` object.

It also fixes the problem with blurry icons on HiDPI screens. As you can see, the IconTheme icon is no longer blurry.

Screenshot_34

P.S. I'm not sure if I copied the file in Automake correctly, so any help is appreciated.

### DIFF
--- a/synfig-studio/images/CMakeLists.txt
+++ b/synfig-studio/images/CMakeLists.txt
@@ -242,7 +242,8 @@ set(IMAGES
 	${ICONS}
 )
 
-set(PIXMAPS_DIR ${SYNFIG_BUILD_ROOT}/share/synfig/icons/classic)
+set(THEME_DIR ${SYNFIG_BUILD_ROOT}/share/synfig/icons/classic)
+set(PIXMAPS_DIR ${THEME_DIR}/128x128)
 
 file(MAKE_DIRECTORY ${PIXMAPS_DIR})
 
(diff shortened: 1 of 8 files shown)


rodolforg:
> Also, I didn't find a way to change the app IconTheme to a custom one.
> However, we can change the IconTheme for the default screen using the `Gtk::Settings` object.

Please see `app.cpp` in #1508 . Does it help?

ice0:
> > Also, I didn't find a way to change the app IconTheme to a custom one.
> > However, we can change the IconTheme for the default screen using the `Gtk::Settings` object.
> 
> Please see `app.cpp` in #1508 . Does it help?

```
	Glib::RefPtr<Gtk::Settings> gtk_settings = Gtk::Settings::get_default();

	if(char *theme_name = getenv("SYNFIG_GTK_THEME"))
		gtk_settings->property_gtk_theme_name() = theme_name;
```

This is the same approach.

ice0:
What do you think about merge #1508?
After that I can add changes from this PR (CMake mostly).

This need to start merging #2589

PR #2632, opened 2022-05-15
Added basic support for Gtk theme

ice0:
This is rebased PR #1508 and CMake improvements from #2624 

P.S. I also excluded `Widget_InterpolationEnum` because I'm not sure if it's needed.


morevnaproject (approved):
Cool!

Issue #1077, opened 2020-01-07
Feature Request: Show/Hide rulers


hellocatfood:
**Synfig version & platform**:
Synfig version 1.3.11 on Ubuntu 19.10

**Issue description**:
Can an option be added to show or hide the rulers? It would save a bit of vertical and horizontal space.

I would suggest that it go in the View menu under the options for Guides and Ctrl + Shift + R as a keyboard shortcut.

(this was discussed in for the forum in [2010](https://forums.synfig.org/t/52-notes/837) and [2014](https://forums.synfig.org/t/konstantins-weekly-report-48/3041)

rodolforg:
@morevnaproject  in order to implement this, should the "arrow" menu-button be moved to the main workarea toolbar?
I was already thinking about implement a "[feature-discovery](https://materializecss.com/feature-discovery.html)"-like effect when user hides the main-menu to show from where you can access main-menu back.
Something could be done in this case also.

morevnaproject:
I think it is okay to hide "arrow" menu-button, because user can access the same menu by right-clicking on empty space of workarea.

"feature-discovery" effect is cool.

hellocatfood:
By feature discovery do you mean something like this?

![hide_rulers](https://user-images.githubusercontent.com/151947/72450160-cad64300-37b1-11ea-9bc8-17d6c6b2c088.gif)

Looks cool but I think having the rulers (or main menu) animate when they're hidden doesn't give any clues how to unhide it.

In the example link you gave the animation shows where a menu originated from and where a user has to click/press to reveal it again. If the rulers are hidden in the same way it doesn't disappear to a "tap target" that can be clicked/tapped on to reveal them again.

rodolforg:
@hellocatfood the link shows a demonstration :) The pop-up (labeled) circle could appear on any event

hellocatfood:
Ah, I see. This is starting to go off track so I'll make a new feature request, but instead of the feature-discovery idea you shard perhaps a Status Bar would be better. There's lots of examples of implementation in GIMP, Inkscape, and Blender

mohamedAdhamc:
Hey guys, would it also be of any use to allow for hiding the vertical and horizontal rulers separately? I believe they should both be hidden together but just making sure.

mohamedAdhamc:
@ice0 can this be safely closed now ?

PR #2634, opened 2022-05-16
feat: Show/Hide rulers 

mohamedAdhamc:
as asked for in #1077 
whats left to do:
1- currently there is no icon for the ruler so I will have to make one myself im guessing.
2- probably would be useful to also add this in the menu as is done for the other similar buttons.

this is a how it works as of now 
before:
before ruler hide
after:
after ruler hide


mohamedAdhamc:
Hey @ice0 , thanks for the review 
Regarding 
1- I had a feeling that this method of implementation could possibly cause problems, I will look into that for sure.

2- problem is that this features goal I believe was to save space, So one would have to not only hide the ruler but also the corresponding grid row/column. I kept looking for a workaround to not have to remove and reinsert the grids rows/colums something like a hide()/show() for the gtk grid columns/rows but there weren't any. So the only way was to manually remove and reinsert the specified rows/columns.

3- yes the duplication here is surely not good I will try to make the gtk::event box a class member then so I wouldnt have to recreate it. Although that requires adding the event box header in the workarea header, but yes I guess it would be worth it, to avoid the duplication.

ice0:
@mohamedAdhamc 

> I kept looking for a workaround to not have to remove and reinsert the grids rows/colums something like a hide()/show() for the gtk grid columns/rows but there weren't any.

Row/column will hide if you hide all widgets in them.

mohamedAdhamc:
> @mohamedAdhamc
> 
> > I kept looking for a workaround to not have to remove and reinsert the grids rows/colums something like a hide()/show() for the gtk grid columns/rows but there weren't any.
> 
> Row/column will hide if you hide all widgets in them.

Oh that's great actually, then that's what needs to be done.

ice0:
1. Looks good, but you can simplify the code even further:
```
	ruler_status = !ruler_status;
	hruler->set_visible(ruler_status);
	...
```
2. I suggest renaming the method to `WorkArea::show_hide_ruler()` to better reflect its purpose.
3. I've noticed that `Gtk::Arrow` is deprecated, so I suggest replacing it with `Gtk::Button` (in a separate PR).

ice0:
> 1- currently there is no icon for the ruler so I will have to make one myself im guessing.

If you'd like, you can ask the artists on the Synfig forums for help.

I also suggest saving the state of the rulers in the settings file.

mohamedAdhamc:
> 1. Looks good, but you can simplify the code even further:
> 
> ```
> 	ruler_status = !ruler_status;
> 	hruler->set_visible(ruler_status);
> 	...
> ```
> 
> 2. I suggest renaming the method to `WorkArea::show_hide_ruler()` to better reflect its purpose.
> 3. I've noticed that `Gtk::Arrow` is deprecated, so I suggest replacing it with `Gtk::Button` (in a separate PR).

All sounds good, will be working on them tonight hopefully. Also yes, thats a great idea I will in the meantime ask in the forums also that way we could get some opinions about the icons I guess.

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1361,6 +1361,16 @@ CanvasView::create_right_toolbar()
 		displaybar->append(*snap_grid);
 	}
 
+	{
+		Gtk::ToggleToolButton *show_ruler = Gtk::manage(new Gtk::ToggleToolButton());
```
ice0, line 1365: I suggest to rename it to `toggle_ruler`. `WorkArea::show_hide_ruler()` too.

mohamedAdhamc: Yes I guess that makes it more unified with the other toggle buttons, will do.
 I'm now working on saving the meta data correctly so that when the file is reopened the button is as left, like the others.

mohamedAdhamc:
hey, @ice0 
Now the setting is saved so when the ruler is hidden in a canvas and reopened it remains hidden and vice versa. Also default for creating new canvas is with ruler, which is the regular of course. I'm assuming this is what you meant yesterday by saving the ruler setting, if not though let me know.

as of right now I believe whats left is:
1- icon 
2- the menu check button for the feature, if it is necessary I'm assuming it is...?
- replacement of deprecated Gtk::Arrow but that's going to be a different pull request 😀.

mohamedAdhamc:
Update: now the check button is added and its synced with the tool button. and the meta data is saved correctly so the ruler position is left as is when reloading the saced canvas.
yet to do :
1- add a ruler logo for the tool button
2- and of course the tooltip

mohamedAdhamc:
Icon and tooltip added. However, I think maybe the icon color needs to be greyer. So i'll just wait for you guy's opinionson this and if so I would just place the icon with the new edited verision.

this is how it looks now 
show ruler icon

ice0:
@morevnaproject 

Do we need an explicit button to toggle the rulers, or would a menu item suffice?
I mean I don't see any reason to overload the UI.

mohamedAdhamc:
Hey, @ice0 so should I remove the button, and just leave the menu item ?

ice0:
> Hey, @ice0 so should I remove the button, and just leave the menu item ?

I suggest to remove the button, we can add it later if users need it.

mohamedAdhamc:
> 
> I suggest to remove the button, we can add it later if users need it.

sounds good will do

mohamedAdhamc:
done, I also retested it to make sure the meta data is saved correctly on reopening the file.

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -2701,6 +2707,18 @@ CanvasView::set_onion_skins()
 	work_area->set_onion_skins(onion_skins);
 }
 
+void
+CanvasView::toggle_show_ruler()
+{
+	if(toggling_show_ruler)
```
ice0, line 2713: How this variable is used?

mohamedAdhamc: Oh yes, thanks for pointing it out. I believe this was necessary when the button was still there to prevent the calling of the function again. I will remove it.

mohamedAdhamc:
@ice0, done. Is anything else left here or is it good to go?

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -2701,6 +2706,14 @@ CanvasView::set_onion_skins()
 	work_area->set_onion_skins(onion_skins);
 }
 
+void
+CanvasView::toggle_show_ruler()
+{
+	work_area->toggle_ruler();
```
ice0, line 2712: Here you can use `work_area->set_ruler_visible(set_ruler_show_toggle...)` and remove `get_ruler_status` and `toggle_ruler` as they are no longer needed.

mohamedAdhamc: sounds good yes thanks again (^_^).

mohamedAdhamc: > Here you can use `work_area->set_ruler_visible(set_ruler_show_toggle...)` and remove `get_ruler_status` and `toggle_ruler` as they are no longer needed.

ok @ice0 , I'm a bit sleepy here so maybe I'm missing something😴, but why is ruler status not needed anymore? It is the flag used for saving the metadata, and get_ruler_status() is its getter because here in canvasview.cpp we are outside the class. Maybe you mean to deal with the meta data here right away? but isn't it better to leave it as it is there as all the other related meta data are saved the same way?

ice0: The flag is already saved inside `set_ruler_show_toggle` widget.
When you call `toggle_show_ruler` you need to toggle this status and just set the visibility for the rulers.
Check this at the morning maybe you'll understand better what I mean.

ice0: > It is the flag used for saving the metadata

Got it. I need to test some ideas in this case.

mohamedAdhamc: > > It is the flag used for saving the metadata
> 
> Got it. I need to test some ideas in this case.

sounds good, please let me know when you reach something so we can get this PR also out of the way.

synfig-studio/src/gui/workarea.h
```diff
@@ -366,6 +373,7 @@ class WorkArea : public Gtk::Grid, public Duckmatic
 	//! Returns the color of the grid
 	const synfig::Color &get_grid_color()const { return Duckmatic::get_grid_color();}
 
+	bool get_ruler_status()const { return ruler_status; }
```
rodolforg, line 376: I suggest you to follow the naming style of guides: `get_show_rulers()`, `set_show_rulers()` , `toggle_rulers()` and `show_rulers`. I chose plural form because there are two rulers: horizontal and vertical.

mohamedAdhamc: Sounds good

mohamedAdhamc:
@rodolforg done

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1577,6 +1577,11 @@ CanvasView::init_menus()
 	{
 		Glib::RefPtr<Gtk::ToggleAction> action;
 
+		rulers_show_toggle = Gtk::ToggleAction::create("toggle-rulers-show", _("Show Rulers")); //also this
+		rulers_show_toggle->set_active(work_area->get_show_rulers()); //of course set the status the correct way
```
rodolforg, line 1581: remove these comments :)

synfig-studio/src/gui/workarea.cpp
```diff
@@ -803,6 +812,21 @@ WorkArea::set_background_rendering(bool x)
...
 }
 
+void
+WorkArea::set_rulers_visible(bool visible)
+{
+	hruler->set_visible(visible);
+	vruler->set_visible(visible);
+	menubutton_box->set_visible(visible);
+}
+void
```
rodolforg, line 822: empty line separating those two methods.

rodolforg (changes requested):
Almost there :)

rodolforg (approved):
@ice0 We need to squash before merge

synfig-studio/src/gui/workarea.cpp
```diff
@@ -803,6 +812,22 @@ WorkArea::set_background_rendering(bool x)
...
+WorkArea::set_rulers_visible(bool visible)
+{
+	hruler->set_visible(visible);
+	vruler->set_visible(visible);
+	menubutton_box->set_visible(visible);
+}
+
+void
+WorkArea::toggle_rulers()
```
ice0, line 824: 1) `toggle_rulers` can be safely replaced and removed:
```
set_rulers_visible(!get_show_rulers())
```
2) I suggest to rename `set_rulers_visible` to `set_show_rulers`, so we get `get_show_rulers`/`set_show_rulers`

mohamedAdhamc: done

ice0 (approved):

ice0:
Merged. Thank you!

mohamedAdhamc:
> Merged. Thank you!

hopefully many more merges to come :)

PR #2641, opened 2022-05-19
Replacement of deprecated Gtk Arrow

mohamedAdhamc:
A quick switch from the deprecated gtk arrow to gtk image.
before :
gtk_arrow_og
after: 
latest_arrow

I manually added the icon myself to my icons folder though so it still has to be added somewhere  in the build files but I'm not at all familiar with c-make or the build process as a whole so if anyone can point it out a bit it would be appreciated and this would be done basically. 
however, what I'm guessing needs to be done should be in makefiles.am file and maybe the icon itself is added as a SIF file. however I'm not sure where I could convert the icon which is a .png file to a SIF file. I will be looking into these but it would save a lot if anyone helps 😄

### DIFF
--- a/synfig-studio/src/gui/iconcontroller.cpp
+++ b/synfig-studio/src/gui/iconcontroller.cpp
@@ -212,6 +212,7 @@ IconController::init_icons(const synfig::String& path_to_icons)
 	INIT_STOCK_ICON(set_outline_color, "set_outline_color." IMAGE_EXT, _("Set as Outline"));
 	INIT_STOCK_ICON(set_fill_color, "set_fill_color." IMAGE_EXT, _("Set as Fill"));
 
+	INIT_STOCK_ICON(right_arrow, "chevron_right_arrow_icon." IMAGE_EXT, _(""));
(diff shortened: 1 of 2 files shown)


ice0:
@mohamedAdhamc 

Don't use Gtk::StockId, they are also deprecated. Instead, use `Gtk::Button` with an image.
Please check the help page for `Gtk::Arrow`, it describes which icon you should use.

mohamedAdhamc:
> @mohamedAdhamc
> 
> Don't use Gtk::StockId, they are also deprecated. Instead, use `Gtk::Button` with an image. Please check the help page for `Gtk::Arrow`, it describes which icon you should use.
> 
> P.S. I mean there is no need to add an image, they are already exist in the theme.

Oh yes I read that it can be replaced with an icon called start symbolic something like that but didn't realize it's already in the theme, if so though then yea it's better to use it. Also why the button though, I mean why not just the gtk image having the needed icon and then its placed in an event box anyway already so the button wouldn't really be needed.

ice0:
> The event box doesn't seem to be deprecated so I'm curious if there's a special reasoning behind that.

This is the right question. 
1. I think it's better to have one object than several.
2. Less code
3. The button has a hover effect to indicate that it is an interactive arrow.

mohamedAdhamc:
implemented via #2824

PR #2645, opened 2022-05-22
Returned back the ability to set a GTK theme using an environment variable

ice0:
SYNFIG_GTK_THEME=Adwaita /path/to/synfigstudio


morevnaproject (approved):
It works now, thank you!

PR #2652, opened 2022-05-27
fix: Missing menu in Synfig when path to plugins contains non-Latin symbols (Windows)

ice0:
Screenshot_49

It shows an error on start:
```
synfig(23672) [10:56:14] info: Init UI Manager...
synfig(23672) [10:56:14] error: building menus and toolbars failed: Error on line 1 char 10067: Document ended unexpectedly with elements still open в?" в??toolbarв?? was the last element opened
```

After adding the output with invalid XML, we can see where the problem
is:
```
<menuitem action='C:/msys64/...<Non-Latin path>...\plugins\add-skeleton-simple/plugin.xml'/>
```

The most interesting thing here is the slashes. According to the Gtk
documentation (https://docs.gtk.org/gtk3/class.UIManager.html):

> The name and action attributes must not contain “/” characters after
> parsing (since that would mess up path lookup) and must be usable
> as XML attributes when enclosed in doublequotes, thus they must
> not “”” characters or references to the " entity.

Unfortunately, I haven't found anything about non-ASCII characters in
`action` attribute, so I propose to use plugin folder name here (instead 
of path to plugin).

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -878,6 +878,14 @@ public:
 
 App::Preferences App::_preferences;
 
+static void add_ui_from_string(const std::string& ui_info) {
+	try	{
+		App::ui_manager()->add_ui_from_string(ui_info);
+	} catch(const Glib::Error& ex) {
+		synfig::error("building menus and toolbars failed: " + ex.what() + "\n\n" + ui_info);
+	}
+}
+
 void
 init_ui_manager()
 {
@@ -1266,21 +1274,12 @@ DEFINE_ACTION("keyframe-properties", _("Properties"))
 
 	#undef DEFINE_ACTION
 
-	try
-	{
-		actions_action_group->set_sensitive(false);
-		App::ui_manager()->set_add_tearoffs(false);
-		App::ui_manager()->insert_action_group(menus_action_group,1);
-		App::ui_manager()->insert_action_group(actions_action_group,1);
-		App::ui_manager()->add_ui_from_string(ui_info);
-		App::ui_manager()->add_ui_from_string(hidden_ui_info);
-
-		//App::ui_manager()->get_accel_group()->unlock();
-	}
-	catch(const Glib::Error& ex)
-	{
-		synfig::error("building menus and toolbars failed: " + ex.what());
-	}
+	actions_action_group->set_sensitive(false);
+	App::ui_manager()->set_add_tearoffs(false);
+	App::ui_manager()->insert_action_group(menus_action_group,1);
+	App::ui_manager()->insert_action_group(actions_action_group,1);
+	add_ui_from_string(ui_info);
+	add_ui_from_string(hidden_ui_info);
 
 	auto default_accel_map = App::get_default_accel_map();
 	for (const auto& accel_item : default_accel_map) {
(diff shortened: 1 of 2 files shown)


rodolforg:
What if some installation has two plugins with same name, but in different folders (example: system and user plugin folders)?

ice0:
> What if some installation has two plugins with same name, but in different folders (example: system and user plugin folders)?

I thought about that too.
The first idea was to use the hash of the file path here, but I don't like that (because we lose a human-readable name).
Another option is to use a static counter variable and increment it for each plugin. So we will get something like `plugin1`, `plugin2`, etc.

Which one is better? Or maybe another idea?

rodolforg:
> but I don't like that (because we lose a human-readable name).

Why does it need to be readable? Is it accessible to read/write?



> Another option is to use a static counter variable and increment it for each plugin. So we will get something like `plugin1`, `plugin2`, etc.

Nice! What about mixing? name + counter?

ice0:
> > but I don't like that (because we lose a human-readable name).
> 
> Why does it need to be readable? Is it accessible to read/write?

I guess not.

> > Another option is to use a static counter variable and increment it for each plugin. So we will get something like `plugin1`, `plugin2`, etc.
> 
> Nice! What about mixing? name + counter?

Done.

synfig-studio/src/gui/pluginmanager.cpp
```diff
@@ -179,7 +179,8 @@ studio::PluginManager::load_plugin( const std::string &file, const std::string &
 {
 	synfig::info("   Loading plugin: %s", etl::basename(plugindir).c_str());
 
-	std::string id = file;
+	static int plugin_count = 0;
+	const std::string id = "plugin" + std::to_string(++plugin_count);
```
rodolforg, line 183: I thought you'd use plugin name + count :)

synfig-studio/src/gui/app.cpp
```diff
@@ -1266,21 +1274,12 @@ DEFINE_ACTION("keyframe-properties", _("Properties"))
...
-		App::ui_manager()->insert_action_group(actions_action_group,1);
-		App::ui_manager()->add_ui_from_string(ui_info);
-		App::ui_manager()->add_ui_from_string(hidden_ui_info);
-
-		//App::ui_manager()->get_accel_group()->unlock();
-	}
-	catch(const Glib::Error& ex)
-	{
-		synfig::error("building menus and toolbars failed: " + ex.what());
-	}
```
rodolforg, line 1283: Is it safe to remove the try-catch?

rodolforg (approved):
It works! \o/

synfig-studio/src/gui/pluginmanager.cpp
ice0: Yes. But in this case, we have to set plugin naming rules (no spaces, only Latin characters, etc.). 
So I think this will be better solution.

synfig-studio/src/gui/app.cpp
ice0: `insert_action_group` should not throw errors.
`add_ui_from_string(...)` is already caught.

synfig-studio/src/gui/pluginmanager.cpp
rodolforg: Got it. Approved

PR #2658, opened 2022-05-28
Refactored python search algorithm

ice0:
- python3 priority increased
- switched to `Glib::getenv(...)` for correct locale handling


morevnaproject (approved):
Great, it removes redundant warning message on my system. Tested.

PR #2667, opened 2022-06-04
Use conan to build synfig

hesham-essam:
To build, you do something like
```
cd conan-recipes/gtk/all
conan export . gtk/3.24.34@
cd ../../gtkmm/all
conan export . gtkmm/3.24.6@
cd ../../../
mkdir build
conan install ..  -if ./modules/ --build=missing
# if on powershell
$env:PKG_CONFIG_PATH = "$($(pwd).Path)/modules"
# if on bash
PKG_CONFIG_PATH = $PKG_CONFIG_PATH:$(pwd)/modules
./modules/activate_run.ps1
# if on powershell
cmake -DCMAKE_MODULE_PATH="$([RegEx]::Escape($(pwd)))/modules" ..
# if on bash
cmake -DCMAKE_MODULE_PATH="$(pwd)/modules" -DCMAKE_BUILD_TYPE=Release ..
cmake --build . --config Release
```
I wrote two powershell scripts to automate this (since currently I tried it only on Windows) `1-conan-export.ps1` and `2-conan-build.ps1`.


ice0:
@hesham-essam 

Can you check with this PR?
https://github.com/synfig/synfig/pull/2622

synfig-core/src/synfig/CMakeLists.txt
```diff
@@ -134,11 +137,11 @@ target_link_libraries(synfig
         ${GLIBMM_LINK_LIBRARIES}
         ${GIOMM_LINK_LIBRARIES}
         ${LIBXML_LINK_LIBRARIES}
-        ${XMLPP_LIBRARIES}
+        ${XMLPP_LINK_LIBRARIES}
```
ice0, line 140: We are trying to support CMake 3.5 for Linux old stable versions.
Can you check if `<XPREFIX>_LDFLAGS` is work here?

ice0:
@hesham-essam 

I suggest to split this PR. 
Extract Conan scripts and recipes into a separate PR, so I can merge them without issues, after that we can continue with CMake.

hesham-essam:
> @hesham-essam
> 
> I suggest to split this PR. Extract Conan scripts and recipes into a separate PR, so I can merge them without issues, after that we can continue with CMake.

I will try making this PR compile correctly first here in the CI, and locally on linux and windows, then I will split things up into multiple PRs if that's okay.

ice0:
> I will try making this PR compile correctly first here in the CI, and locally on linux and windows, then I will split things up into multiple PRs if that's okay.

Sure.

hesham-essam:
@ice0 This builds locally fine on Windows and Linux. Please try it since my environment isn't the most standard one.
First run `1-conan-export`, then `2-conan-build`, then activate the run environment in "conan_build/modules/activate_run.[sh/ps1]".

2-conan-build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/env bash
```
ice0, line 1: Run in Debian 10 container:
```
root@ef5af972deea:~/conan/synfig# ./1-conan-export.sh 
bash: ./1-conan-export.sh: /bin/env: bad interpreter: No such file or directory
```

Same on MacOS. I suggest to use `/usr/bin/env` as a more portable solution.

1-conan-export.ps1
```diff
@@ -0,0 +1,8 @@
+$Cd = Get-Location
```
ice0, line 1: Developer PowerShell for VS 2019 (Windows 10)
```
.\1-conan-export.ps1 : File C:\projects\synfig-conan\synfig\1-conan-export.ps1 cannot be loaded because running scripts is disabled on this system. For more information, see about_Execution_Policies at
https:/go.microsoft.com/fwlink/?LinkID=135170.
At line:1 char:1
+ .\1-conan-export.ps1
+ ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : SecurityError: (:) [], PSSecurityException
    + FullyQualifiedErrorId : UnauthorizedAccess
```

Fixed by:
```
set-executionpolicy remotesigned -Scope CurrentUser
```

Maybe add this as a preparing step before build?

1-conan-export.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/env bash
+
+SCRIPT_DIR="$(readlink --canonicalize $(dirname -- "${BASH_SOURCE[0]:-$0}"))";
```
ice0, line 3: Doesn't work on MacOS:
```
readlink --canonicalize ...
readlink: illegal option -- -
```

ice0:
It stops the build with an error (PS):
```
CMake Error at synfig-studio/src/gui/CMakeLists.txt:36 (add_executable):
  Target "synfigstudio" links to target "PkgConfig::GTKMM" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

The issue is missing `pkg-config` on Windows.

```
+ pkg-config
+ ~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (pkg-config:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

hesham-essam:
@ice0 On windows it seems that the run environment wasn't activated. I added pkgconf as a build dependency so it should've been added to the path in the run environment.
Did you use the script or run cmake without it?
If not, you should run `modules/activate_run.ps1` first.

ice0:
> @ice0 On windows it seems that the run environment wasn't activated. I added pkgconf as a build dependency so it should've been added to the path in the run environment. Did you use the script or run cmake without it? If not, you should run `modules/activate_run.ps1` first.

First I tried to use Developer's PowerShell for VS 2019 to build Synfig.
It fails to build `gettext` library and stops with error `libintl.res.obj : fatal error LNK1112: module machine type 'x64' conflicts with target machine type 'x86'`. The issue was due to the "Developer's PowerShell for VS 2019" being run in 32-bit mode and there is no Developers PowerShell x64 as they did for the command line (I mean `x64 Native Tools Command Prompt for VS 2019`).

Next I tried to use `x64 Native Tools Command Prompt for VS 2019` and it builds all packages fine, but stops when trying to build Synfig.
The problem here is that I can't use `activate_run.ps1` to switch environment from command-line and I missed that `activate_run.bat` is already there.

So I switched to standard PowerShell. Some packages could not be built due to missing CMake (standard PowerShell didn't add VS CMake binary to the PATH).
Fixed this by adding CMake as a build dependency for those packages, so they used Conan's CMake.
After that Synfig builds fine, but there is no GTK icons (it's ok, copied them manually), but it stops on start with error message:
```
...
synfig(5732) [15:56:48] info: Loading Custom Workspace List...
synfig(5732) [15:56:48] info: Init auto recovery...
synfig(5732) [15:56:48] info: Loading Settings...
synfig(5732) [15:56:48] info: Init Tools...
synfig(5732) [15:56:48] info: Checking auto-recover...
synfig(5732) [15:56:48] info: Done.
Fontconfig error: Cannot load default config file: No such file: (null)
```

hesham-essam:
There are a couple of issues with conan that I will fix, then I will come back to this PR. These being:
1. Gettext provided by conan does not define `GETTEXT_MSGMERGE_EXECUTABLE` nor `GETTEXT_MSGFMT_EXECUTABLE`, and it also does not provide `libintl` for Windows. Hopefully I will do a PR for this by tomorrow.
2. Conan's Gtk 3 is broken on Windows, and I believe it could be broken on Linux as well. The problem is mixing shared and static libraries which causes problems with glib. See https://github.com/conan-io/conan-center-index/issues/11022. 
I made these PRs to fix the problem: 
- https://github.com/conan-io/conan-center-index/pull/11040
- https://github.com/conan-io/conan-center-index/pull/11041
- https://github.com/conan-io/conan-center-index/pull/11093
- https://github.com/conan-io/conan-center-index/pull/11171
- https://github.com/conan-io/conan-center-index/pull/11212
- https://github.com/conan-io/conan-center-index/pull/11209
- https://github.com/conan-io/conan-center-index/pull/11208
- https://github.com/conan-io/conan-center-index/pull/11166

hesham-essam:
Conan is in the middle of migrating to v2, which has a new API, so I had to copy and edit a number of recipes. As migration takes place, I will remove these recipes one by one.

This is an overhaul of the PR. My intent is to get rid of environment activation entirely.  I took a page from what vcpkg did, and copied all dlls to synfig's bin directory.

I only tested this on Windows using VS2022, but I will get around to linux and VS2017. To build (using powershell) you do something like:
```
mkdir build
cd build
conan install .. -if . -pr default
cmake --preset=default -Ax64 -DCMAKE_BUILD_TYPE=Release -S .. -B .
cmake --build . -j8 --config Release 
```

Also still missing:
1. Themes
2. gdk-pixbuf loaders.

hesham-essam:
This builds fine on linux locally. To build, you do something like
```
./1-conan-export.sh
mkdir build
cd build
conan install .. -if . -pr default --build=missing
cmake --preset=release -DCMAKE_BUILD_TYPE=Release -S .. -B .
make -j8 
```

PR #2676, opened 2022-06-06
build: Use CMake's PkgConfig imported targets

hesham-essam:
The combination of `<TARGET>_LIBRARIES` and `<TARGET>_LDFLAGS` doesn't work with msvc since cmake doesn't modify all the flags to be msvc-style. `<TARGET>_LINK_LIBRARIES` would work, but it was first documented in cmake 3.12.
The alternative is PkgConfig targets, which was introduced in cmake 3.6, and it is actually more modern-cmake like.


ice0 (approved):

ice0:
Merged. Thank you!

Issue #452, opened 2017-11-02
Color wheel stuck to red when saturation=0


zozorg:
**Synfig version & platform**:
linux 1.3.4

**Issue description**:
When moving the cursor of the color wheel to the grey area ( saturation=0 ), it automatically locks itself to hue=0 ( leaving you mouse hanging over a random color ), and it is not possible to move the hue cursor as long as saturation=0, so you have to slightly move the cursor before you select the hue you need.
This doesn't prevent you from selecting anything, but it can get a bit annoying.

caryoscelus:
Yeah, i remember that one. Pretty annoying.

zozorg:
Posted a [$15 bounty](https://www.bountysource.com/issues/51060128-color-wheel-stuck-to-red-when-saturation-0)

caryoscelus:
Thx for the bounty, i'll look into it. Shouldn't be especially hard to fix.

Btw, i'm not sure if bountysource works atm (see https://github.com/bountysource/core/issues/1147)

morevnaproject:
Thank you for the fix, works like a charm! Merged now.

P.S. Yes, I also have problems authorizing into bountysource. :-/

caryoscelus:
NP :)

> P.S. Yes, I also have problems authorizing into bountysource. :-/

It actually works for me, if i login directly with email/password (i.e. not via github). I don't recall how i registered there in the first place, though.

zozorg:
Bounty awarded! Thanks!

PR #2679, opened 2022-06-10
fix: Wrong color in HSV wheel when color was changed with paint bucket

### LINKED ISSUE #2678: Wrong color in the parameters
**Describe the bug**
When a color is placed on a shape, when you open the color window in the parameters docker, it shows the wrong color different from the one the shape has in the workspace. The same thing happens in the tool docker when clicking on the fill color box the window that opens has the same error.

**To Reproduce**
1. Create any region or shape with area to color.
2. Select any color and with the paint bucket tool paint the shape.
3. Go to the parameters docker and click on the color parameter.
4. In the color window you will see a different color than the one assigned.

**Expected behavior**
It should have the color that was just assigned.

**Screenshots**

https://user-images.githubusercontent.com/45057390/172741819-6d9301e1-9740-4012-ab1d-a6688442b4b2.mp4

**System information:**
 - OS:Arch Linux x64 with Kernel 5.15.45-1-lts
 - Development version:
 - Revision: 20220606
 - Branch: master
 - Revision ID: a18ac7b
 - Built on Jun  7 2022

**Additional comments**
I hope it will be useful for everyone, thanks for the hard work.

ice0:
Fix #2678

`are_close_colors` always returned true.

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_coloredit.cpp
+++ b/synfig-studio/src/gui/widgets/widget_coloredit.cpp
@@ -465,25 +465,23 @@ Widget_ColorEdit::~Widget_ColorEdit()
 {
 }
 
-#define CLIP_VALUE(value, min, max) (value <= min ? min : (value > max ? max : value))
-
 bool are_close_colors(Gdk::RGBA const& a, Gdk::RGBA const& b) {
-	static const int eps = 1;
+	constexpr float epsilon = 0.000001f;
 	if (a == b)
 		return true;
-	return std::fabs(a.get_red()-b.get_red()) <= eps
-		&& std::fabs(a.get_green()-b.get_green()) <= eps
-		&& std::fabs(a.get_blue()-b.get_blue()) <= eps;
+	return std::fabs(a.get_red()-b.get_red()) <= epsilon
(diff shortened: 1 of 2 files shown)


rodolforg:
I guess the main reason was: `Gdk::Color::get_red()` returns `unsigned short`, and `Gdk::RGBA::get_red()` returns `double`.
I wonder if there are more similar cases around.

ice0:
Related PR #2654

rodolforg:
Looking at git history, this method was created by commit 951cd2daf51558e1bef2800987cf2b920bdc48e6 for fixing #452.

Does this bug remain fixed? (I can't test your PR right now).

ice0:
> Looking at git history, this method was created by commit [951cd2d](https://github.com/synfig/synfig/commit/951cd2daf51558e1bef2800987cf2b920bdc48e6) for fixing #452.
> 
> Does this bug remain fixed? (I can't test your PR right now).

The problem is that when we call the `hvsColorWidget->set_current_rgba`, we immediately lose the HUE value, which sets rotation angle for color wheel and the color wheel is reset to zero position.
Unfortunately there is no way to set color wheel using HSV value, so I reverted back this function (`are_close_colors`).

rodolforg (approved):

PR #2687, opened 2022-06-13
refactor: cleanup `get_build_date()` function

ice0:
I forgot to test a reproducible build, but after I did it,
I noticed that if I copy the date from the build and try
 to rebuild it, it says "Unknown build date".

 This fixes the issue.


rodolforg:
@ice0 ping

PR #2690, opened 2022-06-17
fix: Alpha channel resets when changing color in HSV wheel

ice0:
The alpha channel was reset because it was not saved.


rodolforg (approved):
Maybe the method name should be changed to clarify it is called only as a slot for hvsColorWidget change?

ice0:
> Maybe the method name should be changed to clarify it is called only as a slot for hvsColorWidget change?

Agree.

PR #2696, opened 2022-06-18
refactor: removed deprecated std::unary_function and std::binary_function classes

### LINKED ISSUE #2692: Remove/Replace deprecated `std::unary_function` and `std::binary_function`
Since C++11, `std::unary_function` and `std::binary_function` are deprecated.
Since C++17, they were removed from standards.

As we use C++11, we can and should already get rid of it. However, I'm not capable of doing so, as I don't understand it XD

Some alternatives:
- https://stackoverflow.com/questions/57236763/how-to-change-binary-function-to-c17-code
- https://stackoverflow.com/questions/56001160/how-do-i-replace-stdbinary-function-with-something-else-without-pain
- https://stackoverflow.com/questions/28299581/why-does-stdbinary-function-not-have-an-operator-method

By the way, AppVeyor complains a lot about it.

ice0:
`std::unary_function` and `std::binary_function` just need to define 3 types:

`unary_function` provides only two types - `argument_type` and `result_type` -
defined by the template parameters.
(https://en.cppreference.com/w/cpp/utility/functional/unary_function).

`binary_function` provides only three types - `first_argument_type`,
`second_argument_type` and `result_type` - defined by the template parameters.
(https://en.cppreference.com/w/cpp/utility/functional/binary_function)

Based on this code:
```
class bezier_base<synfig::Vector,float> : public std::unary_function<float,synfig::Vector>
{
public:
	typedef synfig::Vector value_type;
	typedef float time_type;
```

we can assume what:
```
using argument_type = time_type;
using result_type = value_type;
```

So I just replaced `argument_type` and `result_type`
with the corresponding types.

Fix #2692

### DIFF
--- a/ETL/ETL/_angle.h
+++ b/ETL/ETL/_angle.h
@@ -475,7 +475,7 @@ struct affine_combo<etl::angle, T>
 };
 
 template <>
-struct distance_func<etl::angle> : public std::binary_function<etl::angle, etl::angle, etl::angle>
+struct distance_func<etl::angle>
 {
 	etl::angle operator()(const etl::angle &a,const etl::angle &b)const
 	{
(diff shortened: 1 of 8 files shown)


rodolforg (approved):
It builds ok, so it should be fine XD

PR #2710, opened 2022-06-24
refactor!: move filepath handling methods to synfig::filesystem::Path

rodolforg:
BREAKING CHANGE: ETL API changed:
- remove `etl::is_separator()`; use `synfig::filesystem::Path::is_separator()` instead
- remove `etl::basename()`; use `synfig::filesystem::Path::filename()` instead
- remove `etl::dirname()`; use `synfig::filesystem::Path::remove_filename()` or `synfig::filesystem::Path::parent_path()` instead
- remove `etl::filename_extension()`; use `synfig::filesystem::Path::extension()` instead
- remove `etl::filename_sans_extension()`; use `synfig::filesystem::Path::relative_path() / synfig::filesystem::Path::stem()`  or `synfig::filesystem::Path::remove_extension()` instead
- remove `etl::is_absolute_path()`; use `synfig::filesystem::Path::is_absolute()` instead
- remove `etl::get_root_from_path()`; use `synfig::filesystem::Path::root_path()` instead
- remove `etl::remove_root_from_path()`; use `synfig::filesystem::Path::relative_path()` instead
- remove `etl::get_root_from_path()`; use `synfig::filesystem::Path::root_path()` instead
- remove `etl::cleanup_path()`; use `synfig::filesystem::Path::lexically_normal()` or `synfig::filesystem::Path::cleanup()` instead
- remove `etl::relative_path()`; use `synfig::filesystem::Path::relative_to()` instead
- remove `etl::absolute_path()`; use `synfig::filesystem::Path::append()` instead
- remove `etl::solve_relative_path()`; use `synfig::filesystem::Path::append()` instead


rodolforg:
Is this interesting? It would clean up `ETL/stringf`

rodolforg:
Should/Could I invest time working on this?

ice0:
> Is this interesting? It would clean up `ETL/stringf`

Hi! Sorry for the late reply.
Yes, this is a good idea!

hesham-essam:
This is a problem not the place for this question, but why not use `std::filesystem` and fall to `boost::filesystem` if c++17 is not available?

rodolforg:
@hesham-essam  We avoid to increase our dependencies.
@ice0 What about our `synfig::FileSystem::Identifier::filename` be a `Path`? If yes, we have to move up the `Path` declaration in the file. Or create a separate file (e.g. `filesystem_path.h`) and `#include` it in `filesystem.h`.

ice0:
> What about our `synfig::FileSystem::Identifier::filename` be a `Path`? If yes, we have to move up the `Path` declaration in the file. Or create a separate file (e.g. `filesystem_path.h`) and `#include` it in `filesystem.h`.

I think it's a good idea

rodolforg:
I'll move  Path to its own file then.
Or should I?

ice0:
> I'll move Path to its own file then.

Yes.

rodolforg:
I finished the "port". It miss iterators (`begin()` and `end()`) and (intentionally) some method overloads.
Our `Path` class will behave differently from the C++17: in the new standard the focus is native encoding. We handle mostly in UTF-8, only needing the native one with OS calls like `fopen()`.

Some few differences are noted in individual commit messages.

I'm thinking about changing some methods' behavior: like force path normalization (old `cleanup_path`) in some cases.
Maybe we should add a "`add_suffix()`" to add a string between filename stem and its extension (useful for multiple-image-export. The alternative would be:
```c++
filesystem::Path(filename).replace_filename(filename.stem() + ".000" + filename.extension());
```
instead of current:
```c++
filename_sans_extension(filename) + ".000" + filename_extension(filename);
```

My proposal is:
```c++
filesystem::Path(filename).add_suffix(".000"); // or add_filename_suffix
```

Anyway, now I can continue to make the current methods in `ETL/stringf` "disappear".
Although it would be good to finish #2491 first, to avoid conflicts and unnecessary rework.

@ice0 can you take a look here?

rodolforg:
(rebased)

rodolforg:
(rebased)

ice0:
@rodolforg 

I suggest merging #2740 first, this will allow testing on all OSes.

rodolforg:
@ice0, do you understand why macOS is failing with Autotools?

ice0:
> @ice0, do you understand why macOS is failing with Autotools? (It is segfault'ing in app_layerduplicate.cpp)

Looks like all tests is ok. Can you share a link to a failed build?

rodolforg:
> Looks like all tests is ok. Can you share a link to a failed build?

I re-ran the job, and it completed fine :)

@ice0 , what is missing now is:
- `begin()`, `end()` iterators
- an static method `from_native()` to construct a Path from envvars and command lines

Do we need them?

rodolforg:
> Can you share a link to a failed build?

https://github.com/synfig/synfig/actions/runs/4589295235/attempts/1

ice0:
> > Looks like all tests is ok. Can you share a link to a failed build?
> 
> I re-ran the job, and it completed fine :)

Ok.

> @ice0 , what is missing now is:
> 
> * `begin()`, `end()` iterators
> * `<<`, `>>` operators
> 
> Do we need them?

I suggest skipping them if they are not needed in the current code.

> Besides, I propose to add two methods:
> 
> ...
> My proposal is:
> 
> ```c++
> filesystem::Path(filename).add_suffix(".000"); // or named as add_filename_suffix()
> ```

Looks good!

rodolforg:
@ice0 last steps here, I hope:
1. Should I create `u8str()` (or `u8_str()`) to simplify `u8string().c_str()`? (remember just calling `c_str()` means `native().c_str()`)
  It would be useful to use as argument for `synfig::strprintf()`.
2. There is `etl::absolute_path(std::string curr_path, std::string path)` and `etl::absolute_path(std::string path)` . How should we replace it? In `filesystem::Path`? In `FileSystemNative`? Other way?
3. At first, I 'll port the remaining ETL/_stringf.h functions as static methods of `synfig::filesystem::Path`.
  In another PR(s), I'll change all filenames variables/members of different functions and classes that are now `synfig::String` or `std::string` to `synfig::filesystem::Path` and get rid of these static methods. What do you think?

Remaining functions to remove:
image

rodolforg:
@ice0, now the only thing that remains on `ETL/stringf` is `ETL_DIRECTORY_SEPARATOR`.
Should I stop this PR here?

ice0:
@rodolforg 
Is it ready to merge?

rodolforg:
@ice0  not how I intended to, but it is ready.
I think it is better to continue the migration/port in other PRs, to be easier to detect possible problems.

ice0 (approved):

ice0:
> @ice0 not exactly how I intended to, but it is ready. I think it is better to continue the migration/port in other PRs, to be easier to detect possible problems.

Agree.

ice0:
Merged. Thank you!

PR #2711, opened 2022-06-25
refactor!: rename etl::clock to synfig::clock

rodolforg:
Moved ETL benchmark test to synfig-core due to `clock` requirement

BREAKING CHANGE: both ETL and synfig API changed:
- `etl::clock` is now `synfig::clock`


ice0 (approved):

ice0:
Merged. Thank you!

PR #2714, opened 2022-06-27
refactor: make use of synfig::clamp instead of std::max(std::min())

rodolforg:
It's more readable this way.
In C++17, we can replace it with `std::clamp()`.


ice0 (approved):

ice0:
Merged. Thank you!

Issue #2528, opened 2022-02-02
Add the posibility of move the paramether values using the mouse


gelatinbomb:
**Is your feature request related to a problem? Please describe.**
I think is fine to click the paramether and then change the value with the keyboard but at the end it's kind of slow and I think it would be more easy to do small changes if can change the values using the mouse.

**Describe the solution you'd like**
The idea is that when I place the mouse on the parameter an then I click and slide the mouse the value of the parameter will move according to the movement of the mouse.

**Additional comments**
One good example of the behavior I would like for the parameter usage is how enve 2d handles it. It´s really comfortable to do small changes on the parameters in enve 2d thanks to that.

As far I read the documentation and experimented by myself I didn't found you can do this on Synfig but I know I could be wrong at this so if the option exist and you know how to do it I would be really thankful if you can share it here.

ice0:
Can you attach a short video showing how it works in enve?

gelatinbomb:
https://imgur.com/a/fRek4Rm

triawarman:
may i add little request, when the input control is active, another way to change its value is using "up key" or "down key"

PR #2717, opened 2022-06-28
[UI Feature 1] Change Parameters with the Mouse

mohamedAdhamc:
This is main feature number 1 in my project.  it was asked for in #2528 .

The first commit is very preliminary (as in includes comments that I write for my self,etc...) it includes
1- switching the mouse cursor to "move" when pressing on the parameters row (will probably have to be a diffrent icon when its actually done). then switching it back to default when released.
2- storing the selected row and tree view selection object and then updating them when changed.

Main things left to do
1- Change the actual values. moving right increases, left decreases. I'm assuming this would most likely be done through the TreeView model.
2- Of course also optimization, clean ups, and changing the mouse cursor to a more suitable icon. Also probably oher things that come up in the way, this is probably one of the "longer" features.

### DIFF
--- a/synfig-studio/src/gui/trees/layertree.cpp
+++ b/synfig-studio/src/gui/trees/layertree.cpp
@@ -46,6 +46,7 @@
 #include <gui/exception_guard.h>
 #include <gui/instance.h>
 #include <gui/localization.h>
+#include <gui/canvasview.h>
 
 #ifdef TIMETRACK_IN_PARAMS_PANEL
 #  include <synfig/timepointcollect.h>
@@ -53,7 +54,6 @@
 
 #include <synfigapp/actions/layerremove.h>
 #include <synfigapp/instance.h>
-#include <synfigapp/action_system.h>
 #include <gui/trees/historytreestore.h>
 
 #include <thread>
@@ -252,24 +252,28 @@ LayerTree::on_param_tree_view_key_press_event(GdkEventKey* event)
 		Glib::RefPtr<Gdk::Screen> scrn = disp->get_default_screen();
 		Glib::RefPtr<Gdk::Window> window = scrn->get_root_window();
 		window->set_cursor(window->get_cursor());
+		etl::loose_handle<Instance> current_instance = App::get_selected_instance();
 
 		param_tree_view().get_window()->set_cursor(default_cursor);
-		synfigapp::Action::System::block_new_history=true;
-		just_finished=false;
-		value_base.set<float>(initial_value); row[param_model.value] = value_base;
-		synfigapp::Action::System::block_new_history=false;
-		while(gtk_events_pending ()) { gtk_main_iteration (); }
+		current_instance->cancel_repeated_action = true;
+		delete group;
+		group = nullptr;
+		current_instance->repeated_action = false;
+
+		while (gtk_events_pending()) {
+			gtk_main_iteration ();
+		}
+		current_instance->cancel_repeated_action = false;
 	}
 	return false;
 }
 bool
 LayerTree::on_motion_notify_event(GdkEventMotion* event)
-{
-
+{	
 	auto t_now = std::chrono::high_resolution_clock::now();
 	double elapsed_time_ms = std::chrono::duration<double, std::milli>(t_now-t_start).count();
 	if(elapsed_time_ms>65){//not a click so start proocessing inc/dec
-	if(synfigapp::Action::System::block_new_history){//block new history is set to true when there is motion/and in general in rapid action movements such as color wheel drag
+	if(App::get_selected_instance()->repeated_action){//repeated_action is set to true when there is motion/and in general in rapid action movements such as color wheel drag
 
 	if(iter && param_mouse_edit){ //to make sure there is already a treeview i.e. not empty and to not apply changes if its just a click
 
@@ -299,7 +303,6 @@ LayerTree::on_motion_notify_event(GdkEventMotion* event)
 
 		first_cord_x= second_cord_x;
 		value_base = value_base_mod;
-		just_finished = true;
 		first_iteration = false;
 	}}}
 
@@ -312,6 +315,8 @@ LayerTree::on_button_press_or_release_event(GdkEventButton* event)//To DO change
 {
 
 	if(event->button==1 && iter){
+
+	etl::loose_handle<Instance> current_instance = App::get_selected_instance();
 
 	//to be able to know the slected column
 	Gtk::TreeModel::Path path;
@@ -339,7 +344,8 @@ LayerTree::on_button_press_or_release_event(GdkEventButton* event)//To DO change
 			param_mouse_edit=true;
 
 		value_base = value_base_test; // storing initial value base
-		synfigapp::Action::System::block_new_history=true;
+		current_instance->repeated_action = true;
+		group = new synfigapp::Action::PassiveGrouper(App::get_selected_canvas_view()->get_instance().get(),_("Set Parameter"),true);
 		t_start = std::chrono::high_resolution_clock::now();
 		first_iteration=true;
 		signal_handled=false;
@@ -357,7 +363,8 @@ LayerTree::on_button_press_or_release_event(GdkEventButton* event)//To DO change
 			param_tree_view().set_cursor(path, *column, true);
 			grab_focus();
 			param_tree_view().get_window()->set_cursor(default_cursor);
-			synfigapp::Action::System::block_new_history=false;
+			current_instance->repeated_action = false;
+			delete group;
 			param_mouse_edit=false;
 			return false;
 		}
@@ -375,16 +382,12 @@ LayerTree::on_button_press_or_release_event(GdkEventButton* event)//To DO change
 //		release_cord= event->x;
 
 
-		if(just_finished)
-			value_base.set<float>(initial_value); row[param_model.value] = value_base;
+		if (group)
+			delete group;
 
-		synfigapp::Action::System::block_new_history=false;
-
-		if(just_finished)
-		value_base.set<float>(value_float); row[param_model.value] = value_base;
+		current_instance->repeated_action = false;
 
 		param_tree_view().get_window()->set_cursor(default_cursor);
-		just_finished = false;
 		param_mouse_edit=false;
 	}}}
 	return false;
--- a/synfig-studio/src/gui/trees/historytreestore.cpp
+++ b/synfig-studio/src/gui/trees/historytreestore.cpp
@@ -217,7 +217,6 @@ HistoryTreeStore::on_redo_stack_cleared()
 void
 HistoryTreeStore::on_new_action(etl::handle<synfigapp::Action::Undoable> action)
 {
- if (!synfigapp::Action::System::block_new_history){ /*if change isn't due to rapid repeated actions show history normally*/
 	Gtk::TreeRow row;
 
 	row=*insert(next_action_iter);
@@ -227,7 +226,7 @@ HistoryTreeStore::on_new_action(etl::handle<synfigapp::Action::Undoable> action)
 	next_action_iter = row;
 	++next_action_iter;
 
-	signal_undo_tree_changed()();}
+	signal_undo_tree_changed()();
 }
 
 void
--- a/synfig-studio/src/gui/trees/layertree.h
+++ b/synfig-studio/src/gui/trees/layertree.h
@@ -39,6 +39,7 @@
 #include <gui/widgets/widget_value.h>
 
 #include <synfigapp/value_desc.h>
+#include <synfigapp/action_system.h>
 
 
 /* === M A C R O S ========================================================= */
@@ -249,9 +250,8 @@ private:
 	float value_float ;
 	bool first_iteration=false;
 	float initial_value;
-	bool just_finished = false;
 	bool param_mouse_edit=false;
-
+	synfigapp::Action::PassiveGrouper* group = nullptr;
 
 }; // END of LayerTree
 


mohamedAdhamc:
Hey @ice0 , @rodolforg 

 so I'm going through with the suggested implementation, and now I'm working on incrementing/decrementing the values. I believe there might be two ways to do this:

1- through getting the value base stored in the tree model and then going all the way back to its original type and then incrementing it. I think this would require more work but the advantage would be that its independent of the "view" columns implementation ( i mean the spin button). However I think it also might cause problems with the spin button maybe not updating etc...

2-through just incrementing the spin button. I think this might be easier however this would be spin button dependent so if in the future the spin button is to be changed to another thing this would have to be updated as well probably.

now the questions are 
1- How to get the c++ value from the ValueBase object.  so then I can increment and update the valueBase
2- How to reach the spin button from the CellRenderer_ValueBase class. which actually turned out to be a bit complicated as I found the original spibutton for example is a member of Widget_ValueBase class which is then used in a valuebase entry class which is then finally used in the CellRenderer_ValueBase class. so it's a bit complex xD.

I will be spending tomorrow mostly trying to answer these two or either as I think after they are answered rest is a lot easier. However, if you guys could maybe give some insight onto them it would help a lot.

Thanks!

mohamedAdhamc:
ok guys so this is driving me nuts xD
there are already implemented get and set functions for the Value Base but I cant seem to use them 
this is the function i'm talking about
function def

and these are the errors:

error 1
error 2

ice0:
> update got it to set too and now clicking on the rows succesfully increments the parameter woohoo xD.

Awesome!

mohamedAdhamc:
Ok so quick update here

what I have now is
1- cursor changes on button press and resets on release
2- data in the treeview row needed is succesfully obtained, incremented/decremented, and set again.
3- inital cursor position which is sent along with the mouse press event is stored
4- new mouse position/positions are also obtained this time through the window .

What's left here is somehow to loop in the button press callback until the button is released. I dont quite know how that could be done yet as the event sent to the button press method isn't updated meaning i cant just check for its type. I'm thinking maybe signals somehow, if anyone has an idea it would be appreciated :-)

overall:
if this last part is done this feature would be 90 percent done maybe just optimizations, and cleanups will be left to do.

rodolforg:
> Basically the main obstacle right now is getting the value of the ValueBase as a float and then setting it back again. if this is done correctly the feature would be almost done.

Our (irritating) way to do it is:
```c++
float data = value.get(float());
```

---
I don't know if this is useful here (I'm sleepy now), but Ducks have `set_edit_immediatelly()` method.

mohamedAdhamc:
> 
> I don't know if this is useful here (I'm sleepy now), but Ducks have `set_edit_immediatelly()` method - but this will pollute undo history.

Yes, actually I have managed to get and set using a class called Value, but this could -possibly- be a neater way to do it. will have to look into it at the end.

ice0:
> 1- cursor changes on button press and resets on release

1. It is better to save and restore the previous cursor, in case it somehow changed here.
2. Please check the history while you change the value - it should not pollute history while you moving mouse - only set the value then you release mouse button. Also pressing the `Esc` key should cancel the change and return the original value.

mohamedAdhamc:
> > 
> Also pressing the `Esc` key should cancel the change and return the original value.

Well, currently ctr+z undoes the change. The way its implemented is that the ValueBase itself of the parameter is changed. There might be something I'm missing here. Maybe you mean this should be done through the spin button?

mohamedAdhamc:
ok guys, so this is how its working so far
moving right increments, moving left decrements. incrementing/decrementing keeps going until button release
(sorry for the horrible recording quality though xD)
https://media.giphy.com/media/tGGfbeFgLtN7dlD2yE/giphy.gif

Issues/things left/ways of improving
1- Currently this method does pollute the history as each iteration which increments/decrements shows up in the history.
     possible solution: Im thinking maybe disable the history somehow for the iterations. I.E makes history only capture first and l
     final change. However, I dont really have an idea how the history is implemented so I'm not sure how difficult it would be.
2- moving further right increases incrementation speed. this shouldnt be time consuming though.
3- increment by smalled amounts but faster (in the video it incremented by 1 every time). Also not time consuming but I 
     wonder what its effect would be on the performance aspect.
4- Oh and also I think the cursor should be diffrent xD. not the one I have now. Any suggestions ?

rodolforg:
@mohamedAdhamc 
1. I think you should receive and parse these events in the cell editor widget itself, not in the TreeView itself. @ice0 ?
2. In order to reduce undo history pollution, I propose you to deal/handle/"create" with 3 signals: `editing-started`, `editing-canceled`, `editing-done`. This way you can apply temporary changes and "easily" discard them, keeping only the final value (if edition was not canceled by pressing `Esc` for example).

ice0:
> Well, currently ctr+z undoes the change. The way its implemented is that the ValueBase itself of the parameter is changed. There might be something I'm missing here. Maybe you mean this should be done through the spin button?

I mean pressing the `Esc` key **while** you are pressing the mouse button and moving it.

> I think you should receive and parse these events in the cell editor widget itself, not in the TreeView itself. @ice0 ?

Yes, this is the next step.

ice0:
@mohamedAdhamc 

Just tested. It should change the value only when you move the mouse.

ice0:
> I have found that gtk has two functions which help in doing this, namely gtk_events_pending(), and gtk_main_iteration(); . Tested the new implementation now the incrementation/decrementation continues until button release . I for a little thought that I would have had to do some multi-threading which would have been overkill xD.

This all happens because you doing this wrong way.

You should subscribe to `signal_motion_notify_event()` and get mouse position from here. Doing endless cycle inside event listener or anywhere in event thread is just a bad practice.

I recommend you to write a simple Gtk application just to better understand how it works.

mohamedAdhamc:
Alright guys will be looking into these and updating you soon.

mohamedAdhamc:
> 
> Just tested. It should change the value only when you move the mouse.

@ice0  Just so I'm sure I understand you correctly here. You mean that even if the mouse is to the right of the initial clicking position, the parameter shouldn't increase. It should only increase while the mouse is moving right correct?

mohamedAdhamc:
@ice0 should I start with working on fixing history pollution first or moving to the cell editor widget instead of treeview ? 

p.s. until I get a response from you guys i will be workking onfixing history pollution more specifacaly first I will be trying to understand how the history works for the parameter layer values then seeing how it could be stopped while increasin with mouse.

ice0:
> @ice0 should I start with working on fixing history pollution first or moving to the cell editor widget instead of treeview ?
> 
> p.s. until I get a response from you guys i will be workking onfixing history pollution more specifacaly first I will be trying to understand how the history works for the parameter layer values then seeing how it could be stopped while increasin with mouse.

Yes, better to start from history.

mohamedAdhamc:
update: I got the history to stop making new actions using a flag, I will be trying to enhance it up a bit and will be commiting before the end of the da hopefully :-)

mohamedAdhamc:
ok so as stated in the commit description. this commit makes only the final changed value appear in the history dock. There probably might be a more sophisticated way of doing this but I think this was the simplest approach I could think of.

so a rough updated version of  whats left:
1-  using the cell editor widget instead of the treeview.
2-  Esc to cancel change, this will propably be kind of related to point one though I believe.

2 shouldnt take much time it would just be capturing the escape press and canceling change. the main thing here is 1.
Could you guys explain more what you meant by one as I'm not really sure what is meant, and im kind of worried it will mean reimplementing a huge chunk of things done xD.

ice0:
> ok so as stated in the commit description. this commit makes only the final changed value appear in the history dock. There probably might be a more sophisticated way of doing this but I think this was the simplest approach I could think of.

There is an easy way to check if this is done right. As you already know, Synfig has a history pollution issue when choosing a color using the color wheel. A good and universal solution should solve both of these problems.
So, I propose the following plan:
1. Solve the color history pollution issue in this PR (this will help you see the common code/solution);
2. Extract the code that fixes the color history issue into the new PR;
3. These changes will be merged into the `master` branch;
4. Rebase this PR to master branch (this will remove already merged changes).

ice0:
> Could you guys explain more what you meant by one as I'm not really sure what is meant, and im kind of worried it will mean reimplementing a huge chunk of things done xD.

First please read about Gtk Inspector (this will help you a lot).
https://wiki.gnome.org/Projects/GTK/Inspector

In short, you need to set an environment variable before running the `./synfigstudio` binary. For example:
```
GTK_DEBUG=interactive ./synfigstudio
```

Now about cell editor.

Here's how a standard Gtk::Grid should look like (I've hidden our custom widget here):

Screenshot_3

This is how it looks with custom widget:

Screenshot_4

Reduced opacity for better understanding:

Screenshot_6

This is how this widget looks in widgets tree (you can see this Gtk Inspector):

Screenshot_5

This widget was implemented using custom Gtk::CellRenderer (https://docs.gtk.org/gtk3/class.CellRenderer.html)
Code is located in `synfig-studio/src/gui/cellrenderer/cellrenderer_value.h`

mohamedAdhamc:
>. So, I propose the following plan:
> 
> 1. Solve the color history pollution issue in this PR (this will help you see the common code/solution);
> 2. Extract the code that fixes the color history issue into the new PR;
> 3. These changes will be merged into the `master` branch;
> 4. Rebase this PR to master branch (this will remove already merged changes).

ok, sounds good. I will be looking to do that but also I'm not sure how complex that could get and possibly divert us from the project at focus. However, it's a bit early to know that I will first spend some time looking into that as yes a global solution for both would be perfect.





> First please read about Gtk Inspector (this will help you a lot). https://wiki.gnome.org/Projects/GTK/Inspector
> 
> In short, you need to set an environment variable before running the `./synfigstudio` binary. For example:
> 
> ```
> GTK_DEBUG=interactive ./synfigstudio
> ```
> 

surprisingly in windows it was as simple as just typing out that command (^_^). great I think this will save time.

mohamedAdhamc:
update:
 so I have been looking at the color wheel history pollution problem. It could have been solved basically the same way as here using a flag but the problem for the color wheel, in particular, is that the color wheel in itself is a part of the `Gtk::ColorSelection` and there doesn't seem to be a way where I could access it from the parent widget if it were possible then maybe I would just use the same approach and connect to its button/key release events to know when the drag stopped (as I dont think it has signal_drag_started(), etc...)

but from reading in issue #451 it seems like @rodolforg solved a similar issue (with I think waypoints ?) but I cant seem to determine where exactly it is?  if its possible here I could just have it done the same way. Right now I have it done using a flag that is checked for in the function which makes the new row in the history tree view.

mohamedAdhamc:
Also, @ice0 regarding the custom widget I still don't really get how you guys want it to be done. do you want it to be so that the drag happens on the widget itself? if so then how would this go along with pressing on the widget and having the spin button show for example? I just want a description of how you guys see it behaving .

ice0:
I need a time to check this.
I will answer you later.

ice0:
@mohamedAdhamc 
Okay, I found a solution.
Please read the `Gtk::ColorSelection` documentation. It already has everything you need to resolve the issue.

mohamedAdhamc:
> Please read the `Gtk::ColorSelection` documentation. It already has everything you need to resolve the issue.

Sounds good I'll be rereading it.

ice0:
> Also, @ice0 regarding the custom widget I still don't really get how you guys want it to be done.

Example:

https://user-images.githubusercontent.com/5604544/178890074-37e47d39-26f3-48c1-8da0-82571347f182.mp4

mohamedAdhamc:
> 
> Example:

ok, now I think I have a much better idea. but just for extra clarity confirmation
1- so basically click without drag allows edit of value by keyboard
2- however click with drag simply makes the change by drag.

I will be trying to get something working by end of tomorrow(meaning in about 30 something hour span) then we can pick on the rest of the details from there hopefully. I think that way would make most sense.

ice0:
> ok, now I think I have a much better idea. but just for extra clarity confirmation 
>1- so basically click without drag allows edit of value by keyboard 
> 2- however click with drag simply makes the change by drag.

Just clarification:
Click - is a short press of the mouse button down and up.
So here we get:
1 - Click - enter edit mode
2 - Mouse down and move - change value

mohamedAdhamc:
update: ok so doing some work of getting in between the hierarchy I managed to get the color wheel widget and then I captured its button release and using basically exactly the same way as I did here I fixed the history pollution problem. I'm now working on optimizing it a bit and will be making a separate PR for the color wheel history fix as suggested (might do it when I wake up though).

mohamedAdhamc:
ok I will keep this comment as a tracker for the thing i find today/do 

1- in `CellRenderer_ValueBase::start_editing_vfunc ` if a nullptr is returned then that means editing isn't possible and the popup widget for editing isnt shown. so now a little playing around with the button press/release event should theoretically allow us to control when it enters edit mode and when not. And to make things even better the function itself also has an event parameter but yesterday it kept giving out segmentation errors but I think it should be of use so I will try with it again.

rodolforg:
> I'm having a bit of trouble following the implementation of the cell editor, i was trying to hook up the usual button press/release signals without much effect. maybe because its very different from all the other widgets I have been dealing with so far

Well, it's a strategy of Gtk to make it more efficient: on a TreeView, only one editor is needed at a time, so it only instance the correspondent cell editor when needed. Otherwise, it uses the correspondent (and simpler) CellRenderer.



> so now a little playing around with the button press/release event should theoretically allow us to control when it enters edit mode and when not.

I think you don't have to. Our cell editor widget is `studio::ValueBase_Entry` , defined in `cellrenderer/cellrenderer_value.cpp`.
It gives our `Widget_ValueBase` as the real editor, a widget that handle `synfig::ValueBase` variables. This widget could track mouse dragging.




> 2- seems that the widget which is sent and which is the one that received the event is the tree view (which is named time track )

No, CellRenderer TimeTrack is the widget that renders the timeline of a layer parameter.

mohamedAdhamc:
> 
> I think you don't have to. Our cell editor widget is `studio::ValueBase_Entry` , defined in `cellrenderer/cellrenderer_value.cpp`. It gives our `Widget_ValueBase` as the real editor, a widget that handle `synfig::ValueBase` variables. This widget could track mouse dragging.
> 

ok, this sounds interesting I will be looking into it.

mohamedAdhamc:
update: its working pretty well actually now like the video sent of aftereffects, im working on cleaning up the code now and handling corner cases and will be pushing the commit in a few hours.

mohamedAdhamc:
ok finally, hopefully, I didn't forget things here when committing xD. looking forward to you guys testing and reviewing it.

1-now in terms of functionality as I understand it here is only one thing left which is escape to cancel the drag
* I will be doing it tomorrow though as soon as I start working it shouldn't take long, also would be a good chance for me to retest this*

Also the way I currently have for blocking hisory pollution is using a flag which I think is fine to block, but what i'm doing to then show the last action is probably not the best. best would probably be to manually emit the signal signal_new_action(). I will also look into that.

mohamedAdhamc:
I have added the same method of filing history pollution used in the color wheel pollution fix and added the ability to press escape for cancelling. I will be testing it out a bit then cleaning up the code and pushing a comit when I'm done.

mohamedAdhamc:
ok, that took a while xD. But I tried my best to test every single case I could think of to make sure that everything in the parameters panel is working as it should be along with the feature. I tested when there were waypoints around and when not and made sure to my best that I checked everything also handling things when they showed up.

I haven't cleaned up the code completely though, will do that tommorow first thing hopefully, and it shouldnt take long all the testing was done today. 

Looking forward to having you guys  test and review it.

synfig-studio/src/gui/trees/layertree.cpp
```diff
@@ -225,6 +238,169 @@ LayerTree::create_layer_tree()
...
+		while(gtk_events_pending ()) { gtk_main_iteration (); }
+	}
+	return false;
+}
+bool
+LayerTree::on_motion_notify_event(GdkEventMotion* event)
+{
+
+	auto t_now = std::chrono::high_resolution_clock::now();
+	double elapsed_time_ms = std::chrono::duration<double, std::milli>(t_now-t_start).count();
```
ice0, line 270: Why are you trying to implement the click behavior yourself? Why not just use standard event listeners?

mohamedAdhamc: I couldn't find anything in the standard listeners regarding the normal clicked event, closest I found was a long click gesture which I believe was also more geared toward touch devices. If there is one though, yes it would be much better.

ice0: `signal_button_press_event`?

ice0: Ok, I need a time to make a test app to test this.

mohamedAdhamc: Alright, sounds good.

rodolforg:
I tested and you are just hiding history pollution on the History panel. But `ctrl+Z` still suffers the problem: every little change at a time.

mohamedAdhamc:
> I tested and you are just hiding history pollution on the History panel. But `ctrl+Z` still suffers the problem: every little change at a time.

Oh yes I haven't updated the history pollution as I was waiting on the history pollution pr. However at the current state if updated the history pollution should work and the undo.

ice0:
@mohamedAdhamc 

Try to merge changes from [color-wheel-history-pollution](https://github.com/mohamedAdhamc/synfig/tree/color-wheel-history-pollution) branch and check if it works.

mohamedAdhamc:
ok, looks like ill have to clean up this branch a bit before seeing how it looks like here, as a previous way was implemented here which conflicts with the new "to be used" way

mohamedAdhamc:
ok @ice0 , i merged the changes and it seems like its working fine (exactly same way as in color wheel). 

while testing some things came into mind though.
1- should the increase/decrease increase in magnitude as the mouse goes further to the right/left without stopping ?
     i.e. the more I keep moving the cursor to the right the faster the increase is.

PR #2719, opened 2022-06-30
fix: Indicate segments where points are disabled in Timetrack Panel (regression)

### LINKED ISSUE #2716: Indicate segments where points are disabled in Timetrack Panel
In old TimeTrack (synfig 1.4.x and earlier) the time spans where a Dynamic List entry is kept off is indicated by hatches.
See it on the left part of screenshot -
screenshot_001

We need to indicate those regions in new Timetrack Panel.

I suggest to indicate those regions with darker color (black color with 50% alpha) -
screenshot_002

rodolforg:
Fix #2716

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_timetrack.cpp
+++ b/synfig-studio/src/gui/widgets/widget_timetrack.cpp
@@ -858,13 +858,45 @@ void Widget_Timetrack::draw_active_point_status(const Cairo::RefPtr<Cairo::Conte
 	const synfig::ValueNode_DynamicList::ListEntry& list_entry = dynamic_list->list[ value_desc.get_index() ];
 	const synfig::ValueNode_DynamicList::ListEntry::ActivepointList& activepoint_list = list_entry.timing_info;
 
+	if (activepoint_list.empty())
+		return;
+
 	const Gdk::RGBA activepoint_color[] = {
 	    Gdk::RGBA("#ff0000"),
 	    Gdk::RGBA("#00ff00")
 	};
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/widgets/widget_timetrack.cpp
```diff
@@ -862,9 +862,39 @@ void Widget_Timetrack::draw_active_point_status(const Cairo::RefPtr<Cairo::Conte
...
+		} else {
+			if (initial_off_time != synfig::Time::end()) {
+				cr->rectangle(+w/2+time_plot_data->get_pixel_t_coord(initial_off_time), row_info.get_geometry().y, -w+time_plot_data->get_delta_pixel_from_delta_t_coord(activepoint.get_time() - initial_off_time), row_info.get_geometry().h);
+				cr->fill();
+				initial_off_time = synfig::Time::end();
+			}
+		}
+	}
+	if (initial_off_time != synfig::Time::end()) {
+		int initial_pixel_index = +w/2+time_plot_data->get_pixel_t_coord(initial_off_time);
```
ice0, line 891: `+w`? What does it mean?

rodolforg: The green or red line width

rodolforg:
I just found out that OFF point only means OFF in that time point o.O Right after it is ON again...

rodolforg:
> I just found out that OFF point only means OFF in that time point o.O Right after it is ON again...

Now it's properly working :)

ice0 (approved):

ice0:
Merged. Thank you!

PR #2727, opened 2022-07-02
fix: layers parameter 'filename' does not accept backslash as directory separator

### LINKED ISSUE #2723: Linux version not understands "\" as directory separator
**Describe the bug**
Linux version of Synfig does not parses "\\" as directory separator (in image filenames).

**To Reproduce**
1. Download and unpack sample archive - [paths.zip](https://github.com/synfig/synfig/files/9032840/paths.zip)
2. Open `paths.sif` file.
3. See image is not displayed -

screenshot_005

In version 1.4.2 it is displayed correctly -

screenshot_004

You can examine image layer, and see its "Filename" property -
screenshot_006

Notice that it uses "\\" as directory separator. This usually happens when file was created in WIndows version of Synfig.

**Expected behavior**
Synfig should treat both "/" and "\\" as directory separators.

rodolforg:
Maybe we should fix them on canvas loading.
Probably it would be good to uniform them on saving too as well
on choosing via Widget_Filename too.

Fix #2723

### DIFF
--- a/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
+++ b/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
@@ -730,7 +730,8 @@ Layer_Freetype::set_shape_param(const String & param, const ValueBase &value)
 */
 	IMPORT_VALUE_PLUS(param_family,
 		{
-			synfig::String family=param_family.get(synfig::String());
+			synfig::String family = FileSystem::fix_slashes(value.get(synfig::String()));
+			param_family.set(family);
 			int style=param_style.get(int());
 			int weight=param_weight.get(int());
 			new_font(family,style,weight);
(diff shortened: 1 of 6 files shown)


rodolforg:
Backport it?

rodolforg:
(force pushed to fix indentation)

ice0 (approved):

ice0:
> Maybe we should fix them on canvas loading.
> Probably it would be good to uniform them on saving too as well
> on choosing via Widget_Filename too.

Or maybe switch to `synfg::filesystem::Path` for all paths?

> Backport it?

Sure.

rodolforg:
> Or maybe switch to `synfg::filesystem::Path` for all paths?

Yes! :) Long journey, but it would be a great solution!

morevnaproject:
> > Backport it?
> 
> Sure.

This doesn't affects 1.4.x branch, so no backporting.

PR #2729, opened 2022-07-03
fix: avoid rebuild ParamTreeStore when animating a parameter

### LINKED ISSUE #386: Parameters panel collapses items on edit
When Animation Mode is enabled, then editing any sub-parameter, which wasn't animated previously will collapse all items in Parameters Panel.

Steps to reproduce:
1. Create Group Layer
2. Turn on Animation Mode
3. Expand Transformation Parameter
4. Edit Angle Parameter
5. -> Transformation parameter collapsing

Original post:

> When editing vertices in a spline, the parameters panel will unpredictably collapse everything after you edit an item, like the tangent values. This doesn't happen all the time, but it does happen more the 75% of the time, and is very frustrating when you're trying to edit a spline with multiple vertices.

See also:
- https://forums.synfig.org/t/weekly-reports-2022/12854/4?u=konstantindmitriev
- https://github.com/synfig/synfig/issues/466
- https://github.com/synfig/synfig/issues/573

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46967891-parameters-panel-collapses-items-on-edit?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321991&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2390: Animating Active Layer Name of Switch Layer crashes Synfig Studio
**Describe the bug**
Animating `Active Layer Name` of the attached file crashes Synfig Studio.

**Step to Reproduce**
1. Open the attached animation file:
[SwitchCrash.zip](https://github.com/synfig/synfig/files/7436429/SwitchCrash.zip)

2. Turn on animate editing mode
3. In the `Layers` window select `Occhiale` > `Occhio SX`
4. In the `Parameters` window change `Active Layer Name`

Synfig Studio crashes consistently.

**Expected behavior**
Animation should be applied as usual.
The crash doesn't happen if animate mode is disabled.

**System information:**
 - OS: Windows
 - OS version: 10
 - Commit ID: 78e045f

**Additional comments**
The `Switch` was duplicated after exporting the `Active Layer Name`.

rodolforg:
Calling rebuild() causes two bugs:
1. Parameters panel collapses items on edit (GH#386), when we
   we have Animation Mode enabled and are editing parameter that
   wasn't previously animated.
   'Solution' seems to be to call queue_refresh()
2. Crash (or emit Gtk warning) when we animate a parameter (like
   previous case) that uses a combobox for editing (GH#2390), as
   Active Layer Name of Switch Layer or Bone Parent in Bones list
   of Skeleton Layer.
   'Solution' seems to be to call queue_rebuild()

However, when a value node is replaced (via RHandle::replace()),
it already calls Node::signal_changed() so the row is refreshed
by other signal slot.

Fix #386
Fix #2390

### DIFF
--- a/synfig-studio/src/gui/trees/layerparamtreestore.cpp
+++ b/synfig-studio/src/gui/trees/layerparamtreestore.cpp
@@ -598,7 +598,26 @@ LayerParamTreeStore::on_value_node_replaced(synfig::ValueNode::Handle /*replaced
 	// failing as a result.  not sure how the tree code has a pointer
 	// rather than a handle - maybe it has a loosehandle, that would
 	// cause the problem I was seeing
-	rebuild();
+
+	// Maybe the comment above is deprecated. -^
+	// It seems we don't need to use the old solution (call queue_refresh()),
+	// neither the previous solution (rebuild(), commented out below).
+	// When a value node is replaced (via RHandle::replace()), it
+	// already calls Node::signal_changed() so the row is refreshed
+	// by other signal slot.
+	//
+	// Calling rebuild() causes two bugs:
+	// 1. Parameters panel collapses items on edit (GH#386), when we
+	//    we have Animation Mode enabled and are editing parameter that
+	//    wasn't previously animated.
+	//    'Solution' seems to be to call queue_refresh()
+	// 2. Crash (or emit Gtk warning) when we animate a parameter (like
+	//    previous case) that uses a combobox for editing (GH#2390), as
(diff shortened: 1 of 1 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

PR #2730, opened 2022-07-03
fix: Synfig hangs when importing palette

### LINKED ISSUE #2720: Palette import bug
**Describe the bug**
Importing GIMP(.gpl) palette into Synfig causes it to freeze.

**To Reproduce**

[Animation-#1.zip](https://github.com/synfig/synfig/files/9019957/Animation-.1.zip)
Import the given palette into synfig. It will freeze.
**Expected behavior**
It should successfully import the palette.

**System information:**
 - OS: Windows 10 64bit
 - OS version:21h2
 - Synfig version: 1.5.1

ice0:
The issue was that the `unget()` method calls the `pbackfail()`
method, which is not implemented, because of this the error
flag is set and the program goes into an infinite loop.

Fix #2720

### DIFF
--- a/synfig-core/src/synfig/filesystemnative.cpp
+++ b/synfig-core/src/synfig/filesystemnative.cpp
@@ -70,6 +70,14 @@ std::istream::pos_type FileSystemNative::ReadStream::seekpos(std::istream::pos_t
 	return ftell(file_);
 }
 
+int FileSystemNative::ReadStream::pbackfail(int ch)
+{
+	if (fseek(file_, -1, SEEK_CUR)) {
+		return EOF;
+	}
+	return FileSystem::ReadStream::underflow();
+}
+
 // WriteStream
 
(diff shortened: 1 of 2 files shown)


synfig-core/src/synfig/filesystemnative.cpp
```diff
@@ -70,6 +70,12 @@ std::istream::pos_type FileSystemNative::ReadStream::seekpos(std::istream::pos_t
 	return ftell(file_);
 }
 
+int FileSystemNative::ReadStream::pbackfail(int ch)
+{
+	fseek(file_, -1, SEEK_CUR);
```
rodolforg, line 75: Would be good to check the returned value of `fseek()`?
`pbackfail()` should return `EOF` in case of failure.

ice0: Done

rodolforg (approved):

PR #2689, opened 2022-06-15
feat: Added support for `Ctr+Enter` to complete text input in Text tool dialog

mohamedAdhamc:
A new clean PR from the orginal in #2588 , which will be closed but I'm just waiting for approval.

PR #2160, opened 2021-05-18
Basic implementation of Keyboard Shortcut Editor [studio]

rodolforg:
Some other features could be done, for example:
- Own button for 'restore default shortcuts only' (instead of together with all other Synfig Studio settings)
- Save and load specific/custom accel map file (it always read/overwrite [config-folder]/accelrc)
- List all available actions (currently it lists only those with a default shortcut set in app.cpp)
- Group actions in a Tree view structure (but I'm lazy for that)
- List with better (simpler) names and maybe tooltips: right now it's the action full path. However, it would require a proper registration on every case, and this PR is just a quick implementation
- Warn user if a shortcut is used for another action

fix #422

PR #2731, opened 2022-07-04
feat: added support for `Ctr+<num>` hotkey to change animation tabs

### LINKED ISSUE #1422: Shortcut for switching between animation windows
Can we have something like `Alt + 1`/`Alt + 2` to switch between animation windows. This is used in firefox to change the tabs. Google chrome uses `Ctrl + num` for the same.
I guess this will be useful for people who have multiple tabs of animation windows open.

mohamedAdhamc:
Fixes #1422
This is feature 3 in my project. I chose to do it before 2 as it's kind of a subset of 2. Maybe they should be both merged into one feature where I just do the changes I did here but as an override in the DocKBook class *see last paragraph*. 

Implementation Description:
Ctr+page num move to the page needed. Also ctr+9 moves to last page. 

p.s. no need to check if the specified page exists as in the decs for the set_current_page() method it states
"If greater than the number of pages in the notebook, nothing will be done."

Also maybe just use a macro instead of basically repeating the same two/three lines nine times? although I don't think it would really affect performance I guess it's down to your preferences here.

Now one last thing, I did a quick search and it doesn't seem that the other notebooks are dockbooks I could be wrong though but they seem to be of different overloaded types. If they are dockbooks also though I could just have this implementation done as an override in the dockbook class.


rodolforg:
`DockBook` is used everywhere a panel exists. In the Synfig Studio main window, there are 7 dockbooks in the default workspace ;)

Regarding the code repetition, you could create a `std::map` that binds a keyval to its page num. Or keep the switch statement, but it would set a variable, ad the function call could happen after it.


> Ctr+page num move to the page needed. Also ctr+9 moves to last page.

Firefox style, I got it ;)

mohamedAdhamc:
> 
> Or keep the switch statement, but it would set a variable, ad the function call could happen after it.

I think that's fairly simple, and a better way of doing this. Will update in a couple hours when I wake up :-)

ice0:
1. Didn't work for me. Pressing `Ctrl+1` didn't switch to first tab.
2. MacOS uses Command instead of Control. So you should use `<Primary>` key in GTK for that.
3.
```
if (event->keyval >= GDK_KEY_1 && event->keyval >= GDK_KEY_9) {
	if (event->keyval == GDK_KEY_9) {
		main_dock_book().set_current_page(-1); //negative moves to last page
	} else {
		main_dock_book().set_current_page(event->keyval - GDK_KEY_1);
	}
}
```

mohamedAdhamc:
> 1. Didn't work for me. Pressing `Ctrl+1` didn't switch to first tab.
> 2. MacOS uses Command instead of Control. So you should use `<Primary>` key in GTK for that.
> 

@ice0 Are these two related ? 

> ```
> if (event->keyval >= GDK_KEY_1 && event->keyval <= GDK_KEY_9) {
> 	if (event->keyval == GDK_KEY_9) {
> 		main_dock_book().set_current_page(-1); //negative moves to last page
> 	} else {
> 		main_dock_book().set_current_page(event->keyval - GDK_KEY_1);
> 	}
> }
> ```

seems a lot better yes.

ice0:
> > 1. Didn't work for me. Pressing `Ctrl+1` didn't switch to first tab.
> > 2. MacOS uses Command instead of Control. So you should use `<Primary>` key in GTK for that.
> 
> @ice0 Are these two related ? (I mean were you using MacOS).

No, not related I testing it on Linux (XFCE).
Just tested on Windows - it works fine. Maybe in XFCE this hotkey is reserved for something?

> If not then was the main window/canvas window in focus ? as if not it wouldn't work yes.

Yes, window is in focus.

> p.s. I'm a bit confused about the `<primary>` key bit. the [documentation ](https://docs.gtk.org/gdk4/flags.ModifierType.html)for the modifier keys used in the events state doesn't seem to include anything about the apple/primary modifier (in the listed members).

You are checking docs for GTK4. GTK3 docs (https://docs.gtk.org/gdk3/flags.ModifierType.html)

Unfortunately, I also didn't find something like the `Primary` modifier. 
We use it as a shortcut key in other places, for example:
```
  const gchar *quit_accels[2] = { "<Primary>q", NULL };

  gtk_application_set_accels_for_action (GTK_APPLICATION (app), "app.quit", quit_accels);
```

> Also in #2689 it was done same as this so is something missing there as well for apple ?

Yes, I forgot to check it :/

Anyway, I realized that it's better to implement it differently.

Please check this PR (https://github.com/synfig/synfig/pull/2160).
Implementing this feature in the same way will allow user to change the shortcuts, and they will also be visible in the shortcuts panel. (Edit->Preferences->Shortcuts)

mohamedAdhamc:
> 
> Anyway, I realized that it's better to implement it differently.
> 
> Please check this PR (#2160). Implementing this feature in the same way will allow user to change the shortcuts, and they will also be visible in the shortcuts panel. (Edit->Preferences->Shortcuts)

Sounds good for users to be able to change the shortcut if they want and also have it visible for them to see. from what I believe I would have to make an action and it would basically be what is called an accelerator ( or maybe Im not quite on the same page here). So the question here is, would it make sense to have nine accelerators for this ? There might be a way not to but id have to look into it then.

Anyways, I will be working on this in the morning.

ice0:
> So the question here is, would it make sense to have nine accelerators for this ? There might be a way not to but id have to look into it then.

I think it's okay to have nine accelerators. If there is place for optimization, we can optimize them after that.

mohamedAdhamc:
Ok so I implemented it in the shortcut panel and will be pushing this commit in a bit. However, should this feature be for only the animation windows or should I just make it general for all docks? for now, I will be committing the one for the animation windows only until I get an opinion from you guys however the switch for all dock books shouldn't take time I already have it done too.

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -194,6 +194,35 @@ MainWindow::init_menus()
 		sigc::ptr_fun(App::edit_custom_workspace_list)
 	);
 
+	//animation tabs
+	action_group->add( Gtk::Action::create("move-to-tab-1", Gtk::StockID("synfig-move-to-tab-1"), _("Move to Tab 1"), _("")),
```
ice0, line 198: You can use `for` cycle here:
```
for (...) {
	action_group->add( Gtk::Action::create("move-to-tab-2", Gtk::StockID("synfig-move-to-tab-2"), _("Move to Tab 2"), _("")),
		sigc::bind<int>(sigc::mem_fun(*this, &MainWindow::move_tab), n)
	);
}
```

Also try to not use `Gtk::StockID` because it's already deprecated.

ice0:
> However, should this feature be for only the animation windows or should I just make it general for all docks? for now, I will be committing the one for the animation windows only until I get an opinion from you guys however the switch for all dock books shouldn't take time I already have it done too.

Since there are no visible indicators of the currently active dock, I suggest implementing it for the animation window only.

synfig-studio/src/gui/mainwindow.cpp
mohamedAdhamc: yes and storing the names in a container sounds good will do.

mohamedAdhamc:
quick update here guys we have a holiday today so unfortunately, I won't be able to do anything today. However, hopefully tomorrow I will try to resolve the points by ice, and continue with other things.

mohamedAdhamc:
done

synfig-studio/src/gui/app.cpp
```diff
@@ -1333,6 +1352,15 @@ App::get_default_accel_map()
 		{"<Control>l",              "<Actions>/canvasview/toggle-grid-snap"},
 		{"<Control>n",              "<Actions>/mainwindow/new"},
 		{"<Control>o",              "<Actions>/mainwindow/open"},
+		{"<Control>1",              "<Actions>/mainwindow/move-to-tab-1"},
```
ice0, line 1355: Use `<Primary>` instead of `<Control>`

mohamedAdhamc: done and actually on windows it works same as control, which makes me wonder why all the "Control" are not just "Primary".

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -194,6 +194,22 @@ MainWindow::init_menus()
 		sigc::ptr_fun(App::edit_custom_workspace_list)
 	);
 
+	//animation tabs
+	std::string arg_1[9] = {"move-to-tab-last","move-to-tab-1","move-to-tab-2","move-to-tab-3",
+						 "move-to-tab-4","move-to-tab-5","move-to-tab-6","move-to-tab-7","move-to-tab-8"};
+	std::string arg_2[9] = {"_(synfig-move-to-tab-last)","_(synfig-move-to-tab-1)","_(synfig-move-to-tab-2)","_(synfig-move-to-tab-3)",
```
ice0, line 200: Can you explain what this means?

synfig-studio/src/gui/app.cpp
```diff
@@ -1057,13 +1057,32 @@ DEFINE_ACTION("help-about",     Gtk::StockID("synfig-about"))
...
+DEFINE_ACTION("move-to-tab-8",  _("synfig-move-to-tab-8"))
+DEFINE_ACTION("move-to-tab-last",  _("synfig-move-to-tab-last"))
 
 //Layout the actions in the main menu (caret menu, right click on canvas menu) and toolbar:
 	Glib::ustring ui_info_menu =
 "	<menu action='menu-file'>"
 "		<menuitem action='new' />"
 "		<menuitem action='open' />"
 "		<menu action='menu-open-recent' />"
+"		<menu action='move-to-tab-1' />"
```
ice0, line 1077: Can you explain why do you add this actions to the menu?

synfig-studio/src/gui/mainwindow.cpp
mohamedAdhamc: these are just the arguments stored in an array so that the actions can be created with the loop.

synfig-studio/src/gui/app.cpp
mohamedAdhamc: before adding this the action wasn't working. i.e the callback wasn't being called so I read that for an action to function properly it must be added as a menu item even if it doesn't show and sure enough that did the trick.

synfig-studio/src/gui/mainwindow.cpp
mohamedAdhamc: I couldn't really find t in the documentation as `Gtk::Action` as a whole seems to be deprecated and should be replaced by `Gio::Action` now. However, from what I saw in Synfigs code base so far, it's used either at times for stock ids, and other times in the _("") notation which I'm assuming then would be some sort of a label.

ice0: > I couldn't really find t in the documentation as Gtk::Action

https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1Action.html

> and other times in the _("") notation which I'm assuming then would be some sort of a label.

Yes. And you still not see the issue here?
https://stackoverflow.com/questions/10566109/fputs-what-does-the-underscore-stand-for

mohamedAdhamc: > 
> https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1Action.html

Yes, this is my go-to usually as it's the official gtkmm3 documentation. However, Gtk::Action::Create method only appears 3 times and none with the same number of arguments --2--.
create
Maybe I'm somehow looking into it wrong ? i'm asking so that i can maybe avoid this in the next time.

>
>Yes. And you still not see the issue here?
https://stackoverflow.com/questions/10566109/fputs-what-does-the-underscore-stand-for

I hadn't seen this post, looking through it. it seems like this notation is necessary for being able to translate the string into other languages so then it seems like I actually have to stick to it, and not what I did here. 

update: I think now i have the get text done same as in DEFINE_ACTION hopefully this is what you mean. and thanks for the insight.

ice0: > Yes, this is my go-to usually as it's the official gtkmm3 documentation. However, Gtk::Action::Create method only appears 3 times and none with the same number of arguments --2--.

I think this should help you: https://en.cppreference.com/w/cpp/language/default_arguments

ice0:
> p.s. I think maybe instead of the current format synfig-move-to-tab-1 it should be something like Move to Tab 1 not sure though.

Yes, this will be much better.

mohamedAdhamc:
done

synfig-studio/src/gui/mainwindow.cpp
mohamedAdhamc: > 
> I think this should help you: https://en.cppreference.com/w/cpp/language/default_arguments

Yes !, that answers multiple questions actually thanks.

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -194,6 +194,22 @@ MainWindow::init_menus()
 		sigc::ptr_fun(App::edit_custom_workspace_list)
 	);
 
+	//animation tabs
+	std::string arg_1[9] = {"move-to-tab-last","move-to-tab-1","move-to-tab-2","move-to-tab-3",
```
ice0, line 198: I think this part can be improved. Something like this:
```
for (int i = 1; i <= 8; ++i) {
	const std::string tab = std::to_string(i);
	action_group->add(Gtk::Action::create("move-to-tab-" + tab, _("Move to Tab ") + tab),
		[this, i]() { main_dock_book().set_current_page(i-1); }
	);
}
action_group->add(Gtk::Action::create("move-to-tab-last", _("Move to Last Tab")),
	[this]() { main_dock_book().set_current_page(-1); }
);
```

ice0:
> done, anything still missing here?

I noticed that switching to the first tab does not switch the layer in the Canvas browser.
For tabs 2-last everything works fine. Can you check it?


Screenshot_7

synfig-studio/src/gui/mainwindow.cpp
mohamedAdhamc: I actually really like this it basically does everything at once (^_^).

mohamedAdhamc:
> I noticed that switching to the first tab does not switch the layer in the Canvas browser. For tabs 2-last everything works fine. Can you check it?

I checked it first by seeing if using another function like prev_page() would also result in page 1 also not being shown in canvasbrowser. which yes it did do that so I was thinking this probably has to do with something in the implementation. so I came across the function i edited, its the signal_switch_page() override. by commenting it all out and basically removing the override none of the new pages show in the canvas browser, so this is necessary for that to happen. 

`this->page_num(*current_page)` yields out zero which is the first pages index, maybe whoever wrote this didn't notice? 
there maybe some far reasoning I do not grasp here but this way switching to page 1 works good.

p.s
I also noticed for me it wasn't just the canvas browser not showing the correct canvas it was also that tab 1 would never grab focus when reached by any of the functions.

ice0:
> `this->page_num(*current_page)` yields out zero which is the first pages index, maybe whoever wrote this didn't notice?

This change was introduced by this commit (d5e01e26) and it looks like just an error.
So I suggest to fix it in new PR.

mohamedAdhamc:
> 
> This change was introduced by this commit ([d5e01e2](https://github.com/synfig/synfig/commit/d5e01e26f58f68e5c7e142c7fcd5d65581ca093c)) and it looks like just an error. So I suggest to fix it in new PR.

ok will remove it from here and make a separate PR for this.

ice0:
> ok will remove it from here and make a separate PR for this.

Didn't notice you already fixed it.
Anyway I see a better way to fix it, so please check documentation for `page_num(...)` method.

> p.s. i might sleep though so if that happens will continue when I wake up :-)

Ok.

synfig-studio/src/gui/app.cpp
```diff
@@ -1057,13 +1057,32 @@ DEFINE_ACTION("help-about",     Gtk::StockID("synfig-about"))
 // actions: Keyframe
 DEFINE_ACTION("keyframe-properties", _("Properties"))
 
+// actions: move to tab
+DEFINE_ACTION("move-to-tab-1",  _("Move to Tab 1"))
```
ice0, line 1061: 1. The title does not match the content of the PR.
2. I don't know why using a loop to generate names in one place you didn't use it in another.

ice0: Every `_()` call adds a string to translation. Less strings to translate - less work need to be done.

mohamedAdhamc: oh ok now it makes much more sense and the numbers dont get translated anyways, perfect!

synfig-studio/src/gui/app.cpp
```diff
@@ -1058,14 +1058,9 @@ DEFINE_ACTION("help-about",     Gtk::StockID("synfig-about"))
...
 // actions: move to tab
-DEFINE_ACTION("move-to-tab-1",  _("Move to Tab 1"))
-DEFINE_ACTION("move-to-tab-2",  _("Move to Tab 2"))
-DEFINE_ACTION("move-to-tab-3",  _("Move to Tab 3"))
-DEFINE_ACTION("move-to-tab-4",  _("Move to Tab 4"))
-DEFINE_ACTION("move-to-tab-5",  _("Move to Tab 5"))
-DEFINE_ACTION("move-to-tab-6",  _("Move to Tab 6"))
-DEFINE_ACTION("move-to-tab-7",  _("Move to Tab 7"))
-DEFINE_ACTION("move-to-tab-8",  _("Move to Tab 8"))
+for (int i = 1; i <= 8; ++i) {
```
ice0, line 1061: please use:
```
for () {
	...
} <- on the new line
```

mohamedAdhamc: alright, and by the way did you guys set the new rules for indentation that was being talked about on the forums yet?

ice0:
Looks good.
Please check conventionalcommits.org to get the idea for PR naming.

Also you can check the latest merged PRs to get an example: https://github.com/synfig/synfig/pulls?q=is%3Apr+is%3Aclosed

mohamedAdhamc:
> Looks good. Please check https://conventionalcommits.org to get the idea for PR naming.
> 
> Also you can check the latest merged PRs to get an example: https://github.com/synfig/synfig/pulls?q=is%3Apr+is%3Aclosed

sounds great! will do.

synfig-studio/src/gui/app.cpp
ice0: I am not sure. 

Anyway I am trying to copy the style from the previous code, in this case the new code will look the same.

synfig-studio/src/gui/app.cpp
```diff
@@ -1057,13 +1057,28 @@ DEFINE_ACTION("help-about",     Gtk::StockID("synfig-about"))
 // actions: Keyframe
 DEFINE_ACTION("keyframe-properties", _("Properties"))
 
+// actions: move to tab
+for (int i = 1; i <= 8; ++i) {
+	const std::string tab = std::to_string(i);
+	DEFINE_ACTION("move-to-tab-" + tab, _("Move to Tab ") + tab)
```
ice0, line 1063: 1. I think may be better to replace the word "Move" with "Switch" or "Jump" as a more established term?

https://support.google.com/chrome/answer/157179?hl=en&co=GENIE.Platform%3DDesktop#zippy=%2Ctab-window-shortcuts

https://workona.com/blog/switch-tabs-chrome-shortcut/#:~:text=The%20keyboard%20shortcut%20Cmd%20%2B%20Shift,switch%20between%20tabs%20in%20Chrome.

2. I noticed the term "Jump to the rightmost tab" in Google help. For right-to-left languages, will the last tab be on the left or right? Also "last" is not the best word here, because the user may be thinking about the last tab he worked on.

mohamedAdhamc: 1- sure, looks like switch is more established and appropriate will be doing this.

2- I think regarding right to left languages (as is Arabic which I  speak) rightmost shouldnt realy be a problem the las tab is still the rightmost tab. I even checked ( to maybe see if in synfig order is reversed upon changing to this type of language -which would be not standard behaviour-) but even in arabic still the rightmost tab is the last tab, so I think its ok. 
p.s. yes last could posssibly cause confusions so rightmost would probably be bettter ill also check how they named it in other browsers.

mohamedAdhamc: done.

synfig-studio/src/gui/app.cpp
rodolforg: There is a "hidden-menu" somewhere else, to avoid showing stuff in the menu.

mohamedAdhamc: hmm, ok seems better to move this there then.

mohamedAdhamc: done, I guess this can now be safely resolved :)

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -194,6 +194,17 @@ MainWindow::init_menus()
 		sigc::ptr_fun(App::edit_custom_workspace_list)
 	);
 
+	//animation tabs
+	for (int i = 1; i <= 8; ++i) {
+		const std::string tab = std::to_string(i);
+		action_group->add(Gtk::Action::create("switch-to-tab-" + tab, _("Switch to Tab ") + tab),
+			[this, i]() { main_dock_book().set_current_page(i-1); }
```
rodolforg, line 201: Remember that issue with signal slots that are lambda functions?
Protect them  (`sigc::track_obj`) or change to a class member function (`sigc::mem_fun()`)

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -194,6 +194,17 @@ MainWindow::init_menus()
...
+		);
+	}
+	action_group->add(Gtk::Action::create("switch-to-rightmost-tab", _("Switch to Rightmost Tab")),
+		[this]() { main_dock_book().set_current_page(-1); }
```
rodolforg, line 205: Same above:
Remember that issue with signal slots that are lambda functions?
Protect them (sigc::track_obj) or change to a class member function (sigc::mem_fun())

synfig-studio/src/gui/mainwindow.cpp
mohamedAdhamc: oh yes of course, done ^_^.

synfig-studio/src/gui/mainwindow.cpp
mohamedAdhamc: yep, done

mohamedAdhamc:
@rodolforg done :)

rodolforg (approved):

ice0:
Merged. Thank you!

PR #2737, opened 2022-07-08
fix: mod_libavcodec didn't compile with MinGW

ice0:
```
error: invalid conversion from 'const AVOutputFormat*' to 'AVOutputFormat*' [-fpermissive]
  259 |                 context->oformat = format;
```

`context->oformat` is `AVOutputFormat*` before FFMPEG 5.0. It builds fine with gcc on Linux, so I missed that.


rodolforg (approved):
Approved as it builds on Linux, Windows and MacOS

PR #2738, opened 2022-07-08
test: create tests for synfig/string_helper.h

rodolforg:


synfig-core/test/string.cpp
```diff
@@ -0,0 +1,187 @@
...
+{
+	ASSERT_EQUAL("Something", trim("Something"))
+}
+
+/* === E N T R Y P O I N T ================================================= */
+
+int main() {
+
+	char* orig_LANG = getenv("LANG");
+	setenv("LANG", "en", 1);
```
ice0, line 155: If this is used for number formatting, perhaps it's better to use `std::setlocale(...)`?

rodolforg: Yes, it is.
But for some reason I can't remember, we force setlocale in the method

ice0: I think because of this comment:
> // MinGW C++ std::locale accepts "C" and "POSIX" it does not support other locales.

But in this case it's safe to use "C" locale for this test.

ice0:
Please check with different locale:
```
$ sudo locale-gen de_DE.UTF-8
$ sudo update-locale 

$ LC_NUMERIC=de_DE.UTF-8 ./string

FFF..FF...........
synfig(32279) [11:17:18] warning: FAILURE:
 test_remove_trailing_zeroes_without_trailing_zeroes_returns_unchanged_string:40
	 - expected 10.0305, but got 10.0305,0

synfig(32279) [11:17:18] warning: FAILURE:
 test_remove_trailing_zeroes_without_trailing_zeroes_and_no_decimal_point_appends_dot_zero:47
	 - expected 102.0, but got 102,0

synfig(32279) [11:17:18] warning: FAILURE:
 test_remove_trailing_zeroes_with_trailing_zeroes_but_no_decimal_point_appends_dot_zero:54
	 - expected 100.0, but got 100,0

synfig(32279) [11:17:18] warning: FAILURE:
 test_remove_trailing_zeroes_with_real_trailing_zeroes_does_remove_them:74
	 - expected 100.3, but got 100.3000,0

synfig(32279) [11:17:18] warning: FAILURE:
 test_remove_trailing_zeroes_with_only_trailing_zeroes_keeps_the_decimal_point_by_default:81
	 - expected 100.0, but got 100.000,0


========================================================================
synfig(32279) [11:17:18] warning: Statistics:
 5 tests failed.
 13 successful tests

```

synfig-core/test/string.cpp
rodolforg: Is this it what you suggest?

ice0: If I understand correctly this should be something like:
`synfig::ChangeLocale change_locale(LC_NUMERIC, "C");`

rodolforg: It won't work.
There is a ` synfig::ChangeLocale change_locale(LC_NUMERIC, "");` inside `synfig::remove_trailing_zeroes()`
https://github.com/synfig/synfig/blob/aa8f1f35bcd67a7aaf765b6c7b3e152a39cfc1af/synfig-core/src/synfig/string_helper.cpp#L42-L56

rodolforg: I wonder if line 47 is still really needed (` 	synfig::ChangeLocale changeLocale(LC_NUMERIC, ""); `).

The non-MinGW32 'version' just get the current locale name (`setlocale(LC_NUMERIC, nullptr)`) and create an object with this locale name. It doesn't actually change the locale.

MinGW32 does not support this strategy, as you pointed out, but `localeconv()` already gets the current locale info. Is there need to force the locale?

Analysing via `git blame` and `git log`, the first usage of `synfig::remove_trailing_zeroes()` is with the commit fd1e5ecdc3cf2cf9cb2ee642ca7f7b89aa04b40b (PR #2102), although the method was introduced (but not used anywhere) in a509ebd64fd7af777130abd9753e2e333ffb8575 (PR #1995). At that time, it didn't force a locale.

Later, to fix the MingW issue, we introduced the ChangeLocale in 3c0f04d55e54459e36b7872812a07772989e74d0 (PR #2157). Maybe we missed it there?

ice0: I am not sure, but maybe this line is needed to read locale set by environment variables. Without it it should always return "C" locale.

https://man7.org/linux/man-pages/man3/setlocale.3.html
> If locale is an empty string, "", each part of the locale that
> should be modified is set according to the environment variables.

Also related discussion: 
https://stackoverflow.com/questions/33117258/c-setlocale-not-working-as-expect

rodolforg: > If locale is a null pointer, setlocale queries the current C locale without modifying it. 
(https://en.cppreference.com/w/c/locale/setlocale)



> Also related discussion:
> https://stackoverflow.com/questions/33117258/c-setlocale-not-working-as-expect

This is about user setting wrong locale name (wrong: `en_US`, correct: `en_US.utf8`)

ice0: > > Also related discussion:
> > https://stackoverflow.com/questions/33117258/c-setlocale-not-working-as-expect
> 
> This is about user setting wrong locale name (wrong: `en_US`, correct: `en_US.utf8`)

He asks two questions. Second was:
> When try to get current LC_TIME, it just print C, but not the value specified from environment variable.

And the comments points the need to read the values from the environment first.

rodolforg: Here are where we already set locale:

https://github.com/synfig/synfig/blob/b836a72d4149552d562863db62efd5fa567fbe51/synfig-core/src/tool/main.cpp#L78-L82

https://github.com/synfig/synfig/blob/b836a72d4149552d562863db62efd5fa567fbe51/synfig-studio/src/gui/main.cpp#L61-L71

ice0: It's still fails with `de_DE` locale (`LC_NUMERIC=de_DE.UTF-8 ./string`). 
Can you check the issue?

rodolforg: What about now?
It's necessary to test on MS Windows...

synfig-core/test/CMakeLists.txt
```diff
@@ -25,12 +25,13 @@ add_test(NAME test_synfig_clock COMMAND test_synfig_clock)
...
 
-#add_executable(test_synfig_string string.cpp)
-#target_link_libraries(test_synfig_string PRIVATE synfig)
-#add_test(NAME test_synfig_string COMMAND test_synfig_string)
+add_executable(test_synfig_string string.cpp)
+target_link_libraries(test_synfig_string PRIVATE synfig)
+add_test(NAME test_synfig_string COMMAND test_synfig_string)
 
 set_target_properties(
         test_synfig_angle test_synfig_benchmark test_synfig_bline test_synfig_bone test_synfig_clock
```
rodolforg, line 33: whoops, forgot to add the test here...

synfig-core/test/string.cpp
ice0: It works on Linux. Test on Windows today.

synfig-core/test/CMakeLists.txt
rodolforg: Fixed

synfig-core/test/string.cpp
rodolforg: > It works on Linux. Test on Windows today.

Could you test it?

ice0: Sorry, been busy testing vcpkg.
Works without any issues.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2746, opened 2022-07-11
build(CMake): moved synfig-core tests to `bin/test` folder

ice0:


rodolforg (approved):
Approving although I don't understand CMake XD

PR #2750, opened 2022-07-14
Color Wheel History Pollution Fix

### LINKED ISSUE #451: Color wheel fills history too much

mohamedAdhamc:
Fixes #451 

Using a flag to know when there are rapid repeated actions (as in here and in #2717 ) then if it's a rapid repeated action only showing in the history panel the very last value.

p.s. I have to go now there is still a small thing left to do for when the color wheel is set using the keys I will do it when im back.

### DIFF
--- a/synfig-studio/src/gui/trees/historytreestore.cpp
+++ b/synfig-studio/src/gui/trees/historytreestore.cpp
@@ -56,6 +56,8 @@ using namespace studio;
 
 /* === M E T H O D S ======================================================= */
 
+bool HistoryTreeStore::block_new_history = false;
+
 static HistoryTreeStore::Model& ModelHack()
 {
 	static HistoryTreeStore::Model* model(nullptr);
@@ -209,6 +211,7 @@ HistoryTreeStore::on_redo_stack_cleared()
 void
 HistoryTreeStore::on_new_action(etl::handle<synfigapp::Action::Undoable> action)
 {
+ if (!block_new_history){ /*if change isn't due to rapid repeated actions show history normally*/
 	Gtk::TreeRow row;
 
 	row=*insert(next_action_iter);
@@ -218,7 +221,7 @@ HistoryTreeStore::on_new_action(etl::handle<synfigapp::Action::Undoable> action)
 	next_action_iter = row;
 	++next_action_iter;
 
-	signal_undo_tree_changed()();
+	signal_undo_tree_changed()();}
 }
 
 void
--- a/synfig-studio/src/gui/widgets/widget_coloredit.cpp
+++ b/synfig-studio/src/gui/widgets/widget_coloredit.cpp
@@ -436,7 +436,7 @@ Widget_ColorEdit::Widget_ColorEdit():
 		Gtk::Box* box_cast_2 = static_cast<Gtk::Box*>( internal_child_2[0] );
 		std::vector<Widget*> internal_child_3 = box_cast_2->get_children(); //internal_child_3[0] is the color wheel widget
 		internal_child_3[0]->signal_button_release_event().connect([&](GdkEventButton *ev){ was_released=true;on_color_changed();return false;},false);
-		internal_child_3[0]->signal_key_release_event().connect([&](GdkEventButton *ev){ was_released=true;on_color_changed();return false;},false);
+		internal_child_3[0]->signal_key_release_event().connect([&](GdkEventKey *ev){ was_released=true;on_color_changed();return false;},false);
 
 		hvsColorWidget->signal_color_changed().connect(sigc::mem_fun(*this, &studio::Widget_ColorEdit::on_color_changed));
 		//TODO: Anybody knows how to set min size for this widget? I've tried use set_size_request(..). But it doesn't works.
--- a/synfig-studio/src/gui/widgets/widget_coloredit.cpp
+++ b/synfig-studio/src/gui/widgets/widget_coloredit.cpp
@@ -436,6 +436,7 @@ Widget_ColorEdit::Widget_ColorEdit():
 		Gtk::Box* box_cast_2 = static_cast<Gtk::Box*>( internal_child_2[0] );
 		std::vector<Widget*> internal_child_3 = box_cast_2->get_children(); //internal_child_3[0] is the color wheel widget
 		internal_child_3[0]->signal_button_release_event().connect([&](GdkEventButton *ev){ was_released=true;on_color_changed();return false;},false);
+		internal_child_3[0]->signal_key_release_event().connect([&](GdkEventButton *ev){ was_released=true;on_color_changed();return false;},false);
 
 		hvsColorWidget->signal_color_changed().connect(sigc::mem_fun(*this, &studio::Widget_ColorEdit::on_color_changed));
 		//TODO: Anybody knows how to set min size for this widget? I've tried use set_size_request(..). But it doesn't works.
@@ -506,7 +507,7 @@ Widget_ColorEdit::on_color_changed()
 	{
 		Gdk::RGBA newColor = hvsColorWidget->get_current_rgba();
 		Color synfigColor;
-		if(was_released){// if there was a button release record this final color in history panel
+		if(was_released){// if there was a button/key release record this final color in history panel
 		HistoryTreeStore::block_new_history=false;
 		Color synfigColorTemp(
 				newColor.get_red()+0.00001,//slight increase doesnt affect colors value but enough to trigger new action signal
--- a/synfig-studio/src/gui/widgets/widget_coloredit.cpp
+++ b/synfig-studio/src/gui/widgets/widget_coloredit.cpp
@@ -47,6 +47,7 @@
 #include <gui/exception_guard.h>
 #include <gui/localization.h>
 
+#include <gui/trees/historytreestore.h>
 
 #endif
 
@@ -331,7 +332,7 @@ Widget_ColorEdit::SliderRow(int left, int top, ColorSlider* color_widget, std::s
 	grid->attach(*label,        left,   top, 1, 1);
 	grid->attach(*color_widget, left+1, top, 1, 1);
 }
-
+bool was_released=false;
 void
 Widget_ColorEdit::AttachSpinButton(int left, int top, Gtk::SpinButton *spin_button, Gtk::Grid *grid)
 {
@@ -427,6 +428,15 @@ Widget_ColorEdit::Widget_ColorEdit():
 		//I use Gtk::ColorSelection widget here.
 		hvsColorWidget = manage(new Gtk::ColorSelection());
 		setHVSColor(get_value());
+
+		//navigating to inner implementation color wheel widget
+		std::vector<Widget*> internal_child = hvsColorWidget->get_children();
+		Gtk::Box* box_cast = static_cast<Gtk::Box*>( internal_child[0] );
+		std::vector<Widget*> internal_child_2 = box_cast->get_children();
+		Gtk::Box* box_cast_2 = static_cast<Gtk::Box*>( internal_child_2[0] );
+		std::vector<Widget*> internal_child_3 = box_cast_2->get_children(); //internal_child_3[0] is the color wheel widget
+		internal_child_3[0]->signal_button_release_event().connect([&](GdkEventButton *ev){ was_released=true;on_color_changed();return false;},false);
+
 		hvsColorWidget->signal_color_changed().connect(sigc::mem_fun(*this, &studio::Widget_ColorEdit::on_color_changed));
 		//TODO: Anybody knows how to set min size for this widget? I've tried use set_size_request(..). But it doesn't works.
 		hvs_grid->attach(*(hvsColorWidget), 0, 4, 1, 1);
@@ -495,15 +505,27 @@ Widget_ColorEdit::on_color_changed()
 	if (!colorHVSChanged)
 	{
 		Gdk::RGBA newColor = hvsColorWidget->get_current_rgba();
-		Color synfigColor(
+		Color synfigColor;
+		if(was_released){// if there was a button release record this final color in history panel
+		HistoryTreeStore::block_new_history=false;
+		Color synfigColorTemp(
+				newColor.get_red()+0.00001,//slight increase doesnt affect colors value but enough to trigger new action signal
+				newColor.get_green(),
+				newColor.get_blue() );
+		synfigColor=synfigColorTemp;}
+		else{
+		HistoryTreeStore::block_new_history=true;
+		Color synfigColorTemp(
 			newColor.get_red(),
 			newColor.get_green(),
 			newColor.get_blue() );
+		synfigColor=synfigColorTemp;}
 		synfigColor = App::get_selected_canvas_gamma().apply(synfigColor);
 		set_value(synfigColor);
 		colorHVSChanged = true; //I reset the flag in setHVSColor(..)
 		on_value_changed();
 	}
+	was_released=false;
 }
 
 void
--- a/synfig-studio/src/gui/trees/historytreestore.h
+++ b/synfig-studio/src/gui/trees/historytreestore.h
@@ -87,6 +87,8 @@ public:
 
 	const Model model;
 
+	static bool block_new_history;
+
 	/*
  -- ** -- P R I V A T E   D A T A ---------------------------------------------
 	*/
--- a/synfig-studio/src/gui/trees/layertree.cpp
+++ b/synfig-studio/src/gui/trees/layertree.cpp
@@ -53,6 +53,10 @@
 
 #include <synfigapp/actions/layerremove.h>
 #include <synfigapp/instance.h>
+#include <gui/trees/historytreestore.h>
+
+#include <thread>
+#include <chrono>
 
 #endif
 
@@ -125,6 +129,7 @@ LayerTree::LayerTree()
 {
 	layer_tree_view().signal_key_press_event().connect(sigc::mem_fun(*this, &LayerTree::on_key_press_event));
 
+
 	create_layer_tree();
 	create_param_tree();
 
@@ -137,9 +142,18 @@ LayerTree::LayerTree()
 	//param_tree_view().get_selection()->set_mode(Gtk::SELECTION_MULTIPLE);
 	layer_tree_view().show();
 	param_tree_view().show();
-
+	//gets the selection object at first
+	selected_param_object = param_tree_view().get_selection(); //hbd
+	// getting the selected row
+	iter = selected_param_object->get_selected();
+	row = *iter;
 	param_tree_view().set_has_tooltip();
 	layer_tree_view().set_has_tooltip();
+
+	param_tree_view().signal_motion_notify_event().connect(sigc::mem_fun(*this, &LayerTree::on_motion_notify_event), false);
+	param_tree_view().signal_button_press_event().connect(sigc::mem_fun(*this, &LayerTree::on_button_press_or_release_event), false);
+	param_tree_view().signal_button_release_event().connect(sigc::mem_fun(*this, &LayerTree::on_button_press_or_release_event), false);
+	selected_param_object->signal_changed().connect(sigc::mem_fun(*this, &LayerTree::change_selection_param), false);
 
 	disable_single_click_for_param_editing = false;
 }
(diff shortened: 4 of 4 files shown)


mohamedAdhamc:
> history-broken.mp4 
> Now history is completely broken. The first change is not recorded in history. After adding a change, pressing `Ctrl+Z` does not undo the change.

Oh, maybe I forgot something in the commit. It was working fine on my machine. I will look into it.

mohamedAdhamc:
@ice0 
it seems weird that it's not working on your machine, I checked, and it's working alright here. I will look into it also.

https://user-images.githubusercontent.com/100296264/179552524-462046aa-20f7-41e2-9aea-e6f0037e6ef5.mp4

p.s. its a little glitichy because of the screen recorder. the windows screen recorder didnt record the new color dialog window.

mohamedAdhamc:
hey, @rodolforg do you think grouping these actions into one group be a good way of fixing this history pollution issue?

ice0:
> hey, @rodolforg do you think grouping these actions into one group is a good way of fixing this history pollution issue?

I don't like the grouping solution because it doesn't solve the problem, it just hides it.
Every group in history can be opened, so this brings us back to the original issue.

Screenshot_19

mohamedAdhamc:
> 
> I don't like the grouping solution because it doesn't solve the problem, it just hides it. Every group in history can be opened, so this brings us back to the original issue.

yes, this is actually very understandable. I will be looking for something better tommorow I'm about to go to sleep now.

In the meanwhile, it would be great if you guys could give me suggestions on what I should be trying to do/ where i should be looking?

mohamedAdhamc:
update: 
ok so finally its working, on drag ending a new action is made, and undoing successfully undoes the color. I will clean up the code and make sure there arent any corner cases im not handling then commit hopefully.

mohamedAdhamc:
ok so a description of what happened.

color history polution should now be fixed for the color wheel. As in dragging the wheel makes no pollution.

However there are two cases I havent handled the pollution for
1- dragging the color wheel using the keyboard
2- color pollution when using the spin buttons in the dialog (i mean clicking on the +/- buttons of the spin button)

I don't think one is necessary to me it felt really unnatural using the keys to adjust the color wheel. If it has to be handled though please let me know.

Two I think wouldn't be related to this PR as these spin buttons arent part of the color wheel. So maybe I can fix them in another PR later if they should be fixed. But I'd rather continue with the issues I have for my project so as to not get carried away too much now during the project.

p.s. i removed some unecessary thing and waiting or it to rebuild to make sure I didnt accidently mess up anything so I wil be comiiting in a couple minutes

synfig-studio/src/gui/trees/historytreestore.cpp
```diff
@@ -209,6 +212,7 @@ HistoryTreeStore::on_redo_stack_cleared()
 void
 HistoryTreeStore::on_new_action(etl::handle<synfigapp::Action::Undoable> action)
 {
+ if (!synfigapp::Action::System::block_new_history){ /*if change isn't due to rapid repeated actions show history normally*/
```
ice0, line 215: Is this part is still needed?

mohamedAdhamc: seems as if it is redundant after the new implementation (because the new action isn't made when history is blocked so this function shouldn't be called when history is blocked), I also tested it to make sure it is.

synfig-studio/src/gui/trees/historytreestore.cpp
```diff
@@ -41,6 +41,8 @@
 #include <synfig/general.h>
 #include <synfig/valuenode.h>
 
+#include <synfigapp/action_system.h>
```
ice0, line 44: It looks like this header is not needed anymore too.

mohamedAdhamc: oh yes definetly, because we are not using the flag anymore here.

ice0:
Pressing `Esc` key doesn't cancel the change:

https://user-images.githubusercontent.com/5604544/182484092-d4309d30-0a29-4f08-b866-eaa90bbbbc23.mp4

mohamedAdhamc:
> Pressing `Esc` key doesn't cancel the change:

Yes actually I didn't handle ESC for cancelling change will do it in a bit (just woke up) shouldn't take long.

mohamedAdhamc:
ok @ice0,  now escape should be working for cancelling change as well.

ice0:
https://user-images.githubusercontent.com/5604544/182746605-19ebe5d0-323b-45c9-b4ee-411de2c6cf3a.mp4

I am added a debug output to `CellRenderer_ValueBase::color_edited`:
```
CellRenderer_ValueBase::color_edited(synfig::Color color, Glib::ustring path)
{
	std::cout << "CellRenderer_ValueBase::color_edited: " << color.get_string() << " block: " << synfigapp::Action::System::block_new_history;
	ValueBase old_value(property_value_.get_value());
	ValueBase value(color);
	if (old_value != value) {
		std::cout << " signal_edited_";
		signal_edited_(path, value);
	}
	std::cout << std::endl;
}
```

and get the next output:
```
CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 1 signal_edited_
CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 0 signal_edited_
CellRenderer_ValueBase::color_edited: #ffffff :  1.000 block: 1 signal_edited_
CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 0 signal_edited_
CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 0 signal_edited_
```

Can you describe what is happening here?
If I understand correctly, there should be only two changes here:
```
CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 1 signal_edited_
CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 0 signal_edited_
```
the first one is for the preview and the second one is for the final change.

mohamedAdhamc:
> ```
> CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 1 signal_edited_
> CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 0 signal_edited_
> CellRenderer_ValueBase::color_edited: #ffffff :  1.000 block: 1 signal_edited_
> CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 0 signal_edited_
> CellRenderer_ValueBase::color_edited: #e62929 :  1.000 block: 0 signal_edited_
> ```

regarding the third line: 
well so basically what we wanted to do here is to prevent the creation of a new action until we were sure that selection was over. so that led to having to accomdate all the changes in on (previous) action. so now all the changes while dragging end up affecting the color parameter of the (previous) action, and so if left as is undo would only get us to the most recent incremental change before the final color, not the previous color of the layer before the drag. So what I did was when I know the drag is over I quickly reset the layers color to the initial color while history is blocked (so the current action has the correct parameter). then after that  new action is made with the desired selected color and history isnt blocked so it shows.

Also quick question, apart from the output it was working correctly wasnt it ?

p.s. other than that there are some duplicate lines in the debug output I'm looking into them to try and see what exactly happend.

ice0:
> regarding the third line: well so basically what we wanted to do here is to prevent the creation of a new action until we were sure that selection was over. so that led to having to accomdate all the changes in one (previous) action. so now all the changes while dragging end up affecting the color parameter of the (previous) action, and so if left as is undo would only get us to the most recent incremental change before the final color, not the previous color of the layer before the drag. So what I did was when I know the drag is over I quickly reset the layers color to the initial color while history is blocked (so the current action has the correct parameter). then after that new action is made with the desired selected color and history isnt blocked so it shows.

I think how to make this behavior more universal. So that the user does not need to think about it.
An example pseudo-code, as I imagine it:

```
synfigapp::Action::System::enable_preview();

void
Widget_ColorEdit::on_hsv_color_changed()
{
	set_color(hsvColorWidget->get_current_rgba());


	// if user stopped adjusting color
	if (!hsvColorWidget->is_adjusting()) {
		synfigapp::Action::System::commit_action();
	}
}

Widget_ColorEdit::on_key_press(int key) {
	if (key == Escape) {
		synfigapp::Action::System::cancel_action();
	}
}
```
 
> Also quick question, apart from the output it was working correctly wasnt it ?

Yes. But I still need to test linked (exported) values.

mohamedAdhamc:
> 
> I think how to make this behavior more universal. So that the user does not need to think about it. An example pseudo-code, as I imagine it:
> 
> ```
> synfigapp::Action::System::enable_preview();
> 
> void
> Widget_ColorEdit::on_hsv_color_changed()
> {
> 	set_color(hsvColorWidget->get_current_rgba());
> 
> 
> 	// if user stopped adjusting color
> 	if (!hsvColorWidget->is_adjusting()) {
> 		synfigapp::Action::System::commit_action();
> 	}
> }
> 
> Widget_ColorEdit::on_key_press(int key) {
> 	if (key == Escape) {
> 		synfigapp::Action::System::cancel_action();
> 	}
> }
> ```

hmm, I kind of see what you mean here but as another view. why I chose the flag approach is that this way the action system already in place doesn't need to be changed. And also user/devoloper woring on regular action doesnt need to worry about the flag at all, only if the devoloper is trying to block some histoy pollution as here then he would use the flag in a similar manner to block making new actions and then settting the initial value before finally setting the needed value and having it show. I did this same methd in _changing parameters with mouse_ and it was working as it is here.

> Yes. But I still need to test linked (exported) values.

alright, sounds good

mohamedAdhamc:
> I'm thinking about adding a new flag/field in `synfigapp::Action`: if it overwrites the values changed by previous (same type of) action or not.
> 

I think this idea sounds good, it would remove the need of me "quickly resetting the parameters intial value" which I did here to have the undo work correctly.

rodolforg:
> However, I'm not sure maybe I'm not completely getting it, but if the intermediate actions don't overwrite the layers parameter then how would they show ?

I meant instead of pushing action in undo-history stack, those overwriting actions (like the main one here, "ValueDescSet") replace the immediately previous one, if the latter is related to the same `ValueDesc` (i.e., changing the exact same parameter).

ice0:
History is broken when multiple objects is selected:

https://user-images.githubusercontent.com/5604544/184590066-30607d7c-f729-4995-a099-91012c46a114.mp4

After undo change not all objects will restore they original colors.

mohamedAdhamc:
> I meant instead of pushing action in undo-history stack, those overwriting actions (like the main one here, "ValueDescSet") replace the immediately previous one, if the latter is related to the same `ValueDesc` (i.e., changing the exact same parameter).

Hmm, so what you mean is... when we had history pollution all action were showing from action 1 to say 15, now what you are saying is that actions show normally but instead they replace each other as in 2 replaces 1, 3 replaces 2, and so on. I think this sounds like a good idea although I dont really know how it would be implemented/"replace" each other xd, maybe I need to go over all the action related classes at once to try to understand them better.

mohamedAdhamc:
> History is broken when multiple objects is selected (after undoing a change, not all objects restore their original colors):
>

this is due to the current way of resetting the original color before setting the final one, in which case here the original is "the original" for only one of the selected. seems like the current fix isnt general enough yet, I will try to go along with what rodolf suggested.

mohamedAdhamc:
Hey guys, unfortunately things have been a bit slow past few days as my grandfather suffered a stroke and we moved him into the hospital and now back home, hopefully starting tomorrow things should go back to normal and I should be back to pace. 

As of now there are 3 features left and everything else is in various review stage. I will be trying to get the remaining features done as quick as possible so I can have all focus on reviewing and even self enhancing certain things to make the review easier for you guys.

ice0:
> this is due to the current way of resetting the original color before setting the final one, in which case here the original is "the original" for only one of the selected. seems like the current fix isnt general enough yet, I will try to go along with what rodolf suggested.

Is any progress here?

mohamedAdhamc:
> 
> Is any progress here?

Not really I was mainly caught up with the other features and reviews. I think to actually get to a fix that would be general enough for everything i'd have to really understand in detail the action system in place. I should get to it soon though (starting tomorrow if things go well with the things I'm supposed to address from reviews).

ice0:
> Not really I was mainly caught up with the other features and reviews.

I suggest to concentrate here, because without completing this PR we cannot continue with #2717

mohamedAdhamc:
> 
> I suggest to concentrate here, because without completing this PR we cannot continue with #2717

ok I will be focusing on this as yes this is critical for #2717.

mohamedAdhamc:
> 
> I meant instead of pushing action in undo-history stack, those overwriting actions (like the main one here, "ValueDescSet") replace the immediately previous one, if the latter is related to the same `ValueDesc` (i.e., changing the exact same parameter).
> 

ok so I have been trying to do this but there is something I dont really understand and is causing lots of problems. 

I tried doing what you suggested which is for these actions to have the new one replace the old one in the undo_action_stack so theoretically we should end up with the very last action in the undo stack and then whichever is before it so for example:

making a new circle layer (let say default color is black)
setting the color to blue using color wheel (now the last action in undo stack should be this one)

here when I undo it only takes me back to the previous action but with the wrong color (slightly lighter blue) not black,

the only time that seems to not happen is when the very first action in the color wheel is pushed onto the undo stack and not removed. So right now I think maybe I could make a passive group with the very first and last action/actions (num_layers * 2)
but even for that I'm not really sure how i would even put them in a passive group with the way they are sent and processed.

any ideas?

mohamedAdhamc:
ok so I went with rodolf's second suggestion which is using a passive group but modified ( because a passive group directly would just be hiding all the many actions not more). Now the passive group only includes the very first and last actions which are necessary for the undo process to work, and importantly it works for an selected number of layers.

mohamedAdhamc:
ok its not yet fully complete I will finish it first thing in the morning tommorow.
whats left to do 
1- ok so now we reduced the number of actions into only two actions, very first and very last (for one layer, for n layers its n*2).
    now in the passive group containing the action, it shows these two actions although its not really important for the user to  
    see both we only need them in the group stack for the undo/redo, so maybe only show the user one action and not have the 
    other one printed out in the group, so for example if its only one layer being edited he sees only one action.
2- I have to adjust the "escape to cancel" to fit in with the new implementation.
3- quite a bit of clean-ups, lots of commented-out code from when I was experimenting diffrent ideas :).

mohamedAdhamc:
Now the three points I mentioned should be done (there may be more cleaning needed though I will have to review the changes).

Quick brief of what is reached so far:
- the large number of actions (history pollution) is reduced to only two actions per layer (initial+final) which are grouped in a passive group so as to become one action in the undo/redo stacks. And also the end-level user only sees the color change as one action (no need to show him an initial and a final action).
- escape to cancel works.
- also most importantly all that's mentioned works for any number of layers :)

synfig-studio/src/synfigapp/action_system.h
```diff
@@ -105,6 +106,8 @@ class System : public etl::shared_object, public sigc::trackable
 	*/
 
 public:
+		static bool repeated_action;
```
ice0, line 109: It's better to move this to private non-static variables.
Each file has its own ActionSystem. If we try to work with them in parallel (from different threads for example), in the case of static members, this will lead to undefined behavior.

synfig-studio/src/gui/widgets/widget_coloredit.cpp
```diff
@@ -427,6 +429,23 @@ Widget_ColorEdit::Widget_ColorEdit():
...
 		setHVSColor(get_value());
+
+		//navigating to inner implementation color wheel widget
+		std::vector<Widget*> internal_child = hvsColorWidget->get_children();
+		Gtk::Box* box_cast = static_cast<Gtk::Box*>( internal_child[0] );
+		std::vector<Widget*> internal_child_2 = box_cast->get_children();
+		Gtk::Box* box_cast_2 = static_cast<Gtk::Box*>( internal_child_2[0] );
+		std::vector<Widget*> internal_child_3 = box_cast_2->get_children(); //internal_child_3[0] is the color wheel widget
+
+		internal_child_3[0]->signal_button_press_event().connect([&](GdkEventButton *ev){ wheel_pressed=true; escape_cancel=false; return false;},false);
```
ice0, line 440: Try to remove `wheel_pressed` variable - you already know that wheel is pressed if `hvsColorWidget->is_adjusting() == true`

synfig-studio/src/synfigapp/action_system.h
mohamedAdhamc: yes, I was initially thinking that because it probably isn't possible to have more than one of these "repeated actions" then a static variable would be enough to be used by each "repeated action". But yes I can see how this could get messy and possibly lead to some undefined behavior.

switched over to _public_ non-static variables. As they need to be set and gotten from outside, however if its better to just keep them private and write setters and getters please let me know :)

synfig-studio/src/gui/widgets/widget_coloredit.cpp
mohamedAdhamc: done

synfig-studio/src/gui/widgets/widget_coloredit.cpp
```diff
@@ -427,6 +427,23 @@ Widget_ColorEdit::Widget_ColorEdit():
...
 		setHVSColor(get_value());
+
+		//navigating to inner implementation color wheel widget
+		std::vector<Widget*> internal_child = hvsColorWidget->get_children();
+		Gtk::Box* box_cast = static_cast<Gtk::Box*>( internal_child[0] );
+		std::vector<Widget*> internal_child_2 = box_cast->get_children();
+		Gtk::Box* box_cast_2 = static_cast<Gtk::Box*>( internal_child_2[0] );
+		std::vector<Widget*> internal_child_3 = box_cast_2->get_children(); //internal_child_3[0] is the color wheel widget
+
+		internal_child_3[0]->signal_button_press_event().connect([&](GdkEventButton *ev){ colorHVSChanged = false;escape_cancel=false; return false;},false);
```
ice0, line 438: Ok, now it should be possible to remove this part too:
```
internal_child_3[0]->signal_button_press_event().connect([&](GdkEventButton *ev){ colorHVSChanged = false;escape_cancel=false; return false;},false);
internal_child_3[0]->signal_button_release_event().connect([&](GdkEventButton *ev){ if((hvsColorWidget->is_adjusting()))wheel_released=true;return false;},false);
```

When you press the mouse down - it emits `signal_color_changed()` event with `is_adjusting() = true` flag.
When you release the mouse button - it emits `signal_color_changed()` event with `is_adjusting() = false` flag.

mohamedAdhamc: removed the button press event, but the release event is needed to diffrentiate between color change from wheel vs color change from elsewhere, as also color change not from color wheel emits `signal_color_changed()` event with `is_adjusting() = false` flag.

ice0: > as also color change not from color wheel emits `signal_color_changed()` event with `is_adjusting() = false` flag.

Yes. In that case `setHVSColor()` -> `signal_color_changed()` is called. 
To distinguish manual color adjustment from the color wheel, the `colorHVSChanged` flag was previously used.
You can add your own flag and use it to test this condition. Next step will be merging this both flags (this will simplify the code).

mohamedAdhamc:
@ice0 done

ice0:
Just tested.
When a color is exported to the project library, two undo steps are required to undo the change.
Demonstration and test file below.

The first circle is a normal circle, the second circle has an exported color value. 
Notice the chain icon on the Parameters panel.

Screenshot_120


https://user-images.githubusercontent.com/5604544/193395745-0ae3c809-bbc0-4f9a-8909-4264fd711d97.mp4


[test-history.zip](https://github.com/synfig/synfig/files/9689695/test-history.zip)

mohamedAdhamc:
@ice0, I tried replicating but its not doing that.

https://user-images.githubusercontent.com/100296264/193407802-2a76a6d6-902f-4917-b0ef-7e65eaebefae.mp4

(here i set the color from the color wheel twice and it works fine, for some reason the windows screen recorder doesnt capture windows on top of windows xD but the point is shown i hope)

However, I noticed something similar earlier but I don't think it's related to exported values. Sometimes an extra action is shown in addition to the change color action group, as shown in the video you sent there was the regular group but one `set layer parameter` action wasn't grouped in it for some reason. 

it happens only fairly rarely though soI will keep on testing, and debugging to see why this happens sometimes.

mohamedAdhamc:
ok @ice0 I believe it should be fixed now :).

ice0:
I think there's room for improvement here, but it looks like the current implementation is leading to a dead end.

What would like to be improved:
1) Make it easier to work with preview from the point of view of the author of the new code
2) Remove PassiveGrouper

In order to make sure that this is possible, I have already made my own implementation.

This is how the changes in `widget_coloredit.cpp` looks like for user (https://github.com/synfig/synfig/commit/f255b220e9170d49c615acb1e681fef880ec0e25)

This is how the history looks like:

https://user-images.githubusercontent.com/5604544/194459118-f8c726c9-8061-4044-91ae-08ae507ca2d7.mp4

I think maybe to leave this PR as it is, and try to implement the same thing from the scratch?

mohamedAdhamc:
> What would like to be improved:
> 
> 1. Make it easier to work with preview from the point of view of the author of the new code
> 2. Remove PassiveGrouper

> I think maybe to leave this PR as it is, and try to implement the same thing from the scratch?

Fair enough, this implementation relies on passive-groupers abilities to do certain things like handling undo history, etc.. so this is in fact very close to its dead end. Will start thinking of a new way that could be better. 

I.e. have the repeated actions replace each other smoothly without segmentation errors :). which would be a bit more complex especially with multiple layers being processed at the same time, as some type of sorting might be needed. But if theres already a simpler core idea please let me know :)

mohamedAdhamc:
ok I think i reached something here, and now i'll begin implementing and make a new PR. So the thing here is that we cant just push the first action into the undo stack as that will only make and undo possible and not redo, same for pushing last action only. so we need to push into the undo stack a group of the first and last action, and then that should make undo/redo work fine and regarding the history tree store we just show the action (first or last) normally without being in an expandable passive group as in the previous implementation.

mohamedAdhamc:
ok I think i reached something here, and now i'll begin implementing and make a new PR. So the thing here is that we cant just push the first action into the undo stack as that will only make and undo possible and not redo, same for pushing last action only. so we need to push into the undo stack a group of the first and last action, and then that should make undo/redo work fine and regarding the history tree store we just show the action (first or last) normally without being in an expandable passive group as in the previous implementation.

mohamedAdhamc:
@ice0, here is what I reached
 
So in order to fix history pollution and at the same time not compromise undo/redo there has to be some sort of group to be stored in the undo action list which contains the very first and very last action. Now we could make a whole new grouper, and call it something like "repeated grouper", or just expand the existing functionality of the passive grouper. Since there are not much core differences between this "repeated grouper" and the "passive grouper" I chose to 
 extend the passive grouper here. 

 Now up until now this is very similar to the last implementation, but with this new commit I have hopefully addressed the issues you mentioned

**1- "Making it easier to work with preview from the point of view of the author of the new code ":**

 Now to be able to use this all one has to do is 2 steps:
- make a passive group (with the repeated_group parameter set to true in the constructor) before the repeated actions start.
- have it destruct when they end. The destruction could either be implicit, or explicit using new and delete if that is what suits the nature of the action more. 
And also this implementation makes it very easy to add escape/key to cancel feature. Just make the escape/key pressed signal  call App::get_selected_instance()->cancel_repeated_action() and destruct the group. 

( This could all also be further enhanced by using only a variable and then when this variable is set to true while it was false we create the passive group under the hood and when its false we destruct it also and escape to cancel is kept the same way by just doing what is mentioned above. But I think this could make things messier, and also the concept of using something like a passive grouper shouldn't be too foreign for someone working on Synfig code so he just uses it as if he is making a regular passive group but just supplies its constructor one extra argument so that under the hood history pollution is taken care of )

**2-  to not use a passive grouper:**
  I assumed what is meant here is to not have an extra grouper shown for the users as was the case with the "change color" grouper. So I removed it and now what is shown is only the action done. So basically the appearance for the 

users is the same as before this PR just without the "pollution"

 Here is a video of how it is working now: 

(at the end I was showing escape to cancel functionality. Also the software I use for recordidng make everything very laggy xd)
https://user-images.githubusercontent.com/100296264/195347809-c2f04092-940c-4bdf-a267-53c6daf39850.mp4

I hope to have fully addresed the things mentioned :) if there is something that I have missed though please let me know.

p.s. I originally was  going to make a new pr as suggested but then due to things stated I found it makes sense to continue here. Anyways we can always revert this last commit and head over to a new PR if its needed :)

PR #2751, opened 2022-07-16
fix: synfig studio was not saving synfigapp settings on quit

rodolforg:
Steps to reproduce the problem:

1. Select Bline Tool
2. Change a parameter in Tool Options Panel (e.g. Opacity or Line Width)
3. Restart Synfig Studio
4. Select Bline Tool and see the change was not saved.

Reason:
Since #2497, we finally use Gtk::Application as base class for `App`.
`App` is destructed before `MainWindow` is closed (basically at start).
So, stuff done on quitting before merging #2497 was not done on the right time now.

As `signal-shutdown` is only called in primary instance of Gtk::Application,
it's the proper moment to delete and finish stuff on application exit.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -1396,6 +1396,7 @@ App::App()
 {
 	add_main_option_entry(Gio::Application::OPTION_TYPE_BOOL, "console", 'c', N_("Opens a console that shows some Synfig Studio output"));
 	signal_handle_local_options().connect(sigc::mem_fun(*this, &App::on_handle_local_options));
+	signal_shutdown().connect(sigc::mem_fun(*this, &App::on_shutdown));
 }
 
 int App::on_handle_local_options(const Glib::RefPtr<Glib::VariantDict> &options)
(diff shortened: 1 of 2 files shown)


rodolforg:
Select Bline Tool and change any parameter in its Tool Options Panel (like Opacity or Line Width).
Restart Synfig Studio.
Select Bline Tool and see the change was not saved.

ice0 (approved):

synfig-studio/src/gui/app.cpp
```diff
@@ -1823,7 +1825,7 @@ App::~App()
 
 	main_window->hide();
 
-	delete main_window;
+	// delete main_window;
```
ice0, line 1828: Looks like we'll need to refactor this part in the future.

ice0: If we move this part to the end of the method, then it works:
```
	process_all_events();
	delete main_window;
```

rodolforg: Great!
Done!

ice0 (approved):

Issue #1458, opened 2020-05-31
Minimalistic theme


Agnyy:
I propose to make a minimalistic theme. Like Adobe. 
I can design it and draw icons, but I need someone to collect the theme from this. I started with a toolbar.  Background color #323232
![image](https://user-images.githubusercontent.com/25137502/83355909-0b9eb500-a36b-11ea-9417-a9791b5995c6.png)  
![image](https://user-images.githubusercontent.com/25137502/83355977-9089ce80-a36b-11ea-99d4-c48326c29bf3.png)

Here's what an instrument might look like when it is on and off.

![image](https://user-images.githubusercontent.com/25137502/83356081-46551d00-a36c-11ea-9d16-093c0da539bf.png)

Agnyy:
![image](https://user-images.githubusercontent.com/25137502/83358426-1dd51f00-a37c-11ea-9433-703da39217a2.png)

Agnyy:
Here are all the tools in this file and they are signed. SVG format 
[Synfig.zip](https://github.com/synfig/synfig/files/4708081/Synfig.zip)

rodolforg:
Nice :D (But I like colored icons too, but they look like outdated as they are now.)

Could you test them with light-background/mode too? (like @libreartist did here https://github.com/synfig/synfig/issues/1041#issuecomment-579783016, but he never showed up anymore)

@Agnyy could you please provide icons for the "actions" (states in Layer panel) too?

-     layer visibility (visible / hidden)
-     layer rendering (enabled / disabled)
-     layer selection locking (locked / unlocked)


@morevnaproject @ice0

FirasH:
Very interesting proposal... I like it!

Agnyy:
@rodolforg Especially for you. Icons on a white background.
![image](https://user-images.githubusercontent.com/25137502/83360431-f1280400-a389-11ea-9c00-39b489dc88e2.png)


[White version.zip](https://github.com/synfig/synfig/files/4708222/White.version.zip)

Agnyy:
There is another idea. Paint the tools in different colors thereby slaving them by type.
![image](https://user-images.githubusercontent.com/25137502/83360832-db680e00-a38c-11ea-854f-e9f7eeb443ae.png)

This has been done in some programs before. This helped to focus on colors.

![image](https://user-images.githubusercontent.com/25137502/83360855-26822100-a38d-11ea-83fd-6027be231448.png)

rodolforg:
[...]

PR #2756, opened 2022-07-19
feat!: allow to lock a layer for click/selection

### LINKED ISSUE #301: Ability to lock layers
Forgive me if this has already been implemented, but I'm unable to find anything (aside from this [old thread](http://synfig.org/forums/viewtopic.php?t=1734)) on the ability to lock layers, so that they cannot be accidentally be selected via the canvas.

This feature is part of any layered graphics program I have ever used, so I'm having a hard time believe this to be an oversight for years. Does Synfig use a different approach/workflow?

### LINKED ISSUE #2629: Lock ability for all layers
Nothing major, just not being able to lock layers that aren't groups so I would accidentally select and/or move something else that is not what I wanted.
To be able to lock all layer types
That it is more obvious that synfig has a feature to lock layers maybe another check box in the layers panel.
**Additional comments**
Add any other comments or screenshots about the feature request here.
If you are going to make it more obvious that synfig has a lock feature by having another check box it would be a great idea to have a  way to tell the difference on what they do maybe 2 more icons.

rodolforg:
Fix #301
Fix #2629

BREAKING CHANGE: synfig and synfigapp API changed:
- add synfig::Layer::set_hitlocked() and synfig::Layer::is_hitlocked()
- add synfigapp actions:
  LayerSetHitLocked, LayerSetHitLockedOn and LayerSetHitLockedOff
- add synfigapp CanvasInterface::signal_layer_hit_locked_changed

Now working too: if a layer inside a group is locked, it succeeds to look group to check below it


rodolforg:
Captura de tela de 2022-07-19 00-04-24

Ugly icons, but as soon as I finish ETL cleaning, I'll study how to swap the icons for those in #1458.

Gilraiser:
Hello, best regards
To save space in the layers dialog, it could be implemented using the menu and colors to mark the lines of the layers that are locked. Here are some images of what it might look like. I hope you like the idea. I think it would contribute to a cleaner interface.
imagen
imagen
imagen

rodolforg:
IMHO I don't think it's good to highlight the locked layers. It isn't intuitive.
Many software use icons to let it clear to user nor distract him/her.

rodolforg:
Good sugestion about unlock all layers.
Do you think it makes sense to create action "lock all"?

Gilraiser:
Hello 
rodolforg I think it makes sense to block them all at the moment that the user thinks he has finished his work inside synfig and wants to start rendering. Or if he wants to start creating new ones, but all the ones he created previously, he knows that he is not going to make any more changes to them. In my opinion, I think the 'Lock all' option is useful.

Gilraiser:
As for my idea of ​​colors and not icons, I came up with a cleaner interface where there would be no need to add one more column to the capable docker. That the option was in the menu and not taking up space unnecessarily. I think there are already many icons on the screen and the menus solve this very well.

Gilraiser:
It wouldn't have to be one color red, it could be another. It could also be a typography change. It is a humble opinion that he left on the table. I thank you very much for your attention and your interest.

rodolforg:
> Or if he wants to start creating new ones, but all the ones he created previously, he knows that he is not going to make any more changes to them.

Nice use case. It indeed is useful.

> I think there are already many icons on the screen and the menus solve this very well.

I think Z depth takes too much space. Is it really need to show or a it would be better be in a tip that shows up with mouse hover?
Besides, I think the menu is already too cluttered...

> It could also be a typography change.
Group and Switch Group layers use this trick already. The only intuitive case to me is the bold layer name for currently selected layer in Switch group.

Gilraiser:
It occurs to me to save space in the layer docker that the Z Depth shows only integers, and when you hover the mouse over it, a tooltip is displayed that shows the number with its decimals.

Another alternative would be to click on the Zdepth column name to toggle between decimals and integers, or just display a decimal.

With all good intentions I insist that doing something, like dimming the text in the row that represents the layer so that it looks disabled when it is locked would be a good way.

I also think that placing the lock, unlock, lock all, and unlock all option in the menu is still a quick way to lock layered editing. In terms of speed, keyboard commands could help. It is my humble opinion.

rodolforg:
> With all good intentions I insist that doing something, like dimming the text in the row that represents the layer so that it looks disabled when it is locked would be a good way.

It could mean layer is disabled (inactive), its rendering is disabled, ... I think it is confusing for user:
- what does layer name in italics mean? disabled rendering on export only
- what does a greyed layer name mean? disabled clicking
- what does the box before layer name mean? disabled rendering on both screen and export

rodolforg:
Here how it is in Blender:
image

Gilraiser:
Colors could be designated. One color to disable render and a different color for the locked layer.

rodolforg:
> Colors could be designated. One color to disable render and a different color for the locked layer.

And another when disabled to render and locked simultaneously?

nahCyzar0520:
I like the progress being made on this. But now there is a question about which indication for locking layers should be done the color indication (red) or a lock icon. I like both ideas but I can't decide.

Gilraiser:
> > Colors could be designated. One color to disable render and a different color for the locked layer.
> 
> And another when disabled to render and locked simultaneously?

The same solution occurs to me with colors. Half of the row is the lock color and the other half is the color of the disabled render. I think it could work.

Gilraiser:
I think it's a good idea. What do you think?
locklayersynfig

rodolforg:
Test file: 
[hit-lock-test.zip](https://github.com/user-attachments/files/31615392/hit-lock-test.zip)

Change what layer(s) is(are) locked to test it.

rodolforg:
@Gilraiser please test it :)

Somebody please make proper icons for this feature! o.O

Svarov-RZM:
@rodolforg Hey, I tried to compile this branch and it failed. Check the log out:

BuildLog

I was building as usual, using 2-build-cmake.sh in my MSYS2 environment. 
It seems that it does not like that GUID class is referenced in ambiguous way, it expects it to be synfig::GUID (I changed the problematic file and it builds). I understand that it's not related to this particular commit but to brush work, but still dropping it here as a note that should be fixed.

I'll get back when I finish testing this PR...

Svarov-RZM:
OK, I tested this functionality and it works well. Very useful feature, thanks @rodolforg!
Now, about icons, they are fine. Could be better, yes, but it's something that can always be addressed later.

Give me some time for code review, there's a lot of it and I am not good with C++.

P.S. Why there's an exclamation mark in subject title? It has "feat!:". not just "feat:" as it would usually be. Is there some secret meaning to this or is it just a misspell?

rodolforg:
> Now, about icons, they are fine. Could be better, yes, but it's something that can always be addressed later.

They aren't fine lol I only got them from other existing icon from Synfig, but they are not consistent.
I wonder if I should do the same strategy (icons) for flagging renderable status (instead of italic).



> P.S. Why there's an exclamation mark in subject title? It has "feat!:". not just "feat:" as it would usually be. Is there some secret meaning to this or is it just a misspell?

Because of a breaking change, but it is irrelevant... we do many and don't signalize it XD
https://www.conventionalcommits.org/en/v1.0.0/

rodolforg:
Rebased and reduced the icon size.
Before:
image
After:
Captura de tela de 2026-09-06 22-21-59

Svarov-RZM:
Found a bug - icon for layer lock doesn't change upon undo/redo action. So when you hit CTRL+Z the lock icon stays the same until you hover mouse cursor over it which is quite unintuitive.

rodolforg:
> Found a bug - icon for layer lock doesn't change upon undo/redo action. So when you hit CTRL+Z the lock icon stays the same until you hover mouse cursor over it which is quite unintuitive.

Fixed! :)

Svarov-RZM:
Alright! I gave it one more test and found not a bug per se, but more like a strange behavior that's pretty inconvenient in my eyes. You see, if there's a shape that is locked and on top of it is another group layer that's unlocked but its blend method set to "Onto" it will also get locked so you won't be able to pick layers from it.

This behavior is unwanted because let's say you made a shape of some object and satisfied with it. Now you want to lock it to avoid unintentional modification and start working on textures but with current implementation this workflow is not possible.

How to reproduce:
* Open attached file:
[BirchLeaf.zip](https://github.com/user-attachments/files/32334874/BirchLeaf.zip)

* Lock the region layer of a leaf
* Try to select any of the sprouts - not possible

rodolforg:
@Svarov-RZM I think it is fixed.

I found another bug with your example: `ctrl` + mouse scroll give weird zooms XD

Svarov-RZM:
> I think it is fixed.

Awesome! Will test everything one more time on weekends and report back.

> I found another bug with your example: `ctrl` + mouse scroll give weird zooms XD

Ohhh, that's not a bug, it's a behavior that was there from the start I believe. What's happening here is that I changed the origin of the canvas to center-bottom:
CanvasProperties
You never see examples like that because I am the only one who uses this functionality. It's useful when designing objects that start from the very bottom (e.g. a flower) to maximize available canvas space, plus you don't have to fiddle the group's origin (purple dot) whenever you import the object into parent canvas. So yeah, zoom is jerky because it's calculated with 'center-bottom' offset, not the usual 'center-center'.

PR #2762, opened 2022-07-21
fix: fixed a bug when saving a file to a Windows drive root path and then reusing the file dialog

hesham-essam:
`is_absolute_path` returns false with directories like `E:`. The following test will fail for example:
```
std::string file = "E:/file.sifz";
std::string file_dir = dirname(file_dir) // file_dir = E:
assert(is_absolute_path(file_dir) == true);
```
I encountered this on Windows where I save a sifz file as `E:/Synfig Animation 1.sifz`, which will cause synfig to record the most recent directory to `E:` in `settings-1.4` file as
```
pref.animation_dir=E:
```
Now if I try to open a file I get an error:
Untitled

### DIFF
--- a/ETL/ETL/_stringf.h
+++ b/ETL/ETL/_stringf.h
@@ -160,8 +160,6 @@ is_absolute_path(const std::string &path)
 #ifdef _WIN32
 	if(path.size()>=3 && path[1]==':' && is_separator(path[2]))
 		return true;
-	else if(path.size()==2 && path[1] == ':')
-		return true;
 #endif
 	if(!path.empty() && is_separator(path[0]))
 		return true;
--- a/ETL/ETL/_stringf.h
+++ b/ETL/ETL/_stringf.h
@@ -129,6 +129,12 @@ dirname(const std::string &str)
 		   return ".";
 	}
 
+#ifdef _WIN32
+	// leaving trailing separator after windows drive name
+	if (std::distance(str.begin(), iter) == 2 && str[1] == ':')
+		iter++;
+#endif
+
 	return std::string(str.begin(),iter);
 }
 
--- a/ETL/ETL/_stringf.h
+++ b/ETL/ETL/_stringf.h
@@ -132,7 +132,7 @@ dirname(const std::string &str)
 #ifdef _WIN32
 	// leave the trailing separator after windows drive name
 	if (std::distance(str.begin(), iter) == 2 && str.size() >= 3 && str[1] == ':' && is_separator(str[2]))
-		iter++;
+		++iter;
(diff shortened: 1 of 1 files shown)


ETL/ETL/_stringf.h
```diff
@@ -160,6 +160,8 @@ is_absolute_path(const std::string &path)
 #ifdef _WIN32
 	if(path.size()>=3 && path[1]==':' && is_separator(path[2]))
 		return true;
+	else if(path.size()==2 && path[1] == ':')
+		return true;
```
hesham-essam, line 164: I would've preferred to replace the first condition entirely, but I chose to be prudent

rodolforg:
I don't think `E:` is an absolute path. o.O `E:\` is. Where did you find this situation you consider it an error?

hesham-essam:
I found this error when building synfig with msvc in the master branch. Here is how to reproduce:
1. Save a file as `E:/file.sifz`
2. Try opening a new file.

The error in the screenshot will appear. The problem here is that `dirname` is used to get the directory of the saved file, and record it so that when the open dialog is executed once again it will go to it. And `dirname("E:/file.sifz") = "E:"` (dirname removes the last forward slash, which seems consistent with the behavior of linux's dirname). Now synfig sees "E:" as relative path, so it appends it to the cwd, so I get `E:/Programming/synfig_build/E:/`.

I tried playing with msvc's `std::filesystem` and interestingly I get:
```
assert(std::filesystem::path("E:/a/b").parent_path() == "E:/a"); // removes trailing sepeartor
assert(std::filesystem::path("E:/a/").parent_path() == "E:/a"); // removes trailing separator and doesn't go one directory up
assert(std::filesystem::path("E:/a").parent_path() == "E:/"); // does not remove trailing separator

assert(std::filesystem::path("E:/a").is_absolute());
assert(std::filesystem::path("E:/").is_absolute());
assert(!std::filesystem::path("E:").is_absolute());
```

So it seems that the behaviour of `dirname` is not consistent with msvc's implementation. Only in the case of `E:/file` does it return dirname to be `E:/`, otherwise it removes the trailing separator.

hesham-essam:
The problem here is that `E:` is a valid filename on linux, but the colon is an invalid character for a filename on Windows. So my solution isn't correct. The better solution would be to change `dirname` behaviour to leave the trailing separator (only in this case) to be consistent with msvc's `std::filesystem`

rodolforg (commented):
Prefer `++iter`

rodolforg (approved):

ice0:
Merged. Thank you!

PR #2763, opened 2022-07-23
feat: Added support for searching in the Parameters Panel

### LINKED ISSUE #1048: Please add search functionality in properties panel
Please add search functionality in properties panel. It'll be helpful in finding properties faster instead of scrolling through a long list of properties.

mohamedAdhamc:
fixes #1048 
**Background:**

 first I just need to give a little background to make it clear why I implemented this way
The default treeview search (impemented in synfig in places like the layers panel) works in a way such that, as you type the first match found is selected, to advance the search there are two options, one is to use keys/scroll  wheel, two is to add more letters to the search. But the way it works is that it only work for the rows which are shown/expanded. 

for example, in the layers panel when its first opened, one is only able to search through the shown rows (the first level). one can not search for something which is beyond the first level. which maybe  fine in places like the layers panel. However in parameters panel its quite important to be able to search further than the first level when needed.

**How my implmentation of the search for the parameters panel works:**

 if searched for item is found in level one its selected, then with keys you -same as default- can move to next match. However, after level one is done the search continues and expands the next levels first branch on the condition that there is a match there. if there is no match no expansion happens

**whats left for this to be ready:**
currently searching for something deeper than two levels of current level doesn't function as it is supposed to be, so i will be working on this tommrow and hopefully then this will be done and ready for review.

### DIFF
--- a/synfig-studio/src/gui/trees/layertree.cpp
+++ b/synfig-studio/src/gui/trees/layertree.cpp
@@ -371,6 +371,12 @@ LayerTree::create_param_tree()
 	}
 #endif	// TIMETRACK_IN_PARAMS_PANEL
 
+
+	param_tree_view().set_enable_search(true);//mod adham //enable_search	true, if the user can search interactively.
+	param_tree_view().set_search_column(param_model.label);//mod adham //column	The column of the model to search in, or -1 to disable searching.
+	param_tree_view().set_search_equal_func(sigc::mem_fun(*this, &LayerTree::search_param_tree));//mod adham
+
+
 	// This makes things easier to read.
 	param_tree_view().set_rules_hint();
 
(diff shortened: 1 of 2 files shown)


mohamedAdhamc:
ok I think now its working exactly like it should be. a match anywhere will be selected no matter its level/depth, also no unneeded paths are expanded in the process.

rodolforg (dismissed):
Initial review only.

mohamedAdhamc:
ok now these comments should be addressed, however with indentation I'm never sure as there isn't a clear set way :'). I tried as much as possible to follow the talked about rules in the forums when redoing it now, if there is something I missed though please let me know.

mohamedAdhamc:
@rodolforg anything else needed here ?

synfig-studio/src/gui/trees/layertree.cpp
```diff
@@ -392,6 +397,39 @@ LayerTree::create_param_tree()
 	//column_time_track->set_visible(false);
 }
 
+bool
+LayerTree::search_param_tree(const Glib::RefPtr<Gtk::TreeModel>& model,int column,const Glib::ustring& search_string,const Gtk::TreeModel::iterator& iter)//make sure it is indeed search string
+{
+	Gtk::TreePath path(iter);
+	if(match_found && (path != match_path))// when match is found dont bother with others
+		return true;
```
rodolforg, line 405: I don't see what this `if` and your state variables help.
I wrote:
```
bool
LayerTree::search_param_tree(const Glib::RefPtr<Gtk::TreeModel>& model,int column,const Glib::ustring& search_string,const Gtk::TreeModel::iterator& iter)//make sure it is indeed search string
{
	static int count1 = 0, count2 = 0;
	warning("b: %i a: %i", count1, count2);
	++count1;
	Gtk::TreePath path(iter);
	if(match_found && (path != match_path))// when match is found dont bother with others
		return true;
	++count2;
	error("\tb: %i a: %i", count1, count2);
```

And it always print both debug messages (warning and error) and the counters are always equal.

synfig-studio/src/gui/trees/layertree.cpp
```diff
@@ -392,6 +397,39 @@ LayerTree::create_param_tree()
...
+
+	Gtk::TreeModel::Row row = *iter;
+	Gtk::TreeModel::Children children = row.children();
+	Glib::ustring substr(search_string.uppercase());
+	Glib::ustring label((*iter)[param_model.label]);
+	label=label.uppercase();
+	if(label.find(substr)!=Glib::ustring::npos){//i.e. match found
+		match_found=false;
```
rodolforg, line 413: comment and variable name/assignment diverge o.O

rodolforg (changes requested):
One thing I noticed: if I search for "Ver" it only access "Vertices" parameter, and Up/Down arrow keys don't let me search for inner "Vertex 00x" subparameters (in a Region layer, for example)

mohamedAdhamc:
> One thing I noticed: if I search for "Ver" it only access "Vertices" parameter, and Up/Down arrow keys don't let me search for inner "Vertex 00x" subparameters (in a Region layer, for example)

I think this is all due to the modified functionality I was trying to implement, which is described in the first comment ,however I think it may be unnecessary and the search functionality would probably be better off working the same way it does in other panels.

synfig-studio/src/gui/trees/layertree.cpp
mohamedAdhamc: this was to implement what was described in initial comment, but I think the functionality I was thinking of wasn't really necessary. Removed.

synfig-studio/src/gui/trees/layertree.cpp
mohamedAdhamc: xD, I don't really remember what I was thinking. Anyways this is remove also now.

rodolforg (approved):

ice0:
I did a short test trying to find the `invert` word.
As you can see, the first letters didn't work because they are used as keyboard shortcuts for tools. 
I'm thinking maybe add a textbox for that?

https://user-images.githubusercontent.com/5604544/219957776-1686df5f-c5c7-4f2f-9433-5762f5becc6f.mp4

rodolforg:
> I did a short test trying to find the `invert` word.

`ctrl` `f` works

mohamedAdhamc:
> > I did a short test trying to find the `invert` word.
> 
> `ctrl` `f` works

Agreed (also just retested to recheck), actually I thought `ctrl` `f` to bring up the pop-up search bar was the usual. It was working fine back on windows, and mint I believe. Also if I am not wrong this is also how it is in other places like layers panel for example.

p.s. However, I noticed in Ubuntu 22.04 (my current main distro) for some weird reason the pop-up search bar doesn't appear on using `ctr` `f` (but behaviour is still working) neither here nor in other places which have similar functionality as in layers panel for example. But then again the windowing seems to be messed up in Ubuntu 22.04 altogether, all the popup menus give out gdk/wayland warnings and behave in weird ways.

PR #2764, opened 2022-07-26
fix: fixed broken synfig.dll debugging in MSVC build

ice0:
While testing the Synfig MSVC build, I noticed that debugging of
functions located in the synfig.dll library does not work.

Visual Studio shows that this is happening because the pdb file
(with debug information) cannot be loaded. 

Screenshot_42

Attempting to load this file manually gives a strange error message:
"No matching symbol file found in this folder."

Screenshot_43

However, if we rebuild the synfig target, then everything started working.

This is happens because the synfig library (synfig.dll) and the
synfig executable (synfig.exe) have the same name and located
in the same directory (bin on Windows).

There are several solutions here:
1) Change the name of the `synfig` executable to something else like `synfig-cli`.

2) Set different output path for library pdb file:
```
set_target_properties(synfig PROPERTIES PDB_OUTPUT_DIRECTORY "lib")
```

3) Rename target in CMake from `synfig` to `libsynfig`

I prefer the last solution, but I'm not sure which one is better.

Your opinion?

### DIFF
--- a/synfig-core/src/synfig/synfig_export.h
+++ b/synfig-core/src/synfig/synfig_export.h
@@ -27,7 +27,7 @@
 
 #ifdef _MSC_VER
 // We need this export only for MSVC. Even on MinGW, it breaks linkning.
-#ifdef synfig_EXPORTS
+#ifdef libsynfig_EXPORTS
 /* We are building this library */
 #define SYNFIG_EXPORT __declspec(dllexport)
 #else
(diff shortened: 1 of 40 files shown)


morevnaproject:
I also prefer last solution. ^__^

hesham-essam:
+1 For the third solution as well.

morevnaproject (approved):

morevnaproject (approved):

synfig-core/src/synfig/synfig_export.h
```diff
@@ -27,7 +27,7 @@
 
 #ifdef _MSC_VER
 // We need this export only for MSVC. Even on MinGW, it breaks linkning.
-#ifdef synfig_EXPORTS
+#ifdef libsynfig_EXPORTS
```
rodolforg, line 30: Is this correct?

ice0: Yes. MSVC automatically adds the `%library_name%_EXPORTS` definition when creating a shared library.

PR #2765, opened 2022-07-26
feat: added support for setting playback interval through movable buttons

### LINKED ISSUE #1493: Better way to set playback area
The program has a way to limit the loading area. 

image

After Effect has a similar mechanism. Only there it works more clearly. 

image

AE has borders with which you can easily edit this zone. The loading area itself is also lighter.

There are also hot keys to set the sides. The side on the left is set with button B, and with the right button N.

image

Here is how it would look with us.

image

mohamedAdhamc:
Quick description of what I did in terms of implementation.
- added the boundary button rectangle
- adjusted the old rectangle to make its width smaller by the value which is the width of the button
- event handling for the button

Things that need opinions:
- Is the button width good or should it be wider also the color and look of it (I chose blue as I think it fits and was the same in the original request) ? 

Feature:

https://user-images.githubusercontent.com/100296264/180999147-dd452bfb-6275-4e55-be36-dbd81072c0d4.mp4

Fix #1493


mohamedAdhamc:
> 1. I propose to add on hover cursor (GDK_LEFT_SIDE/GDK_RIGHT_SIDE). This will show the user that these elements are interactive.
>    https://docs.gtk.org/gdk3/enum.CursorType.html

yes, i think the on hover cursor is a good idea to notify the users that it is movable. But since they are able to be moved in both directions along the x axis why not just this one for both ?
cursor for buttons (3)

mohamedAdhamc:
I added the "ew-resize" cursor, however, if other two cursors are more suitable, I could just switch them up it wouldn't be a big deal.

@ice0 regarding making them as icons. I saw that the icon would have to be rendered into a pixbuf then the pixbuff would be used in the cairo rectangle/context but I dont quite know yet how I would put the icon into the pixbuf, im looking into how to do it, in the documentation it mentions using `Gdk::Pixbuf::create_from_file("");` with the strirng being the path to the image but I dont really know how that would work, any ideas?

mohamedAdhamc:
These are the new icons:
the new icons (2)

Maybe, they would need some more adjustment but that should be just as easy as adjusting and changing the icon file.

mohamedAdhamc:
@ice0 done.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -172,6 +172,20 @@ Widget_Timeslider::Widget_Timeslider():
 			  | Gdk::BUTTON_RELEASE_MASK
 			  | Gdk::BUTTON_MOTION_MASK
 			  | Gdk::SCROLL_MASK );
+
+
+	GdkPixbuf *lower_boundary_pixbuf_c = gtk_icon_theme_load_icon (gtk_icon_theme_get_default (),
```
ice0, line 177: Why do not use Gtkmm methods here?
```
auto icon_theme = Gtk::IconTheme::get_default();
lower_boundary_pixbuf = icon_theme->load_icon("bound_button_icon_lower", 16);
upper_boundary_pixbuf = icon_theme->load_icon("bound_button_icon_upper", 16);
```

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -303,21 +317,51 @@ Widget_Timeslider::on_draw(const Cairo::RefPtr<Cairo::Context> &cr)
...
 				cr->save();
-				cr->rectangle(x0, 0.0, w, (double)get_height());
+				cr->rectangle(x0 + background_adjust, 0.0, w-8, (double)get_height());
 				cr->clip();
 				cr->translate(offset, 0.0);
 				cr->set_source(play_bounds_pattern);
 				cr->paint();
 				cr->restore();
 
+				Glib::RefPtr<Gdk::Pixbuf> icon;
```
ice0, line 336: I changed the icon just to show the problem:
Icon:
Screenshot_59

Problem:
Screenshot_58

Screenshot_60

As you can see, it overlaps with borders.
My suggestion is to first draw the borders and then draw a handle over them.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: oh yes definitely, for some reason I was having some trouble getting the equivalent gtkmm methods right away, so I thought to do it with gtk first then switch over but seems like I forgot xD.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: ok sounds good will do.

mohamedAdhamc:
ok, now when the added icon is 100 x 200 pixels it should be showing fine, I also added a bit of extra buffer width for the rectangle to be safe.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: hmm, the methods work but surprisingly I couldn't find them in either the gtkmm 3.0 docs or the 4.0 ones.

synfig-studio/images/CMakeLists.txt
```diff
@@ -192,6 +192,8 @@ set(ICONS
 	set_outline_color
 	show_grid_icon
 	show_guideline_icon
+	bound_button_icon_lower
+	bound_button_icon_upper
```
rodolforg, line 196: Change the icon names to:
- lower_bound_button_icon
- upper_bound_button_icon

@ice0: "button", "mark" or "handle"?

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -172,6 +174,10 @@ Widget_Timeslider::Widget_Timeslider():
 			  | Gdk::BUTTON_RELEASE_MASK
 			  | Gdk::BUTTON_MOTION_MASK
 			  | Gdk::SCROLL_MASK );
+
+	auto icon_theme = Gtk::IconTheme::get_default();
+	lower_boundary_pixbuf = icon_theme->load_icon("bound_button_icon_lower", 1);
+	upper_boundary_pixbuf = icon_theme->load_icon("bound_button_icon_upper", 1);
```
rodolforg, line 180: There is no need to this "1". Or is there?

synfig-studio/src/gui/widgets/widget_timeslider.h
```diff
@@ -56,9 +56,18 @@ class Widget_Timeslider: public Gtk::DrawingArea
...
+	Glib::RefPtr<Gdk::Cursor> default_cursor = Gdk::Cursor::create(get_display(), "default");
+
+	Glib::RefPtr<Gdk::Pixbuf> lower_boundary_pixbuf;
+	Glib::RefPtr<Gdk::Pixbuf> upper_boundary_pixbuf;
+
 	// last mouse position for dragging
 	double lastx;
 
+	bool move_boundary_button_lower=false;
+	bool move_boundary_button_upper=false;
```
rodolforg, line 69: keep naming consistency: "move_lower_bound_button", "lower_bound_pixbuf" (or change icon name to follow "lower boundary")
I don't know what is more correct in English: bound or boundary though.

synfig-studio/src/gui/widgets/widget_timeslider.h
```diff
@@ -56,9 +56,18 @@ class Widget_Timeslider: public Gtk::DrawingArea
 
 	Cairo::RefPtr<Cairo::SurfacePattern> play_bounds_pattern;
 
+	Glib::RefPtr<Gdk::Cursor> bounds_cursor = Gdk::Cursor::create(get_display(), "ew-resize");
+	Glib::RefPtr<Gdk::Cursor> default_cursor = Gdk::Cursor::create(get_display(), "default");
```
rodolforg, line 60: @ice0 what do you think about this initialization here in header?

synfig-studio/src/gui/widgets/widget_timeslider.h
rodolforg, line 69: spaces surrounding `=`

synfig-studio/src/gui/widgets/widget_timeslider.h
```diff
@@ -75,6 +84,8 @@ class Widget_Timeslider: public Gtk::DrawingArea
 
 	virtual void draw_background(const Cairo::RefPtr<Cairo::Context> &cr);
 
+	void get_bounds_rectangle_dimensions(double& x0, double& x1, double& w ,bool lower);
```
rodolforg, line 87: wrong spaces in last comma

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -303,23 +309,52 @@ Widget_Timeslider::on_draw(const Cairo::RefPtr<Cairo::Context> &cr)
 				double x1 = time_plot_data->get_double_pixel_t_coord(bounds[i][1]);
 				double w = x1 - x0;
 
+				double boundary_dimension=8;
```
rodolforg, line 312: `const`?
spaces surrounding assignment operator `=`

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -303,23 +309,52 @@ Widget_Timeslider::on_draw(const Cairo::RefPtr<Cairo::Context> &cr)
...
+				double background_adjust;
+
+				if(i==0)
+					background_adjust=0;
+				else
+					background_adjust=8;
```
rodolforg, line 318: IMO it's clearer if it was in a single (and simple) line:
```c++
double background_adjust = (i == 0) ? 0 : 8;
```
It's a safer habit too: always initialize a (primitive) variable.

I guess this `8` should actually be `boundary_dimension`.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -303,23 +309,52 @@ Widget_Timeslider::on_draw(const Cairo::RefPtr<Cairo::Context> &cr)
...
+
 				cr->save();
-				cr->rectangle(x0, 0.0, w, (double)get_height());
+				cr->rectangle(x0 + background_adjust, 0.0, w-8, (double)get_height());
```
rodolforg, line 321: should this `8` and the following `8`'s be `boundary_dimension`?

synfig-studio/images/CMakeLists.txt
mohamedAdhamc: true "lower_bound_button_icon" sounds more like an actual phrase, I have a weird habit of making then names have a weird order :".

p.s. will keep it "button" for now and if decided one of the others is better will change then.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: This is the desired icon size.

synfig-studio/src/gui/widgets/widget_timeslider.h
mohamedAdhamc: ok makes sense.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: sure const seems reasonable.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: I believe so, I remember changing them all when changing the size.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: hmm, yes the ternary operator in my opinion also here makes things nice and compact, but I don't usually opt for it as some/many people deem it "bad for readability".

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: actually here I switch up boundary_dimension value depending on lower/upper bound so I think maybe better to name it `boundary_adjust `and have another const for the actual dimension... the 8 in this case

mohamedAdhamc:
Hmm, looks like I messed something up while renaming the icon , will look into it when I get back home/ in the morning.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +388,65 @@ bool
...
 	SYNFIG_EXCEPTION_GUARD_BEGIN()
 	lastx = (double)event->x;
+	move_lower_bound_button = false;
+	move_upper_bound_button = false;
 	return event->button == 1 || event->button == 2;
 	SYNFIG_EXCEPTION_GUARD_END_BOOL(true)
 }
 
+void
+Widget_Timeslider::get_bounds_rectangle_dimensions(double& x0, double& x1, double& w ,bool lower)
```
rodolforg, line 398: 1. You don't need the width parameter.
When you use this method below, you only uses `x0` or `x1`.

2. space-comma issue before `lower`

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +388,65 @@ bool
...
+		 w= x1 - x0;
+	}
+}
+
 bool
 Widget_Timeslider::on_motion_notify_event(GdkEventMotion* event) //for dragging
 {
 	SYNFIG_EXCEPTION_GUARD_BEGIN()
+
+	double current= time_plot_data->get_double_pixel_t_coord(time_plot_data->get_t_from_pixel_coord(event->x));
```
rodolforg, line 417: `current` what? :) Why this conversion followed by reversed conversion?

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +388,65 @@ bool
...
+	double x0,x1,w,z0,z1,width;
+	get_bounds_rectangle_dimensions(x0,x1,w,true);
+	get_bounds_rectangle_dimensions(z0,z1,width,false);
+
+	if((((current <= (x0+w))  && (current >= (x0 + w - 4)) && !move_lower_bound_button) || (move_lower_bound_button))){
```
rodolforg, line 422: 1. I'm curious what this magic `4` is
2. You can definitely rephrase this boolean expression :)
```c++
if (move_lower_bound_button || (current <= x1 && current >= x1 - 4))
```

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +388,65 @@ bool
...
+		if((get_window()->get_cursor() != bounds_cursor))
+		   get_window()->set_cursor(bounds_cursor);
+	}
+	else if((((current <= (z0 + 4))  && (current >= (z0)) && !move_upper_bound_button) || (move_upper_bound_button))){
```
rodolforg, line 426: same here

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +388,65 @@ bool
...
+		if((get_window()->get_cursor() != bounds_cursor))
+		   get_window()->set_cursor(bounds_cursor);
+	}
+	else {
+		if((get_window()->get_cursor() != default_cursor))
```
rodolforg, line 431: I'm not sure if it is necessary to check it before calling `set_cursor`

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: 1- regarding one yes the width isn't necessary to be added but I actually made this function in order not to redo the same exact lines while getting the dimensions at various places so instead I just define the local variable and pass them to this function by reference to have them set up, kind of like a c style function. However, instead I can just manually define the width everywhere I need the coordinates. what do you think ?

2- oops didnt notice that one.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: yes, it seems redundant will remove.

mohamedAdhamc:
unfortunately, we are having a power outage now for the past 3/4 hours or so but they say it will end soon I was just about to push a commit in the very latest feature but power went out and laptop is about to die soon :( . will push the commit when power comes back then resolve the comments here.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: I think the four was the boundary button dimension but when increasing the size I probably forgot to increase as well here, :'' exactly the reason why I shouldn't be using magic numbers, they are difficult to maintain xD.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: @rodolforg  actually now I remember why I did that it was to avoid the constant setting of the cursor, as you see this is in the motion notify event.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: hmm looks like I overcomplicated this for no reason :" I thought the "pixel coordinates" where a bit diffrent from the events coordinate but turns out they are the same.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
rodolforg: The set-cursor method does not check it then? Oh ok

mohamedAdhamc: No in fact you're correct, loooking at the source [code](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkwidget.c#L12144)  line 12165 to be exact 
supposedly `g_set_object()` returns if the value had been changed or not and if not changed the `set_cursor()` method returns. so In fact my extra check here seems redundant will remove now.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +387,58 @@ bool
...
+
 bool
 Widget_Timeslider::on_motion_notify_event(GdkEventMotion* event) //for dragging
 {
 	SYNFIG_EXCEPTION_GUARD_BEGIN()
+
+	double current_x = event->x;
+	double x0,x1,w,z0,z1,width;
+	get_bounds_rectangle_dimensions(x0,x1,w,true);
+	get_bounds_rectangle_dimensions(z0,z1,width,false);
```
rodolforg, line 419: What I meant in previous review is: `w` and `width` are useless here. So, why `get_bounds_rectangle_dimensions()` needs them as (out) parameter. This method is only used in two places here and none of them use `w` :)

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +387,58 @@ bool
...
+		   get_window()->set_cursor(default_cursor);
+
 	double dx = (double)event->x - lastx;
 	lastx = (double)event->x;
 
 	if (!time_plot_data->time_model || get_width() <= 0 || get_height() <= 0)
 		return false;
 
-	Gdk::ModifierType mod = Gdk::ModifierType(event->state);
+	Gdk::ModifierType mod = Gdk::ModifierType(event->state);//mouse buttons are also modifiers
```
rodolforg, line 434: remove comment?

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -338,9 +365,20 @@ Widget_Timeslider::on_button_press_event(GdkEventButton *event) //for clicking
...
 	}
 
+	double current= time_plot_data->get_double_pixel_t_coord(time_plot_data->get_t_from_pixel_coord(event->x));
+	double x0,x1,w,z0,z1,width;
+	get_bounds_rectangle_dimensions(x0,x1,w,true);
+	get_bounds_rectangle_dimensions(z0,z1,width,false);
+
+	if((current <= (x0+w))  && (current >= (x0 + w - boundary_dimension) ))
+		move_lower_bound_button = true;
+	else if((current <= (z0 + boundary_dimension))  && (current >= (z0)))
```
rodolforg, line 378: While it's better than if they were missing, there are many parenthesis here XD
In C++ at least, only the outmost parenthesis here are needed.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -338,9 +365,20 @@ Widget_Timeslider::on_button_press_event(GdkEventButton *event) //for clicking
...
-		time_plot_data->time_model->set_time(time);
+		time_plot_data->time_model->set_time(time);//setting the time which is in time model
 	}
 
+	double current= time_plot_data->get_double_pixel_t_coord(time_plot_data->get_t_from_pixel_coord(event->x));
+	double x0,x1,w,z0,z1,width;
+	get_bounds_rectangle_dimensions(x0,x1,w,true);
+	get_bounds_rectangle_dimensions(z0,z1,width,false);
+
+	if((current <= (x0+w))  && (current >= (x0 + w - boundary_dimension) ))
```
rodolforg, line 376: `x1` ?

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +387,58 @@ bool
 Widget_Timeslider::on_button_release_event(GdkEventButton *event){
 	SYNFIG_EXCEPTION_GUARD_BEGIN()
 	lastx = (double)event->x;
+	move_lower_bound_button = false;
+	move_upper_bound_button = false;
```
rodolforg, line 391: I believe `moving_lower_bound_button` is better

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -338,9 +365,20 @@ Widget_Timeslider::on_button_press_event(GdkEventButton *event) //for clicking
 
 	if (event->button == 1) {
 		Time time = time_plot_data->get_t_from_pixel_coord(event->x);
-		time_plot_data->time_model->set_time(time);
+		time_plot_data->time_model->set_time(time);//setting the time which is in time model
 	}
 
+	double current= time_plot_data->get_double_pixel_t_coord(time_plot_data->get_t_from_pixel_coord(event->x));
```
rodolforg, line 371: surround assignment operator `=` with single space

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -338,9 +365,20 @@ Widget_Timeslider::on_button_press_event(GdkEventButton *event) //for clicking
 
 	if (event->button == 1) {
 		Time time = time_plot_data->get_t_from_pixel_coord(event->x);
-		time_plot_data->time_model->set_time(time);
+		time_plot_data->time_model->set_time(time);//setting the time which is in time model
```
rodolforg, line 368: remove comment

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -303,23 +309,44 @@ Widget_Timeslider::on_draw(const Cairo::RefPtr<Cairo::Context> &cr)
...
+					boundary_adjust=-boundary_dimension;
+					icon = lower_bound_pixbuf;
+				}
+				else{
+					boundary_adjust=-w;
+					icon = upper_bound_pixbuf;
+				}
+
+				cr->save();
+				Gdk::Cairo::set_source_pixbuf(cr, icon, x0+w+boundary_adjust, 0);
```
rodolforg, line 346: spaces surrounding `+`

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +387,58 @@ bool
...
+Widget_Timeslider::get_bounds_rectangle_dimensions(double& x0, double& x1, double& w , bool lower)
+{
+	if(lower){
+		 x0 = time_plot_data->get_double_pixel_t_coord(time_plot_data->lower_ex);
+		 x1 = time_plot_data->get_double_pixel_t_coord(time_plot_data->time_model->get_play_bounds_lower());
+		 w = x1 - x0;
+	}
+	else {
+		 x0= time_plot_data->get_double_pixel_t_coord(time_plot_data->time_model->get_play_bounds_upper());
+		 x1= time_plot_data->get_double_pixel_t_coord(time_plot_data->upper_ex);
```
rodolforg, line 406: spaces ` = `

synfig-studio/src/gui/widgets/widget_timeslider.h
```diff
@@ -75,6 +84,9 @@ class Widget_Timeslider: public Gtk::DrawingArea
 
 	virtual void draw_background(const Cairo::RefPtr<Cairo::Context> &cr);
 
+	void get_bounds_rectangle_dimensions(double& x0, double& x1, double& w , bool lower);
+	const double boundary_dimension =8;
```
rodolforg, line 88: ugh. It is used in more than one method, so we cannot reduce its scope.
However, if we decide to change the default value, it will force recompilation of any file that `#include`s Widget_TimeSlider.h

@ice0  opinion?

rodolforg (changes requested):

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: I think I kinda get you what you mean here, yes this method is used I believe in two places of which this one doesnt need the width and w. I'll just remove it then from the parameters of the funciton.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: oh yes definitely I just comment on random things when I first start working on the feature to understand stuff, and sometimes I forget to delete some of them xD.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: I admit they are too much here :') will remove them.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: excuse all these not spaced out operators I wrote them before knowing about this style xD

mohamedAdhamc:
alright, I spent most of the day today working on the latest feature, I'm now heading out and hopefully, when I come back in a couple of hours I will address the comments before I sleep.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
mohamedAdhamc: oh, yes x1 makes for a smaller condition. I updated it in the other place and forgot to here, will do. 

p.s. now I get what you meant, through x1 w/width arent really ever needed

synfig-studio/src/gui/widgets/widget_timeslider.cpp
ice0: https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1IconTheme.html#a9e9307cefa458eea1e8946d9def2e527

synfig-studio/images/CMakeLists.txt
ice0: I think the handle is more correct, because we don't click on it, we drag it.

synfig-studio/src/gui/widgets/widget_timeslider.h
ice0: I think it's ok, but it might be better to move it to a .cpp file for this reason:

> However, if we decide to change the default value, it will force recompilation of any file that #includes Widget_TimeSlider.h

synfig-studio/src/gui/widgets/widget_timeslider.h
ice0: Agree. 
@mohamedAdhamc Can you explain how it is used?

ice0:
Increased `boundary_dimension` to 80.

It looks like the changes should be mirrored?

Screenshot_69

synfig-studio/src/gui/widgets/widget_timeslider.h
mohamedAdhamc: `boundary_dimension` is the width of the draggable handle.
 It's used in 
1- Widget_Timeslider::on_draw in the process of rendering using cairo to set the dimensions for the bound rectangles and the bound handle.
2- it is also used in the button press and motion notify event to know if the press/ position of the mouse is within the boundaries of the "bound handle"

I think then it should probably be initialized in the constructor

ice0: I think it might be much easier to use `Gtk::Image` for handles?
It supports event handling, knows the position and size, and we can control the boundaries with margins.
Am I missing something here?

mohamedAdhamc:
> Increased `boundary_dimension` to 80.
> 
> It looks like the changes should be mirrored?

I think the problem here is that the boundary dimension was increased but the icon size wasnt, and because the icon is rendered from left to right for the lower handle it is fully drawn then there is a big space left unused to the right of it, and for the upper bound it is drawn fully at the start of the rectangle specified for the icon and then also a large portion is left unused.

I mean increasing the boundary dimension without increasing the width of the icon would always cause issue, the way it is now if I remember correctly is that the icon of size 100x200 pixels is drawn perfectly, so Im guessing in this case increasing the width of the cairo rectangle of the icon from 8 to 80 width would have to be increased from 200 to 2000.

p.s. this is how it looks like
correct size icon bound

synfig-studio/src/gui/widgets/widget_timeslider.h
mohamedAdhamc: > I think it might be much easier to use `Gtk::Image` for handles? It supports event handling, knows the position and size, and we can control the boundaries with margins. Am I missing something here?

well the thing is the whole "timetrack" is drawn with cairo surfaces, the play bounds, the patterns, etc... and in it isnt resizable. so I think it kind of makes sense to have this bound handle be the same as well to be moved in the same manner and for overall consistency. An image may make it easier to be resized and all (even though really that isnt needed the only thing needed if a user would want to replace the icon he just needs to make sure to replace it with an icon of the same size and everything should be fine), and I'm not sure how reliable it would be in terms of being moved around the time track .

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +389,56 @@ bool
...
 	SYNFIG_EXCEPTION_GUARD_BEGIN()
+
+	double current_x = event->x;
+	double x0, x1, w, z0, z1;
+	get_bounds_rectangle_dimensions(x0, x1, true);
+	get_bounds_rectangle_dimensions(z0, z1, false);
+
+	if(moving_lower_bound_handle || (current_x <= x1 && current_x >= x1 - boundary_dimension))
+		   get_window()->set_cursor(bounds_cursor);
+	else if(moving_upper_bound_handle || ((current_x <= (z0 + boundary_dimension))  && (current_x >= (z0))))
```
rodolforg, line 423: space before `(` -> `if (`

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +389,56 @@ bool
...
+	moving_lower_bound_handle = false;
+	moving_upper_bound_handle = false;
 	return event->button == 1 || event->button == 2;
 	SYNFIG_EXCEPTION_GUARD_END_BOOL(true)
 }
 
+void
+Widget_Timeslider::get_bounds_rectangle_dimensions(double& x0, double& x1, bool lower)
+{
+	if(lower){
```
rodolforg, line 401: space before `(` -> `if (`

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -341,6 +371,16 @@ Widget_Timeslider::on_button_press_event(GdkEventButton *event) //for clicking
...
 	}
 
+	double current_x = event->x;
+	double x0, x1, z0, z1;
+	get_bounds_rectangle_dimensions(x0, x1, true);
+	get_bounds_rectangle_dimensions(z0, z1, false);
+
+	if(current_x <= x1 && current_x >= x1 - boundary_dimension)
+		moving_lower_bound_handle = true;
+	else if(current_x <= (z0 + boundary_dimension)  && current_x >= z0)
```
rodolforg, line 381: space before `(` -> `if (`

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -303,23 +312,44 @@ Widget_Timeslider::on_draw(const Cairo::RefPtr<Cairo::Context> &cr)
...
 				cr->stroke();
-				cr->set_source_rgba(0.0, 0.0, 0.0, 0.3);
-				cr->rectangle(x0 + 0.5, 0.5, w - 1.0, (double)get_height() - 1.0);
+				cr->set_source_rgba(0.0, 0.0, 0.0, 0.3);//outer border
+				cr->rectangle(x0 + 0.5 + boundary_adjust, 0.5, w - 1 - boundary_dimension, (double)get_height() - 1.0);
 				cr->stroke();
 				cr->restore();
+
+				Glib::RefPtr<Gdk::Pixbuf> icon;
+				if(i == 0){
```
rodolforg, line 339: space before `(` -> `if (`

rodolforg: and between `){` -> `) {`

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -303,23 +312,44 @@ Widget_Timeslider::on_draw(const Cairo::RefPtr<Cairo::Context> &cr)
...
+					boundary_adjust=-boundary_dimension;
+					icon = lower_bound_pixbuf;
+				}
+				else{
```
rodolforg, line 343: `} else {`

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -349,21 +389,56 @@ bool
...
 }
 
+void
+Widget_Timeslider::get_bounds_rectangle_dimensions(double& x0, double& x1, bool lower)
+{
+	if(lower){
+		 x0 = time_plot_data->get_double_pixel_t_coord(time_plot_data->lower_ex);
+		 x1 = time_plot_data->get_double_pixel_t_coord(time_plot_data->time_model->get_play_bounds_lower());
+	}
+	else {
```
rodolforg, line 405: `} else {`

mohamedAdhamc:
@rodolforg, done :)

p.s. I believe the deadline for the mentor mid- evaluation is sometime tomorrow the 12th. they probably should've sent out an email, but I'm notifying here just in case as my timeline is non-standard.

synfig-studio/src/gui/widgets/widget_timeslider.h
rodolforg: I don't know if it be better, but you could use a `Gtk::Overlay` that contains a `Gtk::Fixed` to then include the `Gtk::Image`s

rodolforg (commented):
One thing that bothers me is that the current time changes when dragging the playback boundaries.
But I don't know if it is really a problem or just me XD

mohamedAdhamc:
> One thing that bothers me is that the current time changes when dragging the playback boundaries. But I don't know if it is really a problem or just me XD

Hmm, I'm not sure either if it could be a problem - that is better left for you guys to decide- but if it is of any help. The reason I thought it may be helpful is while moving the bound knowing the exact place where it is, and if there is already an animation in place it lets you visualise which part of it you are bounding as was shown in the video above.

ice0:
> I think the problem here is that the boundary dimension was increased but the icon size wasnt

1. I thought the boundary dimension is the space between the handle and the borders.
If not, how is it used?

Screenshot_84

> One thing that bothers me is that the current time changes when dragging the playback boundaries.
2. Same for me, but I prefer to hear @morevnaproject opinion here.

3. I suggest to use `synfig::Rect` class to check bounds. 
Example pseudcode:
```
Point cursor_pos(event->x, 0);

Rect low_bound = get_play_bounds_rect(true);
Rect up_bound  = get_play_bounds_rect(false);
```

Before:
```
if (moving_lower_bound_handle || (current_x <= x1 && current_x >= x1 - boundary_dimension))
	   get_window()->set_cursor(bounds_cursor);
else if (moving_upper_bound_handle || ((current_x <= (z0 + boundary_dimension))  && (current_x >= (z0))))
	   get_window()->set_cursor(bounds_cursor);
else
	   get_window()->set_cursor(default_cursor);
```

After:
```
bool moving_handle = moving_lower_bound_handle || moving_upper_bound_handle;
if (moving_handle || low_bound.contains(cursor_pos) || up_bound.contains(cursor_pos))
	get_window()->set_cursor(bounds_cursor);
else
	get_window()->set_cursor(default_cursor);
```

As for me now it's more clear what is going on here.

4. Cursor problem:

https://user-images.githubusercontent.com/5604544/185033173-68859b86-b2b9-4b4a-9462-162cecfd8464.mp4

mohamedAdhamc:
> 1. I thought the boundary dimension is the space between the handle and the borders.
>    If not, how is it used?
> 

Its for the the whole pointed to space plus the space taken up by the handle in the picture, it is supposed to be the space which the handle fills up, and which is movable by the cursor.


> 3. I suggest to use `synfig::Rect` class to check bounds.
>    Example pseudcode:
> 

yes, this seems to make it much more readable and abstract away the implementation a bit, I will be lookign into `synfig::rect` to do it.

rodolforg:
> I think maybe we can change the cursor for moving the bounds, we could use something like what was suggested in the beginning GDK_LEFT_SIDE/GDK_RIGHT_SIDE.

I think it's better now.



> also thought maybe we can alter the color of the guide but this seems like a bad option because the icon is meant so that it can be replaced, so if it is replaced by someone we are not sure if the altered color will function properly.

You could blend the focused handle with another source color to highlight it: https://cairographics.org/operators/

mohamedAdhamc:
> I think it's better now.
> 
> > also thought maybe we can alter the color of the guide but this seems like a bad option because the icon is meant so that it can be replaced, so if it is replaced by someone we are not sure if the altered color will function properly.
> 
> You could blend the focused handle with another source color to highlight it: https://cairographics.org/operators/

hmm ok sounds good, and what about the "time changing when dragging the bounds" should I just disable it?

mohamedAdhamc:
https://user-images.githubusercontent.com/100296264/188481265-6d3c9234-5abd-4c1e-a336-abe382581c8d.mp4

this is the new highlight.

I think I need two opinions here on what to do:
1- regarding the time-slider thing (not sure what its called) should I disable it moving when the bounds are moving ? so far ice0 and rodolforg say yes but we are waiting for morevnaproject opinion.

2- when the button is released if the pointer is on the bound should it be still highlighted or not? the way I have it now is it doesn't so as to show that the guide drag ended but I don't know if that's how it should be.

rodolforg:
2. Maybe a different highlight intensity?

mohamedAdhamc:
> 2. Maybe a different highlight intensity?

I'm not sure... I think it might get a bit confusing. I think to just not complicate things here we should still have it highlighted even if the button is released as long as the cursor is still in the bound, but I'm also not sure.

rodolforg:
It's pretty common UI themes deal with these widget states:
- disabled, enabled, enabled+focused, enabled+mouse hover, enabled + pressed, etc.

See https://tailwindcss.com/docs/hover-focus-and-other-states#pseudo-classes (section "Hover, focus, and active"). The button "Save changes" has a color when left untouched, another when mouse is hovering it, and other color while pressed down.

mohamedAdhamc:
> It's pretty common UI themes deal with these widget states:
> 
> * disabled, enabled, enabled+focused, enabled+mouse hover, enabled + pressed, etc.

Hmm, ok sounds good, so now since the ideas shouldn't be too foreign. I think three different states here would do 
1- button is pressed 
2- button is being hovered on (whether before or after or without a press)
3- button is untouched (original state)

mohamedAdhamc:
@rodolforg done.

rodolforg:
@morevnaproject could you test this PR?

ice0:
@mohamedAdhamc 
Can you rebase it please?

mohamedAdhamc:
@ice0 done.

rodolforg:
@morevnaproject  ping :)

morevnaproject (dismissed):

rodolforg:
Should I merge, @mohamedAdhamc ? XD I think I'll squash all commits then merge. What do you think?

mohamedAdhamc:
> Should I merge, @mohamedAdhamc ? XD

Konstantin approved so I'm assuming things are good :smile: . I'll also give it one final test from my side as extra precaution. 

>I think I'll squash all commits then merge. What do you think?

Yes sure, I think most of the commits were just multiple fixes and adjustments anyways.

rodolforg (approved):

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -351,22 +403,73 @@ bool
...
+Widget_Timeslider::get_bounds_rectangle(bool lower) const
+{
+	if (lower) {
+		 double x1 = time_plot_data->get_double_pixel_t_coord(time_plot_data->time_model->get_play_bounds_lower());
+		 synfig::Rect lower_bound(synfig::Point(x1 - handle_dimension, 0.0), synfig::Point(x1, 0.0));
+		 return lower_bound;
+	} else {
+		 double x0 = time_plot_data->get_double_pixel_t_coord(time_plot_data->time_model->get_play_bounds_upper());
+		 synfig::Rect upper_bound(synfig::Point(x0, 0.0), synfig::Point(x0 + handle_dimension, 0.0));
+		 return upper_bound;
```
rodolforg, line 434: fix indentation... extra space here right before the line contents.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -351,22 +403,73 @@ bool
...
+	bool moving_handle = moving_lower_bound_handle || moving_upper_bound_handle;
+	bool bounds_enabled = time_plot_data->time_model->get_play_bounds_enabled();
+	hovering_on_lower_bound_handle = lower_bound.is_inside(cursor_pos);
+	hovering_on_upper_bound_handle = upper_bound.is_inside(cursor_pos);
+	bool hovering_on_handle = hovering_on_lower_bound_handle || hovering_on_upper_bound_handle;
+
+	if ((moving_handle || hovering_on_handle) && (bounds_enabled))
+		   get_window()->set_cursor(bounds_cursor);
+	else
+		   get_window()->set_cursor(default_cursor);
```
rodolforg, line 455: fix indentation... extra spaces here right before the line contents.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -342,7 +382,19 @@ Widget_Timeslider::on_button_press_event(GdkEventButton *event) //for clicking
 		Time time = time_plot_data->get_t_from_pixel_coord(event->x);
 		time_plot_data->time_model->set_time(time);
 	}
-
+	//if moving timeslider along with handle isnt wanted move these above prev if and add cond.
+	bool bounds_enabled = time_plot_data->time_model->get_play_bounds_enabled();
+	if (bounds_enabled)	{
```
rodolforg, line 387: remove extra spaces before `{`. Use a single one.

synfig-studio/src/gui/widgets/widget_timeslider.cpp
```diff
@@ -342,7 +382,19 @@ Widget_Timeslider::on_button_press_event(GdkEventButton *event) //for clicking
...
+		synfig::Rect lower_bound = 	get_bounds_rectangle(true);
+		synfig::Rect upper_bound =	get_bounds_rectangle(false);
+		synfig::Point cursor_pos(event->x, 0.0);
+
+		if (lower_bound.is_inside(cursor_pos))
+				moving_lower_bound_handle = true;
+		else if (upper_bound.is_inside(cursor_pos))
+				moving_upper_bound_handle = true;
```
rodolforg, line 395: fix indentation... extra tab here right before the line contents.

rodolforg (changes requested):
Final touches then ;)

mohamedAdhamc:
Nice catches, qt creator has the tendency to be really weird with indentations, it wasn't even showing me some of the tabs correctly now xD. 

Btw -sidenote- has there even been a clang file made for synfig ? I don't recall one being mentioned before. Maybe if we make one that only targets indentation for pr specific changes it could be helpful in reviews just to ask people to run it. What do you think ?

rodolforg:
> Btw -sidenote- has there even been a clang file made for synfig ? I don't recall one being mentioned before. Maybe if we make one that only targets indentation for pr specific changes it could be helpful in reviews just to ask people to run it. What do you think ?

I'm trying to create one, but it still has some issues.
[synfig-clang-format.zip](https://github.com/user-attachments/files/20027458/synfig-clang-format.zip)

rodolforg (approved):

rodolforg:
Merged! Thank you for your contribution.

PR #2740, opened 2022-07-09
ci: Enabled tests on CI

ice0:

PR #2769, opened 2022-07-30
test: Disabled failing test of `synfig::Node` class for now

rodolforg:
To enable our tests in the CI framework (PR #2740), we'll temporarily
avoid the test deleting_node_removes_it_as_parent_from_its_children().

I don't know if it is really needed: the problem seems to exist since
ever. Maybe it makes Synfig Studio crashes, maybe not.
More investigation is needed.

I did make a quick fix for it (storing sigc::connection), but I don't
know if it would help or not or even cause more problems.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2775, opened 2022-08-02
refactor: clean `synfig::Target` a bit

rodolforg:
- remove unneeded `#include`
- use `override`
- sort overridden methods in the same order in the different Target-inherited classes
- fix constructor for `Target_Null` and `Target_Null_Tile`
- remove virtually empty `.cpp` files


ice0 (approved):

ice0:
Merged. Thank you!

PR #2778, opened 2022-08-04
fix: hide bone width handles when they are meaningless

rodolforg:
- For Skeleton Deformation layer, it only makes sense for rest mode bones
- For Skeleton layer, bone width is completely useless
  (as for other bone info, e.g Z depth)


morevnaproject (approved):
Works perfectly! Thank you so much!

PR #2780, opened 2022-08-08
refactor: prefer FileSystemNative rather GLib

rodolforg:
less (direct) external dependency

### DIFF
--- a/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
+++ b/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
@@ -34,12 +34,10 @@
 #	include <config.h>
 #endif
 
-#include <synfig/localization.h>
+#include <synfig/filesystemnative.h>
 #include <synfig/general.h>
+#include <synfig/localization.h>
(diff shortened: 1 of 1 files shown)


ice0 (dismissed):

synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp
```diff
@@ -34,8 +34,9 @@
 #	include <config.h>
 #endif
 
-#include <synfig/localization.h>
+#include <synfig/filesystemnative.h>
 #include <synfig/general.h>
+#include <synfig/localization.h>
 #include <synfig/soundprocessor.h>
 
 #include <cstdio>
```
ice0, line 42: Looks like this includes no longer being used.

```
// MSVC
#ifndef F_OK
#define F_OK 0
#endif
```
Can be removed too.

rodolforg: Done

ice0 (approved):

ice0:
Merged. Thank you!

PR #2789, opened 2022-08-10
fix!: mod_imagemagick didn't import anything

rodolforg:
it uses application name "convert" to convert the desired image
into another temporary one with a well-known format (PNG).
This temporary image would then be really imported
(via another module - not very clever; what if it is not loaded?)

The problem is the temporary file is automatically named without any
extension.
In the end, Synfig can't load it, as it depends on the extension to select
the Importer.

Tested with a XPM file because XPM seems to be only loaded by imagemagick
https://filesamples.com/formats/xpm

BREAKING CHANGE: Synfig library API changed:
- Two methods of FileSystemTemporary has a new optional parameter: extension:
 - generate_system_temporary_filename
 - generate_system_temporary_filename_base


ice0 (approved):

ice0:
Merged. Thank you!

Gilraiser:
Thank you very much, I'm sure it will be very helpful for everyone.

PR #2791, opened 2022-08-11
refactor!: add etl::hermite::derivative() and remove ETL/calculus

rodolforg:
BREAKING CHANGE: ETL API changed:
- add etl::hermite::derivative()
- remove etl::derivative() ; not used
- remove etl::derivative<hermite<>>() ; use etl::hermite::derivative()
- remove etl::integral() ; not used


rodolforg:
I believe next step here is to merge ETL/hermite with ETL/bezier, since hermite is just a simple "specialization" class of bezier class.
Then we can move new bezier file to synfig. Nothing on ETL depends on them.

ice0 (approved):

ice0:
Merged. Thank you!

rodolforg:
> next step here is to merge ETL/hermite with ETL/bezier, since hermite is just a simple "specialization" class of bezier class.

@ice0 do you agree? If so, what should be the final filename: bezier or hermite? (Note: both classes will exist in the same file)

ice0:
Fast grep from sources:

```
include.*hermit - 18 matches
include.*bezier - 8 matches
```

First I thought about bezier, but now I am not sure about it.

rodolforg:
Some with <hermite> includes <bezier> too. So there is even less with bezier directly.
Change to hermite, then?

rodolforg:
Or, as the majority of code is in bezier, we keep it :)

ice0:
I think the decision will be more clear when you check second bezier/hermite code.
P.S. I haven't checked them yet

PR #2805, opened 2022-08-17
feat: Faster entry of values in the value field

### LINKED ISSUE #2191: Feature request: Faster entry of values in the value field
Synfig studio v1.4.1 Windows 10 21H1
While clicking on the value field in the properties/ parameters panel. if the whole value is selected upon clicking, then it would provide easier entry of new values, like just clicking the value field and entering the parameter values through the keyboard directly, instead of manually selecting/ deleting the value first and then entering the new value. This feature should be implemented throughout synfig as it would facilitate easier entry of values in the value field.
Animation1

mohamedAdhamc:
Fixes #2191

- done for the distance widget, Aka the spin button with the pixels unit.
- now mostly what remains is to decide on what method would be best for the remaining spin buttons, I think maybe just sub-classing/deriving a class from the spin button would save having to connect signals to each and every spin button, so I think I will probably go with that, will do early in the morning tomorrow.

p.s in the original feature request it was mainly highlighting the need for this there, I'm not quite sure if it should only be done there, but since it was mentioned it was wanted to be universal behavior I'm assuming maybe apply everywhere in the main window? However, I think I will just continue this for the different tools until I get an opinion, and anyways I think only places where it is yet to be implemented is probably some of the entries below and above the zoom dial entry so it shouldnt be a problem.

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_distance.cpp
+++ b/synfig-studio/src/gui/widgets/widget_distance.cpp
@@ -59,6 +59,7 @@ Widget_Distance::Widget_Distance():
 {
 	set_adjustment(adjustment);
 	set_numeric(false);
+	signal_event_after().connect(sigc::mem_fun(*this, &Widget_Distance::after_event));
 }
 
 Widget_Distance::~Widget_Distance()
@@ -99,6 +100,14 @@ Widget_Distance::on_key_release_event(GdkEventKey* event)
 	SYNFIG_EXCEPTION_GUARD_END_BOOL(true)
 }
 
+void
+Widget_Distance::after_event(GdkEvent *event)
+{
+	SYNFIG_EXCEPTION_GUARD_BEGIN()
+	if (event->type == GDK_BUTTON_RELEASE)
+		select_region(0, get_text_length()-2);
+	SYNFIG_EXCEPTION_GUARD_END()
+}
 
 void
 Widget_Distance::set_value(const synfig::Distance &data)
--- a/synfig-studio/src/gui/widgets/widget_distance.h
+++ b/synfig-studio/src/gui/widgets/widget_distance.h
@@ -55,6 +55,7 @@ protected:
 
 	bool on_key_press_event(GdkEventKey* event);
 	bool on_key_release_event(GdkEventKey* event);
+	void after_event(GdkEvent *event);
 
 public:
 	void set_value(const synfig::Distance &data);


ice0:
Now try to select a part of the value with the mouse.

mohamedAdhamc:
> Now try to select a part of the value with the mouse.

oh yes now that seems like a problem, I didn't think about it much as the zoom dial is already functioning like that, but ok I'll look into how to handle that.

mohamedAdhamc:
ok, so now all the tool palettes spinbuttons/entries have this new selection on click. I think now pretty much everywhere in the main windows have this now other than maybe a few spots such as the two entries above, and below the zoom dial, etc.. 

So should I make this global as in also extend to everywhere in the main menu or is it bettter to keep this PR as is... only for the tool palettes?

mohamedAdhamc:
Hmm it was building fine on my machine, must have forgot something when committing, will recheck

synfig-studio/src/gui/widgets/widget_distance.cpp
```diff
@@ -99,6 +100,27 @@ Widget_Distance::on_key_release_event(GdkEventKey* event)
 	SYNFIG_EXCEPTION_GUARD_END_BOOL(true)
 }
 
+void
+Widget_Distance::after_event(GdkEvent *event)
```
ice0, line 104: 1) As you can see, `Widget_Distance`, `Widget_Entry`, `Widget_Spin` have almost the same code, so it looks like bad practice.
2) Why do you select not all the text, but only up to the first digit?

mohamedAdhamc: Regarding 2, all the digits are selected, it is the units which are not selected.

Regarding 1, well we need to connect to same signals in spin, entry, and distance widgets. The only other possible alternative I can think of is to connect instances each time (which would be really bad of course ).

ice0: > Regarding 2, all the digits are selected


https://user-images.githubusercontent.com/5604544/194246772-e8bfc7cb-e9a7-4029-8441-5d96acec3ce9.mp4

> Should the units be selected as well ?

I think yes. Units can be selected too.

mohamedAdhamc: Ok then I"ll just make it select all.

P.s. I thought it did select all digits I remember it working when I tried but maybe the zeros after the digits were an unhandled case anyways, that's not needed now if all should be selected :).

mohamedAdhamc: alright, @ice0 now it selects all. so point one should be done :)

yoSachinkr: @mohamedAdhamc It'd have been amazing if "select all" is enabled in all text fields by default.

https://user-images.githubusercontent.com/50204208/211195419-a5cd88fc-e39e-43f9-b47b-19f6b04d6c69.mp4

ice0:
@mohamedAdhamc 
Can you rebase it, please?

mohamedAdhamc:
Oh god it seems I messed up the rebase :/.

mohamedAdhamc:
Ok, @ice0 I believe now it is rebased and up to date with master branch. It will probably just need a squash on merge as the commit history is messed up.

PR #2809, opened 2022-08-20
feat: reduced displayed floating point precision

### LINKED ISSUE #1708: UI suggestion: Reduce float precision displayed
I'm using 1.3.16 on Ubuntu 18.04 (via AppImage)

**Issue description**:
I'm looking to start doing a 2D animation series and I've been trying to pick out some animation software that doesn't break the bank.  So far I'm looking at Synfig, but I think there are some things that can be done to improve the usability.  As for this ticket:

**When displaying a floating point, don't show the exact number**

Take a look at this:
synfig_precision

For `Z Depth` and `Opacity`, I think showing `1` or `1.0` would suffice, instead of showing my six digits of precision.  Only if a user has entered, say `1.555`, it should then show all of those digits.

The value for `Transformation` is interpolated.  I also don't think that showing such precision is necessary.  Maybe showing a truncated/rounded version would be fine, but then if a user clicks (or hovers over) the field/value, it's true decimal value will be shown.

IIRC, other editing tools that I use (such as DaVinci Resolve) typically show an integer, but support floats if you want them, and sometimes allow up to three decimals of precision.

mohamedAdhamc:
Fixes #1708 

Not really sure if this is a "feat" or a "fix"... anyway, what I have done is basically just remove all the unnecessary trailing zeroes so that the "exact precision" is there when needed, but when not used it is not of hindrance to the user e.g. 1.000000, 0.340000 , etc...

What's left: -reducing the precision in the layers panel, for the z-depth to be exact.

### DIFF
--- a/synfig-core/src/synfig/distance.cpp
+++ b/synfig-core/src/synfig/distance.cpp
@@ -114,9 +114,7 @@ synfig::String
 Distance::get_string(int digits)const
 {
 	digits=synfig::clamp(digits,0,9);
-	String fmt(strprintf("%%.%01df",digits));
-	String str(strprintf(fmt.c_str(),value_));
-	return strprintf("%s%s",remove_trailing_zeroes(str).c_str(),system_name(system_).c_str());
+	return synfig::float_presentation(value_, digits) + system_name(system_);
 }
 
 void
--- a/synfig-core/src/synfig/distance.cpp
+++ b/synfig-core/src/synfig/distance.cpp
@@ -111,16 +111,12 @@ Distance::operator=(const synfig::String& str)
 }
 
 synfig::String
-Distance::get_string(int digits, bool remove_trailing_zeroes_, bool force_decimal_point, bool has_unit)const
+Distance::get_string(int digits)const
 {
 	digits=synfig::clamp(digits,0,9);
 	String fmt(strprintf("%%.%01df",digits));
 	String str(strprintf(fmt.c_str(),value_));
-	if(remove_trailing_zeroes_){
-		return remove_trailing_zeroes(strprintf("%s%s",str.c_str(),system_name(system_).c_str()),
-									   force_decimal_point, has_unit);
-	} else
-	return strprintf("%s%s",str.c_str(),system_name(system_).c_str());
+	return strprintf("%s%s",remove_trailing_zeroes(str).c_str(),system_name(system_).c_str());
 }
 
 void
--- a/synfig-core/src/synfig/distance.cpp
+++ b/synfig-core/src/synfig/distance.cpp
@@ -111,11 +111,15 @@ Distance::operator=(const synfig::String& str)
 }
 
 synfig::String
-Distance::get_string(int digits)const
+Distance::get_string(int digits, bool remove_trailing_zeroes_, bool force_decimal_point, bool has_unit)const
 {
 	digits=synfig::clamp(digits,0,9);
 	String fmt(strprintf("%%.%01df",digits));
 	String str(strprintf(fmt.c_str(),value_));
+	if(remove_trailing_zeroes_){
+		return remove_trailing_zeroes(strprintf("%s%s",str.c_str(),system_name(system_).c_str()),
+									   force_decimal_point, has_unit);
+	} else
 	return strprintf("%s%s",str.c_str(),system_name(system_).c_str());
 }
 
--- a/synfig-core/src/synfig/string_helper.cpp
+++ b/synfig-core/src/synfig/string_helper.cpp
@@ -74,8 +74,8 @@ synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point
 		else
 			result.pop_back();
 	} else {
-			const size_t last_non_zero_pos = result.find_last_not_of('0');
-			result = result.substr(0, std::max(decimal_point_pos+1, last_non_zero_pos) + 1);
+		const size_t last_non_zero_pos = result.find_last_not_of('0');
+		result = result.substr(0, std::max(decimal_point_pos+1, last_non_zero_pos) + 1);
 	}
 	return result;
 }
--- a/synfig-core/src/synfig/string_helper.cpp
+++ b/synfig-core/src/synfig/string_helper.cpp
@@ -39,7 +39,6 @@
 
 #include <algorithm>
 #include <locale>
-#include <iostream>
 
 #include "general.h"
 #endif
--- a/synfig-core/src/synfig/string_helper.cpp
+++ b/synfig-core/src/synfig/string_helper.cpp
@@ -39,6 +39,7 @@
 
 #include <algorithm>
 #include <locale>
+#include <iostream>
 
 #include "general.h"
 #endif
@@ -59,7 +60,7 @@ get_locale_decimal_point()
 }
 
 std::string
-synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point)
+synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point, bool has_unit)
 {
 	std::string result(text);
 	auto decimal_point = get_locale_decimal_point();
@@ -74,8 +75,14 @@ synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point
 		else
 			result.pop_back();
 	} else {
-		const size_t last_non_zero_pos = result.find_last_not_of('0');
-		result = result.substr(0, std::max(decimal_point_pos+1, last_non_zero_pos) + 1);
+		if (has_unit) {
+			const size_t last_non_zero_pos = result.find_last_of("123456789.");
+			const size_t first_unit_pos = result.find_first_not_of("0123456789.-");
+			result = result.substr(0, std::max(decimal_point_pos+1, last_non_zero_pos) + 1) + result.substr(first_unit_pos);
+		} else {
+			const size_t last_non_zero_pos = result.find_last_not_of('0');
+			result = result.substr(0, std::max(decimal_point_pos+1, last_non_zero_pos) + 1);
+		}
 	}
 	return result;
 }
--- a/synfig-core/src/synfig/string_helper.cpp
+++ b/synfig-core/src/synfig/string_helper.cpp
@@ -76,7 +76,9 @@ synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point
 	} else {
 		if (has_unit) {
 			const size_t last_non_zero_pos = result.find_last_of("123456789.");
-			const size_t first_unit_pos = result.find_first_not_of("0123456789.-");
+			std::string str = "0123456789-";
+			str.push_back(decimal_point);
+			const size_t first_unit_pos = result.find_first_not_of(str);
 			result = result.substr(0, std::max(decimal_point_pos+1, last_non_zero_pos) + 1) + result.substr(first_unit_pos);
 		} else {
 			const size_t last_non_zero_pos = result.find_last_not_of('0');
--- a/synfig-core/src/synfig/string_helper.cpp
+++ b/synfig-core/src/synfig/string_helper.cpp
@@ -59,7 +59,7 @@ get_locale_decimal_point()
 }
 
 std::string
-synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point, bool has_unit)
+synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point)
 {
 	std::string result(text);
 	auto decimal_point = get_locale_decimal_point();
@@ -74,16 +74,8 @@ synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point
 		else
 			result.pop_back();
 	} else {
-		if (has_unit) {
-			const size_t last_non_zero_pos = result.find_last_of("123456789.");
-			std::string str = "0123456789-";
-			str.push_back(decimal_point);
-			const size_t first_unit_pos = result.find_first_not_of(str);
-			result = result.substr(0, std::max(decimal_point_pos+1, last_non_zero_pos) + 1) + result.substr(first_unit_pos);
-		} else {
 			const size_t last_non_zero_pos = result.find_last_not_of('0');
 			result = result.substr(0, std::max(decimal_point_pos+1, last_non_zero_pos) + 1);
-		}
 	}
 	return result;
 }
--- a/synfig-core/src/synfig/string_helper.cpp
+++ b/synfig-core/src/synfig/string_helper.cpp
@@ -81,6 +81,14 @@ synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point
 }
 
 std::string
+synfig::float_presentation(Real value, int num_decimals)
+{
+	std::string format = strprintf("%%.%df", num_decimals);
+	std::string text = remove_trailing_zeroes(strprintf(format.c_str(), value));
+	return text;
+}
+
+std::string
 synfig::trim(const std::string& text)
 {
 	return right_trim(left_trim(text));
--- a/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
+++ b/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
@@ -351,7 +351,8 @@ CellRenderer_ValueBase::render_vfunc(
 	{
 		property_text() =
 			data.get(Time()).get_string( get_canvas()->rend_desc().get_frame_rate(),
-														 App::get_time_format());
+				                                         App::get_time_format());
+
 	}
 	else
 	if (type == type_angle)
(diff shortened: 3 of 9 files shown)


synfig-studio/src/gui/trees/layertreestore.cpp
```diff
@@ -199,8 +199,10 @@ LayerTreeStore::get_value_vfunc(const Gtk::TreeModel::iterator& iter, int column
 			if (column == model.index.index())
 				set_gvalue_tpl<int>(value, layer->get_depth());
 			else
-			if (column == model.z_depth.index())
-				set_gvalue_tpl<float>(value, layer->get_true_z_depth(canvas_interface()->get_time()));
+			if (column == model.z_depth.index()){
+				Glib::ustring str = remove_trailing_zeroes(std::to_string(layer->get_true_z_depth(canvas_interface()->get_time())));
+				set_gvalue_tpl<Glib::ustring>(value, str);//the value itself doesnt have trailing zeroes but seems like the renderer does it
```
ice0, line 204: > the value itself doesnt have trailing zeroes but seems like the renderer does it

Please check `std::to_string` documentation.

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
```diff
@@ -402,10 +402,20 @@ CellRenderer_ValueBase::render_vfunc(
 			Distance x( vector[0], Distance::SYSTEM_UNITS ), y( vector[1], Distance::SYSTEM_UNITS );
 			x.convert( App::distance_system, get_canvas()->rend_desc() );
 			y.convert( App::distance_system, get_canvas()->rend_desc() );
-			property_text() = strprintf("%s,%s", x.get_string(real_num_decimals).c_str(), y.get_string(real_num_decimals).c_str());
+			std::string text = strprintf("%s,%s", x.get_string(real_num_decimals).c_str(), y.get_string(real_num_decimals).c_str());
```
ice0, line 405: It looks like it's better to add a flag to `Distance::get_string(int digits, bool remove_trailing_zeroes)`

synfig-studio/src/gui/trees/layertreestore.h
```diff
@@ -85,7 +85,7 @@ class LayerTreeStore : public Gtk::TreeStore
 
 		Gtk::TreeModelColumn<bool>						children_lock;
 
-		Gtk::TreeModelColumn<float> z_depth;
```
ice0, line 88: Better to use `Gtk::TreeViewColumn::set_cell_data_func` to change only column presentation, not value.
https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1TreeViewColumn.html#a3469e1adf42e5932ea123ec33e4ce4e1

synfig-studio/src/gui/trees/layertreestore.cpp
mohamedAdhamc: oh yes, I did check it. this was an old comment before I switched the float into a string, but i'm guessing the same still applies. the CellRendererText renders the values as some form of string which is probably why this all happens.

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
mohamedAdhamc: sounds like a good idea

synfig-studio/src/gui/trees/layertreestore.h
mohamedAdhamc: sounds like a safer approach, ok I'm looking into how to implement it.

mohamedAdhamc: done

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
mohamedAdhamc: done

mohamedAdhamc:
Oops looks like I forgot the address operator when connecting to the cell render function(wonder how it was building locally though xD, maybe it's GitHub desktop thing).

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
```diff
@@ -338,20 +338,20 @@ CellRenderer_ValueBase::render_vfunc(
...
-			property_text() = strprintf(format.c_str(), data.get(Real()));
+			property_text() = remove_trailing_zeroes(strprintf(format.c_str(), data.get(Real())));
 		}
 	}
 	else
 	if (type == type_time)
 	{
 		property_text() =
 			data.get(Time()).get_string( get_canvas()->rend_desc().get_frame_rate(),
-				                                         App::get_time_format());
```
ice0, line 354: I suggest to revert this change.

synfig-studio/src/gui/trees/layertree.cpp
```diff
@@ -392,6 +399,13 @@ LayerTree::create_param_tree()
 	//column_time_track->set_visible(false);
 }
 
+void
+LayerTree::set_z_depth_cell_data(Gtk::CellRenderer *cell, const Gtk::TreeIter& it)
```
ice0, line 403: Circle parameter `Feather` currently displays incorrectly:

Before:
Screenshot_92

Now:
Screenshot_91

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
mohamedAdhamc: oh yes, definitely I moved it around by fault.

synfig-core/src/synfig/string_helper.cpp
```diff
@@ -59,7 +59,7 @@ get_locale_decimal_point()
 }
 
 std::string
-synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point)
+synfig::remove_trailing_zeroes(const std::string& text, bool force_decimal_point, bool has_unit)
```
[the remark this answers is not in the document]
mohamedAdhamc: yes for any string we send which has a unit at the end, the original implementation didn't account for it. I mean we could parse the string to remove the unit before sending and then re-add it (but it may become redundant), also I feel this may be good as everywhere I seen it the unit was part of the string. 

if there's something i'm not noticing here though please let me know :)

synfig-studio/src/gui/trees/layertree.cpp
mohamedAdhamc: oh, I didnt know that the decimal point could be a comma (and now I'm assuming maybe also others), I'll look into it

synfig-studio/src/gui/trees/layertree.cpp
```diff
@@ -194,16 +194,23 @@ LayerTree::create_layer_tree()
 		layer_tree_view().append_column(*column);
 	}
 	{	// --- Z D E P T H ----------------------------------------------------
-		int index;
-		index=layer_tree_view().append_column(_("Z Depth"),layer_model.z_depth);
-		// Set up the Z-Depth label cell-renderer
+		column_z_depth = Gtk::manage( new Gtk::TreeView::Column(_("Z Depth")) );
```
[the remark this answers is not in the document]
mohamedAdhamc: done

synfig-studio/src/gui/trees/layertree.cpp
mohamedAdhamc: ok, I think now it should be fixed, instead of checking for '.' as the decimal point i used the static function used to return the decimal character, but I couldn't test it as I cant find where to change the deximal point (*_*), will update if i find it.

synfig-core/src/synfig/string_helper.cpp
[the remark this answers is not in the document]
ice0: > if there's something i'm not noticing here though please let me know :)

As I see it in the current code - all values are floating point/double converted to string, so they don't have a trailing part.

synfig-studio/src/gui/trees/layertree.cpp
```diff
@@ -195,15 +195,19 @@ LayerTree::create_layer_tree()
...
-		index=layer_tree_view().append_column(_("Z Depth"),layer_model.z_depth);
-		// Set up the Z-Depth label cell-renderer
-
-		column_z_depth=layer_tree_view().get_column(index-1);
-		column_z_depth->set_reorderable();
-		column_z_depth->set_resizable();
-		column_z_depth->set_clickable();
-
-		column_z_depth->set_sort_column(layer_model.z_depth);
+			index=layer_tree_view().append_column(_("Z Depth"),layer_model.z_depth);
```
ice0, line 198: Indent

synfig-core/src/synfig/string_helper.cpp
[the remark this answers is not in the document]
mohamedAdhamc: > > 
> so they don't have a trailing part.

How ? 
example: in `cell_renderervalue.cpp`
`x.get_string(real_num_decimals).c_str();` yields out trailing zeroes in the string which have to be removed using the `remove_trailing_zeroes()` method -and also the has_unit parameter has to be set to true for it to correct remove the zeroes-. And this is so the case for the value of type real which doesnt have a unit and all the other float to string types.

synfig-studio/src/gui/trees/layertree.cpp
mohamedAdhamc: done

synfig-core/src/synfig/string_helper.cpp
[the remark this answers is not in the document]
ice0: > example: in `cell_renderervalue.cpp` line 341

```
Distance x( data.get(Real()), Distance::SYSTEM_UNITS);
x.convert( App::distance_system, get_canvas()->rend_desc() );
property_text() = x.get_string(real_num_decimals,true,true,true).c_str();
```

Distance contains float value without units.
```
Distance::get_string(int digits, bool remove_trailing_zeroes)const
{
        ...
	return strprintf("%s%s",str.c_str(),system_name(system_).c_str()); <- units added here
}
```
so trailing zeroes should be already removed.

synfig-core/src/synfig/distance.cpp
```diff
@@ -111,11 +111,15 @@ Distance::operator=(const synfig::String& str)
...
 
 synfig::String
-Distance::get_string(int digits)const
+Distance::get_string(int digits, bool remove_trailing_zeroes_, bool force_decimal_point, bool has_unit)const
 {
 	digits=synfig::clamp(digits,0,9);
 	String fmt(strprintf("%%.%01df",digits));
 	String str(strprintf(fmt.c_str(),value_));
+	if(remove_trailing_zeroes_){
+		return remove_trailing_zeroes(strprintf("%s%s",str.c_str(),system_name(system_).c_str()),
```
ice0, line 120: You should remove trailing zeroes **before** adding `system_name(...)`

synfig-core/src/synfig/distance.cpp
```diff
@@ -111,11 +111,15 @@ Distance::operator=(const synfig::String& str)
 }
 
 synfig::String
-Distance::get_string(int digits)const
+Distance::get_string(int digits, bool remove_trailing_zeroes_, bool force_decimal_point, bool has_unit)const
```
ice0, line 114: > remove_trailing_zeroes_

Underscore in suffix usually used in private members, so better not use it for arguments.

synfig-core/src/synfig/distance.cpp
mohamedAdhamc: oh, I get what you mean now. yes makes sense.

synfig-core/src/synfig/distance.cpp
mohamedAdhamc: I did so to avoid the naming conflict with the remove_trailing_zeroes() method, I guess I could just rename the variable though.

synfig-core/src/synfig/string_helper.cpp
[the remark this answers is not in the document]
mohamedAdhamc: oh ok now I get it you mean we should remove the trailing zeroes before the units are even added which in turn removes the need for has_unit. will do.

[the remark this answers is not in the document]
mohamedAdhamc: done

synfig-core/src/synfig/distance.cpp
mohamedAdhamc: done

synfig-core/src/synfig/distance.cpp
mohamedAdhamc: removed altogether as trailing zeroes shouldn't be needed anyway.

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
```diff
@@ -402,10 +402,17 @@ CellRenderer_ValueBase::render_vfunc(
...
 			x.convert( App::distance_system, get_canvas()->rend_desc() );
 			y.convert( App::distance_system, get_canvas()->rend_desc() );
-			property_text() = strprintf("%s,%s", x.get_string(real_num_decimals).c_str(), y.get_string(real_num_decimals).c_str());
+			std::string text = strprintf("%s,%s", x.get_string(real_num_decimals).c_str(),
+												  y.get_string(real_num_decimals).c_str());
+			property_text() = text;
 		} else {
 			std::string format = strprintf("%%.%01df,%%.%01df", real_num_decimals, real_num_decimals);
-			property_text() = strprintf(format.c_str(), vector[0], vector[1]);
+			std::string text = strprintf(format.c_str(), vector[0], vector[1]);
```
ice0, line 410: 1) Avoid using keywords such as `final` as variable names.
2) This part can also be simplified. 
3) After completing step 2, check the changes again. You should notice a common part in all of them. I suggest to extract this part to the new function `std::string float_presentation(Real value, int num_decimals = real_num_decimals)`. This will allow you to move all formatting in one place.

ice0: What about 3 step?

ice0: > I was thinking about how to best implement it but Im still not sure how best to do it.

Step 2 can be done in a different way, this should lead you to a better solution.

mohamedAdhamc: @ice0, how about now ? 

p.s. the formatting for the vector was a bit diff than the real it was `%%.%01d` vs `%%.%df` . I'm not sure what the 01 does but from comparison it seems like it added an extra decimal place so I think its probably not needed anyways, if not though we cold add a format argument to `float_presentation` what do you think ?

ice0: Awesome!
This is exactly what I mean.

Next step is to check other places with `%%.%df` expression. For example:
```
if (type == type_angle)
{
	const std::string angle_format = strprintf("%%.%df°", angle_num_decimals);
	property_text() = strprintf( angle_format.c_str(), (Real) Angle::deg( data.get(Angle()) ).get() );
```

```
if (type == type_angle)
{
	property_text() = float_presentation(Angle::deg( data.get(Angle()) ).get(), angle_num_decimals) + "°";
```

>  I'm not sure what the 01 does

Same for me. 
@rodolforg Any ideas here?

mohamedAdhamc: perfect, ok sounds good will do

mohamedAdhamc: done, it seems like it was only there type_angle and transformation but for type_transformation it has two strings before and after so I don't think it would be any good to split it up to three parts.

synfig-core/src/synfig/distance.cpp
```diff
@@ -116,7 +116,7 @@ Distance::get_string(int digits)const
 	digits=synfig::clamp(digits,0,9);
 	String fmt(strprintf("%%.%01df",digits));
 	String str(strprintf(fmt.c_str(),value_));
-	return strprintf("%s%s",str.c_str(),system_name(system_).c_str());
+	return strprintf("%s%s",remove_trailing_zeroes(str).c_str(),system_name(system_).c_str());
```
ice0, line 119: Looks like this can be simplified:
```
return synfig::float_presentation(value_, digits) + system_name(system_);
```

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
ice0: My suggestion:
```
property_text() = x.get_string(real_num_decimals) + "," + y.get_string(real_num_decimals) +
		float_presentation(angle.get(), angle_num_decimals) + "°" +
		sx.get_string(real_num_decimals) + "," + sy.get_string(real_num_decimals);
```

synfig-core/src/synfig/distance.cpp
mohamedAdhamc: done

mohamedAdhamc:
@ice0, done. Is anything else needed here?

ice0 (approved):

ice0:
Merged. Thank you!

mohamedAdhamc:
> Merged. Thank you!

Great, hopefully many more to come :)

PR #2810, opened 2022-08-24
feat: Added support for scrolling through text in text paragraph dialog

mohamedAdhamc:
before:

https://user-images.githubusercontent.com/100296264/186509747-ac2fdcf4-c8d9-4640-b8d2-006553db931c.mp4

After:

https://user-images.githubusercontent.com/100296264/186509792-8eb58113-f5e8-4531-8db7-b08615ce9d78.mp4

One thing I'm not sure about though... with the current implementation if the user manually resized the dialog to a greater vertical size, if he then deletes a line then the dialog is auto-adjusted to the size fitting to the number of lines he has, is this okay? or should it be that if it's manually resized the size of the resize is the new minimum size?

Also I just thought of this, should this auto shrink be only when we are deleting the last line?

p.s. excuse the horrendous video quality I had to compress to upload xD.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -3852,6 +3852,18 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
 		SYNFIG_EXCEPTION_GUARD_END_BOOL(true)
 
 	}), false );
+
+	int previous_num_lines(0);
+
+	text_buffer->signal_changed().connect(sigc::track_obj([&dialog,&text_buffer,&previous_num_lines](){
+			int num_lines=text_buffer->get_line_count();
+			if(num_lines<previous_num_lines)
+				dialog.resize(dialog.get_width(), 2);//any size less than minimum causes a resize to minimum possible size
+
+			previous_num_lines=num_lines;
+
+		}, dialog));
+
 	//text_entry.signal_activate().connect(sigc::bind(sigc::mem_fun(dialog,&Gtk::Dialog::response),Gtk::RESPONSE_OK));
 	dialog.show();
 
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -70,6 +70,7 @@
 #endif
 #include <gtkmm/stock.h>
 #include <gtkmm/textview.h>
+#include <gtkmm/scrolledwindow.h>
 
 #include <gui/app.h>
 #include <gui/autorecover.h>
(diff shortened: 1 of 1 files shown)


ice0:
> One thing I'm not sure about though... with the current implementation if the user manually resized the dialog

Can we detect manual resizing?

mohamedAdhamc:
> 
> Can we detect manual resizing?

not in a direct way, but if needed I think we can use the button_press,button_release, and check_resize signal to get insight onto whether the resize was by mouse (manual). i.e. if between the button press and release there was a resize then this was a manual resize. And to be even more sure we could even make sure that during this no key was pressed.

ice0:
> > Can we detect manual resizing?
> 
> not in a direct way, but if needed I think we can use the button_press,button_release, and check_resize signal to get insight onto whether the resize was by mouse (manual). i.e. if between the button press and release there was a resize then this was a manual resize. And to be even more sure we could even make sure that during this no key was pressed.

If you notice that something is starting to get overcomplicated (like here), then you may just be moving in the wrong direction.
Please check `Gtk::ScrolledWindow` widget.

mohamedAdhamc:
> If you notice that something is starting to get overcomplicated (like here), then you may just be moving in the wrong direction. Please check `Gtk::ScrolledWindow` widget.

I agree, also yes I think I saw scrolled windows being used with text-views so yes I think it'll probably make sense, I'll look into it.

mohamedAdhamc:
@ice0  done.

synfig-studio/src/gui/app.cpp
```diff
@@ -3826,14 +3827,17 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
 
 	Gtk::Label* label = manage(new Gtk::Label(message));
 	label->show();
-	dialog.get_content_area()->pack_start(*label);
+	dialog.get_content_area()->pack_start(*label, false, false);
```
ice0, line 3830: I suggest to use `pack_start(Widget& child, PackOptions options = PACK_EXPAND_WIDGET, guint padding = 0)` override.
It will be more clear what is happening here.

synfig-studio/src/gui/app.cpp
```diff
@@ -3826,14 +3827,17 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
...
 
 	Glib::RefPtr<Gtk::TextBuffer> text_buffer(Gtk::TextBuffer::create());
 	text_buffer->set_text(text);
 	Gtk::TextView text_view(text_buffer);
 	text_view.show();
+	Gtk::ScrolledWindow s_window;
```
ice0, line 3836: Better to rename `s_window` -> `scrolled_window`

synfig-studio/src/gui/app.cpp
```diff
@@ -3826,14 +3827,17 @@ App::dialog_paragraph(const std::string &title, const std::string &message,std::
...
+	s_window.add(text_view);
+	s_window.show();
 
-	dialog.get_content_area()->pack_start(text_view);
+	dialog.get_content_area()->pack_start(s_window);
```
ice0, line 3840: I suggest to increase default window size:

Current:
Screenshot_98

Suggested:
Screenshot_97

synfig-studio/src/gui/app.cpp
mohamedAdhamc: done

synfig-studio/src/gui/app.cpp
mohamedAdhamc: done

mohamedAdhamc:
@ice0  done, this is the new default size

new def size

ice0 (approved):

ice0:
Merged. Thank you!

mohamedAdhamc:
Great! hopefully many more to come :)

PR #2816, opened 2022-09-04
build(vcpkg): `cmake --install` now also installs required dependencies

hesham-essam:
This PR:
1. Creates a separate file `SynfigVcpkg` for properly packaging synfig and its dependencies using vcpkg.
2. Copies synfig's dependencies when installing using cmake.
3. Removes msvc-fonts directory and uses the default configuration files built with fontconfig.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2819, opened 2022-09-08
fix: Synfig CLI does not open file if path contains non-Latin characters (Windows)

ice0:
The problem was double conversion of command line options from local codepage to UTF-8.

Arguments was already converted in main()

### DIFF
--- a/synfig-core/src/tool/optionsprocessor.cpp
+++ b/synfig-core/src/tool/optionsprocessor.cpp
@@ -241,7 +241,11 @@ bool SynfigCommandLineParser::parse(int argc, char* argv[])
 #ifdef GLIBMM_EXCEPTIONS_ENABLED
 	try
 	{
+#ifdef _WIN32
+		context.parse(argv);
+#else
 		context.parse(argc, argv);
+#endif
 	}
 	catch(const Glib::Error& ex)
 	{


rodolforg:
Could MacOS and Linux use this version without `argc` too?

ice0:
> Could MacOS and Linux use this version without `argc` too?

It fails in `parse` with error `free(): invalid pointer`, when using it on standard `argv`.

I think this happens because of this:
> This function is similar to [parse(int& argc, char**& argv)](https://developer-old.gnome.org/glibmm/stable/classGlib_1_1OptionContext.html#a08ab225e3a4be10364f32948a6532726) except that it respects the normal memory rules when dealing with a strv instead of assuming that the passed-in array is the argv of the main function. In particular, strings that are removed from the arguments list will be freed using g_free().

https://developer-old.gnome.org/glibmm/stable/classGlib_1_1OptionContext.html#a04b835ea562e10f28f5a2c6fda7637c3

Also, I'm not sure about the encoding that is used on Linux/macOS, so I just left it as is.

> On Unixes, the argv has no fixed encoding:
>
> a) File names inserted by tab-completion/globbing will occur in argv verbatim as exactly the byte sequences by which they are named on disk. This is true even if those byte sequences make no sense in the current locale.
>
> b) Input entered directly by the user using their IME will occur in argv in the locale encoding. (Ubuntu seems to use LOCALE to decide how to encode IME input, whereas OS X uses the Terminal.app encoding Preference.)

https://stackoverflow.com/questions/5408730/what-is-the-encoding-of-argv

rodolforg (approved):

PR #2820, opened 2022-09-08
feat: ico format added to `mod_imagemagick` list of exported formats

ice0:


rodolforg (approved):
It seems ffmpeg can do it too :)

PR #2821, opened 2022-09-08
build(cmake): fixed build when path contains non-Latin characters (MSYS2)

ice0:
MSYS2/Windows only.
The problem occurred when trying to compile Windows resources using the `winres.exe` tool.

If the path to an icon in a resource contains non-Latin characters, then winres.exe cannot
compile this resource. Solved by using a relative path to the icon.

### DIFF
--- a/synfig-studio/src/gui/CMakeLists.txt
+++ b/synfig-studio/src/gui/CMakeLists.txt
@@ -30,8 +30,9 @@ endif()
 ##
 
 if(WIN32)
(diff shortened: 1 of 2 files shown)


rodolforg:
Just for information: will it broke Autotools building for Windows?

ice0:
> Just for information: will it break Autotools building for Windows?

I didn't find any mention of a `.rc` file in the autotools scripts.
Also unpacking current release archive shows what binary doesn't have embedded icon and app information.
So I think no, it doesn't break autotools.

rodolforg (approved):

PR #2823, opened 2022-09-09
fix: "rule of two" : copy constructor and copy assignment operator

rodolforg:
And copy assignment operator returns reference, not a copy.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2825, opened 2022-09-10
build(vcpkg): Support building synfig using vcpkg on linux

hesham-essam:
This PR adds vcpkg support for linux. Changes include:
1. Adding ports for at-spi2-core and at-spi2-atk since they're required for gtk on Linux. These are the same from the PR https://github.com/microsoft/vcpkg/pull/24934, and should be removed once this PR is merged into vcpkg (not merged until now because of problems regarding CI)
2. Adding `glib` port to fix pkgconfig variable issue. I submitted these PRs to fix this, and once they're merged both the glib port and gdk-pixbuf port should be removed: https://github.com/microsoft/vcpkg/pull/26744 https://github.com/microsoft/vcpkg/pull/26743
3. Added `x64-linux-synfig` (probably a bad name) triplet file inside new folder `vcpkg-triplets`. Triplets are the recommended way for adding per port customization regarding build type.
4. Added a python script (appdeps.py) that mimics what vcpkg's `applocal.ps1` script does on Windows. i.e. it copies all dependencies of a library to enable portability.

To build, you do something like:
```
> cmake -DCMAKE_TOOLCHAIN_FILE=<vcpkg-dir>scripts/buildsystems/vcpkg.cmake -DVCPKG_INSTALLED_DIR=$(pwd)/../../synfig_vcpkg_installed -DCMAKE_BUILD_TYPE=Release -DVCPKG_OVERLAY_PORTS=$(pwd)/../vcpkg-ports  -DVCPKG_OVERLAY_TRIPLETS=$(pwd)/../vcpkg-triplets -DVCPKG_TARGET_TRIPLET="x64-linux-synfig" -DCMAKE_INSTALL_PREFIX=$HOME/synfig ..
> make -j8
> make install
```

Some dependencies require some system libraries, so they will need to be installed as needed.

Similar to windows, installation will create a portable synfig package, although it won't be truly portable in contrast to the case with Windows because of the dependencies on these system libraries.

This probably will be my last PR in my GSoC preiod. I will work on conan support afterwards.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,11 @@ set(STUDIO_VERSION_PATCH 0)
 
 option(ENABLE_TCMALLOC "Enable use of tcmalloc library" OFF)
 
+if(VCPKG_TOOLCHAIN)
+	set(CMAKE_INSTALL_RPATH "\$ORIGIN/../lib")
+	add_definitions(-DSYNFIG_PORTABLE)
+endif()
+
 if(NOT CMAKE_BUILD_TYPE)
 	set(CMAKE_BUILD_TYPE Debug)
 endif()
--- a/autobuild/appdeps.py
+++ b/autobuild/appdeps.py
@@ -1,0 +1,135 @@
+import argparse
+import subprocess
+import shutil
+import os
+
+
+def _change_rpath(lib, rpath):
+    val = subprocess.run(args=["patchelf", "--set-rpath", rpath, lib])
+    if val.returncode != 0:
+        print(f"Error: Failed setting rpath of {lib} to {rpath}!")
+        exit(-1)
+
+
+def change_rpath(libs, rpath):
+    for lib in libs:
+        _change_rpath(lib, rpath)
+
+
+def is_executable_or_shared(file):
+    val = subprocess.run(args=["file", "-L", "--mime-type", file],
+                         capture_output=True)
+    output = val.stdout.decode("utf-8")
+
+    mime_types = [
+        "application/x-sharedlib", "application/x-executable",
+        "application/x-pie-executable"
+    ]
+    for mime_type in mime_types:
+        if mime_type in output:
+            return True
+
+    return False
+
+
+def find_deps(file, libdirs):
+    val = subprocess.run(args=["ldd", file],
+                         env={"LD_LIBRARY_PATH": ":".join(libdirs)},
+                         capture_output=True)
+
+    output = val.stdout.decode("utf-8")
+
+    deps = {}
+    for line in output.splitlines():
+        # ldd writes lines in the form of "<lib-name> => <lib-path> (<addr>)"
+        # or in the form "<lib-name> (<addr>)"
+        # we're only interested in the first form
+        parts = line.split("=>")
+        if len(parts) == 1:
+            continue
+
+        lib_name = parts[0].strip()
+        lib_path = parts[1].split(" (")[0].strip()
+        if lib_path == "not found":
+            lib_path = None
+
+        deps[lib_name] = lib_path
+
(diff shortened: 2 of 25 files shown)


vcpkg-ports/imagemagick/portfile.cmake
```diff
@@ -30,13 +32,46 @@ vcpkg_configure_make(
...
+    --without-perl
+    --without-raqm
+    --without-tiff
+    --without-webp
+    --without-wmf
+    --without-raw # libraw exists but it has a bug with its pkgconfig file
     --disable-docs
 )
 
+# FIXME
```
ice0, line 63: Can you add more detailed explanation here? For FIXME.

CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ set(STUDIO_VERSION_PATCH 0)
 
 option(ENABLE_TCMALLOC "Enable use of tcmalloc library" OFF)
 
+if(VCPKG_TOOLCHAIN)
+	set(CMAKE_INSTALL_RPATH "\$ORIGIN/../lib")
+	add_definitions(-DSYNFIG_PORTABLE)
```
ice0, line 17: Maybe move `include(SynfigVcpkg)` here too?

hesham-essam: I was hesitant to do that at first because it will require me to monkey patch `add_library` and `install` cmake functions. Adding it at the end of the file has the advantage that all the targets have been already defined.

In anyways, I now understand how this is done, so I'm going to do this.

vcpkg-ports/imagemagick/portfile.cmake
hesham-essam: The short story: The proper fix would be to edit glib port to change the rpath of glib's libraries (glib doesn't want me to live in peace :"D)

The long story: Imagemagick depends on Gio (from Glib). Gio privately depends on gmodule (also from Glib). This means that linking against Gio does not need us to link against gmodule (since gio is already linked against gmodule), but instead gmodule has to be in the runtime search path while linking to resolve all symbols. Hence setting the LD_LIBRARY_PATH environment variable.

This isn't a good solution though, since `vcpkg_install_make` builds both debug and build, and the environment variable is only set for release.

hesham-essam:
This builds (and installs) fine locally now on a freshly installed Ubuntu. The following system libraries are required though:
```
git curl nasm bison python3 python3-distutils libx11-dev libxext-dev libxft-dev libegl-dev libdbus-1-dev libxi-dev libxtst-dev libxrandr-dev libgl1-mesa-dev libglu1-mesa-dev mesa-common-dev libxrandr-dev libxxf86vm-dev
```

vcpkg-ports/imagemagick/portfile.cmake
hesham-essam: Vcpkg had the option `VCPKG_FIXUP_ELF_RPATH` which changes the rpath of installed libraries automatically. I used it in our custom triplet and it fixed the issue for me.

CMakeLists.txt
hesham-essam: done

autobuild/appdeps.py
```diff
@@ -0,0 +1,144 @@
+import argparse
```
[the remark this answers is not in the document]
hesham-essam: I made it so that it only copies dependencies in "-L" directories (to avoid copying system dependencies in /usr/lib).
Try something like this
```
python3 ./appdeps.py --outdir ./outd -L<vcpkg-installed-dir>/x64-linux-synfig/lib/ ~/projects/synfig/synfig-master-clean/build/output/Release/bin/synfig
```

PR #2826, opened 2022-09-11
perf: improved file opening speed in Windows

ice0:
getenv is very slow on Windows, so I replaced most of the logging calls with a macro that is disabled by default.

Benchmark test (used a null target to exclude rendering impact):
```
./synfig synfig-core/examples/pirates.sif -o pirates.png -t null
```

All benchmarks made on Release builds.

Windows (MSVC) before:
```
Execution time: 28.352 s
Execution time: 28.345 s
Execution time: 28.391 s
```
Avg: 28362 ms

Windows (MSVC) after:
```
Execution time: 6.140 s
Execution time: 6.341 s
Execution time: 6.201 s
```
Avg: 6227 ms
Improvement: **4.55x faster** (455%)

Windows (MinGW) before:
```
real    0m19.757s
real    0m18.605s
real    0m18.697s
```
Avg: 19019 ms

Windows (MinGW) after:
```
real    0m6.842s
real    0m6.816s
real    0m6.772s
```
Avg: 6810 ms
Improvement: **2,79x faster** (279%)

Linux before:
```
./synfig  -o pirates.png -t null  3,72s user 0,39s system 109% cpu 3,756 total
./synfig  -o pirates.png -t null  3,83s user 0,37s system 109% cpu 3,834 total
./synfig  -o pirates.png -t null  3,71s user 0,42s system 109% cpu 3,778 total
```
Average: 3789 ms

Linux after:
```
./synfig  -o pirates.png -t null  3,62s user 0,38s system 111% cpu 3,583 total
./synfig  -o pirates.png -t null  3,50s user 0,40s system 110% cpu 3,523 total
./synfig  -o pirates.png -t null  3,52s user 0,39s system 108% cpu 3,593 total
```
Average: 3566 ms
Improvement: 0,06x (6%)

macOS before:
```
./synfig synfig-core/examples/pirates.sif -o  -  2,85s user 0,31s system 108% cpu 2,928 total
./synfig synfig-core/examples/pirates.sif -o  -  2,84s user 0,32s system 108% cpu 2,911 total
./synfig synfig-core/examples/pirates.sif -o  -  2,83s user 0,31s system 108% cpu 2,898 total
```
Average: 2912 ms

macOS after:
```
./synfig synfig-core/examples/pirates.sif -o  -  2,57s user 0,31s system 108% cpu 2,642 total
./synfig synfig-core/examples/pirates.sif -o  -  2,57s user 0,31s system 109% cpu 2,642 total
[...]

PR #2828, opened 2022-09-13
revert: getenv is very slow on Windows, so I replaced most of the logging calls with a macro that is disabled by default.

ice0:
Reverts synfig/synfig#2826

Forgot to update the commit message.

Refs: e23b229

### DIFF
--- a/synfig-core/src/synfig/bone.cpp
+++ b/synfig-core/src/synfig/bone.cpp
@@ -63,8 +63,8 @@ Bone::Bone():
 	depth_(0.0),
 	parent_(nullptr)
 {
-	DEBUG_LOG("SYNFIG_DEBUG_NEW_BONES",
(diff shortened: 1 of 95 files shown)


morevnaproject (dismissed):

ice0:
@rodolforg 
Git comment is important for automatic changelog generation.
So I want to revert commit with incorrect message and merge the correct one.
I was thinking about adding an empty commit with a new message, but that's bad for cherry picking.
Maybe you guys have a better idea.

morevnaproject:
@rodolforg The commit message should start with "perf:" (or other tag, to help automatically generate changelog in the future)

rodolforg:
I don't it is worth to add a revert commit and a new one just for a tag.
The history will have 3 commits and the bad one keeps existing.

As we have so few contributors we could do a forced push this time if that's important

ice0:
Agree.

PR #2830, opened 2022-09-16
fix: prevent crash or hang when have problems getting frame on importing

rodolforg:
Problem detected on creating PR #2491, when I wrongly assumed that could redirect stderr to stdout (FFmpeg import does not like it, as it always write to stderr the FFmpeg version info)


ice0 (approved):

ice0:
Merged. Thank you!

PR #2834, opened 2022-09-25
fix: broken opengl build (due to ETL_DIRECTORY_SEPARATOR) and typos

rodolforg:
Since 9c7e20221bd6d2d33db24a90090bf4b05614e58f, `ETL_DIRECTORY_SEPARATOR` is now defined in <ETL/stringf>.
As it will be 'soon' removed anyway, I just changed to `"/"`.

Fixed a couple of typos too.

### DIFF
--- a/synfig-core/src/synfig/rendering/opengl/internal/shaders.cpp
+++ b/synfig-core/src/synfig/rendering/opengl/internal/shaders.cpp
@@ -178,7 +178,7 @@ String
 gl::Shaders::get_shader_path(const String &filename)
 {
 	return get_shader_path()
-		 + ETL_DIRECTORY_SEPARATOR
+		 + "/"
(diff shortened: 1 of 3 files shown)


synfig-core/src/synfig/rendering/opengl/internal/shaders.cpp
```diff
@@ -178,7 +178,7 @@ String
 gl::Shaders::get_shader_path(const String &filename)
 {
 	return get_shader_path()
-		 + ETL_DIRECTORY_SEPARATOR
+		 + "/"
```
ice0, line 181: It is better to use '/'. Just a bit faster.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2837, opened 2022-09-28
build: updated dependencies for Debian 11 based distros

### LINKED ISSUE #2832: 1-setup-linux-native.sh needs to be updated
**Describe the bug**
1-setup-linux-native.sh needs to be updated for recent Debian flavoured distros.
> *python-dev is no longer available* since **Debian 11, Ubuntu 20.10**, and *any distribution derived from those*.
> If you need Python 2 you can try python2-dev instead, but note that Python 2 is no longer supported by the Python community, and many modules are only available for Python 3 in Debian and Ubuntu.
```
See https://unix.stackexchange.com/questions/708493/python-dev-package-installation-error and https://forums.synfig.org/t/having-some-problem-in-building/13456

**To Reproduce**
Just try to run 1-setup-linux-native.sh on one of the indicated distros

**Expected behavior**
Install a compatible package

ice0:
1) We don't need Python development files. We need only python3 binaries for the plugin system.
2) `python3` package is unnecessary, it will be installed as a dependency for `python3-lxml`.

fix #2832

### DIFF
--- a/1-setup-linux-native.sh
+++ b/1-setup-linux-native.sh
@@ -213,8 +213,6 @@ elif [ "$ID_LIKE" == "debian" ] || [ "$ID_LIKE" == "ubuntu" ] || [ "$ID_LIKE" ==
                 libxml2-dev \
                 libtiff5-dev \
                 libmlt-dev libmlt++-dev libmlt-data \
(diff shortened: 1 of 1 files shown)


rodolforg (approved):

rodolforg:
But you kept `python3` instead of only `python3-lxml`

PR #2838, opened 2022-09-28
build(autotools): allow to compile release build with debug info

### LINKED ISSUE #2392: Don’t let `configure` filter out the compiler’s `-g` option to produce debug information
On Void Linux packages are built with gcc’s `-g` option to produce debug information which is then stripped into a separate package.

However, around line 85 in
https://github.com/synfig/synfig/blob/master/ETL/m4/subs.m4
https://github.com/synfig/synfig/blob/master/synfig-core/m4/subs.m4
https://github.com/synfig/synfig/blob/master/synfig-studio/m4/subs.m4
`ETL`, `synfig` and `synfig-studio` filter out that option if a debug build isn’t enabled. That leaves us either to enable debug mode and not have any optimizations or have no debug data at all.

We could patch the corresponding lines out when building the package, but it would be nice to not have to carry a patch around for this all the time.

Do not filter out the `-g` option, like e.g.
```
--- a/m4/subs.m4	2021-07-12 17:45:48.000000000 +0200
+++ b/m4/subs.m4	2021-10-29 14:23:24.271523009 +0200
@@ -85,8 +85,8 @@
 		;;
 		no|*)
 			debug_flags="-DNDEBUG"
-			CXXFLAGS="`echo $CXXFLAGS | sed 's:-g[[a-z-]]*\s::g' | sed 's:-g[[a-z-]]*$::'` $debug_flags"
-			CFLAGS="`echo $CFLAGS | sed 's:-g[[a-z-]]*\s::g' | sed 's:-g[[a-z-]]*$::'` $debug_flags"
+			CXXFLAGS="$CXXFLAGS $debug_flags"
+			CFLAGS="$CFLAGS $debug_flags"
 		;;
 	esac
``` 

**Additional comments**
Maybe there is a good reason for this behaviour which I don’t get right now, but I think someone who is using these flags does so for a specific reason.

Anyway, I just had to ask and thank you in advance for any insights.

ice0:
fix #2392

### DIFF
--- a/ETL/m4/subs.m4
+++ b/ETL/m4/subs.m4
@@ -82,8 +82,8 @@ AC_DEFUN([AC_ARG_DEBUG],
 		;;
 		no|*)
 			debug_flags="-DNDEBUG"
-			CXXFLAGS="`echo $CXXFLAGS | sed 's:-g[[a-z-]]*\s::g' | sed 's:-g[[a-z-]]*$::'` $debug_flags"
-			CFLAGS="`echo $CFLAGS | sed 's:-g[[a-z-]]*\s::g' | sed 's:-g[[a-z-]]*$::'` $debug_flags"
+			CXXFLAGS="$CXXFLAGS $debug_flags"
+			CFLAGS="$CFLAGS $debug_flags"
 		;;
 	esac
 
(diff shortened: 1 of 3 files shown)


rodolforg (approved):

newbluemoon:
Thank you! :)

Issue #2836, opened 2022-09-27
Remove redundant getenv() calls


morevnaproject:
The #2826 PR improved performance by replacing `getenv()` calls.

Still, there are more `getenv()` calls remaining. It would be nice to replace them, especially in render engine parts.

harshvardhan-khachane:
Hello, @morevnaproject  .
I want to solve this issue.

ice0:
@harshvardhan-khachane

Sure, you can take it.

harshvardhan-khachane:
@ice0 While setting up the NetBeans IDE, in the last step (Setting up make configuration, https://synfig-docs-dev.readthedocs.io/en/latest/ide/NetBeans.html), the document says:- 
Under Build -> make section change Build result so that it points to “_debug/build/bin/synfigstudio”:
But I need help finding the  _debug file or the synfigstudio file in the repository. Please help me with this.

ice0:
@harshvardhan-khachane 

What OS you are using?

harshvardhan-khachane:
Windows

harshvardhan-khachane:
> @harshvardhan-khachane
> 
> What OS you are using?



> @harshvardhan-khachane
> 
> What OS you are using?

Windows

harshvardhan-khachane:
> @harshvardhan-khachane
> 
> What OS you are using?

Windows

harshvardhan-khachane:
@ice0 I am using Windows

ice0:
In that case you need to use CMake-Windows-MSYS2 toolchain.

https://synfig-docs-dev.readthedocs.io/en/latest/building/Preparing%20Environment.html

After environment is prepared you need to run:
`1-setup-windows-msys2.sh` to install all dependencies
and
`2-build-msys-cmake.sh`

After build is complete `synfigstudio` for NetBeans should be located in `cmake-build-msys/output/Release/bin/synfigstudio.exe`

Currently autotools build (which is used for netbeans tutorial) is broken, so I will do CMake+NetBeans instruction tomorrow.

harshvardhan-khachane:
> In that case you need to use CMake-Windows-MSYS2 toolchain.
> 
> https://synfig-docs-dev.readthedocs.io/en/latest/building/Preparing%20Environment.html
> 
> After environment is prepared you need to run: `1-setup-windows-msys2.sh` to install all dependencies and `2-build-msys-cmake.sh`
> 
[...]

PR #2843, opened 2022-10-02
Issue #2836

JTucker17:


mohamedAdhamc:
just a thought here ^_^... why not also remove the other redundant calls ?  I believe there are more left.

PR #2849, opened 2022-10-15
perf: avoid double computation

rodolforg:
Reported-by: Svarov (https://forums.synfig.org/t/sunglasses-short-synfig-animation-detailled-analysis-for-optimization/13482)

### DIFF
--- a/synfig-core/src/modules/mod_geometry/advanced_outline.cpp
+++ b/synfig-core/src/modules/mod_geometry/advanced_outline.cpp
@@ -822,7 +822,6 @@ Advanced_Outline::connect_dynamic_param(const String& param, etl::loose_handle<V
 		}
 		if (!x || is_bline_vn) {
 			connect_bline_to_wplist(x);
-			connect_bline_to_dilist(x);
 		}
 		return Layer::connect_dynamic_param(param, x);
 	}
@@ -840,20 +839,7 @@ Advanced_Outline::connect_dynamic_param(const String& param, etl::loose_handle<V
 		else
 			return false;
 	}
-	if(param=="dilist")
-	{
-		if(Layer::connect_dynamic_param(param, x))
-		{
-			DynamicParamList::const_iterator iter(dynamic_param_list().find("bline"));
-			if(iter==dynamic_param_list().end())
-				return false;
-			else if(iter->second && !connect_bline_to_dilist(iter->second))
-				return false;
-			return true;
-		}
-		else
-			return false;
-	}
+	// no special treatment for param == "dilist"
 
 	return Layer::connect_dynamic_param(param, x);
 }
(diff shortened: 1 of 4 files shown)


rodolforg:
@Svarov-RZM  Could you please test it?

synfig-core/src/modules/mod_geometry/advanced_outline.cpp
```diff
@@ -840,20 +839,7 @@ Advanced_Outline::connect_dynamic_param(const String& param, etl::loose_handle<V
...
-			if(iter==dynamic_param_list().end())
-				return false;
-			else if(iter->second && !connect_bline_to_dilist(iter->second))
-				return false;
-			return true;
-		}
-		else
-			return false;
-	}
+	// no special treatment fot param == "dilist"
```
ice0, line 842: `for?`

rodolforg: fixed

Svarov-RZM:
> Could you please test it?

Tested and it's great.
Loading time of my character, that is a very special case, decreased from >1 hour to 625 seconds (~10 minutes). 

Personally, I would remove this check completely because the only reason it will catch something if file is corrupted. And if it's corrupted, then it's over anyway. But it's of course up to you to decide.

Svarov-RZM:
Additionally, I tested my file without the check, just left type comparision like so:
                bool is_bline_vn = false;
                if (x && x->get_type() == type_list) {
                        is_bline_vn = true;
                }
and the loading time decreaed to 442 seconds (~7 minutes). Just to consider.

rodolforg:
> Personally, I would remove this check completely because the only reason it will catch something if file is corrupted.

Or malformed (it can be manually written or by using bad-written plugins). Anyway, it would crash Synfig and user would lost data. And we both know how bad it is :)

Let's try to optimize it in other PR (if we can).

synfig-core/src/modules/mod_geometry/advanced_outline.cpp
```diff
@@ -872,18 +858,3 @@ Advanced_Outline::connect_bline_to_wplist(etl::loose_handle<ValueNode> x)
 	wplist->set_bline(ValueNode::Handle(x));
 	return true;
 }
-
-bool
-Advanced_Outline::connect_bline_to_dilist(etl::loose_handle<ValueNode> x)
```
ice0, line 872: I didn't check the logic, but it looks like dilist is being used in Advanced outlines.

Can they be safely removed?

**tool_skeleton_icon.sif**
```
<layer type="advanced_outline" active="true" exclude_from_rendering="false" version="0.3" desc="outline">
...

<param name="dilist">
  <dilist type="dash_item" loop="false">
    <entry>
      <composite guid="85AA7300D0817205EA3332A673956D17" type="dash_item">
        <offset>
          <real value="0.1000000000"/>
        </offset>
        <length>
          <real value="0.1000000000"/>
        </length>
        <side_before>
          <integer value="4"/>
        </side_before>
        <side_after>
          <integer value="4"/>
        </side_after>
      </composite>
    </entry>
  </dilist>
</param>
<param name="dash_offset">
  <real value="0.0000000000"/>
</param>
```

rodolforg: The code set bline, but this is never read

ice0: Ok. Please rebase then.

rodolforg: Rebased.

ice0:
Merged. Thank you!

PR #2850, opened 2022-10-16
fix: Synfig hangs when `convert.exe` cannot be found

ice0:
I'm not sure if this is correct solution, it might be better to reset the `child_STDIN_Write` value.
Related PR: #2491 

P.S.
@rodolforg 
I didn't notice at the first time:
```
bool is_open() const override { return read_file || child_STDIN_Write != INVALID_HANDLE_VALUE; };
```

Is there should be `child_STDIN_Write`?

P.P.S.
It finds incorrect `c:\windows\system32\convert.exe` binary, so you need to change `convert` to `convert1` or something like that to test.
I will fix this issue next.

### DIFF
--- a/synfig-core/src/synfig/os.cpp
+++ b/synfig-core/src/synfig/os.cpp
@@ -637,8 +637,7 @@ OS::run_async(std::string binary_path, const RunArgs& binary_args, RunMode mode,
 		synfig::warning("couldn't create pipe for %s %s", binary_path.c_str(), binary_args.get_string().c_str());
 		return nullptr;
 	}
-	run_pipe->open(binary_path, binary_args, mode, redir_files);
-	if (!run_pipe->is_open()) {
+	if (!run_pipe->open(binary_path, binary_args, mode, redir_files) || !run_pipe->is_open()) {
 		synfig::warning("couldn't open pipe for %s %s", binary_path.c_str(), binary_args.get_string().c_str());
 		return nullptr;
 	}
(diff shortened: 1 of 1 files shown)


rodolforg:
Oh, I forgot this problem existed before my changes :/

rodolforg:
My suggestion:
At the end of the `open()` method, handle the error of `create_child_process()` instead of just return its value: close pipe stuff and reset the variables to invalid values.

For Unix case, because I made a `return false` instead of `exit()` at the end of child process (next to line `synfig::error(_("Unable to open pipe to %s (exec failed)"), binary_path.c_str());`)

rodolforg:
> it might be better to reset the `child_STDIN_Write` value.

Yes; looking at it now, it is better to do so. It's the `open()` method that (possibly) sets `child_STDERR_Write`, `child_STDOUT_Write`, `child_STDOUT_Read`, `child_STDIN_Read`, ...

So, if `CreateProcessW()` fails, it should close and invalidate them.

rodolforg:
@ice0 please rebase

rodolforg:
> I didn't notice at the first time:
> 
> ```
> bool is_open() const override { return read_file || child_STDIN_Write != INVALID_HANDLE_VALUE; };
> ```

it is a simpler version of `is_readable() && is_writable()`

ice0:
> @ice0 please rebase

It looks like all the changes are already in `master` branch, so I'm closing this PR.

Issue #2276, opened 2021-08-20
Synfig Studio crashes when trying to use it with my drawing tablet under wayland


HoboPrimate:
**Describe the bug**
When using my drawing tablet (Intuos5 touch (S)) with Synfig on Wayland, it imediately crashes when clicking inside the application with the pen. It does not on Xorg.

**To Reproduce**
1. Be inside Gnome under Wayland
2. Plug in your wacom tablet
3. Open synfig studio
4. Pick up your pen and move it inside the program
5. Synfig studio crashes

**Expected behavior**
Be usable like when using xorg

**System information:**
 - OS: Fedora Linux
 - OS version: 34
 - Synfig version: 1.4.1

**Additional comments**
It does not crash under xorg, and crashes with both flatpak (1.4.1) and from fedora's repo (1.4.0).

lsusb shows sees the tablet as being:
Bus 001 Device 007: ID 056a:0026 Wacom Co., Ltd PTH-450 [Intuos5 touch (S)]

Attaching the log of the crash on flatpak (so I use the most recent version of synfig):
[synfig_crash_flatpak.log](https://github.com/synfig/synfig/files/7022961/synfig_crash_flatpak.log)

ice0:
@HoboPrimate Can you check 1.5.0 AppImage version?

[Synfig 1.5.0 AppImage x32](https://github.com/synfig/synfig/releases/download/v1.5.0/SynfigStudio-1.5.0-2021.08.13-linux32-32dd4.appimage)
[Synfig 1.5.0 AppImage x64](https://github.com/synfig/synfig/releases/download/v1.5.0/SynfigStudio-1.5.0-2021.08.13-linux64-32dd4.appimage)

HoboPrimate:
It doesn't crash with that AppImage, and seems to be working great!

HoboPrimate:
I also just tried the latest stable version's AppImage (1.4.2) and it also doesn't crash (flatpak version 1.4.2 crashes)

richardhenwood:
I'm wondering if this is related: Synfig crashes for me when touching the screen of my Dell XPS 13 9370, compiled from source with commit 6e3a4502338b8c32b554d8b42b61a199c70ad113

```
...
synfig(3701) [11:06:14 AM] info: Loading Settings...
synfig(3701) [11:06:14 AM] info: Init Tools...
synfig(3701) [11:06:14 AM] info: Checking auto-recover...
synfig(3701) [11:06:14 AM] info: Done.
[...]

PR #2855, opened 2022-10-18
fix: crash with a bline containing only 2 points that are too close

rodolforg:
prevent crash due to NaN with a bline containing only 2 points that are too close

distance[n-1] would be zero, and make the WidthPoint position to be NaN

That problem is propagated, affects `synfig::hom_to_std()` and a crash happens

How to reproduce: with Draw Tool and a drawing tablet, quickly tap twice (or more in a row) at the same point. Synfig Studio crashes.

### DIFF
--- a/synfig-studio/src/synfigapp/wplistconverter.cpp
+++ b/synfig-studio/src/synfigapp/wplistconverter.cpp
@@ -80,7 +80,7 @@ WPListConverter::operator()(std::list<synfig::WidthPoint> &wp_out, const std::li
 	Point c(*p_iter);
 	points.push_back(c);
 	widths.push_back(*w_iter);
-	p_iter++;
+	++p_iter;
 	for(;p_iter != end; ++p_iter,++w_iter)
 		if (*p_iter != c)
 		{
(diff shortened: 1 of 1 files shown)


ice0:
Possible issue fixed: https://github.com/synfig/synfig/issues/2276

synfig-studio/src/synfigapp/wplistconverter.cpp
```diff
@@ -101,8 +101,12 @@ WPListConverter::operator()(std::list<synfig::WidthPoint> &wp_out, const std::li
 		p1=p2;
 	}
 	// Calculate the normalized cumulative distances
-	for(i=0;i<n;i++)
-		norm_distances.push_back(distances[i]/distances[n-1]);
+	if (n == 1 && synfig::approximate_zero(distances[n-1])) {
+		norm_distances.push_back(0.);
+	} else {
+		for(i=0;i<n;i++)
+			norm_distances.push_back(distances[i]/distances[n-1]);
```
ice0, line 108: What about other points? 
Can this issue affect other points? Like close distance between points 3-4 or something?

rodolforg: The problem is the last cumulative distance (distances[n-1]) being zero, as it is denominator.

ice0 (approved):

synfig-studio/src/synfigapp/wplistconverter.cpp
rodolforg: Do you think I should remove the "n == 1" check?

rodolforg: DOne

ice0:
Merged. Thank you!

PR #2858, opened 2022-10-18
fix: changing interpolation of Canvas parameter changes interpolation for other waypoints

### LINKED ISSUE #2857: Changing interpolation of Canvas parameter changes interpolation for other waypoints
This is an issue related to recently added buttons, which allow to change interpolation - https://github.com/synfig/synfig/pull/2615

If we will select a Group layer and try to change interpolation for waypoints of Canvas parameter, then it changes interpolation for other parameters of this Group layer.

This shouldn't happen because Canvas parameter of group layer represents waypoints of all _child_ layers inside of the group layer.

Demo:
Peek 2022-10-18 21-59

Steps to reproduce:
1. Download sample file - [Synfig Animation 3.sifz.zip](https://github.com/synfig/synfig/files/9811828/Synfig.Animation.3.sifz.zip)
2. Select Group layer
3. Select waypoints which belong to Canvas parameter
4. Click any Interpolation button on the right of TimeTrack
5. bug - all waypoints of Group layer will change.

Expected behavior: only selected waypoints should be changed.

@mohamedAdhamc can you take a look, please? :)

mohamedAdhamc:
quick fix #2857 .
So from what I understood changing the interpolation of a canvas parameter of a group layer should change the interpolation of itself + the waypoints of the grouped layers which have the same time.
The issue as described was that changing the interpolation of a canvas parameter of a group layer changed along with it other parameters of the group layer. This PR stops that from happening (by not allowing the other waypoints of the group to be a part of the waypoint set we have), but I dont really understand this much so there are two points here :
1.  is the current behaviour the intended behaviour ?
2. does this break anything ? 

And this is just in case previous points are good

1.  is using the layers `get_name()` method to know if its a group safe ? or should I check if a layer is itself a group by another way (quickly looking around in the layer class I couldn't really find alternative ways to check if our layer is a group)

@morevnaproject could you please test it? :)

### DIFF
--- a/synfig-core/src/synfig/timepointcollect.cpp
+++ b/synfig-core/src/synfig/timepointcollect.cpp
@@ -87,7 +87,8 @@ synfig::waypoint_collect(std::set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 		const Layer::DynamicParamList& dyn_param_list(layer->dynamic_param_list());
 		Layer::DynamicParamList::const_iterator iter;
 		int ret(0);
-		if (layer->get_name() != "group"){
+		etl::handle<Layer_PasteCanvas> p = etl::handle<Layer_PasteCanvas>::cast_dynamic(layer);
+		if (!p){
 			for(iter=dyn_param_list.begin();iter!=dyn_param_list.end();++iter)
 				ret+=waypoint_collect(waypoint_set,time,iter->second);
 		}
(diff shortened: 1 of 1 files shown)


ice0:
> quickly looking around in the layer class I couldn't really find alternative ways to check if our layer is a group

This is because Groups in Synfig code is a `Layer_PasteCanvas` (for historical reasons).
So to check if layer is a group you can use something like this:
```
if (etl::handle<Layer_PasteCanvas>::cast_dynamic(layer))
```

mohamedAdhamc:
> This is because Groups in Synfig code is a `Layer_PasteCanvas` (for historical reasons). So to check if layer is a group you can use something like this:
> 

Perfect, will change to that way for checking. Now we just need to know if this new behavior is what was asked for, and also that it doesn't break other things.

morevnaproject:
> changing the interpolation of a canvas parameter of a group layer should change the interpolation of itself + the waypoints of the grouped layers which have the same time

Correction: changing interpolation of waypoint in canvas parameter should change waypoints **for child layers only**.

Waypoints in canvas parameter = waypoints of child layers of the group.

PR #2859, opened 2022-10-19
feat: Move "Default interpolation" dropdown to Preferences dialog

mohamedAdhamc:
moved and this is how every thing looks now:
moved to prefrences

But there is one issue (or not) I noticed. the original default interpolation button didnt save the default interpolation, instead each time synfig opens it defaults its value as "clamped". I think now that its in preferences it probably should be saved on reopen, so I will have to implement that still. 

@ice0 , @morevnaproject opinions ?

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -72,7 +72,6 @@
 #include <gui/preview.h>
 #include <gui/states/state_normal.h>
 #include <gui/widgets/widget_canvastimeslider.h>
-#include <gui/widgets/widget_interpolation.h>
 #include <gui/workarea.h>
 
 #include <pangomm.h>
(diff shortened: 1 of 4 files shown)


ice0:
>  I think now that its in preferences it probably should be saved on reopen, so I will have to implement that still.

Agree, good idea.

What about horizontal alignment? As for me left alignment looks better.

image

image

mohamedAdhamc:
> What about horizontal alignment? As for me left alignment looks better.
> 

Yes, looks better will change that. But someone mentioned something important in the issue this PR is linked to so I'm not really sure now if this feature is in fact still "needed" I need confirmation here because the argument that was made convinced me xD.

mohamedAdhamc:
Alright looks like I just got the confirmation from the issue that it is in fact still needed. Will continue here
What was left:
1. Fixing the entries alignment
2. Saving the default interpolation metadata

mohamedAdhamc:
Functionality is done and seems to be working good. But I will review everything again tomorrow hopefully (implementation wise) to make sure everything is good :)

mohamedAdhamc:
The move to preference dialog is no longer needed. But I think it may be kind of useful to have the selected interpolation saved in settings on reopen. so what do you guys think ?
1- have the default interpolation as is in the preferences dialog along with the already in-place widget.
2- directly save the already in-place widgets default interpolation.
3- don't save at all.

I think 2 would probably make the most sense, but I need your opinions :).

rodolforg:
@morevnaproject what about this PR? It let the user choose the default interpolation for the app in the preferences, without removing the widget on the main window.

mohamedAdhamc:
just to clarify. this pr as of right now is still not updated. Currently it moves the widget to the preference dialog. but the widget can be re added to achieve what @rodolforg mentioned. or any other  options.

So, basically when a decision is made it can be easily altered to the wanted behaviour which could for example be like any of those or another :
>>
>1- have the default interpolation as is in the preferences dialog along with the already in-place widget.
2- directly save the already in-place widgets default interpolation.
3- don't save at all.

>

morevnaproject:
Hi! I remember @Svarov-RZM requested to save default interpolation on per-file basis. (i.e. in file metadata?)
And this sounds logical.
In this case we not need this option in Preferences.

@mohamedAdhamc Thank you so much for making contribution and my apologies for this premature request, which made you to do extra work.

mohamedAdhamc:
@morevnaproject no problem :).

PR #2876, opened 2022-11-08
fix: added ImageMagick 7.1 library initialization

ice0:
fixed hang when using `mod_magickpp` with ImageMagick 7.1 library


ice0:
ping @rodolforg @morevnaproject

rodolforg (approved):

PR #2877, opened 2022-11-11
fix: no context menu when right-clicking on waypoint 

### LINKED ISSUE #2875: No context menu when right-clicking on waypoint
This is reproducible in current master branch (8bf9c30).

1. Download and open sample file - 
[wp-right-click-bug.sifz.zip](https://github.com/synfig/synfig/files/9961597/wp-right-click-bug.sifz.zip)
2.  Select group "1".
3. Try right-clickig on waypoints, which I indicated on screenshot below -
screenshot_001

Notice that no menu shown after right-click.

For other waypoints menu show normally.

This seems to be related with #2858 fix.

@mohamedAdhamc can you take a look, please? ^__^

mohamedAdhamc:
Fixes #2875 , and adds on to #2858.

So the issue #2858  was fixing was that editing the canvas parameter of a group changed along with it the groups other parameters of the same time. So the fix was to not allow the canvas waypoint to change any of the groups parameters, which in fact was not general enough as if the group has a group beneath it, then in that case the canvas parameter should change the child groups parameters as well (but not its own).

Here is a video demonstration: (note: I'm changing the interpolation via the menu on right click but for some reason windows recorder doesn't record windows on top the window we are in)

https://user-images.githubusercontent.com/100296264/201393112-a217d37d-2b39-4b4f-bf59-c37c11420179.mp4

@morevnaproject, waiting for you review :) also if there is something I'm missing here please let me know.

### DIFF
--- a/synfig-core/src/synfig/timepointcollect.cpp
+++ b/synfig-core/src/synfig/timepointcollect.cpp
@@ -58,7 +58,8 @@ using namespace synfig;
 int
 synfig::waypoint_collect(std::set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 						 const Time								&time,
-						 const etl::handle<Node>				&node)
+						 const etl::handle<Node>				&node,
+						 bool initial_waypoint)
 {
 	const TimePointSet& timepoint_set(node->get_times());
 
@@ -88,7 +89,7 @@ synfig::waypoint_collect(std::set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 		Layer::DynamicParamList::const_iterator iter;
 		int ret(0);
 		etl::handle<Layer_PasteCanvas> p = etl::handle<Layer_PasteCanvas>::cast_dynamic(layer);
-		if (!p){
+		if ((!p && initial_waypoint) || !initial_waypoint){
 			for(iter=dyn_param_list.begin();iter!=dyn_param_list.end();++iter)
 				ret+=waypoint_collect(waypoint_set,time,iter->second);
 		}
@@ -114,7 +115,7 @@ synfig::waypoint_collect(std::set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 		Canvas::const_iterator iter;
 		int ret(0);
 		for(iter=canvas->begin();iter!=canvas->end();++iter)
-			ret+=waypoint_collect(waypoint_set,time,*iter);
+			ret+=waypoint_collect(waypoint_set,time,*iter, initial_waypoint);
 		return ret;
 	}
 
--- a/synfig-core/src/synfig/timepointcollect.cpp
+++ b/synfig-core/src/synfig/timepointcollect.cpp
@@ -89,7 +89,7 @@ synfig::waypoint_collect(std::set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 		Layer::DynamicParamList::const_iterator iter;
 		int ret(0);
 		etl::handle<Layer_PasteCanvas> p = etl::handle<Layer_PasteCanvas>::cast_dynamic(layer);
-        if ((!p && ignore_dynamic_parameters) || !ignore_dynamic_parameters){
+		if (!p || !ignore_dynamic_parameters){
 			for(iter=dyn_param_list.begin();iter!=dyn_param_list.end();++iter)
 				ret+=waypoint_collect(waypoint_set,time,iter->second);
 		}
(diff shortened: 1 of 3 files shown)


synfig-core/src/synfig/timepointcollect.cpp
```diff
@@ -58,7 +58,8 @@ using namespace synfig;
 int
 synfig::waypoint_collect(std::set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 						 const Time								&time,
-						 const etl::handle<Node>				&node)
+						 const etl::handle<Node>				&node,
+						 bool initial_waypoint)
```
rodolforg, line 62: I suggest to change the parameter name to `ignore_dynamic_parameters`.

rodolforg (changes requested):

synfig-core/src/synfig/timepointcollect.cpp
mohamedAdhamc: ok, makes sense will do.

synfig-core/src/synfig/timepointcollect.cpp
```diff
@@ -88,7 +89,7 @@ synfig::waypoint_collect(std::set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 		Layer::DynamicParamList::const_iterator iter;
 		int ret(0);
 		etl::handle<Layer_PasteCanvas> p = etl::handle<Layer_PasteCanvas>::cast_dynamic(layer);
-		if (!p){
+		if ((!p && initial_waypoint) || !initial_waypoint){
```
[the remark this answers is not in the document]
mohamedAdhamc: hmm, yes but there was a cast here anyways as it is used in line 100, so I just moved the cast higher here so it can be used here and there. is there something i'm missing ?

i mean if this is not the initial waypoint yes the cast wont be needed on line 92 but it will still be needed for line 100. But if it is the intial waypoint then we will benefit from not having to cast twice.

synfig-core/src/synfig/timepointcollect.cpp
```diff
@@ -88,7 +89,7 @@ synfig::waypoint_collect(std::set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 		Layer::DynamicParamList::const_iterator iter;
 		int ret(0);
 		etl::handle<Layer_PasteCanvas> p = etl::handle<Layer_PasteCanvas>::cast_dynamic(layer);
-		if (!p){
+        if ((!p && ignore_dynamic_parameters) || !ignore_dynamic_parameters){
```
rodolforg, line 92: This can be simplified by a absorption law in Boolean Algebra:
https://www.geeksforgeeks.org/properties-of-boolean-algebra/
https://www.allaboutcircuits.com/textbook/digital/chpt-7/boolean-rules-for-simplification/

Indentation is wrong: use Tabs.

synfig-core/src/synfig/timepointcollect.cpp
```diff
@@ -114,7 +115,7 @@ synfig::waypoint_collect(std::set<Waypoint, std::less<UniqueID> >	&waypoint_set,
 		Canvas::const_iterator iter;
 		int ret(0);
 		for(iter=canvas->begin();iter!=canvas->end();++iter)
-			ret+=waypoint_collect(waypoint_set,time,*iter);
+            ret+=waypoint_collect(waypoint_set,time,*iter, ignore_dynamic_parameters);
```
rodolforg, line 118: Indentation is wrong: use Tabs.

synfig-core/src/synfig/timepointcollect.cpp
mohamedAdhamc: nice, will do

synfig-core/src/synfig/timepointcollect.cpp
mohamedAdhamc: thanks for letting me know, i forgot to set indentation setting after switching OS. will do.

mohamedAdhamc:
@rodolforg, done :)

ice0:
> seems that mac builds are failing at install dependencies brew step. For linux I'm not entirely sure whats going on.

Mac build failed due to some network issues.
Other builds (like Linux) are canceled if some builds fails.
This is due to the early fail strategy.

No problem here, I just restart failed builds later.

morevnaproject:
So, is it ready? :)

rodolforg:
> So, is it ready? :)

@morevnaproject if it works as intend, yes :) Code looks fine now ;)

But please squash commits on merging.

mohamedAdhamc:
> Mac build failed due to some network issues. Other builds (like Linux) are canceled if some builds fails. This is due to the early fail strategy.

nice, good to know :)

rodolforg (approved):

ice0:
Merged. Thank you!

PR #2879, opened 2022-11-14
fix(lang): Spanish translation typo

pgilfernandez:
Simple typo fixed in the Spanish translation file

### DIFF
--- a/synfig-studio/po/es.po
+++ b/synfig-studio/po/es.po
@@ -2275,7 +2275,7 @@ msgstr "Nombre:"
 #: ../src/gui/states/state_rectangle.cpp:482
 #: ../src/gui/states/state_star.cpp:595 ../src/gui/states/state_text.cpp:388
 msgid "Layer Type:"
-msgstr "Tipio de Capa:"
+msgstr "Tipo de Capa:"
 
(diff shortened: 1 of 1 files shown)


rodolforg (approved):

ice0:
Hi, Pablo!

Thanks for the PR. We are using Transifex as our primary source for translations, so I need to check if I can merge it without modifications in Transifex.

@morevnaproject 
Can I merge it as-is, or should this be fixed on Transifex first?

morevnaproject:
Yes, we should fix that on Transifex, otherwise it will be lost after next sync with Transifex.

ice0:
> Yes, we should fix that on Transifex, otherwise it will be lost after next sync with Transifex.

What if I merge this PR, and fix it on Transifex? Will this cause a merge conflict?

morevnaproject:
> What if I merge this PR, and fix it on Transifex? Will this cause a merge conflict?

I have made changes to Transifex (to both development and stable branches). No merge conflicts will be after merge.

morevnaproject (approved):

ice0:
Merged. Thank you!

PR #2892, opened 2022-11-17
feat: add transformation type to Shade layer

Svarov-RZM:
This commit adds transformation type to Shade layer (similar to PasteCanvas/Filter group). Things to consider:
* The forward compatibility for Shade layer is broken. Forward compatibility is not a goal of Synfig as far as I know, so should be fine
* I changed the layer's version to 1.0, I don't think it makes sense to keep it in alpha stage (0.\*) when Synfig itself is considered stable (versions 1.*)
* The 'transformation' type was previously PasteCanvas specific, but now it's generic, meaning any layer in the future can utilize this type if needed

Forum discussion: https://forums.synfig.org/t/improvements-to-shade-layer-part-i/13564

### DIFF
--- a/synfig-core/src/modules/lyr_std/shade.cpp
+++ b/synfig-core/src/modules/lyr_std/shade.cpp
@@ -182,6 +182,7 @@ Layer_Shade::get_param_vocab(void)const
 	ret.push_back(ParamDesc("origin")
 		.set_local_name(_("Origin"))
 		.set_is_distance()
+		.set_invisible_duck()
 	);
 	ret.push_back(ParamDesc("transformation")
 		.set_local_name(_("Transformation"))
@@ -191,7 +192,6 @@ Layer_Shade::get_param_vocab(void)const
 		.set_local_name(_("Size"))
 		.set_description(_("Size of the shade"))
 		.set_is_distance()
-		.set_origin("origin")
 	);
 	ret.push_back(ParamDesc("type")
 		.set_local_name(_("Type"))


morevnaproject:
I suggest to make a copy of Shade Layer, internally call it "shade2" and make all changes there.
Old "Shade Layer" should remain active, but I suggest to rename it to "Shade (old)" and hide from layer creation menu.
This way we can keep compatibility.

This is similar approach as @rodolforg did with new TimeTrack panel -  he did two panels and this allowed to smoothly replace one with another.

rodolforg:
I don't think it is necessary.
New shade layers won't work if opening the .sif in previous Synfig Studio app. But why someone would do it?
The proposed code knows how to convert the old parameter ("origin") to the new one.

Svarov-RZM:
> I suggest to make a copy of Shade Layer, internally call it "shade2" and make all changes there.

Well, this isn't a complete rework of the layer, just a simple improvement, so I don't think it is needed to go through all those hoops by duplicating stuff.

> New shade layers won't work if opening the .sif in previous Synfig Studio app

They kinda will, except the 'origin' parameter will be reset to zero, but it will allow user to continue working in old version with some manual adjustments if it's absolutely necessarily to the user.

rodolforg:
@Svarov-RZM, you propose to hide "origin" parameter, and it is not actually used anymore (it is a constant vector `(0,0)`).

Is it possible to remove it?

Svarov-RZM:
> you propose to hide "origin" parameter, and it is not actually used anymore (it is a constant vector `(0,0)`).

No no, I proposed to hide origin's **duck**.
The 'origin' parameter itself is used of course, it's a part of transformation type. You see, without hiding it the duckmatic will override the origin of transformation widget (that little purple dot on the screenshot):
Tr

> Is it possible to remove it?

No, it's needed. Serves the same purpose as origin parameter for PasteCanvas, it shifts the center of transformation (when you rotate/scale the object).

Svarov-RZM:
OK, so I rebased this on current master and everything still works. Had to fix some conflicts since duckmatic.cpp changed which gave me quite a lot of troubles since I am not good with git, but seems fine now. I even got rid of one useless commit (about fixing spaces).

So, is it going to be merged? I am planing to use this functionality in my next animation and I don't want to end up maintaining a fork. If there are some issues with it, please tell me, I'll try to fix them.

PR #2895, opened 2022-11-20
feat: added hotkey for duplicate layer button

### LINKED ISSUE #2891: Hotkey for Duplicate Layer button
`ctrl + d` hotkey for "duplicate layer" function would help speed up artwork creation in Synfig.
image

mohamedAdhamc:
fixes #2891 
ctrl + d and ctrl + shift + d, are both already being used for unselect all layer and ducks. I semi-arbitrarily chose ctrl+u as its what comes after the d in duplicate, and anyway the user can change it. But if there is a better initial key pair please let me know :).

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -1365,7 +1365,7 @@ App::get_default_accel_map()
 		{"<Control>parenleft" ,     "<Actions>/canvasview/decrease-low-res-pixel-size"},
 		{"<Control>parenright" ,    "<Actions>/canvasview/increase-low-res-pixel-size"},
 		{"<Primary>g",              "<Actions>/action_group_layer_action_manager/action-LayerEncapsulate"},
-        {"<Primary>u",              "<Actions>/action_group_layer_action_manager/action-LayerDuplicate"},
+		{"<Primary>u",              "<Actions>/action_group_layer_action_manager/action-LayerDuplicate"},
 		{"<Control><Mod1>parenleft",  "<Actions>/action_group_layer_action_manager/amount-dec"},
 		{"<Control><Mod1>parenright", "<Actions>/action_group_layer_action_manager/amount-inc"},
 		{"equal",                   "<Actions>/canvasview/canvas-zoom-in"},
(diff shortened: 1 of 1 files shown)


yoSachinkr:
Works perfectly. 💯

ice0:
@yoSachinkr @morevnaproject @rodolforg @pgilfernandez @FirasH

Does anyone use the `ctrl+d` combination to select? Maybe deprecate it?

pgilfernandez:
I have never used any combination of those under Synfig but if you ask me I would set `ctrl+a` (or `cmd+a`) for selecting all and `shift+ctrl+a` (or `shift+cmd+a`) for deselecting all.
Nowadays `a` is used for selecting and deselecting but Windows keep using `e` in the File Explorer... anyway I personally prefer `a` as other open source software uses it (such as Blender).

yoSachinkr:
> @yoSachinkr @morevnaproject @rodolforg @pgilfernandez @FirasH
> 
> Does anyone use the `ctrl+d` combination to select? Maybe deprecate it?

Personally I use `ctrl+a` to select all vertices and `shift+a` to deselect all vertices. Also `ctrl+shif+a` to select all layers and `ctrl+shift+d` to deselect all layers.

rodolforg:
> I have never used any combination of those under Synfig but if you ask me I would set `ctrl+a` (or `cmd+a`) for selecting all and `shift+ctrl+a` (or `shift+cmd+a`) for deselecting all.

What about `Ctrl+a` for both selecting all and deselecting all handles?
What about `Ctrl+shift+a` for both selecting all and deselecting all layers?
Blender uses `a` to toggle all/none selected.

This is a way to free `ctrl+d` and `ctrl+shift+d`

mohamedAdhamc:
@rodolforg , I also think this would be the best to have the layer/duck selects be ctrl/shift/a combinations. So the CTR+d is freed for this. I guess I just need confirmation here now and I can change it accordingly.

ice0:
I think it's better to switch the hotkeys in the new PR.

ice0 (approved):

ice0:
Merged. Thank you!

morevnaproject:
> I also think this would be the best to have the layer/duck selects be ctrl/shift/a combinations. So the CTR+d is freed for this.

Yes, I like this too.

In any case, I do not use Ctrl+D for selecting layers. Actually, I never had a need to select all layers. XD

PR #2898, opened 2022-11-20
chore: added /build-cmake/ folder to .gitignore 

pgilfernandez:
Adds `/build-cmake/` folder to **.gitignore** file as it is needed under macOS when compiling with Cmake

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,7 @@ compile
 /_debug/
 /_production/
 /_release/
+/build-cmake/
 /build.conf
 
 # ------------------------------------------------------------------------


rodolforg (dismissed):

rodolforg:
@pgilfernandez we are using https://www.conventionalcommits.org/en/v1.0.0/ conventions. :)
Could you please rewrite your commit message?

ice0:
Now you can check details here:

image

pgilfernandez:
First of all, sorry about it, as I said somewhere, I'm not an experienced coder so there is a high chance that I miss some things and I may don't follow the written standards, not because I don't want but because of my ignorance... sorry again.

In this commit I'm not sure if the "fix:" convention is the correct one as I'm modifying a Git file but, well, let try it. I hope it's fine this way... if not I will try better ;)

rodolforg:
@ice0 I guess it should be chore

ice0:
@rodolforg Done!

pgilfernandez:
and done in the commit

rodolforg (approved):

ice0:
Merged. Thank you!

PR #2921, opened 2022-11-29
fix: allow to customize TimeTrack font via css

ice0:


pgilfernandez:
WOW!
I love it, it works perfectly =D
Thanks!

rodolforg:
Is it timetrack or parameters panel?

pgilfernandez:
> Is it timetrack or parameters panel?

Both, they are synchronized when I change the font size

pgilfernandez:
As a proof, here is an extreme example with font size 20px:
Screenshot 2022-12-03 at 22 47 32

Keyikedalube:
Now column header font size is also affected.

- Before  (v1.5.1)
Screenshot from 2022-12-06 07-01-46

- After (this PR)
Screenshot from 2022-12-06 06-58-08

An alternate hack around to this is using percentages so we can reset the header font size
```css
#timetrack {
    font-size: 90%;
    font-weight: 400;
}

/*
 * The above overrides the default font size to 90%
 * So we need to add 10% more to get the exact system default size
 */
#timetrack header {
    font-size: 110%;
}
```
Screenshot from 2022-12-06 07-57-12

@ice0 GTK CSS supports negation `:not(<selector>)` too. Though I couldn't figure out how exactly to use it.
I tried:
```css
#timetrack:not(header) {
    ...
 }
```

Doesn't reproduce the same hack around output demonstrated above :(
Messaged the official gtk dev team [here](https://matrix.to/#/!aUhETchlgthwWVQzhi:matrix.org/$1670292462134719FmuqD:matrix.org?via=gnome.org&via=matrix.org&via=libera.chat) for solution
Screenshot from 2022-12-06 08-08-28


_PS: Forgive me for the delay, I only read the commit changes the other day and didn't realize this PR was intended to fix the linked issue. Totally forgot to come back and review this._

synfig-studio/src/gui/resources/css/synfig.css
```diff
@@ -97,3 +97,8 @@
 #canvasview {
 	background: @theme_unfocused_bg_color;
 }
+
+#timetrack {
+	font-size: 11px;
```
Keyikedalube, line 102: Using negation `:not(...)` would skip customizing column header fonts:
```css
#timetrack .view:not(header) { ... }
```

Keyikedalube (changes requested):

Keyikedalube:
Why don't we use percentages instead?

from
```css
  font-size: 11px;
```
to
```css
  font-size: 85%;
```
With percentages, our font size is guaranteed to be relative to the system font size.
So someone with a larger screen monitor doesn't get forced to hard-coded size `11px`. Just my opinion.

synfig-studio/src/gui/resources/css/synfig.css
pgilfernandez: I agree.

This way headers are not scaled and it lets users to set the desired size

pgilfernandez:
> Why don't we use percentages instead?

Yes, it's not a bad idea. It works as expected, I would approve the negation `:not(...)`and percentages.

Any other problem with this PR? it would be great it's merged ;)

synfig-studio/src/gui/resources/css/synfig.css
rodolforg: @ice0 ping

ice0: Hi, guys!
Sorry for a late reply.
Something has changed since this commit, so I'm trying to find a new solution.
1) After rebase this is not working anymore.
2) `#timetrack .view:not(header) { ... }` this is not working for me. Maybe I am doing something wrong.
3)
>  `Why don't we use percentages instead?`. 

This is not work how it should be. I will show this in video below.

> Messaged the official gtk dev team [here](https://matrix.to/#/!aUhETchlgthwWVQzhi:matrix.org/$1670292462134719FmuqD:matrix.org?via=gnome.org&via=matrix.org&via=libera.chat) for solution

Unfortunately I can't see the answer.
image


Video 1. Trying to use `:not` selector:

https://github.com/synfig/synfig/assets/5604544/5abbd693-c313-4a4d-9364-91bf3f8e4fd7


Video 2. font not changing in macOS:

https://github.com/synfig/synfig/assets/5604544/661674e7-73a3-4c2c-823e-434ebccbe803

rodolforg: About the second video: it seems your build uses current code (parameter panel font is a fixed value in code)

ice0: > About the second video: it seems your build uses current code (parameter panel font is a fixed value in code)

I don't know how you noticed this, but you're right! I forgot to cherry-pick changes to the current master before testing lol.

rodolforg: I tested here too, and the `not` operator does not seem to work for this case.
And, as you said, the `header` node inherits the `view` :/

So, the only way to not affect the header is using two selectors:

- #parameters_panel
- #parameters_panel header

So we set the font-size in the first selector and the we revert it back in the second one.

```css
#parameters_panel {
  font-size: 70%;
}

#parameters_panel header {
  font-size: 143%; /* 1 / 0.70 */
}

```

So, what is the best option?

rodolforg (approved):
I somebody finds a way to `:not` or a better way to percentages work, please open another PR ;)
This one already does its job :)

synfig-studio/src/gui/resources/css/synfig.mac.css
```diff
@@ -7,3 +7,8 @@
 	box-shadow: none;
 	margin    : 0;
 }
+
+#parameters_panel {
+    font-size: 11pt;
+    font-weight: 300;
```
rodolforg, line 13: @ice0 please use `tab` for indentation here too (Codacy is complaining and the rest of CSS file use it indeed)

Issue #2272, opened 2021-08-17
Feature request: Animated .webp support


yoSachinkr:
Support for .webp image format as well as import and export of animated webp support would be nice.

PR #2933, opened 2022-12-11
feat: Support for animated webp with transparency

BobSynfig:
Feature request: Animated .webp support https://github.com/synfig/synfig/issues/2272

Renders Animated webp via FFMPEG (libwep_anim & libwebp).
Transparency is supported by libwep_anim (default), replaced by gray background with libwebp.
Configuration of lossless/lossy, compression level, quality and preset.

Preliminary support for import (single image only, like for png)

Let's hope my obsession for indentation and vertical alignment will not hurt the eyes of @rodolforg  ;)

### DIFF
--- a/synfig-core/src/modules/mod_webp/mptr_webp.cpp
+++ b/synfig-core/src/modules/mod_webp/mptr_webp.cpp
@@ -1,0 +1,198 @@
+/* === S Y N F I G ========================================================= */
+/*!	\file mptr_webp.cpp
+**	\brief WEBP (via FFMPEG) Importer Module
+**
+**	\legal
+**	Copyright (c) 2002-2005 Robert B. Quattlebaum Jr., Adrian Bentley
+**	Copyright (c) 2007 Chris Moore
+ **	Copyright (c) 2022 BobSynfig
+**
+**	This file is part of Synfig.
+**
+**	Synfig is free software: you can redistribute it and/or modify
+**	it under the terms of the GNU General Public License as published by
+**	the Free Software Foundation, either version 2 of the License, or
+**	(at your option) any later version.
+**
+**	Synfig is distributed in the hope that it will be useful,
+**	but WITHOUT ANY WARRANTY; without even the implied warranty of
+**	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+**	GNU General Public License for more details.
+**
+**	You should have received a copy of the GNU General Public License
+**	along with Synfig.  If not, see <https://www.gnu.org/licenses/>.
+**	\endlegal
+**
+** ========================================================================= */
+
+/* === H E A D E R S ======================================================= */
+
+#ifdef USING_PCH
+#	include "pch.h"
+#else
+#ifdef HAVE_CONFIG_H
+#	include <config.h>
+#endif
+
+#include "mptr_webp.h"
+
+#include <ETL/stringf>
+
+#include <synfig/general.h>
+#include <synfig/localization.h>
+
+#if HAVE_IO_H
+ #include <io.h>
+#endif
+#if HAVE_PROCESS_H
+ #include <process.h>
+#endif
+#if HAVE_FCNTL_H
+ #include <fcntl.h>
+#endif
+#include <iostream>
+#endif
+
+/* === M A C R O S ========================================================= */
+
+using namespace synfig;
+using namespace etl;
+
+/* === G L O B A L S ======================================================= */
+
+SYNFIG_IMPORTER_INIT       ( webp_mptr );
+SYNFIG_IMPORTER_SET_NAME   ( webp_mptr, "webp" );
+SYNFIG_IMPORTER_SET_EXT    ( webp_mptr, "webp" ); // not working look at the main.cpp for list opf available extensions
+SYNFIG_IMPORTER_SET_VERSION( webp_mptr, "0.1"  );
+SYNFIG_IMPORTER_SET_SUPPORTS_FILE_SYSTEM_WRAPPER(webp_mptr, false);
+
+/* === M E T H O D S ======================================================= */
+
+bool webp_mptr::is_animated()
+{
+	return true;
+}
+
+bool
+webp_mptr::seek_to(const Time& time)
+{
+	//if(frame<cur_frame || !file)
+	//{
+		pipe = nullptr;
+
+		// FIXME: 24 fps is hardcoded now, but in fact we have to get it from canvas
+		//float position = (frame+1)/24; // webp didn't work with 0 frame
+		//float position = 1000/24; // webp didn't work with 0 frame
+		const std::string position = time.get_string(Time::FORMAT_NORMAL);
+
+		OS::RunArgs args;
+		args.push_back( { "-ss",      position                              });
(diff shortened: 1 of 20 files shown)


synfig-studio/src/gui/render.cpp
```diff
@@ -439,18 +465,19 @@ RenderSettings::check_target_destination()
 		//otherwise Auto is selected
 		else
 		{
-			std::list<std::string> ext_multi_auto = {{".bmp"}, {".png"},
-					{".jpg"},{".exr"},{".ppm"}};
-	
+			std::list<std::string> ext_multi_auto = {
+				{".bmp"}, {".png"},{".jpg"},{".exr"},{".ppm"},{".webp"}
+			};
+
```
rodolforg, line 471: Not your fault, but: why `std::list` instead of `std::vector`?
why the double curly brackets `{`?

synfig-core/src/synfig/targetparam.h
```diff
@@ -41,26 +42,53 @@ struct TargetParam
...
+
+		sequence_separator("."),
+		offset_x          (0),
+		offset_y          (0),
+		rows              (0),
+		columns           (0),
+		append            (true),
+		dir               (HR),
+
+		//<!-- Added for Webp support
```
rodolforg, line 65: ugh, I guess we should create something like layer param vocab....

rodolforg (commented):
I didn't review it yet, because I have a question: why a whole new module, instead of just adding it to ffmpeg module?

synfig-core/src/modules/mod_webp/mptr_webp.cpp
```diff
@@ -0,0 +1,198 @@
...
+		pipe = nullptr;
+
+		// FIXME: 24 fps is hardcoded now, but in fact we have to get it from canvas
+		//float position = (frame+1)/24; // webp didn't work with 0 frame
+		//float position = 1000/24; // webp didn't work with 0 frame
+		const std::string position = time.get_string(Time::FORMAT_NORMAL);
+
+		OS::RunArgs args;
+		args.push_back( { "-ss",      position                              });
+		args.push_back( {"-i",        filesystem::Path(identifier.filename) });
```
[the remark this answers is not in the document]
BobSynfig: I made a separated module to avoid to introduce complexity and separation of concerns.
Based on ffmpeg module, it has different parameters, also a different configuration dialog.
"Multiplexed code" is adding conditions and removing lisibility (intention -> goal -> result).
That's also the reason I like vertical alignment, then source has to be seen in its globality, not line-by-line.
And a separated module also permits to avoid to mess the original code with such formatting.

synfig-core/src/modules/mod_webp/mptr_webp.cpp
```diff
@@ -0,0 +1,198 @@
...
+	//}
+	return true;
+}
+
+bool
+webp_mptr::grab_frame(void)
+{
+	if (!pipe)
+	{
+		synfig::error(_("unable to open %s"), identifier.filename.c_str());
```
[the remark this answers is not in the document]
BobSynfig: As said before, the import is not the main focus at the moment.
It is basically reworked from the original ffmpeg equivalent module.

synfig-studio/src/gui/render.cpp
BobSynfig: The same as before, from the original code, just reworked a bit the formatting for more lisibility.
If it works before, it will workthe same after following the same pattern.
No need to introduce bug for a simple added feature, and if a refactor can be done, itcan be done afterwards. KISS ;)

ktunkiewicz:
@BobSynfig I'm trying out your branch on Windows 11, got it compiled and tried rendering to webp but couldn't make it work. I'm totally new to Synfig, so maybe I'm doing something wrong. When Synfig compiles, it does not have an ffmpeg copy in the /bin folder, so I tried with ffmpeg binaries from the two official builds available on ffmpeg website.
Both give me this result:
image

Fails like this in UI:
image

Am I doing something wrong, or is this a problem with the code?

ktunkiewicz:
Also, I was going to ask this in the forum, but maybe you know why it renders motion blur with a visible cut-off? It doesn't show that in preview or anywhere else
tyrael png

zeddicus-pl:
Ah, ignore my last question, it's the PIXEL_RENDERING_LIMIT thing

ice0:
> @BobSynfig I'm trying out your branch on Windows 11, got it compiled and tried rendering to webp but couldn't make it work. I'm totally new to Synfig, so maybe I'm doing something wrong. When Synfig compiles, it does not have an ffmpeg copy in the /bin folder, so I tried with ffmpeg binaries from the two official builds available on ffmpeg website. 

MSYS2 build should find ffmpeg from MSYS2 MinGW64 environment.

Can you show output of `which ffmpeg` command from MSYS2 MinGW shell?

zeddicus-pl:
@ice0 yeah, it's indeed in `/mingw64/bin`, but I get similar result with that binary of ffmpeg too.

I got this one now. I **added an error code from GetLastError** into synfig::error output, to see what the reason is:

image

zeddicus-pl:
the above was because the `ffmpeg` that I copied from msys wasn't running correctly, I got it fixed, and now got into pretty much the same situation I had with the ffmpeg binaries I had before:
image


It works all fine with the `ffmpeg` rendering, but the above is visible with `webp` rendering

zeddicus-pl:
Is it only some specific version of ffmpeg that works fine with this?

BobSynfig:
I see vcodec is still mpeg4, it should be libwebp_anim selected by default while selecting the libwebp module.
Try to change the parameters from dialog, it should work after that.
I will see to make the required modification ( it is in Init() of trgt_webp.cpp )

BobSynfig:
@zeddicus-pl Did you do some new test?

zeddicus-pl:
@BobSynfig ah, yes, I did. Works all fine 👍 Sorry, forgot to tell you :)

BobSynfig:
Still no news for merging this feature?

BobSynfig:
How to solve the conflict on synfig-studio/src/gui/dialogs/CMakeLists.txt ?

mohamedAdhamc:
Hey @BobSynfig , I believe a simple way here since it's a very small conflict is to use the "web editor".

It is colored in blue you'll find it placed after the failed appveyor build. You can then simple remove the "==", "<<", ">>" signs and keep what you would like to be kept or added. Then in the top right you will find mark as resolved you can then follow along with it.

A more detailed [description](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-on-github) on github site.

BobSynfig:
So easy after all oO
Thank you my friend @mohamedAdhamc, I learnt something today that I will be able to apply the next times :)

BobSynfig:
>Codacy Static Code Analysis — This pull request quality could be better.

I like a lot the fact that now it is Codacy that prevents the validation of this PR while there are "minor" concerns and that these parts of code are identical to the existing ones in all the other modules...

rodolforg:
It probably (and I hope) only evaluates the modifications the PR commits do.

PR #2938, opened 2022-12-11
fix: use format string instead of string concatenation

rodolforg:
https://www.gnu.org/software/gettext/manual/gettext.html#Preparing-Strings


ice0 (approved):

ice0:
Merged. Thank you!

PR #2941, opened 2022-12-16
fix: smooth move tool is now initialized with the saved value in settings.

### LINKED ISSUE #2940: SmoothMove radius doesn't use last used value
**Describe the bug**
If you set the SmoothMove radius to anything but 1 it works, but when you go back to the tool, even if the last radius value is displayed, it doesn't works till you update the field.
The following video describes it perfectly:

**To Reproduce**
https://user-images.githubusercontent.com/5942369/207865664-dea33d74-6b4d-4c3a-a20a-3380c9e1e979.mp4

**Expected behavior**
If the last radius value is stored, then it should work

**System information:**
 - OS: macOS
 - OS version: 12.6.1
 - Synfig version: 1.5.1

**Additional comments**
It works in the stable 1.4.x branch but it doesn't in 1.5.x

mohamedAdhamc:
fixes #2940

The radius was being saved and loaded correctly in class  `StateSmoothMove_Context`. And it seems that this class is responsible for setting the radius of the `DuckDrag_SmoothMove` class as when the spin buttons `signal_value_changed()` fires it updates `DuckDrag_SmoothMove`s radius. However the problem was when the `DuckDrag_SmoothMove` class is constructed its radius is initialized with 1 and not with the value saved in the settings.

### DIFF
--- a/synfig-studio/src/gui/states/state_smoothmove.cpp
+++ b/synfig-studio/src/gui/states/state_smoothmove.cpp
@@ -274,7 +274,7 @@ StateSmoothMove_Context::~StateSmoothMove_Context()
 
 
 
-DuckDrag_SmoothMove::DuckDrag_SmoothMove():radius(1.0f)
+DuckDrag_SmoothMove::DuckDrag_SmoothMove():radius(synfigapp::Main::get_selected_input_device()->settings().get_value("smooth_move.radius", 1.0))
 {
 }
 


synfig-studio/src/gui/states/state_smoothmove.cpp
```diff
@@ -274,7 +274,7 @@ StateSmoothMove_Context::~StateSmoothMove_Context()
 
 
 
-DuckDrag_SmoothMove::DuckDrag_SmoothMove():radius(1.0f)
+DuckDrag_SmoothMove::DuckDrag_SmoothMove():radius(synfigapp::Main::get_selected_input_device()->settings().get_value("smooth_move.radius", 1.0))
```
ice0, line 277: `synfigapp::Main::get_selected_input_device()` may return null, and this results in an exception.

I suggest to move init part to constructor, something like this:
```
DuckDrag_SmoothMove::DuckDrag_SmoothMove():radius(1.0f)
{
	if (const auto device = synfigapp::Main::get_selected_input_device()) {
		radius = device->settings().get_value("smooth_move.radius", 1.0);
	}
```

mohamedAdhamc: > `synfigapp::Main::get_selected_input_device()` may return null.
oh, I didn't notice this. makes sense to move init to constructor then. done :).

mohamedAdhamc:
Hey @pgilfernandez, could you test to see if it's working properly now :) ?

ice0:
@mohamedAdhamc 
I have already checked, it works as expected.

The only reason it's not merged is because we need to change its title.
Since this fix will be added to the release notes, the title should be clear from the user's point of view.

mohamedAdhamc:
Sounds good.

ice0:
> Sounds good.

Unfortunately, I have not figured out how to rephrase this yet, so if you have ideas, you are welcome.

mohamedAdhamc:
> Unfortunately, I have not figured out how to rephrase this yet, so if you have ideas, you are welcome.

I think maybe a variation of the original issue title might make sense, I will give it a try.

mohamedAdhamc:
@rodolforg, any ideas (as a third opinion) on the wording of this fix :) ? (this seems to be the blocker here)

rodolforg:
Hm... no :(

rodolforg (approved):

mohamedAdhamc:
@ice0 this title is a bit on the longer side but I think it should now be descriptive enough for users to know what the fix is about.

PR #2947, opened 2022-12-18
fix: show/hide main window toolbar should affect both toolbars

rodolforg:
previously it was just toggling the right toolbar visibility


rodolforg:
> previously it was just toggling the right toolbar visibility

Please note that this behavior was never intended. Before #1958, there was a single toolbar and #2634 didn't handle it properly (due to a code misleading in variable naming here).

ice0 (approved):

ice0:
Merged. Thank you!

PR #2949, opened 2022-12-18
refactor: move etl::reference_counter to synfig namespace

rodolforg:


rodolforg:
@ice0 ,
1) do you see any reason to `counter` be a pointer, not an `int` directly?
2) should I code move constructor and move assignment operator?

rodolforg:
> do you see any reason to `counter` be a pointer, not an `atomic<int>` directly?

Ok, I got it. It was to be a shared info, of course :P

ice0 (approved):

ice0:
Merged. Thank you!

PR #2950, opened 2022-12-18
fix: get_binary_path() for FreeBSD, DragonFlyBSD, NetBSD and OpenBSD

rodolforg:
procfs path was wrong

https://stackoverflow.com/questions/1023306/finding-current-executables-path-without-proc-self-exe https://man.openbsd.org/OpenBSD-5.1/mount_procfs.8 https://www.freebsd.org/cgi/man.cgi?query=procfs&sektion=5


rodolforg:
I guess I'll move it to os.h later

ice0 (approved):

ice0:
Merged. Thank you!

PR #2951, opened 2022-12-18
fix: change convert.exe syntax to avoid clash in Windows

coz-eduardo-hernandez:
On Windows (I built with CMake) for some weird reason calling ImageMagick's convert calls Windows/System32/convert.exe . It's odd because my Path env var seems to list the mingw64 directory first, but a solution is using the 'magick' syntax, where you just call 'magick' to convert as per https://imagemagick.org/script/command-line-tools.php

The problem is widely known as per https://stackoverflow.com/questions/3060205/error-invalid-parameter-fom-imagemagick-convert-on-windows , I was getting the same 'Invalid Parameter' error

The patch uses 'magick convert' (legacy syntax), instead of 'magick' for compatibility with ImageMagick 6, since many distros don't have official packages for ImageMagick 7, despite the release being over 5 years old:
*  https://packages.debian.org/search?searchon=sourcenames&keywords=imagemagick
*  https://packages.fedoraproject.org/pkgs/ImageMagick/ImageMagick/
*  https://packages.ubuntu.com/search?keywords=imagemagick

The reason is that v7's syntax is not fully compatible with v6 but it our syntax seems safe

TL;DR:
Instead of using ImageMagick with 'convert' use 'magick convert' since Win has another 'convert' program

### DIFF
--- a/synfig-core/src/modules/mod_imagemagick/mptr_imagemagick.cpp
+++ b/synfig-core/src/modules/mod_imagemagick/mptr_imagemagick.cpp
@@ -106,9 +106,9 @@ imagemagick_mptr::get_frame(synfig::Surface &surface, const synfig::RendDesc &re
 
 	args.push_back(strprintf("png32:%s", target_filename.c_str()));
 
-	bool success = OS::run_sync("convert", args);
+	bool success = OS::run_sync("magick", args);
 	if (!success) {
-		synfig::error(_("Unable to open convert [ImageMagick]"));
+		synfig::error(_("Unable to open magick [ImageMagick]"));
 		return false;
 	}
 
--- a/synfig-core/src/modules/mod_imagemagick/mptr_imagemagick.cpp
+++ b/synfig-core/src/modules/mod_imagemagick/mptr_imagemagick.cpp
@@ -99,6 +99,7 @@ imagemagick_mptr::get_frame(synfig::Surface &surface, const synfig::RendDesc &re
 	}
 
 	OS::RunArgs args;
+	args.push_back("convert");
 	args.push_back(filesystem::Path(filename));
 
 	if (filename_extension == ".psd" || filename_extension == ".xcf")
--- a/synfig-core/src/modules/mod_imagemagick/trgt_imagemagick.cpp
+++ b/synfig-core/src/modules/mod_imagemagick/trgt_imagemagick.cpp
@@ -126,6 +126,7 @@ imagemagick_trgt::start_frame(synfig::ProgressCallback *cb)
 		newfilename = filename;
 
 	OS::RunArgs args;
+	args.push_back({"convert"});
 	args.push_back({"-depth", "8"});
 	args.push_back({"-size", strprintf("%dx%d", desc.get_w(), desc.get_h())});
 	args.push_back(pixel_size(pf) == 4 ? "rgba:-[0]" : "rgb:-[0]");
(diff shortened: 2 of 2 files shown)


ice0:
> It's odd because my Path env var seems to list the mingw64 directory first,

Can you describe your build environment/process?
I also experienced this issue before, but I don't remember how to reproduce it.

> but a solution is using the 'magick' syntax, where you just call 'magick' to convert as per 

Yes, I was also thinking about the same solution, but we can't use it because some OSes (like Debian 10) are still using ImageMagick 6, so they don't have the `magick` binary.

https://pkgs.org/search/?q=imagemagick

BobSynfig:
> As per https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa regardless of what you have in PATH dearest windows looks in the 32-bit system directory first when running CreateProcessW in os.cpp .
On 64bit systems, *System32* contains *64bit* dlls, *SysWOW64* contains the *32bit* ones ;)

The implied precedence for system directories is due to the risk of injection of dlls which would be loaded first from the directory of the application.
The same concern was affecting ffmpeg on Windows, you could get inspiration from the [mod_ffmpeg](https://github.com/synfig/synfig/blob/master/synfig-core/src/modules/mod_ffmpeg/trgt_ffmpeg.cpp) module to deal with this.

Other fact that I noticed from 1.4.3 and 1.5.1: there is no trace of "convert" or "magick" binaries while "convert.exe" was present in 1.2.2

coz-eduardo-hernandez:
mod_ffmpeg works by relying on the packager; it assumes ffmpeg.exe will always be in the same directory as synfig.exe, but that fails on the CygWin / MinGW environment

I could implement a fix for imagemagick (traverse PATH looking for convert) but given the inconsistencies with other external dependencies like ffmpeg it feels like a hack. Instead I'm going to work on Synfig by starting from the base. Since that needs majority agreement, I made a new post in the forums. I hope that Konstantin can also chime in too for those general decisions

rodolforg:
> Other fact that I noticed from 1.4.3 and 1.5.1: there is no trace of "convert" or "magick" binaries while "convert.exe" was present in 1.2.2

@morevnaproject

rodolforg:
@coz-eduardo-hernandez please edit commit messages and PR title to use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

coz-eduardo-hernandez:
Since my 'fix' is not convenient at this time, I have no problem if the pull request is closed; an issue could be created for the convert.exe clash instead if desired. That being said, if it would be better to leave this pull request open as a reminder that the clash still happens, that's also fine with me. I'm not really sure what would fit the devs workflow best.

rodolforg:
> > but a solution is using the 'magick' syntax, where you just call 'magick' to convert as per
> 
> Yes, I was also thinking about the same solution, but we can't use it because some OSes (like Debian 10) are still using ImageMagick 6, so they don't have the `magick` binary.

What about test "magick" first? If it doesn't exist, we try traditional "convert" then.

rodolforg:
@coz-eduardo-hernandez @ice0 
what do you think about this: https://github.com/rodolforg/synfig/commit/d78e4f51d29adc634751b71fda97b131d97a24e7 ?

ice0:
Thank you guys!
Merged https://github.com/synfig/synfig/pull/3416 with Rodolfo additions.

PR #2960, opened 2022-12-28
refactor: cleanup module.cpp

ice0:


synfig-core/src/synfig/module.cpp
```diff
@@ -80,51 +71,46 @@ Module::subsys_init(const String &prefix)
...
-	book_=new Book;
+
 	return true;
 }
 
 bool
-Module::subsys_stop()
+synfig::Module::subsys_stop()
 {
-	delete book_;
```
rodolforg, line 118: `book_.clear()`?

ice0: Yeah, I thought about it. But `synfig::Module` is a singleton-like object, and `book_` is a static variable, so I'm not sure it's necessary.

rodolforg (approved):
I thought that `USE_CF_BUNDLES` was somehow related to MacOS...

ice0:
> I thought that `USE_CF_BUNDLES` was somehow related to MacOS...

Same for me, but Google shows only our code for this keyword, so I just removed it.

rodolforg:
I just did a `git grep -iE "CF_?BUND"` and files:

- autobuild/osx/app-template/Contents/Info.plist
- synfig-osx/launcher/Synfig Studio.pbproj/project.pbxproj

have some stuff related to `<key>CFBundle[...]`
I don't understand anything about it, but you know, own and use MacOs. So, approved ;)

PR #2961, opened 2022-12-28
Replaced the getenv() calls

harshvardhan-khachane:
For issue #2836


rodolforg:
Sorry, this isn't a solution.
You seem to blindly replace all `getenv()` matches, without even compiling or properly reading the original issue, as @ice0  said there.
So I'll close this PR. Feel free to create a new one when you understand better the issue and want to try again to properly solve it.

PR #2962, opened 2022-12-28
refactor: replaced deprecated `signal_pressed()`  with `signal_clicked()`

mohamedAdhamc:
Documentation recommended `signal_button_press_event()` but `signal_clicked()` also does the job and even has a similar prototype.
https://docs.gtk.org/gtk3/signal.Button.pressed.html

p.s. I'm not sure though but it seems like these buttons aren't even used. Should they just be removed ?... I saw a button with the same name made in the refresh toobar method.

### DIFF
--- a/synfig-studio/src/gui/states/state_draw.cpp
+++ b/synfig-studio/src/gui/states/state_draw.cpp
@@ -736,7 +736,7 @@ StateDraw_Context::StateDraw_Context(CanvasView* canvas_view):
 	options_grid.set_margin_bottom(0);
 	options_grid.show_all();
 
-	fill_last_stroke_button.signal_pressed().connect(
+	fill_last_stroke_button.signal_clicked().connect(
 		sigc::mem_fun(*this, &StateDraw_Context::fill_last_stroke));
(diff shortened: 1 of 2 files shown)


ice0:
@mohamedAdhamc 

> p.s. I'm not sure though but it seems like these buttons aren't even used. Should they just be removed ?... I saw a button with the same name made in the refresh toolbar method.

Can you try attaching these buttons to `options_grid` and check if they working?
Or you can attach them, and we can ask someone from community to check them.

I am usually vote for deleting old and non-working code, but we need to check it first.

mohamedAdhamc:
hey @ice0, i meant that yes they do work but they are not used... as in they do not show up in the tool. (Actually there are other boxes and buttons in state_lasso especially that are not being used at all). But especially for the `fill_last_stroke_button` its not used and there is also another button with an almost identical name doing the same exact thing in `refresh_tool_options()`.

p.s. I tried to attach them and they seem to working fine they are just not used.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2967, opened 2022-12-30
refactor: removed deprecated `Gtk::VBox` and `get_vbox()`

mohamedAdhamc:
- replaced deprecated Gtk::VBox with Gtk::Box
- replaced deprecated get_v_box() with get_content_area()

before:
vectorizerBefore

after:
vectorizerAfter

also tested before and after layout and expansion is the same as it was.


ice0 (approved):

ice0:
Merged. Thank you!

PR #2968, opened 2023-01-01
unhide using xpath

3l3gant-cod3s:
if you find it interesting…

### DIFF
--- a/synfig-studio/plugins/view-unhide-all-layers/view-unhide-all-layers.py
+++ b/synfig-studio/plugins/view-unhide-all-layers/view-unhide-all-layers.py
@@ -20,6 +20,8 @@ def process(filename):
 	for inactive_layer in root.findall('layer[@active="false"]'):
 		inactive_layer.set('active', 'true')
 	tree.write(filename, encoding='UTF-8')
+
+if len(sys.argv) < 2:
 	sys.exit()
 else:
 	process(sys.argv[1])
--- a/synfig-studio/plugins/view-unhide-all-layers/view-unhide-all-layers.py
+++ b/synfig-studio/plugins/view-unhide-all-layers/view-unhide-all-layers.py
@@ -2,6 +2,7 @@
 
 #
 # Copyright (c) 2012 by Konstantin Dmitriev <ksee.zelgadis@gmail.com>
+# Copyright (c) 2022 by Erwan le Gall <synfig@elegant.codes>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


ice0:
@3l3gant-cod3s 

Hi! What about speed? Can you test it on big file?

3l3gant-cod3s:
I'm beginning to use Synfig so I don't have a heavy .sif file at hand. I tried with [this one](https://bonedaddy.net/pabs3/log/2008/01/31/synfig-january-challenge/tshirt.sifz) (1.6 Mio) but there's no noticeable difference between the 2 implementations.

Etree is an very efficient implementation (and the xpath expression only extract inactive layers).
cf https://eli.thegreenplace.net/2012/03/15/processing-xml-in-python-with-elementtree
> Finally, there's xml.etree.ElementTree (from now on, ET in short). It provides a lightweight Pythonic API, backed by an efficient C implementation, for parsing and creating XML. Compared to DOM, **ET is much faster** [[3]](https://eli.thegreenplace.net/2012/03/15/processing-xml-in-python-with-elementtree#id10) and has a more pleasant API to work with. Compared to SAX, there is ET.iterparse which also provides "on the fly" parsing without loading the whole document into memory. **The performance is on par with SAX**, but the API is higher level and much more convenient to use; it will be demonstrated later in the article.

There's some other advantages too: it's more robust as it is guaranteed that you are modifying an “active” attribute of a “layer” element (and not this attribute on another element).

3l3gant-cod3s:
I'll be pleased to test it on an heavy .sif/.sifz file :)

mohamedAdhamc:
Hey @3l3gant-cod3s, if you don't have big files to try on, I believe you can ask for one in the forums I believe some may find it interesting and could send you files to test on.

3l3gant-cod3s:
Hi @mohamedAdhamc and @ice0 !
Sorry, I’ve discovered the synfig-examples folder :/
(I said I was brand new to Synfig)
I wrote a hide-all (1726) layers plugin ;)
I tested unhiding-all against pirates.sif and there's no noticeable differences between Konstantin's implementation and mine (~15s each on my 10-years computer).

3l3gant-cod3s:
Using a more rigorous strategy, I have "instrumented" both codes (K’s and mine) and mine is ~10× slower (3s versus .26s) for the core operation (changing the states of layers).
But the whole operation takes ~12-15s. :)
(it's probably redrawing that takes a long time for such a bunch of layers)
NB @BobSynfig made a comment on my request in the plugin forum.

ice0:
@3l3gant-cod3s 

Docs say ElementTree is deprecated:
https://docs.python.org/3/library/xml.etree.elementtree.html

BobSynfig:
Deprecated because [vulnerable](https://docs.python.org/3/library/xml.html#xml-vulnerabilities) to some attacks, but still widely used
Instead we could use [defusedxml](https://pypi.org/project/defusedxml/) (need to add it as dependency)

3l3gant-cod3s:
@ice0 it's xml.etree.**c**ElementTree module that is deprecated. I used xml.etree.ElementTree module (without a "c").
@BobSynfig I patched with defusedxml instead of xml.etree, works like a charm.

PR #2971, opened 2023-01-14
fix: crash on moving a layer following a bline reference

rodolforg:
https://forums.synfig.org/t/reference-conversion-crash/13774

Reported-by: woody

### DIFF
--- a/synfig-studio/src/gui/duckmatic.cpp
+++ b/synfig-studio/src/gui/duckmatic.cpp
@@ -606,7 +606,18 @@ Duckmatic::update_ducks()
 			{
 				synfig::Real radius = 0.0;
 				synfig::Point point(0.0, 0.0);
-				ValueNode_BLine::Handle bline(ValueNode_BLine::Handle::cast_dynamic(bline_vertex->get_link("bline")));
+				ValueNode_BLine::Handle bline;
+				if (auto vn = bline_vertex->get_link("bline")) {
+					if (vn->get_name() == "reference") {
+						vn = LinkableValueNode::Handle::cast_static(vn)->get_link("link");
+					}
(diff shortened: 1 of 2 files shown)


synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -606,7 +606,18 @@ Duckmatic::update_ducks()
 			{
 				synfig::Real radius = 0.0;
 				synfig::Point point(0.0, 0.0);
-				ValueNode_BLine::Handle bline(ValueNode_BLine::Handle::cast_dynamic(bline_vertex->get_link("bline")));
+				ValueNode_BLine::Handle bline;
```
ice0, line 609: Maybe extract this as a method?
Also similar code in duck.cpp:291
```
ValueNode_BLine::Handle bline = ValueNode_BLine::Handle::cast_dynamic(bline_vertex->get_link("bline"));
```

rodolforg: I can't see where I could place such method. :/
I'll implement another way

ice0: How about something like this?
https://github.com/synfig/synfig/commit/0d5287a235a055774982d70bb14e29919460021f

rodolforg: Wow! Got it! :D

rodolforg: Done!

ice0: Can I squash commits (for easy porting)?

rodolforg: Sure

ice0 (approved):

ice0:
Merged. Thank you!

PR #2974, opened 2023-01-15
fix: preview dialog now shows the correct start and end times

### LINKED ISSUE #2969: Wrong time in Preview if canvas start time is different from zero
**Description**
Preview window doesn't calculate "Start" and "End" time if start time of the canvas is anything different from zero. Although it does render it correctly, so it's just visuals. Still, a bit confusing.

**How to reproduce:**
* Open new composition
* Go to "Canvas->Properties->Time"
* Set "Start Time" to 2s, "End Time" to 5s
* Hit "Preview" and you will see something this:
Preview
Notice how under "Time settings" times are not displayed correctly.

**Expected behavior:**
Preview window should display start time as 2s and end time as 5s.

**System information:**
Not important.

**Affected versions:**
Bug is present in all versions.

mohamedAdhamc:
fixes #2969 
But I haven't tested it enough I think so I will leave it as draft until I get enough time to do so.
@Svarov-RZM could you try it :)

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -3415,9 +3415,6 @@ CanvasView::on_preview_option()
 		RendDesc &r = canv->rend_desc();
 		if(r.get_frame_rate())
 		{
-			float beg = r.get_time_start();
-			float end = r.get_time_end();
-
 			Dialog_PreviewOptions *po = dynamic_cast<Dialog_PreviewOptions *>( get_ext_widget("prevoptions") );
 			if(!po)
 			{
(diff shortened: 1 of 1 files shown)


Svarov-RZM:
> Svarov-RZM could you try it :)

Of course.
I tested it with different time ranges and FPS and it works as expected now.

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -3415,9 +3415,8 @@ CanvasView::on_preview_option()
 		RendDesc &r = canv->rend_desc();
 		if(r.get_frame_rate())
 		{
-			float rate = 1/r.get_frame_rate();
-			float beg = r.get_time_start() + r.get_frame_start()*rate;
-			float end = r.get_time_start() + r.get_frame_end()*rate;
+			float beg = r.get_time_start();
```
ice0, line 3418: You can go further and remove these variables because they were only used once:
```
po->set_begintime(r.get_time_start());
```

mohamedAdhamc: makes sense, done :)

ice0 (approved):

ice0:
Merged. Thank you!

PR #2988, opened 2023-01-24
fix: crash occurs under Wayland when using tablet

alexhaskell:
GTK gives Synfig a GdkEvent containing a GdkDevice with the ID "Master pointer for ..." when using a tablet device. The device is not found, causing a crash elsewhere.

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -3403,6 +3403,10 @@ CanvasView::on_input_device_changed(GdkDevice* device)
 
 	InputDevice::Handle input_device;
 	input_device = synfigapp::Main::select_input_device(gdk_device_get_name(device));
+
+	if (!input_device) {
+		return;
+	}
 	App::dock_toolbox->change_state(input_device->get_state(), true);
 	process_event_key(EVENT_INPUT_DEVICE_CHANGED);
 }
--- a/synfig-studio/src/gui/workarea.cpp
+++ b/synfig-studio/src/gui/workarea.cpp
@@ -1083,7 +1083,7 @@ WorkArea::on_drawing_area_event(GdkEvent *event)
 	// Handle input stuff
 	if (event->any.type==GDK_MOTION_NOTIFY)
 	{
-		GdkDevice *device = event->motion.device;
+		GdkDevice *device = gdk_event_get_source_device(event);
 		modifier = Gdk::ModifierType(event->motion.state);
 
 		// Calculate the position of the
@@ -1148,7 +1148,7 @@ WorkArea::on_drawing_area_event(GdkEvent *event)
 		event->any.type==GDK_3BUTTON_PRESS ||
 		event->any.type==GDK_BUTTON_RELEASE )
 	{
-		GdkDevice *device = event->button.device;
+		GdkDevice *device = gdk_event_get_source_device(event);
 		modifier = Gdk::ModifierType(event->button.state);
 		drawing_area->grab_focus();
 
--- a/synfig-studio/src/synfigapp/main.cpp
+++ b/synfig-studio/src/synfigapp/main.cpp
@@ -365,9 +365,13 @@ synfigapp::Main::add_input_device(const synfig::String id, InputDevice::Type typ
 InputDevice::Handle
 synfigapp::Main::find_input_device(const synfig::String id)
 {
+	static const char MASTER_POINTER_FOR[] = "Master pointer for ";
+	static const size_t MASTER_POINTER_FOR_LENGTH = sizeof (MASTER_POINTER_FOR) - 1;
+	const size_t offset = id.rfind(MASTER_POINTER_FOR, 0) == 0 ? MASTER_POINTER_FOR_LENGTH : 0;
+
 	std::list<InputDevice::Handle>::iterator iter;
 	for(iter=input_devices_.begin();iter!=input_devices_.end();++iter)
-		if((*iter)->get_id()==id)
+		if(id.compare(offset, std::string::npos, (*iter)->get_id()) == 0)
 			return *iter;
 	return 0;
 }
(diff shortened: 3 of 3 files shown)


rodolforg:
Usually a crash is due to:
* a null pointer dereference
* a freed pointer dereference

So, if you really want to fix the *crash* itself, you have to find one of these problems.
As you said, `find_input_device()` returns null (please replace it to `nullptr`, by the way).
If you don't use any IDE, use `git grep` to search this string `find_input_device` to check where it is called and investigate if/where it can be dereferenced without checking for null.

alexhaskell:
@rodolforg  Done.

The only place that a null from find_input_device() was not properly handled was in CanvasView::on_input_device_changed() calling select_input_device().

rodolforg:
Great! 
And the string "Master pointer for" is still needed, right?

alexhaskell:
Yes, "Master pointer for" seems to be necessary on Wayland.

I'm currently running into a problem where the device name is not distinguished between the pen and eraser. Both show up as "Master pointer for Wacom Intuos Pro L Pen". The input devices editor lists the eraser as "Wacom Intuos Pro L Pen (Eraser)". That's a bug for another time it seems.

rodolforg:
It should appear a pen and a eraser item

alexhaskell:
I can't find where "(Eraser)" is added by Synfig. Do you know which file?

I added a logging statement that prints the device name from gdk_get_device_name() in WorkArea::on_drawing_area_event() (motion and click events). Both the pen and eraser have the same name.

rodolforg:
> I can't find where "(Eraser)" is added by Synfig. Do you know which file?

Neither do I. LOL.  I thought I saw it yesterday... Sorry



> I added a logging statement that prints the device name from gdk_get_device_name() in WorkArea::on_drawing_area_event() (motion and click events). Both the pen and eraser have the same name.

Did you enabled them in Input Device dialog? 
Sorry I can't help you here: I don't have any kind of these devices, I don't use Wayland and I don't even know what "pen" "eraser" and "cursor" are XD

alexhaskell:
Strangely, the eraser is marked as "disabled" in the Device Editor, but I can still use it in synfig. Changing it from "disabled" to "screen" or "window" does not get saved when I close the dialog. Nothing in the dialog that I do gets saved, every device gets reset.

Core Keyboard: Screen
Core Pointer: Screen
Wacom Intuos L Pen (Eraser): Disabled
Wacom Intuos L Pen: Screen
Wayland Keyboard: Disabled
Wayland Pointer: Disabled

rodolforg:
Did you test on 1.5.1 official version too?

alexhaskell:
Same thing happens in 1.5.1. No changes to devices gets saved.

ice0:
Hi, guys!
@alexhaskell 
Usually, if you add a hardcoded string or something like that, then there must be another solution.

In this case, the reason for this problem is that Synfig uses source (slave) devices, and events come from a point (master) device.
This looks like this:
```
carlos@sacarino:~$ xinput list
⎡ Virtual core pointer                          id=2    [master pointer  (3)]
⎜   ↳ Virtual core XTEST pointer                id=4    [slave  pointer  (2)]
⎜   ↳ Wacom ISDv4 E6 Pen stylus                 id=10   [slave  pointer  (2)]
⎜   ↳ Wacom ISDv4 E6 Finger touch               id=11   [slave  pointer  (2)]
⎜   ↳ SynPS/2 Synaptics TouchPad                id=13   [slave  pointer  (2)]
⎜   ↳ TPPS/2 IBM TrackPoint                     id=14   [slave  pointer  (2)]
⎜   ↳ Wacom ISDv4 E6 Pen eraser                 id=16   [slave  pointer  (2)]
⎣ Virtual core keyboard                         id=3    [master keyboard (2)]
    ↳ Virtual core XTEST keyboard               id=5    [slave  keyboard (3)]
    ↳ Power Button                              id=6    [slave  keyboard (3)]
    ↳ Video Bus                                 id=7    [slave  keyboard (3)]
    ↳ Sleep Button                              id=8    [slave  keyboard (3)]
    ↳ Integrated Camera                         id=9    [slave  keyboard (3)]
    ↳ AT Translated Set 2 keyboard              id=12   [slave  keyboard (3)]
    ↳ ThinkPad Extra Buttons                    id=15   [slave  keyboard (3)]
```

So to get the original device you need to use `gdk_event_get_source_device(...)` on original GDK event.
In that case you will get the correct name for the device.

Please check this document for details: https://docs.gtk.org/gdk3/class.DeviceManager.html

alexhaskell:
Thanks @ice0! It looks like gdk_event_get_source_device() is what we wanted. I completely missed that when @rodolforg mentioned that page the first time.

Pushing my changes now.

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -3403,6 +3403,10 @@ CanvasView::on_input_device_changed(GdkDevice* device)
 
 	InputDevice::Handle input_device;
 	input_device = synfigapp::Main::select_input_device(gdk_device_get_name(device));
+
+	if (!input_device) {
+		return;
+	}
```
rodolforg, line 3409: Please keep this crash-avoider! :+1:

rodolforg:
> Strangely, the eraser is marked as "disabled" in the Device Editor, but I can still use it in synfig. Changing it from "disabled" to "screen" or "window" does not get saved when I close the dialog. Nothing in the dialog that I do gets saved, every device gets reset.
> 
> Core Keyboard: Screen Core Pointer: Screen Wacom Intuos L Pen (Eraser): Disabled Wacom Intuos L Pen: Screen Wayland Keyboard: Disabled Wayland Pointer: Disabled

This problem still happens, right?

alexhaskell:
The input dialog still doesn't work. I'll take a look at that.

I have a fix for a bug regarding tablet coordinates that I uncovered after I got past the crash. I'll be opening another pull request unless either of you want it added to this pull request.

ice0:
> The input dialog still doesn't work. I'll take a look at that.

Maybe it's just not finished. You can check the git history for more details.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2991, opened 2023-01-27
fix: drawing tablet didn't work on Wayland

alexhaskell:
Even though gdk_device_get_axis() returns true for a drawing tablet, the x/y values it gives are always 0.0. Pressure is correct, however.


ice0:
I tested this PR (both with #2988) and all works fine.

Tested OS:
- Fedora 37/Wayland
- Mint 21/Xorg
- macOS 12.6
- Windows 10

The only issue I noticed is a strange GTK critical warning on macOS, but it doesn't seem to be related to this PR.

ice0 (approved):

ice0:
Merged. Thank you!

Issue #2993, opened 2023-01-28
Save & record width/height toggle state too for Canvas Resize dialog


Keyikedalube:
> Do we want the link and unlink state to be saved when the dialog changes are applied? Canvas properties doesn't. It reverts to toggled state when:

I think yes. It's not noticeable in Canvas properties because they're complex, but here I noticed it right away.

_Originally posted by @ice0 in https://github.com/synfig/synfig/issues/1895#issuecomment-1407260313_

PR #2997, opened 2023-01-30
fix: save widht/height toggle state for Canvas Resize dialog

Keyikedalube:
WIP fix for #2993

### DIFF
--- a/synfig-studio/src/gui/dialogs/canvasresize.cpp
+++ b/synfig-studio/src/gui/dialogs/canvasresize.cpp
@@ -167,8 +167,8 @@ void CanvasResize::on_dialog_shown()
 	if (is_image_checked) set_image_flags(toggle_ratio_wh->get_active());
 
 	// Always enable width/height link
-	set_canvas_flags(true);
-	refresh_wh_toggle_widgets();
+//	set_canvas_flags(true);
+//	refresh_wh_toggle_widgets();
 }
 
 void CanvasResize::on_rend_desc_changed()


synfig-studio/src/gui/dialogs/canvasresize.cpp
```diff
@@ -167,8 +167,8 @@ void CanvasResize::on_dialog_shown()
 	if (is_image_checked) set_image_flags(toggle_ratio_wh->get_active());
 
 	// Always enable width/height link
-	set_canvas_flags(true);
-	refresh_wh_toggle_widgets();
+//	set_canvas_flags(true);
+//	refresh_wh_toggle_widgets();
```
Keyikedalube, line 171: This was the only way to default turn on width/height link

But the side effect was the code disregards user choice and always toggles the link back on when the dialog is shown. Code imitated from Canvas Properties class.

synfig-studio/src/gui/dialogs/canvasresize.cpp
```diff
@@ -166,7 +167,19 @@ void CanvasResize::on_dialog_shown()
 {
 	if (is_image_checked) set_image_flags(toggle_ratio_wh->get_active());
 
-	// Always enable width/height link
+	/*
```
ice0, line 170: 1) I suggest moving this comment next to the variable declaration so it can be found from anywhere.
2) Maybe name this variable something like `switch_changed_manually`? This should better describe it.
3) It looks like this code could be simplified somehow, but I need to test some ideas. So I will back tomorrow.

rodolforg: @Keyikedalube ping =]

Keyikedalube: Waiting for improvement highlighted in point 3 LOL

mohamedAdhamc:
Hey @Keyikedalube, I was testing out this PR.

From what I understand originally like in canvas properties dialog the link is always active regardless if it was toggled by the user before changes are applied to the dialog. Now this PR should be saving the state so that when it's toggled and changes are applied and then I reopen it the toggle state is "saved"... unfortunately this is not the behaviour I was getting while testing :(. The width/height toggle was always set to linked on dialog reopen.

synfig-studio/src/gui/dialogs/canvasresize.cpp
```diff
@@ -195,7 +210,11 @@ void CanvasResize::on_action_signal_response(int response_id)
 		refresh_content();
 		break;
 	case OKAY:
-		if (new_width == old_width && new_height == old_height) break;
+		if (new_width == old_width && new_height == old_height)
+			if (is_toggled == was_toggled && is_toggle_unique) break;
```
mohamedAdhamc, line 214: `is_toggled` is of type "bool" and `was_toggled` is an "int". Is this intended ? 

c and cpp are quite permissive about comparisons of bools and ints but in general I think it should be better avoided probably unless absolutely needed.

Keyikedalube: To be on the safe side we could cast `was_toggled` to `bool`

Keyikedalube:
> unfortunately this is not the behavior I was getting while testing :(

I just got back to this PR and tested my changes. Weird logic noted.

Editing width value while toggled and then untoggling it before saving should save the last action (untoggled) but it's not. On reopening the dialog the link still remains toggled.

I got time to look into this and possibly fix the issue tomorrow.

> * it works when the very first time the resize dialog is opened the link isn't toggled

Because Canvas properties and Resize dialogs each have different renddesc objects.

The former resizes content by default. The latter resize the canvas.

So even before the user edits the value on Resize dialog, renddesc object is already temporarily modified and waiting to be saved. Before the save state is finalized, toggle link state is a mess and the solution I came up with was to track `is_toggle_unique` variable.

synfig-studio/src/gui/dialogs/canvasresize.cpp
mohamedAdhamc: How about just making `was_toggled` itself a boolean if it's only going to be used as such ?

synfig-studio/src/gui/dialogs/canvasresize.cpp
Keyikedalube: Yikes! I missed step 2...
Step 1 done

Keyikedalube:
Now turning off/on toggle will keep its button un/toggled unless the user hits:
- **Cancel** (dialog action button) or 
- **Close** (dialog title bar)

This is because temporary changes to the dialog is refreshed for the above two events.
Safe way is to toggle and hit **OK** to keep the button toggled ie., save the toggle state for next resize.

Thanks in advance to all those who take their time to test this PR and report issues :)

PR #2999, opened 2023-02-03
build: change exported cmake targets and provide a way to find deps

theartful:
I wanted to write a program that links against installed synfig, but the default generated cmake config file links against synfig's dependencies targets without providing a way to find them.

Now something like this works:
```cmake
find_package(synfig)
target_link_libraries(main synfig::libsynfig)
```

I also took the liberty to change the cmake installation folder from "synfig-core" to "synfig", but I'm not sure if that's okay.

### DIFF
--- a/synfig-core/src/synfig/CMakeLists.txt
+++ b/synfig-core/src/synfig/CMakeLists.txt
@@ -149,13 +149,28 @@ install(
     RUNTIME DESTINATION bin
 )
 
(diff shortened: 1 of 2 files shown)


ice0 (approved):

ice0:
Looks good to me.

ice0:
Merged. Thank you!

PR #3001, opened 2023-02-05
fix: check type equality before replacing rhandles

theartful:
I was reading up on `etl::rshared_object`, and found this potential bug. If we have a number of rhandles pointing to the same object but with different types in the hierarchy tree, it is possible to replace the rhandle  with a parent class, which will cause rhandles with child classes to be in undefined state.
Example:
```C++
class Object : public etl::rshared_object
{
private:
    int some_data_to_make_class_not_empty = 0;
};

class Child : public Object
{
private:
    int* data_ptr;

public:
    Child() { data_ptr = new int(10); }
    ~Child() { delete data_ptr; }

    int data() { return *data_ptr; }
};

void buggy() {
    etl::handle<Child> obj{new Child()};
    etl::handle<Object> obj2{new Object()};

    etl::rhandle<Object> obj_replaceable1{obj};
    etl::rhandle<Child> obj_replaceable2{obj};

    // prints 10
    std::cout << "obj_replaceable2->data = " << obj_replaceable2->data()
              << '\n';

    obj_replaceable1.replace(obj2);

    // will crash
    std::cout << "obj_replaceable2->data = " << obj_replaceable2->data()
              << '\n';
}
```

### DIFF
--- a/ETL/ETL/_handle.h
+++ b/ETL/ETL/_handle.h
@@ -319,6 +319,10 @@ public:
 	template<typename U>
 	bool type_equal() const
 	{ return typeid(*obj) == typeid(U); }
+
+	template <typename U>
+	bool type_equal(const etl::handle<U> handle) const
+	{ return typeid(*obj) == typeid(*handle.get()); }
 }; // END of template class handle
 
 // ========================================================================
@@ -589,9 +593,13 @@ public:
 //		value_type*& obj(handle<T>::obj); // Required to keep gcc 3.4.2 from barfing
 		assert(obj);
 		assert(x.get()!=obj);
+		assert(this->type_equal(x));
 
 		if(x.get()==obj)
 			return 0;
+
+		if(!this->type_equal(x))
+			return -1;
 
 		rhandle<value_type> *iter;
 		rhandle<value_type> *next;


rodolforg (commented):
`replace()` could return only a natural number (positive integer or zero), but now it can return `-1`.

So we should:
- check what calls `rhandle::replace()` and how it uses the returned value to fix/adapt it if necessary; OR
- just return 0 even on replacement error due to this value type difference you detected.

theartful:
What about throwing an exception? I feel like crashing knowing the reason is better than failing silently and causing hard to track bugs.

rodolforg:
> I feel like crashing knowing the reason is better than failing silently and causing hard to track bugs.

Don't say that for users that loss hours of work due to crashing ;)

ice0:
> > I feel like crashing knowing the reason is better than failing silently and causing hard to track bugs.
> 
> Don't say that for users that loss hours of work due to crashing ;)

But it will crash even without an exception. At least we will know where it happens.
I also thought about logging it with a GTK fatal warning, in this case we can enable the crash using an environment variable.

rodolforg:
So, what should be done? @ice0 :)

ice0:
I suggest to log error for that case and continue to work as it was before.
This will help us understand if the problem exists without changing the behavior.
`assert` is ok too for Debug builds.

Issue #2953, opened 2022-12-21
Feature request: "add waypoint" menu entry everywhere


pgilfernandez:
**Is your feature request related to a problem? Please describe.**
When adding waypoints, it's always needed to first activate the "animation mode" and then edit values. There is not a fast and easy way of adding waypoints just by going into the Parameters panel.
<img width="392" alt="params" src="https://user-images.githubusercontent.com/5942369/208893507-ad1bc1be-dc45-4555-939a-d6627b73959f.png">

Only when you have an already animated parameter, you have that "add waypoint" option in the secondary mouse button menu of that parameter and you can add a waypoint no matter if you are in animation mode or not (I like it).
<img width="390" alt="params_animated" src="https://user-images.githubusercontent.com/5942369/208893542-1774c90b-5183-48ad-be8f-4d1cd6e8398c.png">

**Describe the solution you'd like**
I think it would be very handy to have that "add waypoint" menu entry always available and in all "animatable" parameters (I mean, if animation is not allowed for that parameter then it makes sense not to include it).
It would allow users to add waypoints out of the animation mode and, more useful, to add a waypoint with the current value in parameters that haven't still been animated (and are allowed to be animated, indeed).

Finally, it would be handy to have a keyboard shortcut to add a waypoint with the selected parameter alla Blender as well (where you hit "i" and it adds an interpolation keyframe).

**Describe alternatives you've considered**
none

**Additional comments**
Thanks!

morevnaproject:
Yes, we need that ^__^

mohamedAdhamc:
sounds interesting, i'd like to work on this but life is quite busy rn xD. Hopefully though I will get going with it soon enough :).

mohamedAdhamc:
At last, I managed to find enough time to look further into this. I will continue tommorow hopefully, but so far the biggest road block here is that in order to add the waypoint there has to be a value node initialized in the `WaypointSetSmart` action, the only way I know of which does this is through the `ValueDescSet` action which is initiated when the user creates the very first waypoint for a parameter when in animate mode. 

How the valuenode could be initialized for the `WaypointSetSmart` in a clean way ? I'm not really sure, and the ideas that came to mind at first all seem like they would be very patchy. I will continue exploring all this maybe there is something I'm missing. But @ice0 @rodolforg any suggestions would be appreciated :).

PR #3005, opened 2023-02-13
chore: fix typo in comment in `valuedescset.h`

mohamedAdhamc:
Found it while working on #2953. The method creates a valuedescset action and adds it to the stack (which seems to be the undoable action list but I didn't look too much as that wasnt directly related to what I was doing).


ice0 (approved):

ice0:
Merged. Thank you!

PR #3006, opened 2023-02-15
feat: always show "add waypoint" menu for animatable parameters

### LINKED ISSUE #2953: Feature request: "add waypoint" menu entry everywhere

mohamedAdhamc:
fixes #2953 

Implementation description:
For the first add waypoint action the action is redirected to a `ValueDescSet` action. then rest are regular add waypoint actions.

### DIFF
--- a/synfig-studio/src/synfigapp/actions/valuedescset.cpp
+++ b/synfig-studio/src/synfigapp/actions/valuedescset.cpp
@@ -157,6 +157,7 @@ Action::ValueDescSet::set_param(const synfig::String& name, const Action::Param 
 	if(name=="add_waypoint_initial" && param.get_type()==Param::TYPE_BOOL)
 	{
 		add_waypoint_initial = param.get_bool();
+		return true;
 	}
 	if(name=="value_desc" && param.get_type()==Param::TYPE_VALUEDESC)
 	{
@@ -807,7 +808,7 @@ Action::ValueDescSet::prepare()
 
 	// If we are in animate editing mode or adding the first waypoint for a parameter from parameter menu
 	// TODO: Can we replace local_value to value after all parameters will be converted into ValueBase type?
-	if((add_waypoint_initial ||(( animate || (get_edit_mode()&MODE_ANIMATE)))) && !value_desc.get_static())
+	if((add_waypoint_initial ||( animate || (get_edit_mode()&MODE_ANIMATE))) && !value_desc.get_static())
 	{
 		ValueNode_Animated::Handle& value_node(value_node_animated);
 		// If this value isn't a ValueNode_Animated, but
--- a/synfig-studio/src/synfigapp/actions/valuedescset.cpp
+++ b/synfig-studio/src/synfigapp/actions/valuedescset.cpp
@@ -98,7 +98,8 @@ Action::ValueDescSet::ValueDescSet():
 	time(0),
 	recursive(true),
 	animate(false),
-	lock_animation(false)
+	lock_animation(false),
+	add_waypoint_initial(false)
 {
 }
 
@@ -153,6 +154,10 @@ Action::ValueDescSet::is_candidate(const ParamList &x)
 bool
 Action::ValueDescSet::set_param(const synfig::String& name, const Action::Param &param)
 {
+	if(name=="add_waypoint_initial" && param.get_type()==Param::TYPE_BOOL)
+	{
+		add_waypoint_initial = param.get_bool();
+	}
 	if(name=="value_desc" && param.get_type()==Param::TYPE_VALUEDESC)
 	{
 		value_desc=param.get_value_desc();
@@ -800,9 +805,9 @@ Action::ValueDescSet::prepare()
 		}
 	}
 
-	// If we are in animate editing mode
+	// If we are in animate editing mode or adding the first waypoint for a parameter from parameter menu
 	// TODO: Can we replace local_value to value after all parameters will be converted into ValueBase type?
-	if((animate || (get_edit_mode()&MODE_ANIMATE)) && !value_desc.get_static())
+	if((add_waypoint_initial ||(( animate || (get_edit_mode()&MODE_ANIMATE)))) && !value_desc.get_static())
 	{
 		ValueNode_Animated::Handle& value_node(value_node_animated);
 		// If this value isn't a ValueNode_Animated, but
--- a/synfig-studio/src/synfigapp/actions/valuedescset.cpp
+++ b/synfig-studio/src/synfigapp/actions/valuedescset.cpp
@@ -98,8 +98,7 @@ Action::ValueDescSet::ValueDescSet():
 	time(0),
 	recursive(true),
 	animate(false),
-	lock_animation(false),
-	add_waypoint_initial(false)
+	lock_animation(false)
 {
 }
 
@@ -154,11 +153,6 @@ Action::ValueDescSet::is_candidate(const ParamList &x)
 bool
 Action::ValueDescSet::set_param(const synfig::String& name, const Action::Param &param)
 {
-	if(name=="add_waypoint_initial" && param.get_type()==Param::TYPE_BOOL)
-	{
-		add_waypoint_initial = param.get_bool();
-		return true;
-	}
 	if(name=="value_desc" && param.get_type()==Param::TYPE_VALUEDESC)
 	{
 		value_desc=param.get_value_desc();
@@ -806,9 +800,9 @@ Action::ValueDescSet::prepare()
 		}
 	}
 
-	// If we are in animate editing mode or adding the first waypoint for a parameter from parameter menu
+	// If we are in animate editing mode
 	// TODO: Can we replace local_value to value after all parameters will be converted into ValueBase type?
-	if((add_waypoint_initial ||( animate || (get_edit_mode()&MODE_ANIMATE))) && !value_desc.get_static())
+	if((animate || (get_edit_mode()&MODE_ANIMATE)) && !value_desc.get_static())
 	{
 		ValueNode_Animated::Handle& value_node(value_node_animated);
 		// If this value isn't a ValueNode_Animated, but
--- a/synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
+++ b/synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
@@ -115,7 +115,9 @@ Action::WaypointSetSmart::is_candidate(const ParamList &x)
 			// We need either a waypoint or a time.
 			(x.count("waypoint") || x.count("time")) &&
 			// Parameter has to be animatable
-			!(x.find("value_desc")->second.get_value_desc()).get_static());
+			!(x.find("value_desc")->second.get_value_desc()).get_static() &&
+			//no expandable parameters
+			!(LinkableValueNode::Handle::cast_dynamic(x.find("value_desc")->second.get_value_desc().get_value_node())));
 }
 
 bool
@@ -185,6 +187,10 @@ Action::WaypointSetSmart::is_ready()const
 void
 Action::WaypointSetSmart::calc_waypoint()
 {
+	if(!value_node){
+		redirect_value_desc_set_action();
+		return;
+	}
 	Time time=waypoint.get_time();
 	try
 	{
@@ -369,21 +375,8 @@ Action::WaypointSetSmart::enclose_waypoint(const synfig::Waypoint& waypoint)
 void
 Action::WaypointSetSmart::prepare()
 {
-	if(!value_node){//this is the first addition of a waypoint, redirect to valuedesctset action
-		synfigapp::Action::Handle action(synfigapp::Action::create("ValueDescSet"));
-		if(!action)
-		{
-			return;
-		}
-
-		action->set_param("canvas",get_canvas());
-		action->set_param("canvas_interface",get_canvas_interface());
-		action->set_param("time",waypoint.get_time());
-		action->set_param("value_desc",value_desc);
-		action->set_param("new_value",value_desc.get_value(waypoint.get_time()));
-		action->set_param("animate",true);
-
-		add_action(action);
+	if(!value_node){
+		redirect_value_desc_set_action();
 		return;
 	}
 
@@ -476,3 +469,21 @@ Action::WaypointSetSmart::prepare()
 
 	throw Error(_("Unable to determine how to proceed. This is a bug."));
 }
+void
+Action::WaypointSetSmart::redirect_value_desc_set_action()
+{
+	synfigapp::Action::Handle action(synfigapp::Action::create("ValueDescSet"));
+	if(!action)
+	{
+		return;
+	}
+	action->set_param("canvas",get_canvas());
+	action->set_param("canvas_interface",get_canvas_interface());
+	action->set_param("time",waypoint.get_time());
+	action->set_param("value_desc",value_desc);
+	action->set_param("new_value",value_desc.get_value(waypoint.get_time()));
+	action->set_param("animate",true);
+
+	add_action(action);
+}
+
--- a/synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
+++ b/synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
@@ -85,6 +85,7 @@ Action::WaypointSetSmart::get_param_vocab()
 
 	ret.push_back(ParamDesc("value_desc",Param::TYPE_VALUEDESC)
 		.set_local_name(_("ValueDesc"))
+		.set_optional()
 	);
 
 	ret.push_back(ParamDesc("value_node",Param::TYPE_VALUENODE)
@@ -118,6 +119,8 @@ Action::WaypointSetSmart::is_candidate(const ParamList &x)
 	return (candidate_check(get_param_vocab(),x) &&
 			// We need either a waypoint or a time.
 			(x.count("waypoint") || x.count("time")) &&
+			// make sure a valuedesc exists
+			(x.find("value_desc") != x.end()) &&
 			// Parameter has to be animatable
 			!(x.find("value_desc")->second.get_value_desc()).get_static() &&
 			//no expandable parameters
@@ -138,8 +141,10 @@ Action::WaypointSetSmart::set_param(const synfig::String& name, const Action::Pa
 
 	if(name=="value_desc" && param.get_type()==Param::TYPE_VALUEDESC)
 	{
-		value_desc=param.get_value_desc();
+		if(!param.get_value_desc().is_valid())
+			return false;
 
+		value_desc=param.get_value_desc();
 		return true;
 	}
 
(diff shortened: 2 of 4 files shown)


mohamedAdhamc:
@pgilfernandez , and @morevnaproject I believe the feature is now finally implemented. It would be great if both or either of you test the functionality : ).

pgilfernandez:
I'll try to test is this week and I will let you know,

Thanks!

synfig-studio/src/synfigapp/actions/valuedescset.cpp
```diff
@@ -98,7 +98,8 @@ Action::ValueDescSet::ValueDescSet():
 	time(0),
 	recursive(true),
 	animate(false),
-	lock_animation(false)
+	lock_animation(false),
+	add_waypoint_initial(false)
```
rodolforg, line 102: couldn't the `animate` parameter be used?

mohamedAdhamc: Looks like it definitely can, done.

mohamedAdhamc:
p.s. I just noticed I missed something. Waypoints cant just be added directly to expandable parameters like for example vertices, vertex001, and tangent1. I'm not sure exactly how to determine if the parameter is expandable (includes other parameters in it) from the value description or the parameter list but I will look into it tommorow. Also maybe not all expandable parameters cant have waypoints added directly I am not sure maybe there are special cases. I'll have to double check that as well.

pgilfernandez:
I finally had time to try it out: it looks amazing.
I was able to add a waypoint everywhere I looked and wanted =)

Congratulations! it's a really handy feature.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -111,10 +112,12 @@ bool
 Action::WaypointSetSmart::is_candidate(const ParamList &x)
 {
 	return (candidate_check(get_param_vocab(),x) &&
-			// We need an animated valuenode.
-			ValueNode_Animated::Handle::cast_dynamic(x.find("value_node")->second.get_value_node()) &&
 			// We need either a waypoint or a time.
-			(x.count("waypoint") || x.count("time")));
+			(x.count("waypoint") || x.count("time")) &&
+			// Parameter has to be animatable
+			!(x.find("value_desc")->second.get_value_desc()).get_static() &&
```
rodolforg, line 118: You can replace `x.find("value_desc")->second` with `x.at("value_desc")`

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -111,10 +112,12 @@ bool
...
+			//no expandable parameters
+			!(LinkableValueNode::Handle::cast_dynamic(x.find("value_desc")->second.get_value_desc().get_value_node())));
```
rodolforg, line 120: You can replace `x.find("value_desc")->second` with `x.at("value_desc")`

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -162,13 +173,10 @@ Action::WaypointSetSmart::set_param(const synfig::String& name, const Action::Pa
...
 {
-	if(!value_node)
-		synfig::error("Missing value_node");
-
 	if(waypoint.get_time()==(Time::begin()-1))
 		synfig::error("Missing waypoint");
 
-	if(!value_node || waypoint.get_time()==(Time::begin()-1))
+	if(waypoint.get_time()==(Time::begin()-1))
 		return false;
```
rodolforg, line 180: merge

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -452,3 +469,21 @@ Action::WaypointSetSmart::prepare()
 
 	throw Error(_("Unable to determine how to proceed. This is a bug."));
 }
+void
+Action::WaypointSetSmart::redirect_value_desc_set_action()
```
rodolforg, line 473: rename to `redirect_to_value_desc_set_action` (or `redirect_to_action_value_desc_set`?)

rodolforg (changes requested):
1. You didn't add `value_desc` to param_vocab.
2. I wonder if it shouldn't be a new action, instead of changing WaypointSetSmart.
What do you think, @ice0 ?

mohamedAdhamc:
> 
>     2. I wonder if it shouldn't be a new action, instead of changing WaypointSetSmart.
>        What do you think, @ice0 ?

Until we get a reply I just thought I could explain my reasoning as to why I did it this way...

So basically this "add_waypoint" action is the same action as the past "add_waypoint" action. Alll this PR does and what was requested (in a very brief way) was to allow the very same action to be available without having to be in animate edit mode. So in this case regarding functionality it makes sense to only modify the pre-existing action to allow for its use outside of animate edit mode. Otherwise if this is a new action we would basically have two actions that do the same exact thing but one is used outside of animate editing mode and the other from inside animate edit mode.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: woops yes ofc, done.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: ok, makes sense.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: x (ParamList) is a multimap, so we can't really use `at()` for element access here.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: same as above

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
rodolforg: Oh, I forgot this is a multimap :+1:

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -128,6 +135,14 @@ Action::WaypointSetSmart::set_param(const synfig::String& name, const Action::Pa
 
 		return static_cast<bool>(value_node);
 	}
+
+	if(name=="value_desc" && param.get_type()==Param::TYPE_VALUEDESC)
+	{
+		value_desc=param.get_value_desc();
+
+		return true;
+	}
```
rodolforg, line 144: You should check if param.get_value_desc() is a valid ValueDesc (it should refer to a value node, right?)
So, only store it to `value_desc` if it is a valid one.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -83,8 +83,13 @@ Action::WaypointSetSmart::get_param_vocab()
 {
 	ParamVocab ret(Action::CanvasSpecific::get_param_vocab());
 
+	ret.push_back(ParamDesc("value_desc",Param::TYPE_VALUEDESC)
+		.set_local_name(_("ValueDesc"))
+	);
```
rodolforg, line 88: If it didn't exist before, it should be optional, right?

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -111,10 +116,12 @@ bool
 Action::WaypointSetSmart::is_candidate(const ParamList &x)
 {
 	return (candidate_check(get_param_vocab(),x) &&
-			// We need an animated valuenode.
-			ValueNode_Animated::Handle::cast_dynamic(x.find("value_node")->second.get_value_node()) &&
 			// We need either a waypoint or a time.
-			(x.count("waypoint") || x.count("time")));
+			(x.count("waypoint") || x.count("time")) &&
+			// Parameter has to be animatable
+			!(x.find("value_desc")->second.get_value_desc()).get_static() &&
```
rodolforg, line 122: You should check if "value_desc" key exists before using it.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -162,13 +177,10 @@ Action::WaypointSetSmart::set_param(const synfig::String& name, const Action::Pa
 bool
 Action::WaypointSetSmart::is_ready()const
 {
-	if(!value_node)
-		synfig::error("Missing value_node");
-
```
rodolforg, line 167: You should check if either `value_node` or `value_desc` is set.

synfig-studio/src/synfigapp/actions/waypointsetsmart.h
```diff
@@ -59,9 +59,12 @@ class WaypointSetSmart :
 
 	void calc_waypoint();
 	void enclose_waypoint(const synfig::Waypoint& waypoint);
+	void redirect_to_value_desc_set_action();
 
 	std::set<synfig::Time> times;
 
+	ValueDesc value_desc;
+
```
rodolforg, line 67: keep it next to `value_node` declaration.

rodolforg (changes requested):

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: forgot to do will do with next commit.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: oh yes, definetly. done :)

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: hmm yes I was thinking about that but I saw it wasn't checked in other places but yes this is safer will do :).

synfig-studio/src/synfigapp/actions/waypointsetsmart.h
mohamedAdhamc: done

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: hmm well none are explicitly needed for the action now. It used to be that a value_node was needed for it, but now in the case that there is no value_node it is made when we redirect to the value desc set action.

rodolforg: It is. Without `value_desc` set, it is impossible to call ValueDescSet action.
So either valuenode or valudesc must be set.

mohamedAdhamc: Fair point. I meant that not always are we going to be calling the ValueDescSet action, as far as I remember it only needs to be called if we are creating the initial waypoint for the parameter. But fair enough I see your point.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: hmm I tried to do so by checking `parent_is_value_node()` but it always return false. Any suggestions ?

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
rodolforg: That's why we need to check if one of them (wither `value_node` or `value_desc`) is set.
We cannot proceed if we don't have any of these info.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
rodolforg: `is_valuenode()` is enough (I just checked `get_valuenode()` implementation ;) )

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: oh yes definitely, we need at least one of the to proceed.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: hmm `is_value_node` is giving issues (probably due to the extra stuff it calls other than `is_valid`). while just the `!` operator was working (it only checks `is_valid`). What do you think ?

rodolforg: `is_valid()` isn't enough, because you need (line 124):
``` c++
!(LinkableValueNode::Handle::cast_dynamic(x.find("value_desc")->second.get_value_desc().get_value_node())));
```

mohamedAdhamc: Then how about something like 
`if (!value_desc || !value_desc.get_value_node())` then we return false.
because using `is_value_node` was just constantly returning false.

mohamedAdhamc: > `is_valid()` isn't enough, because you need (line 124):
> 
> ```c++
> !(LinkableValueNode::Handle::cast_dynamic(x.find("value_desc")->second.get_value_desc().get_value_node())));
> ```

Hey @rodolforg, I got a little carried away by some stuff but I looked back into this. 

On line 124 I am basically checking to see if this `valuedesc` is referring to a linkable valuenode. If it is then it referring to a an expandable parameter to which add waypoint shouldn't be applied to. What I mean is when `get_value_node()` returns  a null pointer the cast fails which is what is required anyways so `get_value_node()` doesn't necessarily have to return something valid.

Now going back to our topic.For the very first waypoint added the `value_desc` doesnt contain a valuenode, so using `get_value_node()` or `is_value_node()` always returns false the very first time a waypoint is added to a parameter. 

This is why I think it's sufficient to just check with is_valid() because we really don't need the value desc to refer to a valuenode. If its a parameter without any waypoints it initially doesnt but then when it has waypoints it does.

morevnaproject:
I tested the feature and it works well for me. But I haven't looked at the code

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: @rodolforg ping ^_^.

rodolforg: So, shouldn't you check ifnit isn't a valuenode?

mohamedAdhamc: > So, shouldn't you check ifnit isn't a valuenode?

oh ok adding a check to make sure it isn't a valuenode would work.

mohamedAdhamc: Done.
At first I was a bit confused as why it might be needed. But I think it makes sense now, if it is not the first time (i.e. value desc is a value node) the we don't need the value desc because we will not need a value desc set action.

mohamedAdhamc:
@rodolforg anything left here ^_^ ?

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -128,6 +138,17 @@ Action::WaypointSetSmart::set_param(const synfig::String& name, const Action::Pa
 
 		return static_cast<bool>(value_node);
 	}
+
+	if(name=="value_desc" && param.get_type()==Param::TYPE_VALUEDESC)
+	{
+		std::cout<<"value desc called"<<std::endl;
```
rodolforg, line 144: please remove this debug line

mohamedAdhamc: Oops, I missed this one. Done.

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -162,14 +182,17 @@ Action::WaypointSetSmart::set_param(const synfig::String& name, const Action::Pa
 bool
 Action::WaypointSetSmart::is_ready()const
 {
-	if(!value_node)
-		synfig::error("Missing value_node");
+	//we either need a valuedesc for the inital waypoint added
+	//or a valuenode to add more waypoints to the parameter
+	if(!value_desc && !value_node){
+		synfig::error("Missing both value_desc and value_node");
```
rodolforg, line 188: 1. Surround error message with `_()` for translation purposes.
2. Rephrase: "Either value_desc or value_node must be set"

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
```diff
@@ -162,14 +182,17 @@ Action::WaypointSetSmart::set_param(const synfig::String& name, const Action::Pa
...
+		return false;
+	}
 
-	if(waypoint.get_time()==(Time::begin()-1))
+	if(waypoint.get_time()==(Time::begin()-1)){
 		synfig::error("Missing waypoint");
```
rodolforg, line 193: Surround error message with `_()` for translation purposes.

rodolforg (changes requested):
Final requests :)

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: Oh yes thanks for noticing I completely forgot. Done : ).

synfig-studio/src/synfigapp/actions/waypointsetsmart.cpp
mohamedAdhamc: Done

rodolforg (approved):

rodolforg:
Merged. Thank you!

PR #3009, opened 2023-02-22
feat: white background in new documents by default

### LINKED ISSUE #2506: White background in new documents by default
Some newbie users say it would be more user-friendly if new documents in Synfig will have white background by default.

I suggest to make relevant option enabled by default - 
screenshot_001

What do you think?

mohamedAdhamc:
fixes #2506

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -287,7 +287,7 @@ int    studio::App::number_of_threads = std::thread::hardware_concurrency();
 String studio::App::navigator_renderer;
 String studio::App::workarea_renderer;
 
-String        studio::App::default_background_layer_type  = "none";
+String        studio::App::default_background_layer_type  = "solid_color";
 synfig::Color studio::App::default_background_layer_color =
 	synfig::Color(1.000000, 1.000000, 1.000000, 1.000000);  //White
 String        studio::App::default_background_layer_image = "undefined";
--- a/synfig-studio/src/gui/dialogs/dialog_setup.cpp
+++ b/synfig-studio/src/gui/dialogs/dialog_setup.cpp
@@ -259,8 +259,8 @@ Dialog_Setup::create_document_page(PageInfo pi)
 	 *  fps   [_]                    [FPS]
 	 *  size  H[_]xW[_]      [resolutions]
 	 * DEFAULT BACKGROUND
-	 * (*) None (Transparent)
-	 * ( ) Solid Color         [colorbutton]
+	 * ( ) None (Transparent)
+	 * (*) Solid Color         [colorbutton]
 	 * ( ) Image               [file path  ]
 	 *
 	 *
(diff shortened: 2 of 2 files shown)


ice0:
Opinion is divided, so I think it might be better to use a different approach?
For example, add an option to installer?

Also I see a problem in the current approach.
The default color for new shapes is white. And it looks like this behavior can be confusing for users.


https://user-images.githubusercontent.com/5604544/221065556-b7001411-d2d0-4d59-bd9d-53e43cfd9859.mp4

mohamedAdhamc:
> Opinion is divided, so I think it might be better to use a different approach? For example, add an option to installer?
>
But then this would only affect windows (perhaps mac also not sure) users right ? I have always built synfig from source but I double checked (downloaded from website) the Linux applmage builds don't seem to have installers.
 
> Also I see a problem in the current approach. The default color for new shapes is white. And it looks like this behavior can be confusing for users.

hmm, I also noticed this. It would especially affect new users since their installation comes with all the defaults. How about maybe we switch the default outline color with the default fill color so that he default fill color becomes black ?

yoSachinkr:
Well, I never supported this feature request. 👀

ice0:
> Well, I never supported this feature request. 👀

@morevnaproject mentioned that he receives questions from new users: why the background looks like a checkerboard and why the background is gray when rendering. 
It is likely that there could be alternative solutions to address this issue.

yoSachinkr:
> > Well, I never supported this feature request. 👀
> 
> @morevnaproject mentioned that he receives questions from new users: why the background looks like a checkerboard and why the background is gray when rendering. It is likely that there could be alternative solutions to address this issue.

I think proper documentation regarding canvas settings is a better solution than changing the defaults because those settings are pretty simple to change in the **Preferences** dialog box. It is just that they're not aware of it.

mohamedAdhamc:
>>
>I think proper documentation regarding canvas settings is a better solution than changing the defaults because those settings are pretty simple to change in the **Preferences** dialog box. It is just that they're not aware of it.

But do new users typically follow along with the documentation ?
Also, I think this goes both way (settings being simple to change). It's easy to change so current users can just not use it. But for new users which from what I understand this would make more sense to them it would be available as soon as they start the program.

Anyways, personally I also don't have a strong opinion here...

p.s. I accidentally edited the comment instead of replying to it again xd, somethings weird when replying from a phone.

yoSachinkr:
> But do new users typically follow along with the documentation ?

I think Synfig needs a start screen just like blender has with links to documentation and quick canvas settings.
image

mohamedAdhamc:
>> 
> @mohamedAdhamc I think Synfig needs a start screen just like blender has with links to documentation and quick canvas settings. 

I think that's a great idea and it probably would also be a good ux enhancement. Also some of these look really good. I say we open a new issue with this request. I personally would've started working on it, but right now i'm a bit busy with understanding and testing out things in the layer optimization project. However, if noone gets to it by the time I have some more time on my hands I would definitely love to work on it : )

PR #3012, opened 2023-02-24
refactor: use RadioToolButton instead of ToggleToolButton for tools dock

alexhaskell:
Multiple calls (4+) to Dock_Toolbox::change_state_() occur because ToggleToolButton::set_active(), inside Dock_Toolbox::set_active_state(), would cause clicked signals.

This commit reduces the number of calls to just 2. Once for the previous tool button and a second for the new tool button. This is the normal behavior of RadioToolButtons.

### DIFF
--- a/synfig-studio/src/gui/docks/dock_toolbox.cpp
+++ b/synfig-studio/src/gui/docks/dock_toolbox.cpp
@@ -146,34 +146,9 @@ void Dock_Toolbox::read_layout_string(const std::string& params) const
 void
 Dock_Toolbox::set_active_state(const synfig::String& statename)
 {
-	std::map<synfig::String,Gtk::ToggleToolButton *>::iterator iter;
-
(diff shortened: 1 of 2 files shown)


alexhaskell:
Working on a as-yet-unreported bug that I found and came across some strange behavior that I at first thought was the cause, but turned out to not be the case.

I'm not entirely sure to call this a "fix" or "refactor". Since it seems like incorrect behavior I called it a "fix".

ice0:
> I'm not entirely sure to call this a "fix" or "refactor". Since it seems like incorrect behavior I called it a "fix".

Hi! We are looking from users the point of view. So if it doesn't change the behavior, it's better to name it refactoring. 
Also fixes are included in the Release notes, and refactorings are not.

ice0 (approved):

ice0:
Merged. Thank you!

Issue #3014, opened 2023-02-25
[ GSOC23 ]  Optimization of internal operations with layers


mohamedAdhamc:
Hey all,
Similar to what I did last year and what is recommended in the docs. I'm opening this issue where I will be writing about my progress breaking down the project and updating my progress in it. And if I finally get to take on the project when the time comes then it can serve a purpose similar to last years issue. Also, I will just be updating this comment whenever I add something so as not to overload anyone subscribed to all issues. I might (probably) also take on some open issues in the mean time related to layers in general which I will link to this issue.

Now without further ado considering progress:

part 1 - introducing a caching mechanism to the set_time function:

I have implemented a very basic demo mechanism which works for the basic layer types and it's already showing some promise
I tested rendering of a simple file of about 5 circle layers with only one waypoint for a couple of the parameters (radius, opacity, origin, feather)
without anything cached:
![Screenshot from 2023-02-25 17-19-29](https://user-images.githubusercontent.com/100296264/221371890-4316d5d8-f258-4a74-994b-ed37cb2cdc2e.png)
with caching:
![Screenshot from 2023-02-25 17-20-26](https://user-images.githubusercontent.com/100296264/221371899-3abfbbdd-2c32-4d47-9ec8-fc33f0400578.png)
(not the best way to test something like this but note the slight difference in speed)
Now of course the benefits of this caching should probably be more apparent with bigger complex files when its fully implemented, and not only in the process of rendering but while working on the animation it self  as well and playing it to test it etc... so hopefully this is just a very very brief start : ), I will continue to understand the current system in more detail and think of different approaches etc...

mohamedAdhamc:
I will use this comment to track my progress during GSoC so I thought it would be helpful to layout the structure for it now :)

**Select Tool:**
1- Move Functionality: #3036 status: in progress 
2- Scale Functionality: 
3- Rotate Functionality:
4- Mirror Functionality:

**Fixing UI lags/Issues:**
1- Benchmarking/Testing various internal layer operations to see if caching optimzations make sense:
    A- [if it does] optimizing 'set_time()'
    B- [if it does]  Introducing a Caching algorithm in the process of layer to task conversion. 
2- Alternatively (or also if there is time) 
    A - Brush Tool Crashes Synfig: 
    B-  Crash on clicking on rows of Library Panel: #3039  status: merged.
    C- Crash when manipulating Duplicate Layer/ValueNode:


**Additional Features:**
[...]

PR #3015, opened 2023-02-27
refactor: remove deprecated `Widget::override_color()`

mohamedAdhamc:
Found this while looking through the code for #3014 . It also says this function heavily slows playback so I thought might as well replace it. 

I noticed there is also a synfig.mac.css, not sure if that is used for mac instead of the synfig.css or ig it is just to define an extra class for a fix. If it's not the latter I might have to redefine the classes there as well.

p.s. 
about four months back the whole style context class was deprecated from GTK and now instead for example it should just be
`versionlabel->add_class("color-white");`
the `add_class` method was moved directly to the widget base class, but it doesn't seem to be ported yet to gtkmm. I checked the docs for even the latest verision and it wasn't there.

p.s. 2
I don't know why the verision label is white while the background is white as well, maybe it's meant like so because its disabled ? Anyways, I just ported the original white color as is.

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -2148,13 +2148,11 @@ CanvasView::on_time_changed()
 	{
 		KeyframeList::iterator iter;
 		if (get_canvas()->keyframe_list().find(time, iter)) {
-			// Widget::override_color() is deprecated since Gtkmm 3.16: Use a custom style provider and style classes instead.
-			// This function is very slow!
-			current_time_widget->override_color(Gdk::RGBA("#FF0000"));
+			current_time_widget->get_style_context()->remove_class("color-black");
+			current_time_widget->get_style_context()->add_class("color-red");
 		} else {
-			// Widget::override_color() is deprecated since Gtkmm 3.16: Use a custom style provider and style classes instead.
-			// This function is very slow!
-			current_time_widget->override_color(Gdk::RGBA(0));
+			current_time_widget->get_style_context()->remove_class("color-red");
+			current_time_widget->get_style_context()->add_class("color-black");
 		}
 
 		// Shouldn't these trees just hook into
@@ -2542,9 +2540,6 @@ CanvasView::play_async()
 	if (timeout < 10) timeout = 10;
 
 	framedial->toggle_play_pause_button(is_playing());
-	// Widget::override_color() is deprecated since Gtkmm 3.16: Use a custom style provider and style classes instead.
-	// Also, this function is heavily slowdowns playback.
-	//current_time_widget->override_color(Gdk::RGBA(0));
 
 	soundProcessor.clear();
 	canvas_interface()->get_canvas()->fill_sound_processor(soundProcessor);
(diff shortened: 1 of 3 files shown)


ice0:
> It also says this function heavily slows playback so I thought might as well replace it.

I'm not sure why this is happening (if so). Maybe it's redrawing the GUI and that requires the canvas to be redrawn?

> I noticed there is also a synfig.mac.css, not sure if that is used for mac instead of the synfig.css or ig it is just to define an extra class for a fix. If it's not the latter I might have to redefine the classes there as well.

Yes, it is used on Mac and imports `synfig.css`, so usually no changes are required.

https://github.com/synfig/synfig/blob/313878d4d12a32f24b533d0832a6660b63ee149a/synfig-studio/src/gui/resources/css/synfig.mac.css#L1


> about four months back the whole style context class was deprecated from GTK and now instead for example it should just be versionlabel->add_class("color-white");
the add_class method was moved directly to the widget base class, but it doesn't seem to be ported yet to gtkmm. I checked the docs for even the latest verision and it wasn't there.

Some time ago, the author of the `gtkmm` library wanted to stop work on it (https://discourse.gnome.org/t/gtkmm-abandonment/8906), but it looks like he is back. 

`gtkmm` is not maintained by the GTK team, so it looks like he just doesn't have enough time to keep it up to date.

> I don't know why the verision label is white while the background is white as well, maybe it's meant like so because its disabled ?

I'm thinking maybe because the label background is grey by default?

mohamedAdhamc:
> I'm not sure why this is happening (if so). Maybe it's redrawing the GUI and that requires the canvas to be redrawn?

I am not sure either I tried setting some break points with gdb and following the execution (with override_color and without) and I didn't notice anything unusual... However I think in either cases it is still deprecated and should be replaced (unless i'm missing something here)

>Yes, it is used on Mac and imports synfig.css, so usually no changes are required.

perfect !

>    Some time ago, the author of the gtkmm library wanted to stop work on it (https://discourse.gnome.org/t/gtkmm-abandonment/8906), but it looks like he is back.

Oh yes I remember seeing him talking about that. Glad he is back, tbh I don't really know what could happen if the eventual end of gtkmm comes... but on the flip side it could be finally the needed sign to start to switch over to pygtk, but that would probably be a lot of work.

> I'm thinking maybe because the label background is grey by default?

Screenshot from 2023-03-01 20-53-29



I don't think so, here I changed it to red and the background is white, when it is white (the norm) it is basically invisible. Maybe it is because this label is kind of redundant ? At the very top the version shows after synfig studio. But in this case shouldn't the label be removed as it is not used.

p.s. I noticed a slight issue in App::init this splash screen is shown a couple lines before App::apply_gtk_settings() which sets up the css file so for a very tiny amount of time that it takes the color is the default for the label (black) until the css file is properly initialized and it's effect shows. On normal startup I could'nt notice it as the time is a fraction of a second but I noticed it while executing line by line. I think the label should be removed altogether as it seems like it's redundant and not used but i'm not sure.

p.s. I accidentally edited the comment at first instead of replying to it.

ice0:
> At the very top the version shows after synfig studio. But in this case shouldn't the label be removed as it is not used.

I think it's still needed. This is how it looks on the splash screen in the Release version.

image

ice0 (approved):

ice0:
Merged. Thank you!

Issue #3017, opened 2023-03-09
[GSOC23] Performance optimization


BharatSahlot:
Hello everyone,
Following my post on the [forums](https://forums.synfig.org/t/gsoc-2023-google-summer-of-code-2023/13859/3?u=b23v), I will be using this issue for updating my progress(before GSOC and hopefully during GSOC) on the project and asking for help.

For the past few days, I have been reading the source code. I am starting to understand how the rendering process works and have some ideas to improve the progress.

To get myself more familiar with the source code, I have started with the task 
- `add flag to synfig CLI binary for performance test (it should measure the time taken to render the file, and allow to set the number of repeats)`

Thanks

BharatSahlot:
Hey @ice0,
I have been working on understanding the Renderer system, but I'm running into some difficulties. Specifically, I am having trouble with the Task, Renderer, and RenderQueue classes, as it appears that they are interdependent in some way. Additionally, I'm struggling to understand the various types of Tasks. If you have any insights or guidance on this matter, I would be very grateful for your help. Thanks in advance!

mohamedAdhamc:
> Specifically, I am having trouble with the Task, Renderer, and RenderQueue classes, as it appears that they are interdependent in some way. 

Just thought to mention this, you can check out [dev docs rendering section](https://synfig-docs-dev.readthedocs.io/en/latest/common/structure.html#rendering-process) it has a general overview on a lot of these concepts that is quite helpful.

BharatSahlot:
I have read it, but I was hoping for a little more detail. Thanks tho!

ice0:
Unfortunately, the people who originally designed this mechanism are not available, so I will try to explain it as I understand it (which means I could be wrong).

**Task**

`Task` is a basic abstraction for rendering something. It can be either software or hardware or contain sub-tasks. 
Sometimes a `Task` is did nothing and just calls other sub-tasks to do the actual work.

Currently, all tasks are calculated in software and have the suffix SW in their name. 
For example, `TaskLumaKeySW`.
https://github.com/synfig/synfig/blob/03c9c20882700a2b4bc7c2305d3a074e9915f14d/synfig-core/src/modules/mod_filter/lumakey.h#L66

**RenderQueue**

`RenderQueue` is a thread pool that receives `Task` and parallelizes its execution across multiple threads if possible.

Thread list member:
[...]

PR #3018, opened 2023-03-09
feat: add a CLI flag(--set-repeats) for rendering the same target multiple times

BharatSahlot:
Following my post #3017 , I decided to implement the feature.

Since synfig cli already had a flag for printing benchmark time, I only added the feature to render the same target multiple times.
The total and average time are printed.

### DIFF
--- a/synfig-core/src/tool/joblistprocessor.cpp
+++ b/synfig-core/src/tool/joblistprocessor.cpp
@@ -264,23 +264,41 @@ void process_job (Job& job)
 	else
 	{
 		VERBOSE_OUT(1) << _("Rendering...") << std::endl;
-		std::chrono::system_clock::time_point start_timepoint =
-            std::chrono::system_clock::now();
 
-		// Call the render member of the target
-		if(!job.target->render(&p))
-			throw (SynfigToolException(SYNFIGTOOL_RENDERFAILURE, _("Render Failure.")));
+		bool should_print_benchmarks = SynfigToolGeneralOptions::instance()->should_print_benchmarks();
+		double dur = 0.f;
+		int repeats = SynfigToolGeneralOptions::instance()->get_repeats();
 
-		if(SynfigToolGeneralOptions::instance()->should_print_benchmarks())
-        {
-            std::chrono::duration<double> duration =
-                std::chrono::system_clock::now() - start_timepoint;
+		for(int i = 0; i < repeats; i++)
+		{
+			std::chrono::system_clock::time_point start_timepoint =
+				std::chrono::system_clock::now();
 
-            std::cout << job.filename.c_str()
-                      << _(": Rendered in ")
-                      << duration.count()
-                      << _(" seconds.") << std::endl;
-        }
+			// Call the render member of the target
+			if(!job.target->render(&p))
+				throw (SynfigToolException(SYNFIGTOOL_RENDERFAILURE, _("Render Failure.")));
+
+			if(should_print_benchmarks)
+			{
+				std::chrono::duration<double> duration =
+					std::chrono::system_clock::now() - start_timepoint;
+
+				dur += duration.count();
+			}
+		}
+
+		if(should_print_benchmarks)
+		{
+			std::cout << job.filename.c_str()
+				<< _(": Rendered ")
+				<< repeats
+				<< _( " times in ")
+				<< dur
+				<< _(" seconds.")
+				<< _(" Average time per render: ")
+				<< dur / repeats
+				<< "." << std::endl;
+		}
 	}
 
 	VERBOSE_OUT(1) << _("Done.") << std::endl;
--- a/synfig-core/src/tool/joblistprocessor.cpp
+++ b/synfig-core/src/tool/joblistprocessor.cpp
@@ -266,13 +266,14 @@ void process_job (Job& job)
 		VERBOSE_OUT(1) << _("Rendering...") << std::endl;
 
 		bool should_print_benchmarks = SynfigToolGeneralOptions::instance()->should_print_benchmarks();
-		double dur = 0.f;
+
+		double total_duration = 0.f;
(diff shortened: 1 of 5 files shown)


BharatSahlot:
For some reason a extra file '\' was added to the commit. Removing it.

synfig-core/src/tool/joblistprocessor.cpp
```diff
@@ -264,23 +264,41 @@ void process_job (Job& job)
...
 	{
 		VERBOSE_OUT(1) << _("Rendering...") << std::endl;
-		std::chrono::system_clock::time_point start_timepoint =
-            std::chrono::system_clock::now();
 
-		// Call the render member of the target
-		if(!job.target->render(&p))
-			throw (SynfigToolException(SYNFIGTOOL_RENDERFAILURE, _("Render Failure.")));
+		bool should_print_benchmarks = SynfigToolGeneralOptions::instance()->should_print_benchmarks();
+		double dur = 0.f;
```
ice0, line 269: I suggest to rename it to `total_duration` or `duration_total`

synfig-core/src/tool/joblistprocessor.cpp
```diff
@@ -264,23 +264,41 @@ void process_job (Job& job)
...
+		int repeats = SynfigToolGeneralOptions::instance()->get_repeats();
+
+		for(int i = 0; i < repeats; i++)
+		{
+			std::chrono::system_clock::time_point start_timepoint =
```
ice0, line 274: I suggest to use `steady_clock` instead of `system_clock`, because it's monotonic.

synfig-core/src/tool/joblistprocessor.cpp
```diff
@@ -264,23 +264,41 @@ void process_job (Job& job)
...
-                      << _(" seconds.") << std::endl;
-        }
+		if(should_print_benchmarks)
+		{
+			std::cout << job.filename.c_str()
+				<< _(": Rendered ")
+				<< repeats
+				<< _( " times in ")
+				<< dur
+				<< _(" seconds.")
```
ice0, line 297: I suggest using milliseconds for benchmarks as it is a more appropriate unit of measurement in this case.
For example current output on test file:
```
./solid.sifz: Rendered 10 times in 0.00730508 seconds. Average time per render: 0.000730508.
```

Suggested:
```
./solid.sifz: Rendered 10 times in 7.30508 ms. Average time per render: 0.730508.
```

Also I notice repeated output:
```
synfig(13480) [09:54:40] info: 	tl: [-4,000000,2,250000], br: [4,000000,-2,250000], focus: [0,000000,0,000000]
./solid.sifz ==> ./solid.null: DONE
./solid.sifz ==> ./solid.null: DONE
./solid.sifz ==> ./solid.null: DONE
./solid.sifz ==> ./solid.null: DONE
./solid.sifz ==> ./solid.null: DONE
./solid.sifz ==> ./solid.null: DONE
./solid.sifz ==> ./solid.null: DONE
./solid.sifz ==> ./solid.null: DONE
./solid.sifz ==> ./solid.null: DONE
./solid.sifz ==> ./solid.null: DONE
./solid.sifz: Rendered 10 times in 8.02305 ms. Average time per render: 0.802305.
./synfig --repeats=10 --benchmarks --time=1 --target=null ./solid.sifz  0,04s user 0,03s system 97% cpu 0,076 total
```

It's better to fix this.

BharatSahlot: Yes, I wasn't sure what to do with this. I guess I will try to just keep the render progress on the same one line for each render? Like it is for one render.

BharatSahlot: image

What should I do in case like this? @ice0

BharatSahlot:
I thought of rendering to a null target. I have implemented that, but that will result in more time required for the last run when the images are actually outputted.
I couldn't find a way to cleanly reset a target.

synfig-core/src/tool/joblistprocessor.cpp
ice0: Hi! Sorry for a late answer.
I suggest considering the entire benchmark as a single cycle. If this will introduce a lot of changes, when leave it as-is.

ice0:
> I thought of rendering to a null target. I have implemented that, but that will result in more time required for the last run when the images are actually outputted.
> I couldn't find a way to cleanly reset a target.

I think there is no need to do that.
We need the ability to set a render target from command line just to check where the performance was affected.
As you may have noticed, I switched to the null target from the command line when it was necessary.

./synfig --repeats=10 --benchmarks --time=1 **--target=null** ./solid.sifz

synfig-core/src/tool/joblistprocessor.cpp
BharatSahlot: Hi, sorry I dont understand what you mean by
> considering the entire benchmark as a single cycle

Like in the progress callback, I should modify it to account for number of repeats? Like,

```cpp
total = frames_count * repeats;
current_frame *= times_repeated; // times repeated is a local variable in RenderProgress
```

BharatSahlot:
Okay, that makes sense.

synfig-core/src/tool/joblistprocessor.cpp
ice0: Yes. But I am not sure about that decision.
I think clean output is preferable, but multiple output also makes sense.
So I suggest to merge this PR as is, and test that decision in another PR. 
What do you think?

BharatSahlot: Yes that makes sense. I will make another PR for that then.

ice0 (approved):

BharatSahlot:
Do I have to do anything else before this is merged?

ice0:
Merged. Thank you!

PR #3020, opened 2023-03-12
refactor: remove useless return statement at the end of a function

mohamedAdhamc:
Just caught my eye... It is not really affecting anything , compiler takes care of it anyways. But just for the sake of a cleaner code base ig :)

p.s. I have been collecting various typos and little things I noticed which seem off during the past week/2 weeks while going through the code in various places. I will wait a bit longer with the typos though just so that they could all be in one compact PR.


ice0 (approved):

ice0:
Merged. Thank you!

PR #3024, opened 2023-03-15
feat: cleaner cli render progress output

BharatSahlot:
Follow up for this https://github.com/synfig/synfig/pull/3018#discussion_r1136616032

### DIFF
--- a/synfig-core/src/tool/renderprogress.cpp
+++ b/synfig-core/src/tool/renderprogress.cpp
@@ -58,7 +58,10 @@ bool RenderProgress::amount_complete(int current_frame, int frames_count)
 
     std::ostringstream outputStream;
 
-    const bool isFinished = (current_frame == frames_count);
+	if(current_frame >= frames_count) repeated_++;
+
+	const int total_repeats = SynfigToolGeneralOptions::instance()->get_repeats();
+	const bool isFinished = (repeated_ >= total_repeats);
     if (!isFinished)
     {
         // avoid reporting the progress too often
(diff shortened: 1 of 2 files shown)


synfig-core/src/tool/renderprogress.cpp
```diff
@@ -69,15 +72,18 @@ bool RenderProgress::amount_complete(int current_frame, int frames_count)
...
 
+		int total_frames_count = total_repeats * frames_count;
+		int r_current_frame = current_frame + frames_count * repeated_;
+
         int percentage_completed = 100;
-        if (frames_count > 0)
+        if (total_frames_count > 0)
         {
-            percentage_completed = 100 * current_frame / frames_count;
+            percentage_completed = (100 * r_current_frame) / total_frames_count;
```
BharatSahlot, line 81: I am not sure if this should be percentage completed for the current render or for the whole thing.

ice0: I think the whole think is preferable.
Can you check remaining time calculation too?

```
./solid.sifz ==> ./solid.null: Iteration 5374: Frame 1 of 1 (53%). Remaining time: 0s 
```

BharatSahlot: Sure, I will fix the remaining time to display the total remaining time as well. I made the changes, will commit after testing.

ice0 (approved):

ice0:
Merged. Thank you!

PR #2831, opened 2022-09-20
perf: improved performance by reducing memory copying when working with valuenodes

ice0:

PR #3027, opened 2023-03-21
fix: added a missing `init_children_vocab()` call in ValueNode constructors

ice0:
It turned out that some necessary initialization of the parameter name dictionary (`children_vocab`) is missing in certain places. 
Previously, it worked because a new dictionary was always created when calling `link_count()` or `get_children_vocab()` method (changes made in #2831).
After the changes, some ValueNodes can work incorrectly because the parameter dictionary became empty when accessing it.

This commit fixes this issue.

P.S. 
There are also ValueNodes with a dynamic set of parameters (`ValueNode_BLine`,`ValueNode_DynamicList`, `ValueNode_StaticList`), but they should work correctly because the parameter dictionary is repopulated when the reindex() method is called.

Related PR: #2831

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_blinereversetangent.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_blinereversetangent.cpp
@@ -62,6 +62,7 @@ REGISTER_VALUENODE(ValueNode_BLineRevTangent, RELEASE_VERSION_0_61_08, "blinerev
 ValueNode_BLineRevTangent::ValueNode_BLineRevTangent(Type &x):
 	LinkableValueNode(x)
 {
+	init_children_vocab();
 }
 
 ValueNode_BLineRevTangent::ValueNode_BLineRevTangent(const ValueNode::Handle &x):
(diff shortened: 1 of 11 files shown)


ice0:
@mohamedAdhamc @BharatSahlot 

Guys, if you have time, could you also review the top 3-4 pull requests? 
Other contributors is busy right now, so this will help a lot.

BharatSahlot:
Okay, will do.

rodolforg:
I created a test, and it passed:

```
	synfig::Main Main("");

	auto duplicate = synfig::ValueNodeRegistry::create("duplicate", synfig::Real());
	ASSERT_EQUAL(3, duplicate->link_count());
	auto duplicate2 = duplicate->clone(nullptr);
	ASSERT_EQUAL(3, synfig::LinkableValueNode::Handle::cast_static(duplicate2)->link_count());
	
	auto list = synfig::ValueNode_StaticList::create(std::vector<synfig::Real>{1.});
	ASSERT_EQUAL(1, list->link_count());
	auto new_entry = list->create_list_entry(0);
	list->add(new_entry);
	ASSERT_EQUAL(2, list->link_count());
```

Maybe we should add something like it in `synfig-core/test` ?

rodolforg (approved):

ice0:
> Maybe we should add something like it in `synfig-core/test` ?

Sure. I will do it.

PR #3029, opened 2023-03-22
fix: "Embed layer" doesn't embed ValueBase nodes

### LINKED ISSUE #426: "Embed layer" doesn't embed ValueBase nodes
linux 1.3.4

**Issue description**:
When using "embed layer", valuebase nodes do appear in the newly created canvas, as they should ; but they have no influence, the values actually used are still those of the original file.

mohamedAdhamc:
fixes #426 ,
Name of the original issue is a little bit confusing ?!. But basically the problem was that "embed layer" is supposed to make the embedded layer independent of the original layer (this includes exported parameters). Now this works correctly for layers of the imported canvas, but it works incorrectly for when these very same layers are enclosed by a group (layer_pastecanvas).

description of what I did:
After some investigating I realized in `ValueDescExport` `scan_layers()` wasn't correctly scanning all the layers for the exported canvas. This was because it was only scanning the top layers, but some of these layers can be groups and have layers inside them. So I just allowed for it to scan all the layers of the canvas. This seems to fix the issue but i'm not sure if breaks stuff so I'll keep it as draft until I check more.

### DIFF
--- a/synfig-studio/src/synfigapp/actions/valuedescexport.cpp
+++ b/synfig-studio/src/synfigapp/actions/valuedescexport.cpp
@@ -49,6 +49,8 @@
 
 #include <synfigapp/localization.h>
 
+#include <synfig/layers/layer_pastecanvas.h>
+
 #endif
 
 using namespace synfig;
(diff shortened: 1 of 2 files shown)


ice0:
@morevnaproject 
Can you test it?

mohamedAdhamc:
It seems like we've got one tester here so far. link to forums [comment](https://forums.synfig.org/t/next-synfig-release/14305/9?u=mohamed.adhamc).

PR #3033, opened 2023-03-25
fix: sigsegv when `LANGUAGE`/`LANG` environment variable is not set

ice0:
Synfig Studio crashes when trying to run on macOS.

`getenv()` can return NULL, so we pass NULL as a `std::string&` reference.

Example code:
```
#include <string>

bool is_empty(const std::string& str) {
    return str.empty();
}

int main() {
    is_empty(nullptr);
    return 0;
}
```

Output:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string: construction from null is not valid
```

Fixed by switching to Glibmm function. In this case, Glibmm returns an empty string.

Linked PR: #2998

### DIFF
--- a/synfig-core/src/synfig/os.cpp
+++ b/synfig-core/src/synfig/os.cpp
@@ -37,6 +37,8 @@
 
 #include <cstdarg>
 
+#include <glibmm/miscutils.h>
+
 #include <synfig/filesystem.h>
 #include <synfig/general.h>
 #include <synfig/localization.h>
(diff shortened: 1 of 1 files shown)


rodolforg:
Ugh. Sorry, my bad.

Can't we check for null (assigning it to a temp variable first) instead of adding Glib methods?

ice0:
> Ugh. Sorry, my bad.
> 
> Can't we check for null (assigning it to a temp variable first) instead of adding Glib methods?

I am not sure if this is better, but ok.

rodolforg (approved):

PR #3034, opened 2023-03-25
Opengl build

BharatSahlot:
@ice0 I managed to build with opengl but I think I am copying the shaders to the wrong directory, so I am trying to fix that.

### DIFF
--- a/synfig-core/src/synfig/rendering/opengl/internal/CMakeLists.txt
+++ b/synfig-core/src/synfig/rendering/opengl/internal/CMakeLists.txt
@@ -1,0 +1,28 @@
+target_sources(libsynfig
+    PRIVATE
+		"${CMAKE_CURRENT_LIST_DIR}/antialiasing.cpp"
+		"${CMAKE_CURRENT_LIST_DIR}/buffers.cpp"
+		"${CMAKE_CURRENT_LIST_DIR}/context.cpp"
+		"${CMAKE_CURRENT_LIST_DIR}/environment.cpp"
+		"${CMAKE_CURRENT_LIST_DIR}/framebuffers.cpp"
+		"${CMAKE_CURRENT_LIST_DIR}/misc.cpp"
+		"${CMAKE_CURRENT_LIST_DIR}/predeclarations.cpp"
+		"${CMAKE_CURRENT_LIST_DIR}/samplers.cpp"
+		"${CMAKE_CURRENT_LIST_DIR}/shaders.cpp"
+)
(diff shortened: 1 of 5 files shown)


synfig-core/src/synfig/rendering/opengl/internal/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+	"${CMAKE_CURRENT_LIST_DIR}/glsl/antialiased_textured_rect_fragment.glsl"
+	"${CMAKE_CURRENT_LIST_DIR}/glsl/antialiased_textured_rect_vertex.glsl"
+	"${CMAKE_CURRENT_LIST_DIR}/glsl/blend_fragment.glsl"
+	"${CMAKE_CURRENT_LIST_DIR}/glsl/color_fragment.glsl"
+	"${CMAKE_CURRENT_LIST_DIR}/glsl/simple_vertex.glsl"
+	"${CMAKE_CURRENT_LIST_DIR}/glsl/texture_fragment.glsl"
+	"${CMAKE_CURRENT_LIST_DIR}/glsl/texture_vertex.glsl"
+)
+
+add_custom_command(TARGET libsynfig POST_BUILD
```
ice0, line 27: In this case, it will not update the shader if the shader is changed but the executable is not, because it only runs after the build is done. If executable sources not changed, then CMake doesn't rebuild it.

This is how this can be improved:
```CMake

set(RENDERING_OPENGL_INTERNAL_GLSL_FILES
	"glsl/antialiased_textured_rect_fragment.glsl"
	"glsl/antialiased_textured_rect_vertex.glsl"
	"glsl/blend_fragment.glsl"
	"glsl/color_fragment.glsl"
	"glsl/simple_vertex.glsl"
	"glsl/texture_fragment.glsl"
	"glsl/texture_vertex.glsl"
)

foreach (SHADER IN ITEMS ${RENDERING_OPENGL_INTERNAL_GLSL_FILES})
	configure_file("${CMAKE_CURRENT_LIST_DIR}/${SHADER}" "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${SHADER}" COPYONLY)
endforeach ()
```

`configure_file` updates files if they have changed. Also directory will be created automatically.

P.S. Also don't forget to add `install` command.

BharatSahlot: Okay, I will update this.

PR #3036, opened 2023-03-26
feat: select tool (move functionality)

mohamedAdhamc:
While working on the proposal for #3014 I was working on the "select tool" part and I ended up implementing a very quick prototype for it.
This is how it is functioning:
 
[Kooha-2023-03-26-11-32-53.webm](https://user-images.githubusercontent.com/100296264/227768622-50c28d31-7047-42de-9be7-9a9b108ebebe.webm)

basically click anywhere on the layer to select and drag it and the layer will be moved according to the cursors motion.

this prototype is missing a few things though mainly (list is not exhaustive):
1- no need to show handles/ducks here. Maybe just an outline or something but other than that I believe they should be hidden 
    when using this tool since the user doesnt use them to move but instead "drags the layer". Also should this state just totally  
    ignore the show/hid handle buttons?
2- icons for this new tool
3- should groups always be given selection priority over the layers grouped by them ?
4- a lot of refactoring

p.s. 
commit history is a mess because I made this off of another branch, wasn't expecting to end up with a prototype so quick xD. I will try to squash all these 5 commits in the history because they are really just one commit.

### DIFF
--- a/synfig-studio/src/gui/states/Makefile_insert
+++ b/synfig-studio/src/gui/states/Makefile_insert
@@ -12,6 +12,7 @@ STATES_HH = \
 	states/state_polygon.h \
 	states/state_rectangle.h \
 	states/state_rotate.h \
+        states/state_select.h \
 	states/state_scale.h \
 	states/state_sketch.h \
 	states/state_smoothmove.h \
@@ -32,10 +33,11 @@ STATES_CC = \
 	states/state_fill.cpp \
 	states/state_gradient.cpp \
 	states/state_mirror.cpp \
-	states/state_normal.cpp \
 	states/state_polygon.cpp \
 	states/state_rectangle.cpp \
 	states/state_rotate.cpp \
+        states/state_normal.cpp \
+        states/state_select.cpp \
 	states/state_scale.cpp \
 	states/state_sketch.cpp \
 	states/state_smoothmove.cpp \
--- a/synfig-studio/src/gui/states/state_normal.cpp
+++ b/synfig-studio/src/gui/states/state_normal.cpp
@@ -35,16 +35,17 @@
 #endif
 
 #include <gui/states/state_normal.h>
+#include <states/state_select.h>
+
+#include <synfig/general.h>
 
 #include <gui/app.h>
-#include <gui/canvasview.h>
 #include <gui/event_keyboard.h>
 #include <gui/event_layerclick.h>
 #include <gui/event_mouse.h>
 #include <gui/docks/dialog_tooloptions.h>
 #include <gui/docks/dock_toolbox.h>
 #include <gui/localization.h>
-#include <gui/workarea.h>
 
 #include <synfig/angle.h>
 
@@ -70,39 +71,6 @@ const int GAP = 3;
 StateNormal studio::state_normal;
 
 /* === C L A S S E S & S T R U C T S ======================================= */
-
-class DuckDrag_Combo : public DuckDrag_Base
-{
-	synfig::Vector last_move;
-	synfig::Vector drag_offset;
-	synfig::Vector center;
-	synfig::Vector snap;
-
-	synfig::Angle original_angle;
-	synfig::Real original_mag;
-
-	std::vector<synfig::Vector> last_;
-	std::vector<synfig::Vector> positions;
-
-
-	bool bad_drag;
-	bool move_only;
-
-	bool is_moving;
-
-public:
-	CanvasView* canvas_view_;
-	bool scale;
-	bool rotate;
-	bool constrain;
-	DuckDrag_Combo();
-	void begin_duck_drag(Duckmatic* duckmatic, const synfig::Vector& begin);
-	bool end_duck_drag(Duckmatic* duckmatic);
-	void duck_drag(Duckmatic* duckmatic, const synfig::Vector& vector);
-
-	etl::handle<synfigapp::CanvasInterface> get_canvas_interface()const{return canvas_view_->canvas_interface();}
-};
-
 
 class studio::StateNormal_Context : public sigc::trackable
 {
--- a/synfig-studio/src/gui/states/state_select.cpp
+++ b/synfig-studio/src/gui/states/state_select.cpp
@@ -1,0 +1,378 @@
+/* === S Y N F I G ========================================================= */
+/*!	\file state_select.cpp
+**	\brief Template File
+**
+**	\legal
+**	Copyright (c) 2002-2005 Robert B. Quattlebaum Jr., Adrian Bentley
+**	Copyright (c) 2007, 2008 Chris Moore
+**	Copyright (c) 2009 Nikita Kitaev
+**
+**	This file is part of Synfig.
+**
+**	Synfig is free software: you can redistribute it and/or modify
+**	it under the terms of the GNU General Public License as published by
+**	the Free Software Foundation, either version 2 of the License, or
+**	(at your option) any later version.
+**
+**	Synfig is distributed in the hope that it will be useful,
+**	but WITHOUT ANY WARRANTY; without even the implied warranty of
+**	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+**	GNU General Public License for more details.
+**
+**	You should have received a copy of the GNU General Public License
+**	along with Synfig.  If not, see <https://www.gnu.org/licenses/>.
+**	\endlegal
+*/
+/* ========================================================================= */
+
+/* === H E A D E R S ======================================================= */
+
+#ifdef USING_PCH
+#	include "pch.h"
+#else
+#ifdef HAVE_CONFIG_H
+#	include <config.h>
+#endif
+
+#include <gui/states/state_select.h>
+
+#include <synfig/general.h>
+
+#include <gui/app.h>
+#include <gui/canvasview.h>
+#include <gui/event_keyboard.h>
+#include <gui/event_layerclick.h>
+#include <gui/event_mouse.h>
+#include <gui/docks/dialog_tooloptions.h>
+#include <gui/docks/dock_toolbox.h>
+#include <gui/localization.h>
+#include <gui/workarea.h>
+
+#include <synfig/angle.h>
+
+#include <synfigapp/main.h>
+
+#endif
+
+/* === U S I N G =========================================================== */
+
+using namespace synfig;
+using namespace studio;
+
+/* === M A C R O S ========================================================= */
+
+#ifndef EPSILON
+#define EPSILON	0.0000001
+#endif
+
+const int GAP = 3;
+
+/* === G L O B A L S ======================================================= */
+
+StateSelect studio::state_select;
+
+/* === C L A S S E S & S T R U C T S ======================================= */
+
+class studio::StateSelect_Context : public sigc::trackable
+{
+	CanvasView* canvas_view_;
+
+	etl::handle<DuckDrag_Select> duck_dragger_;
+	Gtk::Grid options_grid;
+	Gtk::Label title_label;
+	bool pressed = false;
+
+public:
+
+	bool get_lock_animation_flag()const
+	{
+		return get_canvas_view()
+			&& get_canvas_view()->get_work_area()
+			&& get_canvas_view()->get_work_area()->get_lock_animation_mode();
+	}
+	void set_lock_animation_flag(bool x)
+	{
+		if ( get_canvas_view()
+		  && get_canvas_view()->get_work_area()
+		  && get_canvas_view()->get_work_area()->get_lock_animation_mode() != x )
+		{
+			get_canvas_view()->get_work_area()->set_lock_animation_mode(x);
+			get_canvas_view()->get_work_area()->queue_draw();
+		}
+	}
+
+	StateSelect_Context(CanvasView* canvas_view);
+
+	~StateSelect_Context();
+
+	CanvasView* get_canvas_view()const{return canvas_view_;}
+	etl::handle<synfigapp::CanvasInterface> get_canvas_interface()const{return canvas_view_->canvas_interface();}
+	synfig::Canvas::Handle get_canvas()const{return canvas_view_->get_canvas();}
+	WorkArea * get_work_area()const{return canvas_view_->get_work_area();}
+
+	Smach::event_result event_refresh_handler(const Smach::event& x);
+	Smach::event_result event_stop_handler(const Smach::event& x);
+	Smach::event_result event_undo_handler(const Smach::event& x);
+	Smach::event_result event_redo_handler(const Smach::event& x);
+	Smach::event_result event_refresh_ducks_handler(const Smach::event& x);
+	Smach::event_result event_mouse_button_down_handler(const Smach::event& x);
+//	Smach::event_result event_mouse_motion_handler(const Smach::event& x);
+	Smach::event_result event_mouse_release_handler(const Smach::event& x);
+	Smach::event_result event_refresh_tool_options(const Smach::event& x);
+	void refresh_tool_options();
+	Smach::event_result event_layer_click(const Smach::event& x);
+
+
+};	// END of class StateSelect_Context
+
+/* === M E T H O D S ======================================================= */
+
+StateSelect::StateSelect():
+	Smach::state<StateSelect_Context>("select")
+{
+	insert(event_def(EVENT_STOP,&StateSelect_Context::event_stop_handler));
+	insert(event_def(EVENT_REFRESH,&StateSelect_Context::event_refresh_handler));
+	insert(event_def(EVENT_REFRESH_DUCKS,&StateSelect_Context::event_refresh_ducks_handler));
+	insert(event_def(EVENT_UNDO,&StateSelect_Context::event_undo_handler));
+	insert(event_def(EVENT_REDO,&StateSelect_Context::event_redo_handler));
+	insert(event_def(EVENT_WORKAREA_MOUSE_BUTTON_DOWN,&StateSelect_Context::event_mouse_button_down_handler));
+	insert(event_def(EVENT_REFRESH_TOOL_OPTIONS,&StateSelect_Context::event_refresh_tool_options));
+//	insert(event_def(EVENT_WORKAREA_MOUSE_MOTION,		&StateSelect_Context::event_mouse_motion_handler));
+//	insert(event_def(EVENT_WORKAREA_MOUSE_BUTTON_DRAG,	&StateSelect_Context::event_mouse_motion_handler));
+	insert(event_def(EVENT_WORKAREA_MOUSE_BUTTON_UP,	&StateSelect_Context::event_mouse_release_handler));
+	insert(event_def(EVENT_WORKAREA_LAYER_CLICKED,&StateSelect_Context::event_layer_click));
+}
+
+StateSelect::~StateSelect()
+{
+}
+
+void* StateSelect::enter_state(studio::CanvasView* machine_context) const
+{
+	return new StateSelect_Context(machine_context);
+}
+
+
+StateSelect_Context::StateSelect_Context(CanvasView* canvas_view):
+	canvas_view_(canvas_view),
+	duck_dragger_(new DuckDrag_Select())
+{
+	duck_dragger_->canvas_view_=get_canvas_view();
+
+	// Toolbox widgets
+	title_label.set_label(_("Select Tool"));
+	Pango::AttrList list;
+	Pango::AttrInt attr = Pango::Attribute::create_attr_weight(Pango::WEIGHT_BOLD);
+	list.insert(attr);
+	title_label.set_attributes(list);
+	title_label.set_hexpand();
+	title_label.set_halign(Gtk::ALIGN_START);
+	title_label.set_valign(Gtk::ALIGN_CENTER);
+
+	// Toolbox layout
+	options_grid.attach(title_label,
+		0, 0, 1, 1);
+
+	options_grid.set_border_width(GAP*2);
+	options_grid.set_row_spacing(GAP);
+	options_grid.set_margin_bottom(0);
+	options_grid.show_all();
+
+	refresh_tool_options();
+
+	get_work_area()->set_allow_layer_clicks(true);
+	get_work_area()->set_duck_dragger(duck_dragger_);
+
+	App::dock_toolbox->refresh();
+}
+
+void
+StateSelect_Context::refresh_tool_options()
+{
+	App::dialog_tool_options->clear();
+	App::dialog_tool_options->set_widget(options_grid);
+	App::dialog_tool_options->set_local_name(_("Select Tool"));
+	App::dialog_tool_options->set_icon("tool_select_icon");
+}
+
+
+
+StateSelect_Context::~StateSelect_Context()
+{
+	get_work_area()->reset_cursor();
+
+	App::dialog_tool_options->clear();
+
+	App::dock_toolbox->refresh();
+}
+
+DuckDrag_Select::DuckDrag_Select(){
+
+}
+
+void
+DuckDrag_Select::begin_duck_drag(Duckmatic* duckmatic, const synfig::Vector& offset)
+{
+	is_moving = false;
+	last_move=Vector(1,1);
+
+	const DuckList selected_ducks(duckmatic->get_selected_ducks());
+	DuckList::const_iterator iter;
+
+	bad_drag=false;
+
+//		drag_offset=duckmatic->find_duck(offset)->get_trans_point();
+		drag_offset=offset;
+		//snap=drag_offset-duckmatic->snap_point_to_grid(drag_offset);
+		//snap=offset-drag_offset_;
+		snap=Vector(0,0);
+
+	// Calculate center
+	Point vmin(100000000,100000000);
+	Point vmax(-100000000,-100000000);
+	//std::set<etl::handle<Duck> >::iterator iter;
+	positions.clear();
+	int i;
+	for(i=0,iter=selected_ducks.begin();iter!=selected_ducks.end();++iter,i++)
+	{
+		Point p((*iter)->get_trans_point());
+		vmin[0]=std::min(vmin[0],p[0]);
+		vmin[1]=std::min(vmin[1],p[1]);
+		vmax[0]=std::max(vmax[0],p[0]);
+		vmax[1]=std::max(vmax[1],p[1]);
+		positions.push_back(p);
+	}
+	center=(vmin+vmax)*0.5;
+	if((vmin-vmax).mag()<=EPSILON)
+		move_only=true;
+	else
+		move_only=false;
+
+
+	synfig::Vector vect(offset-center);
+	original_angle=Angle::tan(vect[1],vect[0]);
+	original_mag=vect.mag();
+}
+
+Smach::event_result
+StateSelect_Context::event_refresh_tool_options(const Smach::event& /*x*/)
+{
+	refresh_tool_options();
+	return Smach::RESULT_ACCEPT;
+}
+
+Smach::event_result
+StateSelect_Context::event_stop_handler(const Smach::event& /*x*/)
+{
+	//synfig::info("STATE SELECT: Received Stop Event");
+	canvas_view_->stop();
+	return Smach::RESULT_ACCEPT;
+}
+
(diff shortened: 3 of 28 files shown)


mohamedAdhamc:
For some odd reason one of the builds here is failing even though the only change was that I made a constructor explicit o_O.

rodolforg:
The error message says what you shall do here: add `src/gui/states/state_select.cpp` to `po/POTFILES.in`.

image

mohamedAdhamc:
> The error message says what you shall do here: add `src/gui/states/state_select.cpp` to `po/POTFILES.in`.
> 

Oh thanks for notifying :) I guess I was too confused as to why it only started happening now to actually see the error.

mohamedAdhamc:
Hey @rodolforg , There are some things here that I want your opinions about.

When moving layers depending on their type some of the handles are necessary to be able to see where they are being moved. I mean the handles that are related to the outlines of the layers.
For example:
1- radius handle for circle layer, and star layer
2- position handle for rectangle and polygon

And also in general how do you think this tool should deal with handles. Is it best to just completely hide handles while using the select tool or is this not necessary ?
(the internal working of the tool can utilize ducks to move the layer and apply transformations but they don't have to be shown at the user level)

But then back to the original point if we don't show any ducks/handles then maybe we can just make new "outline ducks" that are always visible when using this tool. does this sound good ?

rodolforg:
Yes, a "bounding box" should be enough to show when a layer is selected with the  selection tool on.

Just like LibreOffice Draw, Inkscape and any other normal graphics software lol

mohamedAdhamc:
Quick update here:

-General- 
Finally I'm done with my exams and will start resuming/starting work for my project.

-technical-
The major parts remaining here now are the following:
1- A sort of "bounding box/polygon/circle/curve" to show while the item is being dragged. This is probably the biggest thing here, depending on the type of layer it needs a different shape of outline that matches it. I'm still not exactly sure of the best way to implement that but I'm working on it currently and experimenting with different strategies. Any suggestions here would be greatly appreciated.

2- Moving a layer which is a part of a group moves the whole group: this shouldn't take much time.

3- refactoring/ use actual icons and cursors: This also shouldn't be much of a problem. It's just that I usually try to get to a working prototype quickly so I sometimes leave certain parts for later refactoring for those reasons.

mohamedAdhamc:
Bounding thing seems to be done. Which means that mostly only little tweaks and refactoring are needed. If things go well maybe this will be ready for review in about a day or two.

[Screencast from 06-18-2023 11:17:11 AM.webm](https://github.com/synfig/synfig/assets/100296264/f795381f-abb9-43fb-b81f-769267c09a42)

rodolforg:
> A sort of "bounding box/polygon/circle/curve" to show while the item is being dragged.

If not well defined, it can be a bounding rectangle, as used in other software.
You can get this via `Layer::get_bounding_rect()` or `Layer::get_full_bounding_rect()`.



> 2- Moving a layer which is a part of a group moves the whole group: this shouldn't take much time.

Hm... It depends. User may want to move that particular element only.
You could check if user want to move the group or the element only:
- by reading the `param_children_lock` of `Layer_PasteCanvas` (group, switch group, filter group); or
- by adding a select tool setting that can override the `param_children_lock` (i.e., no matter the `param_children_lock` , Select would behave as set by user in Tool Options Panel)

mohamedAdhamc:
> 
> If not well defined, it can be a bounding rectangle, as used in other software. You can get this via `Layer::get_bounding_rect()` or `Layer::get_full_bounding_rect()`.

Well, what do you think of the current bounding outline (in the very last video) ? I think that's a bit better than just a rectangle for all shapes because you can see better how the layer will be after or moved. What do you think ?


> > 2- Moving a layer which is a part of a group moves the whole group: this shouldn't take much time.
> 
> Hm... It depends. User may want to move that particular element only. You could check if user want to move the group or the element only:
> 
>     * by reading the `param_children_lock` of `Layer_PasteCanvas` (group, switch group, filter group); or
> 
>     * by adding a select tool setting that can override the `param_children_lock` (i.e., no matter the `param_children_lock` , Select would behave as set by user in Tool Options Panel)

hmm, so basically have this setting be user customizable ? I like this idea will do.

rodolforg:
> I think that's a bit better than just a rectangle for all shapes because you can see better how the layer will be after being moved. What do you think ?

I did find it is good! :)

pgilfernandez:
Nice feature!
Do you want me too design the icon or you plan to do it yourself?

I would like to suggest that this move functionality works with multiple selection as well, is that planned or already planned?

mohamedAdhamc:
> Nice feature! Do you want me too design the icon or you plan to do it yourself?

Hello! yes that would be awesome if you can design the icons. Actually I was going to ask you if you would have the time to do so :). 

p.s. I think for now what is needed is just an icon for the select tool, but later on depending on how the other transformations are implemented there may be additional icons needed I'm not sure yet at this point.

> I would like to suggest that this move functionality works with multiple selection as well, is that planned or already planned?

Oh yes definitely, but I think depending on how complex it may get (I'm assuming it shouldn't but I haven't looked yet) maybe it could be done after the initial tool implementation is done.

mohamedAdhamc:
Ok, another quick update here.

It seems everything is getting closed to finished here in the last commit i added:
- a checkbutton to choose whether the select tool "prioritizes groups"
- handle buttons are now insensitive when the select state is on

So now what remains till this is ready for review is mostly:
- refactoring and cleaning up this pr

I should be done with that by tomorrow (hopefully if everything goes as expected).

And for it to be "complete":
- using the icon pgilfernandez designs when he is done with it.

But I was wondering also @rodolforg, this tool has also a couple transformations I would like to add to it. stuff like (rotate, scale, mirror, etc...). So i wanted an opinion on two things:
1- I was thinking of making separate PRs based on this one. Then whenever this is merged the other ones which would be ready  
     to be merged can be merged then. Does this sound good ?
2- For the other transformations as we already have separate tools for them. What do you think of having them as states for this 
    select tool. So inside the select tool you would choose a state, one of (move, rotate, scale, etc...). The advantage here is that
    this would allow for this tool to be introduce alongside the other vertex tools as an alternative without causing issues.

pgilfernandez:
> p.s. I think for now what is needed is just an icon for the select tool, but later on depending on how the other transformations are implemented there may be additional icons needed I'm not sure yet at this point.

OK, let's do it

Thinking about these new tools and how they differentiate from the previous ones I have a question: do they really add new functionalities or they are just a different approaches? what I mean is if there is room for have them both or just one in the near future?
For instance, what you are designing as "selection tool" looks to me as a "move tool", a better approach to the actual "move tool" indeed... but, why to keep both?

mohamedAdhamc:
> 
> OK, let's do it

great : ).
 
> Thinking about these new tools and how they differentiate from the previous ones I have a question: do they really add new functionalities or they are just a different approaches? what I mean is if there is room for have them both or just one in the near future? For instance, what you are designing as "selection tool" looks to me as a "move tool", a better approach to the actual "move tool" indeed... but, why to keep both?

Well that is a bit similar to what I was thinking. Some tools like "move", "mirror", etc... could just be replaced. I think for these we might just need more opinions. 

But other than these I think the current transform tool is quite handy, it allows quick edits to layers by directly using the handles which correspond to the layers parameters, this is probably one of the few good things about vertex tools. you can do a lot with the same tool.

rodolforg:
> For instance, what you are designing as "selection tool" looks to me as a "move tool", a better approach to the actual "move tool" indeed... but, why to keep both?

They are not the same thing. The following tools are focused on vector image editing, by handling its vertices (although we can use for other handles):
image

What you called "Move" tool is actually "Smooth Move" Tool.

mohamedAdhamc:
> What do you think?

Thanks for the idea/sketches :). I like the idea that for example rotate, scale, etc.. would be differentiated from  their vertex/vector counter parts with something like a vector point. 

But there maybe something that I'm not getting. I believe these are the icons for the cursors `Gdk::FLEUR`, `Gdk::SIZING`, etc... . I don't think these need to be changed  (at least for now). Actually I'm not sure at all about creating custom icons for cursor and the issues that may come along with it. I would say Rodolfo would probably know more about this. But I think that could be an idea better left for another PR hopefully after this tool is merged. For now I think just using the same cursors would suffice.

So for now at least I think here we need an icon for the "select" tool itself in the "classic" style.
Screenshot from 2023-06-22 16-47-24

pgilfernandez:
> So for now at least I think here we need an icon for the "select" tool itself in the "classic" style.

No problem, I understand. I was going to far, hehehe, anyway it was a good "brainstorm" in case this new tool gets its cousins for rotating, scaling and mirroring.

I'll design the "select" tool by now and I'll let you know pretty soon.

pgilfernandez:
Here you have a first version, I have also updated the
tool_select_icon
[tool_select_icon.sif.zip](https://github.com/synfig/synfig/files/11835551/tool_select_icon.sif.zip)
 Smooth move icon to make them differentiate a little bit more. Try them and let me know:
tool_smooth_move_icon
[tool_smooth_move_icon.sif.zip](https://github.com/synfig/synfig/files/11835553/tool_smooth_move_icon.sif.zip)

mohamedAdhamc:
> Here you have a first version, I have also updated the Smooth move icon to make them differentiate a little bit more. Try them and let me know

Great. Also I like using a brighter color for the circle in the middle as it makes it more noticeable especially since the icon is usually very small in the program. Actually I don't think I ever noticed the original icon had a circle in the middle xD.

I will try them and let you know. I think I only have a small concern that since they are very similar icons it would be a bit difficult for the eye to differentiate between them (although I maybe wrong).

mohamedAdhamc:
Screenshot from 2023-07-02 22-28-16

Here are the new icons provided by @pgilfernandez.

Also @pgilfernandez if you would like whenever you have the time you can send me the icons you sent at the beginning representing the "move" "rotate" "scale" "mirror" transformations (it would be really great). They will be used inside the tool panel of the select tool. You can check #3114 to see what I mean.

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -462,7 +513,9 @@ Duckmatic::start_duck_drag(const synfig::Vector& offset)
 		duck_dragger_->begin_duck_drag(this,offset);
 
 	//drag_offset_=offset;
-	drag_offset_=find_duck(offset)->get_trans_point();
+	//in the case of using the select tool there are no ducks visible
+	if (find_duck(offset))
```
ice0, line 517: It's better to store result in a variable so that the function is not called twice.

synfig-studio/src/gui/states/state_select.cpp
```diff
@@ -0,0 +1,560 @@
...
+	Pango::AttrList list;
+	Pango::AttrInt attr = Pango::Attribute::create_attr_weight(Pango::WEIGHT_BOLD);
+	list.insert(attr);
+	title_label.set_attributes(list);
+	title_label.set_hexpand();
+	title_label.set_halign(Gtk::ALIGN_START);
+	title_label.set_valign(Gtk::ALIGN_CENTER);
+
+	//should words be capitalized like this ?
+	prioritize_groups_label.set_label(_("Priotirize Group Selection"));
```
ice0, line 164: `Priotirize`

synfig-studio/src/gui/states/state_select.cpp
```diff
@@ -0,0 +1,560 @@
...
+		}
+	}
+
+	if (rotate)
+	{
+		Angle::deg angle(Angle::tan(vect[1],vect[0]));
+		angle=original_angle-angle;
+		if (constrain)
+		{
+			float degrees = angle.get()/15;
```
ice0, line 360: Better to extract magic number 15 to some constexpr variable.

synfig-studio/src/gui/states/state_select.cpp
```diff
@@ -0,0 +1,560 @@
...
+
+			Vector x(positions[i]-center),p;
+
+			p[0]=cosine*x[0]+sine*x[1];
+			p[1]=-sine*x[0]+cosine*x[1];
+			if(scale)p*=mag;
+			p+=center;
+			(*iter)->set_trans_point(p, time);
+		}
+		for(i=0,iter=selected_movement_ducks.begin();iter!=selected_movement_ducks.end();++iter,i++)
```
ice0, line 379: Looks like duplicated code.

synfig-studio/src/gui/states/state_select.cpp
```diff
@@ -0,0 +1,560 @@
...
+
+Smach::event_result
+StateSelect_Context::event_mouse_release_handler(const Smach::event& x){
+	return Smach::RESULT_OK;
+}
+
+Smach::event_result
+StateSelect_Context::event_layer_click(const Smach::event& x)
+{
+	const EventLayerClick& event(*reinterpret_cast<const EventLayerClick*>(&x));
```
ice0, line 530: Why just not reinterpret it directly?
```
const EventLayerClick& event(reinterpret_cast<const EventLayerClick&>(x));
```

ice0:
Also, it didn't work as I expected. Can you check if anything is broken?

synfig-studio/src/gui/states/state_select.cpp
mohamedAdhamc: Oh yes, actually almost everything in `begin/end/duck_drag` is the same except for like 4/5 lines in each. I actually here inherit from the `DuckDrag_Combo` and overload these three methods just to make the same 4/5 changes in each. 

I was actually thinking about just making maybe the base `DuckDrag_Combo` more general to avoid having to overload them again with almost the same implementation. So ok it looks like that would be the best here.

mohamedAdhamc:
> Also, it didn't work as I expected. Can you check if anything is broken?

One thing I knew was still remaining was that the edges/bezier sometimes are draged if you click on them instead of the whole shape being dragged. I will fix this I don't think it should take much.

Was that it or what else exactly wasn't working when you tried ?

synfig-studio/src/gui/states/state_select.cpp
mohamedAdhamc: done.

synfig-studio/src/gui/states/state_select.cpp
mohamedAdhamc: done

mohamedAdhamc:
Ok @ice0, it should now be working as expected. I tested it for all the circle/rectangle/star/polygon/gradient layers with all their types.

It would be great if you can retest now and tell me if everything is working as expected.

yoSachinkr:
Just wondering if we can multi select and move as well. 🤔 @mohamedAdhamc

mohamedAdhamc:
> Just wondering if we can multi select and move as well. thinking @mohamedAdhamc

:smile:   you're actually the second to ask about this. As of now no... I was thinking of leaving it for another PR to be added onto this after this would be approved. But I think it will probably be just a small-ish adjustment so I will work on adding it in this PR.

pgilfernandez:
> Also @pgilfernandez if you would like whenever you have the time you can send me the icons you sent at the beginning representing the "move" "rotate" "scale" "mirror" transformations (it would be really great). They will be used inside the tool panel of the select tool. You can check #3114 to see what I mean.

Hi there!
I have been out some days ;)
Here you have the icons (there's room for improvements or changes, just use them as a first version):
tool_mirror_icon
[tool_mirror_icon.sif.zip](https://github.com/synfig/synfig/files/12001155/tool_mirror_icon.sif.zip)
tool_mirror-vector_icon
[tool_mirror-vector_icon.sif.zip](https://github.com/synfig/synfig/files/12001157/tool_mirror-vector_icon.sif.zip)
tool_rotate_icon
[tool_rotate_icon.sif.zip](https://github.com/synfig/synfig/files/12001160/tool_rotate_icon.sif.zip)
tool_rotate-vector_icon
[tool_rotate-vector_icon.sif.zip](https://github.com/synfig/synfig/files/12001167/tool_rotate-vector_icon.sif.zip)
tool_scale_icon
[tool_scale_icon.sif.zip](https://github.com/synfig/synfig/files/12001168/tool_scale_icon.sif.zip)
tool_scale-vector_icon
[tool_scale-vector_icon.sif.zip](https://github.com/synfig/synfig/files/12001170/tool_scale-vector_icon.sif.zip)

You can also change the naming as you wish... I'm not happy with the one I used =p

Cheers

mohamedAdhamc:
> Hi there! I have been out some days ;) Here you have the icons (there's room for improvements or changes, just use them as a first version): 

Great thanks a lot :).

mohamedAdhamc:
@yoSachinkr now this tool has multiple selection of layers just hold control to select different layers : ).

yoSachinkr:
@mohamedAdhamc Everything works like a charm 🤩
There're few issues which I can think at the moment:
* Moving shapes, moves the origin as well. This has no issues in normal uses, but it can cause a bit of trouble specifically while using **skeleton layer** with **link to skeleton** functionality.
* Select tool doesn't work with text layer.

rodolforg:
@mohamedAdhamc, as @yoSachinkr said, moving selected objects by dragging them should move their origin parameter (and that should always be named as "origin"). That would allow to move Group layers too.

In another PR, it would be cool to show/create handles as other software do:
- in the corners and in the edges to resize it (at least for shapes and group-like layers). Shift (or Ctrl?) should constraint the resize.
- in the corners but shifted a bit away for rotating
- the origin, but not for moving the shape/layer, but for changing the origin itself (the pivot point)

mohamedAdhamc:
> @mohamedAdhamc, as @yoSachinkr said, moving selected objects by dragging them should move their origin parameter (and that should always be named as "origin"). That would allow to move Group layers too.

Hmm, but the select tool already does this. If there is an origin in the layer it moves it. Actually I think @yoSachinkr was suggesting to not move the origin as it may cause issues with bones. What do you think ?

> In another PR, it would be cool to show/create handles as other software do:
> 
>     * in the corners and in the edges to resize it (at least for shapes and group-like layers). Shift (or Ctrl?) should constraint the resize.
> 
>     * in the corners but shifted a bit away for rotating
> 
>     * the origin, but not for moving the shape/layer, but for changing the origin itself (the pivot point)

Great so as a final decision handles for the extra transformations instead of what I had done in the other 2 (rotate/mirror PRs) ?

pgilfernandez:
> Hmm, but the select tool already does this. If there is an origin in the layer it moves it. Actually I think @yoSachinkr was suggesting to not move the origin as it may cause issues with bones. What do you think ?

I like both behaviors, but I prefer to move the origin. BTW, could you include boths by adding a setting in the "Tool options" panel to enable/disable that transformation?

yoSachinkr:
> Actually I think @yoSachinkr was suggesting to not move the origin as it may cause issues with bones. What do you think ?

Yes, the origin shouldn't move and stay at (0,0) to not cause any issues with skeleton layer. @mohamedAdhamc @rodolforg

mohamedAdhamc:
> 
> I like both behaviors, but I prefer to move the origin. BTW, could you include boths by adding a setting in the "Tool options" panel to enable/disable that transformation?

Well I believe yes it could be done. But the question is would it be useful to users or just confuse them ? Also shouldn't origins have to somewhat relate to the place of the layer. Like what is the point of the origin if it's just the same point for all layers xD ? (maybe i'm missing something )

> Yes, the origin shouldn't move and stay at (0,0) to not cause any issues with skeleton layer. @mohamedAdhamc @rodolforg

How about I just make an exception to allow skeletons to work with the select tool (if it's possible xD) instead of adjusting the functionality of the origin for all layer types ? 

But I think this may take a little time I have to look more into it then I will be back reporting what I reached.

yoSachinkr:
> How about I just make an exception to allow skeletons to work with the select tool

@mohamedAdhamc If this is possible it'll fix every problem new users face while using bones in Synfig.
For context:

https://github.com/synfig/synfig/assets/50204208/d1f2f6bc-0785-4980-8a33-f45c04a968f1

rodolforg:
Wouldn't it be an error of "Skeleton layer"? Actually linking vertices to skeleton.

mohamedAdhamc:
I will be using this comment to reply to the issues pointed out by @yoSachinkr as I'm investigating them right now.

> 
>     * Moving shapes, moves the origin as well. This has no issues in normal uses, but it can cause a bit of trouble specifically while using **skeleton layer** with **link to skeleton** functionality.

1- I think the general consensus here is that it's good for origins to also move with the layer. However, if it's absolutely needed that for some layer we dont move the origin then we can do what pgilfernandez recommended by having an option to move it or not.

2- Regarding skeleton tool honestly I have no clue. Actually I personally think skeleton tool is best to be used with the regular transform tool. As I would assume it would be quite difficult to have the same level of control over it without the handles.

>     * Select tool doesn't work with text layer.

It does work it's just that you have to click on the text itself, but the fact that the text layer doesn't have a sort of outline does make it a bit confusing indeed but I think this is unrelated to the select tool specifically.

[Screencast from 07-13-2023 11:52:58 PM.webm](https://github.com/synfig/synfig/assets/100296264/7af237ea-b09f-4af1-86ee-e31da3e796d0)


 
>     * Constraining the movement of shapes using select tool by holding the "shift" key results in offset of constrains in the canvas.

It is working for some layers but not others. I am still investigating it.

yoSachinkr:
> I will be using this comment to reply to the issues pointed out by @yoSachinkr as I'm investigating them right now

@mohamedAdhamc My understanding says when a shape is linked to the bone, the origin of the bone becomes the de facto origin of the shape and since positions of all the vertices are calculated wrt to the origin so when the origin of the shape moves somehow the bone is unable to incorporate the offset in origin and assumes the origin of the shape is still at zero and calculates the relative positions of all the vertices as it was without factoring in the offset.

rodolforg:
@yoSachinkr The bone linking completely ignores the Origin parameter, that's way it currently needs to be (0, 0), otherwise it makes unexpected deformations.

It is the same reason Bones must be on the same canvas the linked layers are (or keep the group layers at (0,0) too).

We can "easily" fix it if, after linking vertices to bone/skeleton, user doesn't change the layer origin – just as it is now.

rodolforg (changes requested):
I suggest you go by a different approach.

Instead of changing `Duckmatic` to get the "selected movement ducks", you should prevent the tool/state to call `CanvasView::queue_rebuild_ducks()`, as some of them do. Polygon state creates its own ducks (only) while user is creating shape. See its `refresh_ducks()` and `insert(event_def(EVENT_REFRESH_DUCKS,&StatePolygon_Context::event_refresh_handler));`

Also, you seem to look for layer that have vertex-type ducks (added by `Duckmatic::add_ducks_layers(synfig::Canvas::Handle canvas, std::set<synfig::Layer::Handle>& selected_layer_set, CanvasView::Handle canvas_view, synfig::TransformStack& transform_stack, int *out_transform_count)` what you should avoid calling).
For now, let's focus on layers with Origin parameter. Then, include those exceptions like Bitmap and Rectangle layers, whose position is defined by top-left and bottom-right coordinates.

Allow multiple selection by adding/removing layers by `ctrl` clicks.

In another PR, it would be very good to let it selection via dragging boxes (those layers whose bounding box is contained by the dragging box)

mohamedAdhamc:
Hello!

> 
> Instead of changing `Duckmatic` to get the "selected movement ducks", you should prevent the tool/state to call `CanvasView::queue_rebuild_ducks()`, as some of them do. Polygon state creates its own ducks (only) while user is creating shape. See its `refresh_ducks()` and `insert(event_def(EVENT_REFRESH_DUCKS,&StatePolygon_Context::event_refresh_handler));`

I think maybe you confused this PR with the other PR for adding rotation handles ? Because yes having the select tool create its own ducks only when the user is in the select state would be better than what I have done here. However in this PR I don't create any ducks.

What happens and why I edited duckmatic was to be able to store the ducks of the selected layers regardless of their visibility (note that the duck show/hide buttons are all disable when entering the select state) and various other things. I store them in a container 'selected_movement_ducks' (maybe another name would be more appropriate) then whenever I want any of the ducks of the currently selected layers I use the method I added `get_selected_movement_ducks`.

This is neccessary as the current 'selected_ducks' container only has the ducks which are "selected" which has the condition of the ducks type being visible. So even if I use `select_all_ducks` method it only selects the ducks which are visible. And this is not good because we always need all the selected layer ducks regardless if they are shown/hidden as we are simple using them to apply transformations to the layer (in the select tool they are always hidden).

What do you think ?
 
> Also, you seem to look for layer that have vertex-type ducks (added by `Duckmatic::add_ducks_layers(synfig::Canvas::Handle canvas, std::set<synfig::Layer::Handle>& selected_layer_set, CanvasView::Handle canvas_view, synfig::TransformStack& transform_stack, int *out_transform_count)` what you should avoid calling). For now, let's focus on layers with Origin parameter. Then, include those exceptions like Bitmap and Rectangle layers, whose position is defined by top-left and bottom-right coordinates.

What I'm currently doing is that after I have all the ducks of the selected layers. To move I do the following
 - iterate over the selected ducks if there are position ducks available I use them to drag the layer.
 - if there are no position ducks available then I just use all the other selected ducks to move
 
This approach so far (hopefully there isn't an exception I havent noticed) works for all types of layers. Is there an issue with it ? 

I suppose you are mentioning that in fact for layers with an origin parameter the origin duck by itself is sufficient to move the layer we don't also need the other position ducks, but whether we only use the origin duck or the other position ducks along with it we end up moving the layer. Exactly like how you can move the layer by only dragging the origin or by selecting the origin as well as the other position ducks and dragging. They are both equivalent but selecting all the position ducks as well is just more general as if we have multiple layers selected one could have an origin and the other just position ducks etc... . 

> Allow multiple selection by adding/removing layers by `ctrl` clicks.

Yes, I have done so multiple selection now works by using holding control while selecting the layers and dragging.
 
> In another PR, it would be very good to let it selection via dragging boxes (those layers whose bounding box is contained by the dragging box)

Yes I have also thought this is a good feature to be added as other programs as well do it. I think it shouldn't also take too much time to implement. I will leave it till after we finish up this PR and the other/s with bonus transformations (rotate/scale).

synfig-studio/src/gui/states/Makefile_insert
```diff
@@ -32,10 +33,11 @@ STATES_CC = \
 	states/state_fill.cpp \
 	states/state_gradient.cpp \
 	states/state_mirror.cpp \
-	states/state_normal.cpp \
 	states/state_polygon.cpp \
 	states/state_rectangle.cpp \
 	states/state_rotate.cpp \
```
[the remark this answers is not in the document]
mohamedAdhamc: oops didn't notice. will fix

synfig-studio/src/gui/states/state_normal.cpp
```diff
@@ -35,16 +35,17 @@
 #endif
 
 #include <gui/states/state_normal.h>
+#include <states/state_select.h>
```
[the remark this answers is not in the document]
mohamedAdhamc: previously I used a separate duck dragger which inherited from Duck Drag Combo. But then it was visible then it was no longer needed and Duck Drag Combo was sufficient to use. I forgot to remove the file though. Thanks for noticing will do.

mohamedAdhamc:
> Instead of changing `Duckmatic` to get the "selected movement ducks"

I realized the need for this in the (rotate) PR, and yes now it is much cleaner I feel.

rodolforg:
OK.

Let's go wild. What about create a new (intermediate) layer called `Layer_2D` or `Layer_2DNode` or `Layer_Node2D` (derived from `Layer_Composite)` that would have _the_ great parameters "origin" and "transform"?.
`Layer_Shape`, `Layer_PasteCanvas` and `Layer_Bitmap` would derive from this class, instead of `Layer_Composite` directly.

The Select Tool would work like this:
- User clicks on the work area.
  -  If it hits a layer that has a parameter named "origin" (and it is a vertex), it can be directly moved.
  -  If it hits a layer that has a parameter named "transform" (and it is a transform), it can be rotated and scaled too.

That way:
- we can remove current "transform duck", if we want to do so; 
- we embrace the Star, Circle, Rectangle, Polygon, Text etc. layers, Group, and Switch Group too.
- avoid being able to "move" the Blur layer parameter "size" as we sometimes can with this PR
- do not need to automatically add the Switch layer to every image we import, even when we don't want to use the switch feature.

As Skeleton Layer derives from `Layer_Shape`, maybe we should hide these "origin" and "transform" parameters.
Maybe we should do it too for `Layer_CompositeFork`. What do you think? @mohamedAdhamc @morevnaproject @ice0

Issue #3019, opened 2023-03-11
Crash after rendering file


morevnaproject:
**Describe the bug**
I've got a random crash, when rendering file from commandline.
The crash happens randomly, but it's more-less reproducible with specific file on specific machine.

Here is traceback with GDB (debug version):
```
$ gdb -batch -ex "run" -ex "bt" --args /home/konstantin/sources/synfig/_debug/build/bin/synfig /home/konstantin/1-work.m/pepper-and-carrot-ep04/scenes/007/007-final.sifz -o /tmp/file.avi --end-time=2
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffef891700 (LWP 7890)]
[New Thread 0x7fffef090700 (LWP 7891)]
No LADSPA plugins were found!

Check your LADSPA_PATH environment variable.
[New Thread 0x7fffc307d700 (LWP 7892)]
[New Thread 0x7fffc287c700 (LWP 7893)]
[New Thread 0x7fffc207b700 (LWP 7894)]
[New Thread 0x7fffc187a700 (LWP 7895)]
synfig(7886) [13:04:28] информация: rendering threads 4
synfig(7886) [13:04:28] информация: Loading lyr_std..
synfig(7886) [13:04:28] информация: Loading lyr_freetype..
synfig(7886) [13:04:29] информация: Loading mod_geometry..
synfig(7886) [13:04:29] информация: Loading mod_gradient..
synfig(7886) [13:04:29] информация: Loading mod_particle..
synfig(7886) [13:04:30] информация: Loading mod_example..
synfig(7886) [13:04:30] информация: Loading mod_gif..
synfig(7886) [13:04:30] информация: Loading mod_imagemagick..
synfig(7886) [13:04:30] информация: Loading mod_magickpp..
synfig(7886) [13:04:30] информация: Loading mod_mng..
synfig(7886) [13:04:30] информация: Loading mod_noise..
synfig(7886) [13:04:31] информация: Loading mod_filter..
synfig(7886) [13:04:31] информация: Loading mod_ffmpeg..
synfig(7886) [13:04:31] информация: Loading mod_bmp..
synfig(7886) [13:04:31] информация: Loading mod_dv..
synfig(7886) [13:04:31] информация: Loading mod_png..
synfig(7886) [13:04:31] информация: Loading mod_ppm..
[...]

PR #3037, opened 2023-03-27
fix: a crash on program exit if bones were used

ice0:
`ValueNode_Bone_Root` does not have a GUID and does not belong to any canvas.
Calling `get_guid()` for `ValueNode_Bone_Root` leads to an attempt to insert it into
a canvas (it calls `global_node_map()` for this purpose).
And if `global_node_map()` is already destroyed at that moment, it leads to a crash.

Since `ValueNode_Bone_Root` is also a static variable, it can be destroyed later than `global_node_map()`.
I moved this variable inside `get_root_bone()` method, which changed the initialization and destruction order.

I also added a check for `is_root()` for more correct behavior.

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_bone.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_bone.cpp
@@ -76,7 +76,6 @@ static int bone_counter;
 // static map<ValueNode_Bone::Handle, Matrix> animated_matrix_map;
 static Time last_time = Time::begin();
 
-static ValueNode_Bone_Root::Handle rooot;
 
 /* === P R O C E D U R E S ================================================= */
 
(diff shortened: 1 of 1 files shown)


rodolforg (approved):

rodolforg:
@morevnaproject does it fix the issue https://github.com/synfig/synfig/issues/3019 ?

mohamedAdhamc:
Apart from Konstantin's issue...

I just tested this by making a simple file and using bones:

before (master):
crash on exit with following error
`Application appears to have terminated successfully`
``synfigstudio: /home/mohamed/synfig/synfig-core/src/synfig/node.cpp:86: void {anonymous}::GlobalNodeMap::remove(const synfig::GUID&, synfig::Node*): Assertion `i != map.end() && i->second == node' failed.``
`Aborted (core dumped)`

after (this branch):
no longer crashes on exit.

morevnaproject:
I have tested and confirming - #3019 is fixed with this PR! :tada: :tada: :tada:

PR #3041, opened 2023-04-03
refactor: prefer Layer_*::Handle and alike rather etl::handle

rodolforg:
Instead of etl::handle<Layer_*>, etl::handle<const Layer_*> and etl::loose_handle<Layer_*>

Reason: etl::handle and alike will be moved to synfig namespace (and maybe one day it'll use an C++ standard smart pointer). It'll avoid future changes everywhere.


ice0 (approved):

ice0:
Merged. Thank you!

PR #3043, opened 2023-04-07
refactor: code cleanup in Targets

rodolforg:
And maybe it helps performance, if new array size is equal to previous one, because resize() wouldn't perform anything in this case.

### DIFF
--- a/synfig-core/src/modules/mod_bmp/trgt_bmp.cpp
+++ b/synfig-core/src/modules/mod_bmp/trgt_bmp.cpp
@@ -154,8 +154,6 @@ bmp::bmp(const char *Filename, const synfig::TargetParam& params):
 	multi_image(false),
 	file(nullptr),
 	filename(Filename),
-	buffer(nullptr),
-	color_buffer(nullptr),
 	pf()
 {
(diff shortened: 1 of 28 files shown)


rodolforg:
And macOS keeps randomly failing in app_layerduplicate :(
image

synfig-core/src/modules/mod_bmp/trgt_bmp.cpp
```diff
@@ -205,8 +201,6 @@ bmp::end_frame()
 {
 	if(file)
 		fclose(file);
-	delete [] color_buffer;
-	color_buffer=0;
```
rodolforg, line 209: I'm not sure if we have to replace both lines with `color_buffer.clear()`

ice0 (dismissed):

synfig-core/src/modules/mod_bmp/trgt_bmp.cpp
rodolforg: Done

rodolforg:
Rebased.

ice0 (approved):

ice0:
Merged. Thank you!

PR #3050, opened 2023-04-12
fix: possible invalid memory access on replacing valuenodes

### LINKED ISSUE #3049: macOS keeps randomly failing in app_layerduplicate :(
macOS keeps randomly failing in app_layerduplicate
image

_Originally posted by @rodolforg in https://github.com/synfig/synfig/issues/3043#issuecomment-1500520118_

rodolforg:
As `dyn_param_list` is a reference to a vector, and the vector can change in the loop, it can have invalid "iterators"/items.

Possible solutions:
1. `dyn_param_list` be a copy not a reference
2. Implement a `Layer::replace_dynamic_param()`
3. Store all the changes in another variable and postpone the actual changes until we finish the loop

The first one can create a "large" copy.

The second one is interesting, but I `git grep` the code and it does not seem to be needed anywhere else XD The existent disconnections are for replacing with a (static) ValueBase.

Therefore, I opted for the 3rd solution. It is lighter than option 1 and less 'intrusive' than the second one.

fix #3049


ice0:
Merged. Thank you!

PR #3052, opened 2023-04-17
refactor: remove unused members of `CanvasView`

mohamedAdhamc:
Noticed this while working on #3046. 

They are supposed to be the two initial rows of the library panel. Maybe the original author thought they would be used here. But they are neither used nor initialized with the correct rows that their name indicates.


rodolforg (approved):

ice0:
Merged. Thank you!

PR #3055, opened 2023-04-19
build: update packages before installing

ice0:
test if it still fails with newer packages


ice0:
Unfortunately, this did not help :(

PR #3060, opened 2023-04-30
refactor: rename method to get_link_description to avoid ambiguity

rodolforg:
There were two overloaded methods, and all of their arguments have default values:

* LinkableValueNode::get_description(bool show_exported_name = true)
* LinkableValueNode::get_description(int index = -1, bool show_exported_name = true)

The first one is an override of the `ValueNode::get_description` method.

There was an unnecessary ambiguity here. So I renamed the second method to `LinkableValueNode::get_link_description()` to make it explicit, and `index` argument is not optional now.

### DIFF
--- a/synfig-core/src/synfig/valuenode.h
+++ b/synfig-core/src/synfig/valuenode.h
@@ -333,20 +333,20 @@ private:
 
 public:
 
-	virtual ValueBase operator()(Time t)const;
+	ValueBase operator()(Time t) const override;
 
-	virtual String get_name()const;
+	String get_name() const override;
 
-	virtual String get_local_name()const;
+	String get_local_name() const override;
 
-	String get_string()const;
+	String get_string() const override;
 
-	virtual ValueNode::Handle clone(etl::loose_handle<Canvas> canvas, const GUID& deriv_guid=GUID())const;
+	ValueNode::Handle clone(etl::loose_handle<Canvas> canvas, const GUID& deriv_guid = GUID()) const override;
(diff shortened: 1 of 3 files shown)


synfig-core/src/synfig/valuenode.h
```diff
@@ -417,10 +417,9 @@ class LinkableValueNode : public ValueNode
 	//! Returns a Loose Handle to the Value Node based on the link's name
 	ValueNode::LooseHandle get_link(const String &name)const { return get_link(get_link_index_from_name(name)); }
 	//! Return a full description of the linked ValueNode given by the index
-	String get_description(int index = -1, bool show_exported_name = true)const;
-	//! Return a full description of the linked ValueNode given by the index
-	//! Proper overload of the inherited function
-	String get_description(bool show_exported_name = true)const;
+	String get_link_description(int index, bool show_exported_name = true)const;
+	//! Return a full description of this linkable ValueNode
+	String get_description(bool show_exported_name = true)const override;
```
rodolforg, line 422: I added `override` here, so `gcc` now warns about other override methods not being tagged as such.
Should I change them too?

ice0: I think yes, it's better to tag them too.

rodolforg: Done.

ice0 (approved):

ice0:
Merged. Thank you!

Issue #3063, opened 2023-05-06
"make distcheck" fails on synfig-core


morevnaproject:
I started preparing 1.5.2 release, but hit the following error when running `autobuild/synfigstudio-release.sh` script (the error happens after "make distcheck" command in synfig-core) -

```
...
make[3]: Leaving directory '/home/konstantin/sources/synfig/_release/synfig-core/synfig-1.5.1/_build/sub/examples'
make[2]: Leaving directory '/home/konstantin/sources/synfig/_release/synfig-core/synfig-1.5.1/_build/sub/examples'
Making check in po
make[2]: Entering directory '/home/konstantin/sources/synfig/_release/synfig-core/synfig-1.5.1/_build/sub/po'
INTLTOOL_EXTRACT="/usr/bin/intltool-extract" XGETTEXT="/usr/bin/xgettext" srcdir=../../../po /usr/bin/intltool-update --gettext-package synfig --pot
rm -f missing notexist
srcdir=../../../po /usr/bin/intltool-update -m
if [ -r missing -o -r notexist ]; then \
  exit 1; \
fi
make[2]: Leaving directory '/home/konstantin/sources/synfig/_release/synfig-core/synfig-1.5.1/_build/sub/po'
Making check in test
make[2]: Entering directory '/home/konstantin/sources/synfig/_release/synfig-core/synfig-1.5.1/_build/sub/test'
make  angle benchmark bezier bline bone clock filesystem_path keyframe node pen reference_counter string surface_etl
make[3]: Entering directory '/home/konstantin/sources/synfig/_release/synfig-core/synfig-1.5.1/_build/sub/test'
  CXX      angle.o
  CXX      benchmark.o
  CXX      hermite.o
  CXX      bline.o
../../../test/hermite.cpp:29:10: fatal error: test_base.h: No such file or directory
 #include "test_base.h"
          ^~~~~~~~~~~~~
compilation terminated.
../../../test/angle.cpp:30:10: fatal error: test_base.h: No such file or directory
 #include "test_base.h"
          ^~~~~~~~~~~~~
compilation terminated.
Makefile:789: recipe for target 'hermite.o' failed
make[3]: *** [hermite.o] Error 1
make[3]: *** Waiting for unfinished jobs....
Makefile:789: recipe for target 'angle.o' failed
make[3]: *** [angle.o] Error 1
../../../test/bline.cpp:36:10: fatal error: test_base.h: No such file or directory
 #include "test_base.h"
[...]

PR #3064, opened 2023-05-06
ci: make a full build with distcheck when using Autotools

morevnaproject:
With this change the CI should do a full build + generate release tarballs from current source. This operation includes "make distcheck" command and we will know for sure that it will run smoothly if this operation will be in CI. 

This check will help to avoid issues which are delay release (like this one - https://github.com/synfig/synfig/issues/3063).


morevnaproject:
We have to wait until https://github.com/synfig/synfig/issues/3063 will get fixed.

rodolforg:
@morevnaproject it seems the problem is a regex in the `/usr/local/bin/intltool-update` script.

ice0:
> @morevnaproject it seems the problem is a regex in the `/usr/local/bin/intltool-update` script.

Yes. This is `intltool-update` issue.
https://github.com/Homebrew/linuxbrew-core/issues/18358

It doesn't work with new Perl.

The Debian team made a patch for it (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=826471), but Brew didn't :(

@rodolforg 
Perhaps we should consider moving from intltool. I'm still not sure if `gettext` is better, but it looks like we have no choice here.

P.S.
There are other options here:
1) Skip `intltool-update` on macOS
2) Maybe we can replace `intltool-update` with a different tool.

rodolforg:
It seems intltool isn't updated/maintained since 2015!

https://freedesktop.org/wiki/Software/intltool/

rodolforg:
Maybe version 0.51 works?

rodolforg:
There is a difference between this PR and original code that I don't know if you realized: original code tests the .tar.gz package.
Is it okay?

PR #3066, opened 2023-05-06
fix: update Tool tip when user changes its shortcut at runtime

rodolforg:
Otherwise, the shortcut in tooltip would be updated only when software restarts.


Keyikedalube (dismissed):

rodolforg:
@Keyikedalube thank you for reviewing :+1: 

But now I noticed the typos in the original code I copied and pasted inside the lambda function :/ I'll fix it.

rodolforg:
Done

PR #3067, opened 2023-05-06
fix: toolbox panel wasn't updated on switching tools via shortcuts

rodolforg:
Related to: PR #3012 (313878d4d12a32f24b533d0832a6660b63ee149a)

### DIFF
--- a/synfig-studio/src/gui/docks/dock_toolbox.cpp
+++ b/synfig-studio/src/gui/docks/dock_toolbox.cpp
@@ -146,10 +146,23 @@ void Dock_Toolbox::read_layout_string(const std::string& params) const
 void
 Dock_Toolbox::set_active_state(const synfig::String& statename)
 {
-	changing_state_=true;
+	changing_state_ = true;
 
 	synfigapp::Main::set_state(statename);
-	changing_state_=false;
+
+	try {
+		for (const auto& item : state_button_map) {
+			if (item.first == statename && !item.second->get_active()) {
+				item.second->set_active(true);
+				break;
+			}
+		}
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/docks/dock_toolbox.cpp
```diff
@@ -146,10 +146,23 @@ void Dock_Toolbox::read_layout_string(const std::string& params) const
...
+		for (const auto& item : state_button_map) {
+			if (item.first == statename && !item.second->get_active()) {
+				item.second->set_active(true);
+				break;
+			}
+		}
+	} catch (...) {
+		changing_state_ = false;
+		throw;
+	}
```
Keyikedalube, line 163: This `catch (...)` block is for string "none", right? So the program don't crash...

Keyikedalube (approved):

synfig-studio/src/gui/docks/dock_toolbox.cpp
rodolforg: Well, I just copied from original code before it was deleted XD https://github.com/synfig/synfig/commit/313878d4d12a32f24b533d0832a6660b63ee149a

Keyikedalube: The code is very safe then 😂

Because I didn't exactly get what try block is catching LOL.

rodolforg: I think it's because `set_state()`, that deals with GUI stuff... Oh well, I'll leave it as is XD

PR #3069, opened 2023-05-07
build: fix `make distcheck` build

ice0:
Fixed:
- "files left after uninstall" error for "distuninstallcheck" target

Backported fix from 1.4 branch: https://github.com/synfig/synfig/commit/e94ea8d6447f39547d762db42e4345258a62a9ae

- ERROR: files left in build directory after distclean: ./po/.intltool-merge-cache.lock

It looks like bug in `libtoolize`: https://bugs.launchpad.net/intltool/+bug/1712194
Fixed by rewriting generated `Makefile.in.in` file.

- bline.cpp:36:10: fatal error: test_base.h: No such file or directory

I'm not sure which way is preferable: using `EXTRA_DIST` or adding `test_base.h` to build the target?

### DIFF
--- a/synfig-core/bootstrap.sh
+++ b/synfig-core/bootstrap.sh
@@ -46,7 +46,7 @@ AUTOPOINT='intltoolize --automake --copy' autoreconf --force --install --verbose
 #       is fixed in intltool
 
 echo "patching po/Makefile.in.in..."
-sed 's/itlocaledir = $(prefix)\/$(DATADIRNAME)\/locale/itlocaledir = $(datarootdir)\/locale/' < po/Makefile.in.in > po/Makefile.in.in.tmp
+sed 's/itlocaledir = $(prefix)\/$(DATADIRNAME)\/locale/itlocaledir = $(datarootdir)\/locale/;s/rm -f .intltool-merge-cache/rm -f .intltool-merge-cache .intltool-merge-cache.lock/' < po/Makefile.in.in > po/Makefile.in.in.tmp
 # -- force didn't work under MacOS
 mv -f po/Makefile.in.in.tmp po/Makefile.in.in
 
--- a/synfig-core/test/Makefile.am
+++ b/synfig-core/test/Makefile.am
@@ -47,3 +47,4 @@ string_SOURCES=string.cpp
 
 surface_etl_SOURCES=surface_etl.cpp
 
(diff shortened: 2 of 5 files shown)


morevnaproject:
Thank you! It works now.

Still, I think it would be more correct to have the following content in `synfig-core/test/Makefile.am`:

```
MAINTAINERCLEANFILES=Makefile.in
AM_CXXFLAGS=@CXXFLAGS@ @ETL_CFLAGS@ @SYNFIG_CFLAGS@ -I$(top_builddir) -I$(top_srcdir)/src
AM_LDFLAGS = \
	../src/synfig/libsynfig.la \
	@SYNFIG_LIBS@

check_PROGRAMS=$(TESTS)

TESTS = \
	angle \
	benchmark \
	bezier \
	bline \
	bone \
	clock \
	filesystem_path \
	keyframe \
	node \
	pen \
	reference_counter \
	string \
	surface_etl

angle_SOURCES=angle.cpp test_base.h

benchmark_SOURCES=benchmark.cpp

bezier_SOURCES=hermite.cpp

bone_SOURCES=bone.cpp

bline_SOURCES=bline.cpp test_base.h

clock_SOURCES=clock.cpp

filesystem_path_SOURCES=filesystem_path.cpp test_base.h

keyframe_SOURCES=keyframe.cpp test_base.h

node_SOURCES=node.cpp test_base.h

pen_SOURCES=pen.cpp test_base.h

reference_counter_SOURCES=reference_counter.cpp test_base.h

string_SOURCES=string.cpp test_base.h

surface_etl_SOURCES=surface_etl.cpp test_base.h
```

(this is because EXTRA_DIST is usually  used for non-source files)

rodolforg:
I vote for.... EXTRA_DIST.

According to AutoMake docs:
> Still, sometimes there are files which must be distributed, but which are not covered in the automatic rules. These files should be listed in the EXTRA_DIST variable. You can mention files from subdirectories in EXTRA_DIST.

https://www.gnu.org/software/automake/manual/1.7.6/html_node/Dist.html

So, it is a valid way and we add a single line (for Makefile.am in test folder).
Otherwise, it would be strange to place in every _SOURCES line or in one only.

morevnaproject:
Okay! ^__^

morevnaproject (approved):

PR #3078, opened 2023-05-14
refactor!: rename action "KeyframeToggl" as "KeyframeToggle"

rodolforg:
and remove unused #include

BREAKING CHANGE: synfigapp API changed:
- action "KeyframeToggl" renamed as "KeyframeToggle"

### DIFF
--- a/synfig-studio/po/POTFILES.in
+++ b/synfig-studio/po/POTFILES.in
@@ -366,7 +366,7 @@ src/synfigapp/actions/keyframeset.cpp
 src/synfigapp/actions/keyframeset.h
 src/synfigapp/actions/keyframesetdelta.cpp
 src/synfigapp/actions/keyframesetdelta.h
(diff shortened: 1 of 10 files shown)


mohamedAdhamc (commented):
How about also renaming the action file "keyframetoggl.cpp/.h" it appears to have the same typo ?

rodolforg:
> How about also renaming the action file "keyframetoggl.cpp/.h" it appears to have the same typo ?

@ice0 ?

ice0:
> > How about also renaming the action file "keyframetoggl.cpp/.h" it appears to have the same typo ?
> 
> @ice0 ?

I think yes, it's better to rename it.

rodolforg:
Done

> I think yes, it's better to rename it.

ice0 (approved):

ice0:
Merged. Thank you!

PR #3079, opened 2023-05-14
refactor(Time): change epsilon_() method to constant; remove c_str()

rodolforg:
Time::c_str() :
 - not used
 - not needed
 - wrong (returned value is invalid, due to be from a temp value)

### DIFF
--- a/synfig-core/src/synfig/time.h
+++ b/synfig-core/src/synfig/time.h
@@ -70,7 +70,7 @@ public:
 private:
 	value_type value_;
 
-	static value_type epsilon_() { return static_cast<value_type>(0.0005); }
+	static const value_type epsilon_;
 
 public:
 	Time(): value_() { }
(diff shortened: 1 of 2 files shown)


synfig-core/src/synfig/time.h
```diff
@@ -70,7 +70,7 @@ class Time
 private:
 	value_type value_;
 
-	static value_type epsilon_() { return static_cast<value_type>(0.0005); }
+	static const value_type epsilon_;
```
ice0, line 73: constexpr?

rodolforg: It didn't work for ubuntu 20.
Thwre is a difference for constexpr in static members on c++14 and c++17.
So I changed to classic `const`.

ice0 (approved):

ice0:
Merged. Thank you!

PR #3081, opened 2023-05-16
fix: uninitialized value in ValueNode_DynamicList::ListEntry::amount_…

rodolforg:
…at_time()

fix Coverity Defect ID: 379193


ice0:
I'm thinking maybe it's better to initialize it on creation?
```
bool rising = false;
```

because it is still not initialized here:
```
try	{ prev_iter=find_prev(t); }
catch(...) { return find_next(t)->state?1.0f:0.0f; }
```

rodolforg:
Yes, sure! My mistake.

Done

PR #3085, opened 2023-05-19
refactor: prefer prefix ++ operators for non-primitive types (part 2)

rodolforg:
And prefer the style of the `for` loop as discussed in the forums https://forums.synfig.org/t/trying-to-define-our-code-style/13097.
```c++
for (int a = 0; a < 30; ++a) {
}
```
(spaces surrounding operators, spaces before `(` and after and `;`, and bracket layout)

### DIFF
--- a/synfig-core/src/synfig/filecontainerzip.cpp
+++ b/synfig-core/src/synfig/filecontainerzip.cpp
@@ -314,7 +314,7 @@ FileContainerZip::HistoryRecord FileContainerZip::decode_history(const String &c
 				{
 					String s;
 					xmlpp::Element::NodeList list = (*i)->get_children();
-					for(xmlpp::Element::NodeList::iterator j = list.begin(); j != list.end(); ++j)
+					for (xmlpp::Element::NodeList::iterator j = list.begin(); j != list.end(); ++j)
 						if (dynamic_cast<xmlpp::TextNode*>(*j))
 							s += dynamic_cast<xmlpp::TextNode*>(*j)->get_content();
 					history_record.prev_storage_size = strtoll(s.c_str(), nullptr, 10);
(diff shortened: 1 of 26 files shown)


rodolforg:
Maybe it should be perf, not refactor... lol

Keyikedalube:
Do you want help creating the coding guideline doc page on synfig dev site? I can do that in the evening today if you want :)

synfig-core/src/synfig/filecontainerzip.cpp
```diff
@@ -343,8 +343,7 @@ void FileContainerZip::read_history(std::list<HistoryRecord> &list, FILE *f, fil
 	bool found = false;
 	HistoryRecord history_record;
 
-	for(int i = read_size - sizeof(EndOfCentralDirectory); i >= 0; i--)
-	{
+	for (int i = read_size - sizeof(EndOfCentralDirectory); i >= 0; i--) {
```
Keyikedalube, line 346: Postfix seems fine to me logically unless the rh value is needed immediately.

Below you changed all to prefixes. Here I guess you missed the `i++` :P

Keyikedalube (approved):

synfig-core/src/synfig/filecontainerzip.cpp
rodolforg: Some "good practices" guides recommend always prefer pre-increment operator in `for` loops. The reason is to not make mistakes and not forget to use it when is not a primitive type.

Keyikedalube: I saw some Internet post about pre increment for user data types saving memory. Never realized post might make non-primitive inefficient dude to temp obj creation.

Your PR was kinda a good CS revision for me LOL
Glad I reviewed and learned something along the way 😅

PR #3090, opened 2023-05-23
fix: find the second closest guide no matter the guide list order

rodolforg:
Before this PR, it wouldn't get any second closest guide if it appears after the best match.


mohamedAdhamc:
Thanks for noticing this case went completely over my head when implementing.

I tested it in both master vs this branch
the test: I just tested snapping to intersection of two lines. One time from one of them the other time from the other. It was noticeable that for one of that there was no snap happening. This pr fixes that successfully.

ice0 (approved):

PR #3091, opened 2023-05-27
refactor: brush size increase/decrease buttons now use theme icons

rodolforg:
As we don't have any specific icon for these actions, we use current +/- signs as fallback if icons do not exist in current icon theme

Requested-by: pablogil in https://forums.synfig.org/t/help-code-changes-in-order-to-add-an-icon/13585/19


ice0 (approved):

ice0:
Merged. Thank you!

PR #3107, opened 2023-06-17
fix: avoid crash if icon theme does not provide type_bool_icon file

rodolforg:
or anything like that.

To reproduce:
- launch Synfig with a nonexistent icon theme folder as theme

It can happen if:
- user by manually edit the settings file to a nonexistent icon theme;or
- user set an existent icon theme, but deletes the icon theme folder without change the icon theme before


ice0 (approved):

ice0:
Merged. Thank you!

PR #3113, opened 2023-06-20
refactor: color blending function

BharatSahlot:
Results in a small speedup of 100-200ms on most example files in 1080p. Sometimes the speedup is more visible like for eye example, the speedup was of 3secs.

### DIFF
--- a/synfig-core/src/synfig/color/colorblendingfunctions.h
+++ b/synfig-core/src/synfig/color/colorblendingfunctions.h
@@ -37,87 +37,85 @@
 
 namespace synfig {
 
-typedef Color (*blendfunc)(Color &,Color &,float);
+typedef Color (*blendfunc)(Color, Color, float);
 
 template <class C>
-C blendfunc_COMPOSITE(C &src,C &dest,float amount)
+C blendfunc_fg(C fg, C bg, float amount)
 {
-	//c_dest'=c_src+(1.0-a_src)*c_dest
-	//a_dest'=a_src+(1.0-a_src)*a_dest
-
-	float a_src=src.get_a()*amount;
-	float a_dest=dest.get_a();
-	const float one(C::ceil); 
-
-	// if a_arc==0.0
-	//if(fabsf(a_src)<COLOR_EPSILON) return dest;
-
-	// Scale the source and destination by their alpha values
-	src*=a_src;
-	dest*=a_dest;
-
-	dest=src + dest*(one-a_src);
-
-	a_dest=a_src + a_dest*(one-a_src);
-
-	// if a_dest!=0.0
-	if(fabsf(a_dest)>COLOR_EPSILON)
-	{
-		dest/=a_dest;
-		dest.set_a(a_dest);
-	}
-	else
-	{
-		dest=C::alpha();
-	}
-	assert(dest.is_valid());
-	return dest;
+	return fg;
 }
 
 template <class C>
-C blendfunc_STRAIGHT(C &src,C &bg,float amount)
+C blendfunc_bg(C fg, C bg, float amount)
 {
-	//a_out'=(a_src-a_bg)*amount+a_bg
-	//c_out'=(((c_src*a_src)-(c_bg*a_bg))*amount+(c_bg*a_bg))/a_out'
+	return bg;
+}
 
-	// ie: if(amount==1.0)
-	//if(fabsf(amount-1.0f)<COLOR_EPSILON)return src;
+template <class C>
+C blendfunc_COMPOSITE(C fg, C bg, float amount)
+{
+	const float a_fg = fg.get_a() * amount;
+	const float a_bg = bg.get_a();
 
-	C out;
+	const float one(C::ceil); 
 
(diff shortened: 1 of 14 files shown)


BharatSahlot:
@rodolforg @ice0 I am not sure why the Autotools run is failing, it looks like its not due to my changes.

synfig-core/src/synfig/color/colorblendingfunctions.h
```diff
@@ -361,7 +359,7 @@ C blendfunc_OVERLAY(C &a,C &b,float amount)
 
 
 template <class C>
-C blendfunc_HARD_LIGHT(C &a,C &b,float amount)
```
rodolforg, line 364: Is it faster passing by value instead of by reference (it should be by const reference, actually)?

synfig-core/src/synfig/color/colorblendingfunctions.h
```diff
@@ -37,87 +37,85 @@
...
-	dest=src + dest*(one-a_src);
+	const float one(C::ceil); 
 
-	a_dest=a_src + a_dest*(one-a_src);
+	const float a = a_fg + a_bg * (one - a_fg);
 
-	// if a_dest!=0.0
-	if(fabsf(a_dest)>COLOR_EPSILON)
+	if(fabsf(a) > COLOR_EPSILON)
 	{
```
rodolforg, line 65: prefer this style:

```c++
if (fabsf(a) > COLOR_EPSILON) {
```

(space before parenthesis and curly bracket in the same line)

synfig-core/src/synfig/color/colorblendingfunctions.h
BharatSahlot: Yes its faster to pass by value in this case. It isnt const ref because the colors are actually changed but the changed colors are not used where this function is called, so ref doesn't serve a purpose.
I tried switch instead of vtable but it was slower for some reason.

synfig-core/src/synfig/color/colorblendingfunctions.h
BharatSahlot: Okay, will keep in mind and make the change.

synfig-core/src/synfig/color/colorblendingfunctions.h
rodolforg: Oh, indeed! Great catch!

synfig-core/src/synfig/color/colorblendingfunctions.h
rodolforg: Any updates?

BharatSahlot: Sorry, I was working on the opengl thing and forgot about this change. Will make it today and push.

rodolforg (approved):

ice0:
@BharatSahlot 

Can you add test measurements (before/after) with test cases so this can be reproduced?

For example: https://github.com/synfig/synfig/pull/2831

BharatSahlot:
> @BharatSahlot
> 
> Can you add test measurements (before/after) with test files so this can be reproduced?
> 
> For example: #2831

Sure. I don't have window and mac to test, I can only test on linux.

ice0:
> Sure, but I don't have window and mac to test, I can only test on linux.

No problems I will do tests on Windows/macOS, after you will done with Linux.

BharatSahlot:
The performance varies between files, in the example files I saw increase anywhere between 0.7% to 26%. Below is the time taken by some example files on linux release build. `A` is this branch and `B` is the master branch. I used [./test_null.sh](https://github.com/BharatSahlot/synfig_test/blob/main/test_null.sh) for testing.

```
Rendering files in ./files/examples. Width: 1920, Height: 1080
===================================
Testing ./files/examples/backdrop.sif
Time taken by A: 964.966
Time taken by B: 1192.06
Speed increase: 23.533
===================================
Testing ./files/examples/eye.sif
Time taken by A: 25778.8
Time taken by B: 28997.7
Speed increase: 12.486
===================================
Testing ./files/examples/pirates.sif
Time taken by A: 882.835
Time taken by B: 1033.49
Speed increase: 17.064
===================================
Testing ./files/examples/sparkle.sif
Time taken by A: 1133.12
Time taken by B: 1251.25
Speed increase: 10.425
===================================
Testing ./files/examples/z_depth_test.sif
Time taken by A: 412.215
Time taken by B: 494.878
Speed increase: 20.053
```

Time taken by all files can be found here: https://pastebin.com/besNtVJF

rodolforg:
@BharatSahlot As I can see in the test_null.sh you are using different renderers (gl vs. software).
Shouldn't you compare by using the same renderer with/without this PR changes?

BharatSahlot:
> @BharatSahlot As I can see in the test_null.sh you are using different renderers (gl vs. software). Shouldn't you compare by using the same renderer with/without this PR changes?

Hi @rodolforg , when I test this branch I use the same renderer(default/software). That change was because I was testing my opengl renderer, I was just lazy and wrote it there directly.

PR #3114, opened 2023-06-24
feat: select tool (rotate functionality)

mohamedAdhamc:
Builds on #3036.

I have made this branch based on that one. Theoretically that one would be edited a couple more time then merged. At which point I would rebase this branch with master and everything should be good.

Also Since I made this branch based on the previous one it had a lot of "un-necessary" commits in its history. I squashed most of them but the remaining ones were a bit of a pain as they had other commits in between them. Anyways this all shouldn't matter much as this pr would be squashed as a whole on merge.

### DIFF
--- a/synfig-core/src/synfig/context.h
+++ b/synfig-core/src/synfig/context.h
@@ -58,7 +58,7 @@ class Time;
 class Rect;
 
 /*!	\class IndependentContext
-**	\brief IndependentContext is a class to warp the iterator for a double queue of layers
+**	\brief IndependentContext is a class to wrap the iterator for a double queue of layers
 * (that is the CanvasBase).
 **	\see Layer, Canvas, CanvasBase, Context */
 class IndependentContext: public CanvasBase::const_iterator
--- a/synfig-core/src/synfig/layer.cpp
+++ b/synfig-core/src/synfig/layer.cpp
@@ -610,7 +610,7 @@ Layer::set_time(IndependentContext context, Time time)const
 {
 	Layer::ParamList params;
 	Layer::DynamicParamList::const_iterator iter;
-	// For each parameter of the layer sets the time by the operator()(time)
+	// For each parameter of the layer sets the value by the operator()(time)
 	for(iter=dynamic_param_list().begin();iter!=dynamic_param_list().end();iter++)
 		params[iter->first]=(*iter->second)(time);
 	// Sets the modified parameter list to the current context layer
--- a/synfig-core/src/synfig/os.cpp
+++ b/synfig-core/src/synfig/os.cpp
@@ -36,6 +36,7 @@
 #include "os.h"
 
 #include <cstdarg>
+#include <cstdlib>
 
 #include <synfig/filesystem.h>
 #include <synfig/general.h>
@@ -76,6 +77,36 @@ class RunPipeUnix;
 #endif
 
 /* === P R O C E D U R E S ================================================= */
+
+#ifdef _WIN32
+
+std::string
+utf16_to_utf8(const wchar_t* src, size_t src_length = 0)
+{
+	std::string utf8;
+
+	if (!src)
+		return utf8;
+
+	if (src_length == 0)
+		src_length = wcslen(src);
+
+	const int length = WideCharToMultiByte(CP_UTF8, 0, src, src_length, 0, 0, nullptr, nullptr);
+	utf8.resize(length + 1);
+
+	WideCharToMultiByte(CP_UTF8, 0, src, src_length, &utf8[0], length, nullptr, nullptr);
+	utf8[length] = 0;
+
+	return utf8;
+}
+
+std::string
+utf16_to_utf8(const std::wstring& src)
+{
+	return utf16_to_utf8(src.c_str(), src.size());
+}
+
+#endif
 
 /* === M E T H O D S ======================================================= */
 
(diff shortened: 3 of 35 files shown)


mohamedAdhamc:
This is the state of the rotate functionality as of now.

[Screencast from 06-29-2023 08:19:35 PM.webm](https://github.com/synfig/synfig/assets/100296264/0eda0749-6c9c-4af0-82b8-725d0234757e)

A couple notes:
- The rotate cursor should be working. I just have an issue on my machine with some of the gdk cursors not being available.
- As was visible the goal is that this tool would have all these functionalities that the user would pick from. Of course there will 
  also be icons in there for the different functionalities.
- I believe for the basic functionality this should now be done (Although code wise it still needs a bit more time refactoring/enhancing). But one functionality that I believe would also be really helpful but was not previously implemented.
Is to have a "center handle" that the user can move through gui so that he has more control over the center of rotation. However I'm not yet sure if that should be left instead to another PR after this is merged. Lightweight PRs can be merged faster and reviewed quicker.

mohamedAdhamc:
Things here are mostly done (some specific parts are broken as I need to rebase this branch onto the main select tool branch but that is still changing). Whenever the first PR is approved/merged I think it may take a day or max two and this should be ready for review.

rodolforg:
When I proposed the Select tool, I meant to implement the regular Selection tool other "graphic" software do, like Inkscape, or even LibreOffice Writer/Impress, MS Office Word/PowerPoint.
You click an object and can drag it around the paper/screen.

As a plus, like those software do, it could show handles to scale or rotate.
The intention was to be more intuitive by being similar to other well-known software.

As I see in the video, you are reimplementing our Transform tool with all of its feature into a pseudo-new tool that does the same, except you don't need to select all the handles. Am I right?

What do you think, @ice0 @morevnaproject ?

mohamedAdhamc:
> When I proposed the Select tool, I meant to implement the regular Selection tool other "graphic" software do, like Inkscape, or even LibreOffice Writer/Impress, MS Office Word/PowerPoint. You click an object and can drag it around the paper/screen.
>

And this is what was done in the (move functionality) correct ?
 
> As a plus, like those software do, it could show handles to scale or rotate. The intention was to be more intuitive by being similar to other well-known software.
> 
> As I see in the video, you are reimplementing our Transform tool with all of its feature into a pseudo-new tool that does the same, except you don't need to select all the handles. Am I right?

Well, the way I understood it was that the need for a "select" tool was born out of the fact that in Synfig all the tools are vertex/handle editing tools where you don't deal with the layers as a whole but deal with their vertices and handles. 

For example what we (I hope) agree on, the move functionality. Using the other vertex-editing tools (transform) the layers can be moved but only through "handles". Instead using the select tool they are just dragged.

Similarly I thought about adding these other functionalities in a similar way. For example rotatation without having to select or deal with handles kind of like [Gimp](https://www.youtube.com/watch?v=ewVUxvTfK9Q) (starting from 1:11). I was also thinking of adding a handle later on for the center of rotation like in the video that can be dragged so that the COR is user adjustable. As of now it's mostly just always the average of highest and lowest x and y handle coordinates of the layer.

Also mirror is in a similar fashion.

So if I understand correctly what you are proposing (for the bonus rotate/scale/mirror) is instead of having them all be done in this way without handles at all. We would just add handles for rotation + scale + mirror ? Am I following along correctly or missing something :) ?

Also if this is what you mean aren't we just going back to basically the same transform (vertex editing space). I mean now we are also applying transformations (scale, rotate, etc..) through handles.

mohamedAdhamc:
progress update: 

So it seems that it is better to go forward with Rodolfo's suggestion for the extra functionalities. I'm currently working on adding handles for rotation at the corners (but a little further), and also (maybe) after that to make a handle for the center of rotation so that it is movable.

Then after this probably in another PR I can add handles for scaling and then the select tool would be basically done. With this new approach I don't really think mirror is needed much.

mohamedAdhamc:
@rodolforg Here is a prototype I have reached 

[Screencast from 07-16-2023 03:55:00 PM.webm](https://github.com/synfig/synfig/assets/100296264/dd274ac0-d2fb-4fbc-90f0-2bdb330c3f21)

I want to know your opinion on a few things
1- Regarding the handles. What do you think is better ?
     -  To use rotate icons kind of like inkscape. 
image
     - Or if it would just be better to "draw" the handle diffrently ?  

2- Since the bounding box doesn't rotate the handles are always at the corners of it which kind of looks weird at times. Do you think maybe the handles can start at the corners of the bounding box but then remain where they are moved or should I just leave it as it is now ?

p.s.
What is left now mostly is:
- icons/diffrent shape for the rotation handles
- moveable center of rotation handle/drawing
- code cleanups/refactor (I will leave this part till after the first select tool pr is merged as I need to rebase this on it first anyways).

rodolforg:
> - Or if it would just be better to "draw" the handle diffrently ?

We could draw differently as Inkscape and many other software do. For now, we could use our "blue" handle, used typically for rotation handles.



> 2- Since the bounding box doesn't rotate the handles are always at the corners of it which kind of looks weird at times. Do you think maybe the handles can start at the corners of the bounding box but then remain where they are moved or should I just leave it as it is now

Always in the bounding box corners. Remember to add constrained rotation (every 15 degrees) if `Shift` is pressed.

---

PS: These handles should be shown as a plus in the select tool.

mohamedAdhamc:
> We could draw differently as Inkscape and many other software do. For now, we could use our "blue" handle, used typically for rotation handles.

Ok, so blue handles it is for now : ).

> Always in the bounding box corners. Remember to add constrained rotation (every 15 degrees) if `Shift` is pressed.

 Yes I think it's better this way too. And noted I will also work on adding constrained rotation.

> PS: These handles should be shown as a plus in the select tool.

You mean like having options to either show them or not in the tool ? Actually I was thinking of having toggle buttons much like the ones in the layer creation tools. We could use the icons provided by pgilfernandez and they could be used to control which handles are visible rotation, scaling, (maybe also rotation center handle), etc... 

Or did you mean something else ?

rodolforg:
> Or did you mean something else ?

I think it's better to always be shown for the selected layer when on Select Tool.

PR #3118, opened 2023-07-05
feat: select tool (mirrror functionality)

mohamedAdhamc:
Added basic UI for now, but I plan to have the functionality done and working in the next 2/3 days if everything goes well. However this and the rotate functionality can't really truly be ready for review until the main select tool pr is finalized meaning approved/merged. Because all the functionalities depend on the main select tool PR.


mohamedAdhamc:
Looks like we no longer need rotate so I will close this for now at least. If we in the future need it for some reason then we can reopen it.

mohamedAdhamc:
Looks like we no longer need rotate so I will close this for now at least. If we in the future need it for some reason then we can reopen it.

PR #3122, opened 2023-07-13
refactor: make argument name consistent in method declaration and definition

mohamedAdhamc:
Nothing major just something I noticed while going through this area.


rodolforg (approved):

rodolforg:
Maybe `style` instead of `refactor`?

mohamedAdhamc:
> Maybe `style` instead of `refactor`?

Not sure also but I think since a "name" was changed after all that makes it more a `refactor` :thinking: .

ice0:
Merged. Thank you!

PR #3136, opened 2023-07-21
feat: add constrained rotation in rotate tool

mohamedAdhamc:
@rodolforg suggested I add constrained rotation in the select tool and I realized the rotate tool is missing it (although transform tool already has it) so I thought it wouldn't hurt to add it to the rotate tool as well.

### DIFF
--- a/synfig-studio/src/gui/states/state_rotate.cpp
+++ b/synfig-studio/src/gui/states/state_rotate.cpp
@@ -43,6 +43,7 @@
 #include <gui/duck.h>
 #include <gui/localization.h>
 #include <gui/workarea.h>
+#include <gui/event_keyboard.h>
 
 #include <synfig/angle.h>
 #include <synfig/general.h>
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/states/state_rotate.cpp
```diff
@@ -86,16 +87,19 @@ class DuckDrag_Rotate : public DuckDrag_Base
 
 	bool bad_drag;
 	bool move_only;
+	bool rotate;
```
mohamedAdhamc, line 90: Seems I added this by mistake will remove

mohamedAdhamc:
Hey @rodolforg , or @ice0. This should be quite simple, I remember using pretty much the same implementation for constrain used in the transform tool. Could you check it out ? : )

mohamedAdhamc:
on a second thought I think it would be quite helpful to allow the user to change the value constrained by from the gui (in another pr ) 
here
Screenshot from 2023-12-12 01-52-09
for both the select tool and rotate too (after/if this gets merged).

PR #3140, opened 2023-07-22
fix: preferences dialog should list only themes with index.theme file

### LINKED ISSUE #3115: Color theme dropdown may no list all folders
**Describe the bug**
It's not a bug but I think it's something to have a look at... and if you let me say it, not a very difficult thing:
As you see in the picture that follows, the Color theme dropdown lists all folders in `/SynfigStudio.app/share/synfig/icons/`, as everything is listed even hidden OS folders are included.

Apart from preventing this folders to be listed, I would suggest that a simple filter is applied in order to list just "icon theme" folders and not any other one that could be placed there. As an example, as "icon themes" now should include a file called _index.theme_ inside, the filter could check if that file exists and if true check the parent folder as a real "icon theme".

**To Reproduce**
1. Go to 'Preferences'
2. Click on 'Color theme' dropdown
3. have a look at the list

**Expected behavior**
Only real "icon themes" should be listed, not all folders

**Screenshots**
theme_list

**System information:**
 - OS: I guess all OS are affected
 - OS version: all
 - Synfig version: 1.5.1

**Additional comments**
Additionally, It would be nice that in the dropdown, instead of listing folder names, it lists the full "name" variable placed in the _index.theme_ file:
```

[Icon Theme]

Name=Classic
Comment=Synfig classic icon theme
Hidden=true
Directories=128x128
Example=synfig_icon

[128x128]
Size=128
MinSize=12
MaxSize=128
Context=Actions
Type=Scalable
```

rodolforg:
Otherwise it is an invalid theme folder.

fix #3115

### DIFF
--- a/synfig-studio/src/gui/dialogs/dialog_setup.cpp
+++ b/synfig-studio/src/gui/dialogs/dialog_setup.cpp
@@ -38,6 +38,9 @@
 
 #include <gui/dialogs/dialog_setup.h>
 
+#include <glibmm/fileutils.h>
+#include <glibmm/keyfile.h>
+
 #include <gtkmm/accelmap.h>
 #include <gtkmm/filechooserdialog.h>
 
@@ -50,6 +53,7 @@
 #include <gui/autorecover.h>
 #include <synfig/threadpool.h>
 
+#include <synfig/general.h>
 #include <synfig/rendering/renderer.h>
 
 #include <synfigapp/main.h>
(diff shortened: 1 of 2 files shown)


synfig-studio/src/gui/dialogs/dialog_setup.cpp
```diff
@@ -784,13 +788,43 @@ Dialog_Setup::create_interface_page(PageInfo pi)
...
+		std::string full_filename = ResourceHelper::get_themes_path() + "/" + dir + "/index.theme";
+
+		if (!FileSystemNative::instance()->is_file(full_filename))
+			continue;
+
+		try {
+			Glib::KeyFile theme_file;
+			theme_file.load_from_file(full_filename);
+			if (theme_file.has_group("Icon Theme")) {
+				if (theme_file.has_key("Icon Theme", "Hidden")) {
```
ice0, line 803: Why not just use something like?
```
if (theme_file.get_boolean("Icon Theme", "Hidden")) {
	continue;
}
```

rodolforg: Because [the documentation says it can raise exception](https://developer-old.gnome.org/glibmm/stable/classGlib_1_1KeyFile.html#a0d383ad579647505ff3376e93591aaa7).
https://gitlab.gnome.org/GNOME/glibmm/-/blob/glibmm-2-58/glib/src/keyfile.ccg#L173-L183

ice0: Lol. Got it.
Should I squash it? Or rebase it please.

rodolforg: It's okay to squash it :)

ice0 (approved):

ice0:
Merged. Thank you!

PR #3149, opened 2023-07-27
perf: optimize `set_time` method

mohamedAdhamc:
**Brief:** 

This is one of the parts in my project that was included but in a bit of a different manner. So basically what was agreed on was that since this "optimization" could possibly not have much improvement, then it would be done if it is found to have improvement. 

So basically right now I'm thinking of which ways would be best to test the performance here to see the effect of this optimization. If you have any ideas/suggestions it would be great if you can share them :).

**Feature brief:**

So basically the whole idea is that the 'set_time' method does some "redundant" calculations. Basically whenever the time we are at changes (or change a parameter) 'set_time' is called on each layer and then each layer recalulates all its (animated) values at the current time.

So for example if in our canvas we have 10 layers and in the animation each of the ten layers has origin, color, and radius parameters that are animated then when the animation is playing at each frame the current of value of these parameters is calculated according to the rule set for them (from the valuenodes () operator). Now with the current caching already in place (green stuff) that only happens the first time it is played. If you play for a second time without changing anything no problem is present and nothing is recalculated.

The problem comes when any parameter of any of the layers changes. For example if we only change the color of one of the layers then instead of (what this optimization does) recalculating only these new color values. Everything is recalculated.

However of course now the important question which comes up that would sort of only be answered by doing tests/benchmarks is that if retrieving the cached value from the map would be faster than simply recalculating the value.

If we talk theory then for the unordered map we are using 
>Average case: constant.
>Worst case: linear in container size

And for recalculating the value. We could say (at least as a general statement I haven't looked too deeply into each valuenodes oprator () overload) that it is always constant in time. However what we kind of want to figure out is how much this time is, and if it would be faster/slower than using a cache.

### DIFF
--- a/synfig-core/src/synfig/layer.cpp
+++ b/synfig-core/src/synfig/layer.cpp
@@ -610,8 +610,31 @@ Layer::set_time(IndependentContext context, Time time)
 	Layer::ParamList params;
 	Layer::DynamicParamList::const_iterator iter;
 	// For each parameter of the layer sets the value by the operator()(time)
-	for (iter = dynamic_param_list().begin(); iter != dynamic_param_list().end(); ++iter)
-		params[iter->first]=(*iter->second)(time);
+	for (iter = dynamic_param_list().begin(); iter != dynamic_param_list().end(); ++iter){
+		bool paramInMap = parameter_cache.find(iter->first) != parameter_cache.end();
+
+		//if parameter has changed remove it from the parameter list
+		if(paramInMap && iter->second->valueNodeChanged){
+			parameter_cache.erase(iter->first);
+		}
+
+		if (iter->second->valueNodeChanged)
+			iter->second->valueNodeChanged = false;
+
+		if(iter->first != "bline"){
+			//if this time is alread cached
+			if(paramInMap && parameter_cache[iter->first].find(time.get_string()) !=
+							  (parameter_cache[iter->first]).end()){
+				params[iter->first]=(parameter_cache[iter->first])[time.get_string()];
+			} else {
+				(parameter_cache[iter->first])[time.get_string()] = (*iter->second)(time);
+				params[iter->first]=(parameter_cache[iter->first])[time.get_string()];
+			}
+		} else{
+			params[iter->first]=(*iter->second)(time);
+		}
+	}
+
 	// Sets the modified parameter list to the current context layer
 	const_cast<Layer*>(this)->set_param_list(params);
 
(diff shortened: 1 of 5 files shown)


mohamedAdhamc:
So here is a little test I made. It's nothing "formal"  but I think it's a bit of a good indication that this maybe a good optimization (still not sure xd).

The code to test it (just using std::chrono to measure the time for set_time):
```
std::chrono::duration<double, std::milli> durationRender;

void
Layer::set_time(IndependentContext context, Time time)
{
	//clear time when we first start the rendering
	if(time == Time::zero()){
		std::cout<<"restarting duration render"<<std::endl;
		std::cout<<"previous duration is : "<<durationRender.count()<<std::endl;
		durationRender = std::chrono::duration<double, std::milli>::zero();
		std::cout<<"----------------/n--------------/n"<<std::endl;
	}

	auto t1 = std::chrono::high_resolution_clock::now();
	Layer::ParamList params;
	Layer::DynamicParamList::const_iterator iter;
	// For each parameter of the layer sets the value by the operator()(time)
	for (iter = dynamic_param_list().begin(); iter != dynamic_param_list().end(); ++iter){
		bool paramInMap = parameter_cache.find(iter->first) != parameter_cache.end();

		//if parameter has changed remove it from the parameter list
		if(paramInMap && iter->second->valueNodeChanged){
			parameter_cache.erase(iter->first);
		}

		if (iter->second->valueNodeChanged)
			iter->second->valueNodeChanged = false;

		if(iter->first != "bline"){
			//if this time is alread cached
			if(paramInMap && parameter_cache[iter->first].find(time.get_string()) !=
							  (parameter_cache[iter->first]).end()){
				params[iter->first]=(parameter_cache[iter->first])[time.get_string()];
				std::cout<<iter->first<<"cach"<<std::endl;
			} else {
				(parameter_cache[iter->first])[time.get_string()] = (*iter->second)(time);
				params[iter->first]=(parameter_cache[iter->first])[time.get_string()];
				std::cout<<iter->first<<"calc"<<std::endl;
			}
		} else{
			params[iter->first]=(*iter->second)(time);
			std::cout<<iter->first<<"calc"<<std::endl;
		}
	}

	auto t2 = std::chrono::high_resolution_clock::now();
	durationRender += std::chrono::duration<double, std::milli>(t2-t1);

	// Sets the modified parameter list to the current context layer
	const_cast<Layer*>(this)->set_param_list(params);

	set_time_mark(time);

	set_time_vfunc(context, time);
}
```

the file tested: [testFourMovingCirclesCached.zip](https://github.com/synfig/synfig/files/12233628/testFourMovingCirclesCached.zip)

Tests:

test1
- no caching : 197.153
- with caching : 151.507

second time 
- no caching : 197.05
- with caching : 147.774

third time (doubled number of layers)
- no caching : 298.789
- caching : 268.669

PR #3153, opened 2023-08-03
feat: added the ability to edit position of pivot point for guides

### LINKED ISSUE #2593: Feature request: Allow rotation of guides in canvas
Right now guides in synfig can either be horizontal or vertical. 
If guides could be rotated in  any angle (as in inkscape), it could ease the process of creation of shapes and animation as well.
Also it would allow to create perspective lines ( which is quite helpful in creating good artwork)
PS: Please also provide a way to place guides at precise coordinates, kind of like this
image

### LINKED ISSUE #3083: Add the ability to edit the pivot for guides from the guide edit dialog.
Related: check #2777.

As is stated in the title, the goal here is to be able to set the coordinates for the pivot point (x, y) from the guide editor dialog. A guide is composed of a pivot and an angle. This should be simple to implement just get the new pivot point from the dialog and modify the guides pivot.

Afterdie:
closes #3083
closes #2593

### DIFF
--- a/.github/workflows/conventional-commits-style.yml
+++ b/.github/workflows/conventional-commits-style.yml
@@ -16,14 +16,14 @@ jobs:
     name: Conventional commits style check
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Install conventional commits linter
         run: npm install --save-dev @commitlint/config-conventional @commitlint/cli
 
       - name: Configure commitlint
         run: |
-          echo "module.exports = {extends: ['@commitlint/config-conventional']}" > commitlint.config.js
+          echo "module.exports = {extends: ['@commitlint/config-conventional'], rules: {'subject-case': [0, 'never']}}" > commitlint.config.js
 
       - name: Lint current pull request title
         run: jq --raw-output ".pull_request.title" "$GITHUB_EVENT_PATH" | npx commitlint --verbose
--- a/.github/workflows/synfig-ci.yml
+++ b/.github/workflows/synfig-ci.yml
@@ -47,20 +47,16 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v3
-
-      - name: Prepare ccache timestamp
-        id: ccache_timestamp
-        run: echo "::set-output name=timestamp::`date "+%Y%m%d-%H%M%S"`"
+      - uses: actions/checkout@v4
 
       - name: Download ccache archive
-        id: ccache-archive
-        uses: actions/cache@v3
+        id: ccache-archive-restore
+        uses: actions/cache/restore@v4
         with:
           path: .ccache
-          key: ${{ matrix.os }}-ccache-${{ matrix.toolchain }}-${{ steps.ccache_timestamp.outputs.timestamp }}
+          key: synfig-ccache-${{ matrix.os }}-${{ matrix.toolchain }}
           restore-keys:
-            ${{ matrix.os }}-ccache-${{ matrix.toolchain }}-
+            synfig-ccache-${{ matrix.os }}-${{ matrix.toolchain }}
 
       - name: Install dependencies (Brew)
         if: runner.os == 'macOS'
@@ -105,3 +101,12 @@ jobs:
 
       - name: ccache statistics
         run: ccache --show-stats
+
+      - name: Save cache (only for master branch)
+        id: ccache-archive-save
+        if: github.ref_name == 'master'
+        uses: actions/cache/save@v4
+        with:
+          path: .ccache
+          key: synfig-ccache-${{ matrix.os }}-${{ matrix.toolchain }}
+
--- a/.github/workflows/synfig-stable.yml
+++ b/.github/workflows/synfig-stable.yml
@@ -34,20 +34,16 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v3
-
-      - name: Prepare ccache timestamp
-        id: ccache_timestamp
-        run: echo "::set-output name=timestamp::`date "+%Y%m%d-%H%M%S"`"
+      - uses: actions/checkout@v4
 
       - name: Download ccache archive
-        id: ccache-archive
-        uses: actions/cache@v3
+        id: ccache-archive-restore
+        uses: actions/cache/restore@v4
         with:
           path: .ccache
-          key: ${{ matrix.os }}-ccache-${{ matrix.toolchain }}-${{ steps.ccache_timestamp.outputs.timestamp }}
+          key: synfig-ccache-${{ matrix.os }}-${{ matrix.toolchain }}
           restore-keys:
-            ${{ matrix.os }}-ccache-${{ matrix.toolchain }}-
+            synfig-ccache-${{ matrix.os }}-${{ matrix.toolchain }}
 
       - name: Install dependencies (Brew)
         if: runner.os == 'macOS'
@@ -78,3 +74,12 @@ jobs:
 
       - name: ccache statistics
         run: ccache --show-stats
+
+      - name: Save cache (only for master branch)
+        id: ccache-archive-save
+        if: github.ref_name == 'master'
+        uses: actions/cache/save@v4
+        with:
+          path: .ccache
+          key: synfig-ccache-${{ matrix.os }}-${{ matrix.toolchain }}
+
--- a/.github/workflows/synfig-tests.yml
+++ b/.github/workflows/synfig-tests.yml
@@ -31,7 +31,7 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: "Synfig Studio (Check appdata.xml)"
         run: |
--- a/1-setup-linux-native.sh
+++ b/1-setup-linux-native.sh
@@ -62,7 +62,7 @@ fi
 # 2 - Install the required packages
 echo "Checking dependencies..."
 
-if ([ "$ID_LIKE" == "fedora" ] && [ VERSION_ID > 22 ]); then
+if ([ "$ID_LIKE" == "fedora" ] && [ "$VERSION_ID" > 22 ]); then
     PKG_LIST="git \
             intltool \
             libpng-devel \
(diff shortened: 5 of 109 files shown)


synfig-studio/src/gui/dialogs/dialog_guide.cpp
```diff
@@ -96,8 +98,35 @@ Dialog_Guide::Dialog_Guide(Gtk::Window& parent, etl::handle<synfig::Canvas> canv
...
+	posGrid->get_style_context()->add_class("dialog-secondary-content");
+	posGrid->set_row_spacing(6);
+	posGrid->set_column_spacing(8);
+	
+
+	Gtk::Label* xPosLabel = manage(new Gtk::Label(_("_X:"), true));
+	Gtk::Label* yPosLabel = manage(new Gtk::Label(_("_Y:"), true));
+	x_widget=manage(new Gtk::SpinButton(x_adjustment,15,2));
+	x_widget->show();
+	y_widget=manage(new Gtk::SpinButton(y_adjustment,15,2));
```
rodolforg, line 118: Please use spaces surrounding the equal sign, like you did for `Gtk::Label` vars.

synfig-studio/src/gui/dialogs/dialog_guide.cpp
```diff
@@ -96,8 +98,35 @@ Dialog_Guide::Dialog_Guide(Gtk::Window& parent, etl::handle<synfig::Canvas> canv
...
+	y_widget->show();
+	
+	posGrid->attach(*xPosLabel, 0, 0, 1, 1);
+	posGrid->attach(*x_widget, 1,0,1,1);
```
rodolforg, line 122: space after comma

synfig-studio/src/gui/dialogs/dialog_guide.cpp
```diff
@@ -137,6 +166,12 @@ Dialog_Guide::on_ok_or_apply_pressed(bool ok)
 		curr_guide->angle = synfig::Angle::rad(angle_widget->get_value());
 	}
 
+	if(curr_guide->point[0] != (x_widget->get_value())/60){
+		curr_guide->point[0] =  (x_widget->get_value())/60;
+	}
+	if(curr_guide->point[1] != (y_widget->get_value())/60){
+		curr_guide->point[1] =  (y_widget->get_value())/60;
+	}
```
rodolforg, line 174: 1. why the `if` condition? After those lines you force `curr_guide->point[0]` to be `x_widget->get_value())/60` (and similar to `curr_guide->point[1]`)
2. prefer the style `if (condition) {` (note the spaces I 'wrote')
3. why so many spaces after assignment sign?
4. why the parenthesis surrounding `x_widget->get_value()`?
5. the magic number 60 is true for the default canvas settings. To get the actual value, you have to look to RendDesc value of Canvas.

synfig-studio/src/gui/dialogs/dialog_guide.cpp
```diff
@@ -160,4 +195,6 @@ Dialog_Guide::init_widget_values()
 		angle_widget->set_value(synfig::Angle::deg(curr_guide->angle).get());
 	else
 		angle_widget->set_value(curr_guide->angle.get());
+	x_widget->set_value((curr_guide->point[0])*60);
+	y_widget->set_value(curr_guide->point[1]*60);
```
rodolforg, line 199: The magic number `60` is true for the default canvas settings. To get the actual value, you have to look to `RendDesc` value of current `Canvas`.

rodolforg (changes requested):

synfig-studio/src/gui/dialogs/dialog_guide.cpp
Afterdie: i tried to follow the convention for the if conditions 
image
should i make changes?

ice0: > should i make changes?

@Afterdie 

I suggest skipping this for now because I also didn't understand that part ("After those lines you force...").
@rodolforg Can you explain it?

rodolforg: @ice0 no matter whether the `if` condition is true or false, at the end the variable value will be that one

ice0: This part is clear, but where is the value setting?
After that code only:

```
	if (ok)
		hide();
	else
		current_work_area->get_drawing_area()->queue_draw();
```

So I don't understand where these values can be set.

Afterdie: I think what @rodolforg meant to say was that why perform a check to see if the value has changed why not just set it to `x_widget->get_value()*60` which is valid but for example if the user changes only the angle, theres no need to change the pivot point too if its the same value.

rodolforg: Sorry, I meant there is no need for the `if` line. Only its contents directly

synfig-studio/src/gui/dialogs/dialog_guide.cpp
Afterdie: By default do you mean the aspect ratio, physical dimensions? im confused as to what im supposed to use from renddesc i see resolutions, height and width but the workarea ruler doesnt seem to be affected by these properties what property exactly should i be working with here?

Afterdie: @rodolforg 
```RendDesc rend_desc = rend_desc();``` 
```float factor = std::fabs((rend_desc.get_w()/2)/rend_desc.get_tl()[0]);```

File: synfig-studio/src/gui/dialogs/dialog_guide.cpp [hunk 5]
```diff
@@ -160,4 +191,9 @@ Dialog_Guide::init_widget_values()
 		angle_widget->set_value(synfig::Angle::deg(curr_guide->angle).get());
 	else
 		angle_widget->set_value(curr_guide->angle.get());
+
+	synfig::RendDesc &rend_desc(canvas->rend_desc());
+	factor = std::fabs((rend_desc.get_w()/2)/rend_desc.get_tl()[0]);
```
rodolforg, line 196: I know it is strange, but we may have different factors for X- and Y- axes.
So please compute the factors for both directions.

Btw, the computation may be wrong as you wrote. You are assuming everything is symmetric (and it usually is), but it may not be. It should be how many pixels it has in a direction divided by how many synfig unities it has in the same direction (Right - Left coordinates).

rodolforg (changes requested):
Nice! Improving your good work!

rodolforg:
One thing to note: I'll edit your PR description to include the text "Fix #ISSUE_NUMBER" to [autoclose](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) if PR is merged ;)

rodolforg:
Maybe you should let user choose the metric unit too : pixel or synfig (vector) one

synfig-studio/src/gui/dialogs/dialog_guide.cpp
Afterdie: totally forgot about the y part ill make the changes

Afterdie:
> Maybe you should let user choose the metric unit too : pixel or synfig (vector) one

would that be useful though ? when in the guide dialog the user wont be able to see the synfig scale so they cant really tell where the pivot is going to be placed. 
should i add it?

synfig-studio/src/gui/dialogs/dialog_guide.cpp
```diff
@@ -160,4 +191,10 @@ Dialog_Guide::init_widget_values()
 		angle_widget->set_value(synfig::Angle::deg(curr_guide->angle).get());
 	else
 		angle_widget->set_value(curr_guide->angle.get());
+
+	const synfig::RendDesc& rend_desc(canvas->rend_desc());
+	x_factor = std::fabs(rend_desc.get_w()/(rend_desc.get_tl()[0]-rend_desc.get_br()[0]));
+	y_factor = std::fabs(rend_desc.get_w()/(rend_desc.get_tl()[1]-rend_desc.get_br()[1]));
```
ice0, line 197: `rend_desc.get_w()` is what correct? not `get_h()`?

rodolforg:
>  user wont be able to see the synfig scale so they cant really tell where the pivot is going to be placed.

Some users have interface set to show synfig units, not pixels...

Afterdie:
@rodolforg ive added all the systems is this approach good? also how do i convert these systems back to meter?

synfig-studio/src/gui/dialogs/dialog_guide.cpp
```diff
@@ -165,8 +165,19 @@ Dialog_Guide::on_ok_or_apply_pressed(bool ok)
...
 
-	curr_guide->point[0] = x_widget->get_value()/x_factor;
-	curr_guide->point[1] = y_widget->get_value()/y_factor;
+	switch (App::distance_system)
+	{
+		case 0: //units system
+			curr_guide->point[0] = x_widget->get_value();
+			curr_guide->point[1] = y_widget->get_value();
+			break;
+		case 1: //pixel system
```
rodolforg, line 174: Instead of using integer literals, use enum "named" items:
https://github.com/synfig/synfig/blob/12d62c42c8e31ae99b925badb84e0fee2f3a27ec/synfig-core/src/synfig/distance.h#L46-L64

rodolforg:
> @rodolforg ive added all the systems is this approach good? also how do i convert these systems back to meter?

you can convert the values with `Distance` class methods, like `get()`, `units()`, `meters()`, `convert()`, etc.

Afterdie:
> you can convert the values with `Distance` class methods, like `get()`, `units()`, `meters()`, `convert()`, etc.
Thanks i think meters() works best here

rodolforg:
What d o you think about using the other methods to the other conversions you needed (and already done)?

Afterdie:
> 

using these functions could probably remove the x_factor variables but im extremely confused about how these functions work the ones that have (&RendDesc rend_desc) as arguments like convert() im getting the syntax wrong. could please help with this

Afterdie:
> What d o you think about using the other methods to the other conversions you needed (and already done)?

used convert() and removed the switch code ig this is faster

yoSachinkr:
If pivot position of guides can also snap to grids, it'd be amazing.

synfig-studio/src/gui/dialogs/dialog_guide.h
```diff
@@ -59,7 +60,11 @@ class Dialog_Guide : public Gtk::Dialog
 	WorkArea* current_work_area;
 
 	Gtk::SpinButton* angle_widget;
+	Gtk::SpinButton* x_widget;
+	Gtk::SpinButton* y_widget;
```
rodolforg, line 64: Maybe you should use `Widget_Distance` :)

Your code does not allow user to user another distance measure unit other than the current one set on preferences.

rodolforg (changes requested):

Afterdie:
@rodolforg the point to pixel conversion and vice versa isnt working for some reason. Might be an issue with .convert() method but I can figure it out.

rodolforg (approved):

rodolforg:
It's ok. The reason is that Canvas dimensions (set in RendDesc) is initially 72dpi. So, pixels and points give the same value.

Issue #1161, opened 2020-02-14
Using Brush Tool crashes Synfig consistently


FirasH:
**Synfig version & platform**:
Latest from master: 474fc9
Using openSUSE Leap 15.1

**Issue description**:
1. Start Synfig Studio
2. Select the "Brush Tool"
3. Click anywhere on canvas: a dialog will pop-up with title "Please choose an image file"
4. Click on "Cancel" or "Load"

Synfig crashes

mohamedAdhamc:
yep, just noticed this also, was going to report it but found this issue.

PR #3157, opened 2023-08-04
fix: brush tool crashes synfig if no layer is selected in the dialog

mohamedAdhamc:
Fixes one of the two issues in #1161

Opening brush tool and then clicking cancel in the file chooser dialog crashed Synfig (before this pr). The brush tool still crashes Synfig when used though on correct files I will work on that in another PR. 

Current state after fix:
[Screencast from 08-04-2023 06:00:38 PM.webm](https://github.com/synfig/synfig/assets/100296264/8e418927-d3c2-4ac3-a5a3-b7fdaa941651)


ice0 (approved):

ice0:
Merged. Thank you!

Issue #3160, opened 2023-08-05
"Current time" field text doesn't change to white color when in dark theme


pgilfernandez:
**Describe the bug**
If you switch to dark theme, the "current time" field text. doesn't change to white color as expected (the rest of UI fields do it)

**To Reproduce**
1. Go to 'Preferences'
2. Set on 'Dark UI theme'
3. Close the Preferences window
4. Check that the text color of "current time" didn't change

**Expected behavior**
It should change to white color as other fields text do.

**Screenshots**
<img width="192" alt="current_frame" src="https://github.com/synfig/synfig/assets/5942369/596bd372-3a7b-4dd2-87d0-076fbf49b0c3">

**System information:**
 - OS: macOS
 - OS version: 12.6.7
 - Synfig version: 1.5.1

**Additional comments**
none

PR #3163, opened 2023-08-06
fix: current time widget color in dark theme

ice0:
Before:
image

After:
image

fix: https://github.com/synfig/synfig/issues/3160

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -2078,11 +2078,9 @@ CanvasView::on_time_changed()
 	{
 		KeyframeList::iterator iter;
 		if (get_canvas()->keyframe_list().find(time, iter)) {
-			current_time_widget->get_style_context()->remove_class("color-black");
 			current_time_widget->get_style_context()->add_class("color-red");
 		} else {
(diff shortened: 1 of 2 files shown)


rodolforg (commented):
Do we have to add/remove any class besides the `color-red`?

Anyway, it's good to remove the `color-black` from synfig.css, as it was used only by this removed code.

ice0:
> Do we have to add/remove any class besides the `color-red`?

Do you mean here? I think no, only `color-red`.

> Anyway, it's good to remove the `color-black` from synfig.css, as it was used only by this removed code.

Done.

rodolforg:
> Do you mean here? I think no, only `color-red`.

So we need to add/remove the class "label" here?

ice0:
> So we need to add/remove the class "label" here?

Moved it outside.

rodolforg (dismissed):
I thought we didn't need to add the label class at all, but it is fine as it is :)

ice0:
> I thought we didn't need to add the label class at all, but it is fine as it is :)

Just checked. Yes, no difference, so I removed it.

rodolforg (approved):

PR #3166, opened 2023-08-06
fix: wrong Copy and Cut icon names

pgilfernandez:
- Cut icon name had a minor typo
- Copy icon was referring to "gtk-copy" and it's not valid (at least in macOS), so I guess the correct one is "edit-copy" as happens with Cut and Paste ones.

### DIFF
--- a/synfig-studio/src/gui/actionmanagers/layeractionmanager.cpp
+++ b/synfig-studio/src/gui/actionmanagers/layeractionmanager.cpp
@@ -201,7 +201,7 @@ LayerActionManager::LayerActionManager():
 	);
 	action_copy_=Gtk::Action::create_with_icon_name(
 		"copy",
-		"gtk-copy", _("_Copy"), _("Copy")
+		"edit-copy", _("_Copy"), _("Copy")
 	);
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/actionmanagers/layeractionmanager.cpp
```diff
@@ -191,7 +191,7 @@ LayerActionManager::LayerActionManager():
 {
 	action_cut_=Gtk::Action::create_with_icon_name(
 		"cut",
-		"edit-cut", _("Cu_t"), _("Cut")
```
rodolforg, line 194: This is not a typo ;) It is a [mnemonic](https://en.wikipedia.org/wiki/Mnemonics_(keyboard))

pgilfernandez: Hahaha, I didn't think it was gonna be that... sorry.
I reverted it ;-)

rodolforg (approved):

ice0:
Merged. Thank you!

PR #3173, opened 2023-08-08
feat: layer Motion Blur now can create a non blurred trail

### LINKED ISSUE #3172: Feature request: Opacity parameter for Motion blur layer
The problem is explained [here](https://forums.synfig.org/t/motion-blur-without-decay/14278) (pictures in there), but in short: I would like to create a trail of duplicates (or subsamples) that follow the parent one but with the option to control its opacity.

The best option I found it to use the "motion blur" layer, set the subsampling type to "constant" and play with the factor and aperture. As I need the subsamples to be opaque (or as opaque as the parent) I can't control that, and that's why the request.

The workaround I'm using now is adding a "duplicate" layer and setting the "to" parameter high till I see the result is opaque. The con is that this solution makes the rendering so so slow.

**Additional comments**
Maybe there is a better solution as Kai in Synfig forums suggested, but, it doesn't behave exactly the same, it's so hard to animate and in some scenarios it's not valid at all while the suggestion would do the trick.

Thanks!

rodolforg:
set "Subsampling Type" to "None".

fix #3172


ice0 (approved):

ice0:
Merged. Thank you!

Issue #3174, opened 2023-08-10
Wrong button labels in Custom Workspace editor 


pgilfernandez:
**Describe the bug**
Actual Custom Workspaces Editor looks like:
<img width="342" alt="workspaces_editor" src="https://github.com/synfig/synfig/assets/5942369/89b7944a-8e14-4cf4-a716-de3edc7a3bbb">

**To Reproduce**
1. Open Window / Workspace / Edit workspace...
2. Look at the buttons labels

**Expected behavior**
I guess the right labels would be just "Delete" and "Close"

**System information:**
 - OS: macOS
 - OS version: 12.6.7
 - Synfig version: 1.5.1

**Additional comments**
none

rodolforg:
Problem is here: synfig-studio/src/gui/resources/ui/dialog_workspaces.glade

pgilfernandez:
@rodolforg thanks for the tip. I made a simple PR to fix it: https://github.com/synfig/synfig/pull/3175

Thanks

pgilfernandez:
close as completed in https://github.com/synfig/synfig/pull/3175

PR #3175, opened 2023-08-11
fix: revert Custom Workspaces Editor button labels to GTK ones

pgilfernandez:
Simple fix for https://github.com/synfig/synfig/issues/3174 as @rodolforg suggested.

Thanks

### DIFF
--- a/synfig-studio/src/gui/resources/ui/dialog_workspaces.glade
+++ b/synfig-studio/src/gui/resources/ui/dialog_workspaces.glade
@@ -104,7 +104,7 @@
                 </child>
                 <child>
                   <object class="GtkButton" id="workspaces_delete_button">
-                    <property name="label">gtk-delete</property>
+                    <property name="label" translatable="yes">_Delete</property>
                     <property name="visible">True</property>
                     <property name="sensitive">False</property>
                     <property name="can_focus">True</property>
@@ -119,7 +119,7 @@
                 </child>
                 <child>
                   <object class="GtkButton" id="workspaces_close_button">
-                    <property name="label">gtk-close</property>
+                    <property name="label" translatable="yes">_Close</property>
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="receives_default">True</property>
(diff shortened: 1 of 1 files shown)


pgilfernandez:
Umm, why the "conventional commits style" fails? uppercase after "fix:"?

pgilfernandez:
> Umm, why the "conventional commits style" fails? uppercase after "fix:"?

I respond myself: no verb or action was added...

rodolforg:
Maybe replace directly with Close and Delete strings?

pgilfernandez:
I just reverted the changes to the ones that worked.
If you prefer to change to just text strings then we should change the "Edit" button as well.

What do you want to do?

ice0:
> I just reverted the changes to the ones that worked. If you prefer to change to just text strings then we should change the "Edit" button as well.

I don't see `Edit` button. Do you mean `Rename` button?

> What do you want to do?

At first I thought that these labels would be automatically translated, but unfortunately they are not, they are still in English after changing the language.
So yes, it's better to just rename it as Rodolfo suggested.

pgilfernandez:
> I don't see `Edit` button. Do you mean `Rename` button?

Yes, I mean `Rename` button that is gtk-rename but afterwards the text really applied is `Edit` (at least in Synfig 1.4.1):

workspaces_141


> So yes, it's better to just rename it as Rodolfo suggested.

Ok, I will rebase and add the new modification... just to be sure, I do these changes:
`gtk-delete` turns into just `Delete` string
`gtk-rename` turns into just `Edit`string
`gtk-close` turns into just `Close`string

And the strings will be translatable or is it needed any other work?

rodolforg:
> `gtk-rename` turns into just `Edit`string

`Rename` is the best choice here. You don't **edit** the workspace layout, only the given name.



> And the strings will be translatable or is it needed any other work?

You have to add an xml attribute `translatable` like this:
`<property name="label" translatable="yes">The Button Label</property>`

pgilfernandez:
Well, I think it's done... I hope I did it right.
Rebased as well =)

rodolforg (approved):

ice0:
Merged. Thank you!

PR #3184, opened 2023-08-19
feat: added shortcut key for deleting guides 

Afterdie:
Pressing the esc key deletes the guide that is being hovered on. smol shortcut should come in handy

### DIFF
--- a/synfig-studio/src/gui/workarea.cpp
+++ b/synfig-studio/src/gui/workarea.cpp
@@ -976,6 +976,10 @@ WorkArea::on_key_press_event(GdkEventKey* event)
 	if (guide_highlighted && (event->keyval == GDK_KEY_Control_L || event->keyval == GDK_KEY_Control_R))
 		rotate_guide=true;
 
+	if (event->keyval == GDK_KEY_x){
+		//deleting logic goes here
+	} 
+
 	auto event_result = canvas_view->get_smach().process_event(
 		EventKeyboard(EVENT_WORKAREA_KEY_DOWN, event->keyval, Gdk::ModifierType(event->state)));
 	if (event_result != Smach::RESULT_OK)
--- a/synfig-studio/src/gui/workarea.cpp
+++ b/synfig-studio/src/gui/workarea.cpp
@@ -1043,7 +1043,7 @@ WorkArea::on_key_release_event(GdkEventKey* event)
 	if (event_result != Smach::RESULT_OK)
 		return true;
 
-	if (event->keyval == GDK_KEY_x && guide_highlighted)
+	if (event->keyval == GDK_KEY_Escape && guide_highlighted)
 		get_guide_list().erase(curr_guide);
 
 	// Other possible actions if current state doesn't accept the event but not forbids it
--- a/synfig-studio/src/gui/workarea.cpp
+++ b/synfig-studio/src/gui/workarea.cpp
@@ -976,10 +976,6 @@ WorkArea::on_key_press_event(GdkEventKey* event)
 	if (guide_highlighted && (event->keyval == GDK_KEY_Control_L || event->keyval == GDK_KEY_Control_R))
 		rotate_guide=true;
 
-	if (event->keyval == GDK_KEY_x){
-		//deleting logic goes here
-	} 
-
 	auto event_result = canvas_view->get_smach().process_event(
 		EventKeyboard(EVENT_WORKAREA_KEY_DOWN, event->keyval, Gdk::ModifierType(event->state)));
 	if (event_result != Smach::RESULT_OK)
@@ -1046,6 +1042,9 @@ WorkArea::on_key_release_event(GdkEventKey* event)
 		EventKeyboard(EVENT_WORKAREA_KEY_UP, event->keyval, Gdk::ModifierType(event->state)) );
 	if (event_result != Smach::RESULT_OK)
 		return true;
+
+	if (event->keyval == GDK_KEY_x && guide_highlighted)
+		get_guide_list().erase(curr_guide);
 
 	// Other possible actions if current state doesn't accept the event but not forbids it
 	// - currently none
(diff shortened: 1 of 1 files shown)


ice0:
Hi, Akshat! 

Can you record a short video showing the process?
I just checked and the guide is not removed after pressing `Esc` key.
If I press `Esc` multiple times, Synfig crashes.

Afterdie:
@ice0 added a redraw event should be working as expected now 👍

ice0:
> @ice0 added a redraw event should be working as expected now 👍

Now it's better, but it still crashes if I press `Esc` second time.

synfig-studio/src/gui/workarea.cpp
```diff
@@ -1043,6 +1043,13 @@ WorkArea::on_key_release_event(GdkEventKey* event)
 	if (event_result != Smach::RESULT_OK)
 		return true;
 
+	if (event->keyval == GDK_KEY_Escape && guide_highlighted)
+	{
+		get_guide_list().erase(curr_guide);
+		guide_highlighted = !guide_highlighted;
```
rodolforg, line 1049: based on the `if` condition, just set it to `false`, because the only way this line is reached is `guide_highlighted` being `true`.

Besides, `curr_guide` should now be properly set.

rodolforg (changes requested):

synfig-studio/src/gui/workarea.cpp
Afterdie: couldnt understand the curr_guide part could you explain or give an example. In the case where guide is dragged into the ruler it just erases(curr_guide)

rodolforg: The variable scope is not only the method `on_key_release_event()` and after the erase, it stores an invalid iterator, as the container changed size.
Pointing to `get_guide_list().end()` is fine.

ice0:
Looks much better, but I noticed two more issues.
1) The guide is not deleted if the user pressed the `Esc` key immediately after the guide was created.
2) Maybe update the current guide after deleting the current one?
In the example below, after removing the horizontal guide, the vertical guide is not selected until the user moves the mouse.


https://github.com/synfig/synfig/assets/5604544/48e9aef7-e77c-48ed-a9f7-d7e33c35820d

Afterdie:
> Looks much better, but I noticed two more issues.
> 
> 1. The guide is not deleted if the user pressed the `Esc` key immediately after the guide was created.
> 2. Maybe update the current guide after deleting the current one?
>    In the example below, after removing the horizontal guide, the vertical guide is not selected until the user moves the mouse.
> 
>  guide_issues.mp4

fixed both the issues. Im not sure if it the queue_draw() is in the right place tho 🤔

synfig-studio/src/gui/workarea.cpp
```diff
@@ -1749,6 +1760,7 @@ WorkArea::on_hruler_event(GdkEvent *event)
 			set_drag_mode(DRAG_GUIDE);
 			curr_guide = get_guide_list().insert(get_guide_list().begin(),{synfig::Point(((1.0/2.0)*(drawing_area->get_window()->get_width())*get_pw())+ get_window_tl()[0], 0),
 																			   synfig::Angle::rad(0)});
+			guide_highlighted = true;
```
rodolforg, line 1763: Why did you write this here?

rodolforg, line 1763: Why did you write this here?

synfig-studio/src/gui/workarea.cpp
```diff
@@ -1031,12 +1031,23 @@ bool
...
 	if (event_result != Smach::RESULT_OK)
 		return true;
 
+	if (event->keyval == GDK_KEY_Escape && guide_highlighted)
+	{
+		get_guide_list().erase(curr_guide);
+		drawing_area->queue_draw();
+		GuideList::iterator iter = find_guide(get_cursor_pos(), radius);
+		curr_guide = iter;
+		drawing_area->queue_draw();
```
rodolforg, line 1047: 1. Why do you need two `queue_draw()` calls in a row? Although it won't make it redraw twice, it has no point.
2. After you remove a guide from `GuideList`, you have to save the metadata, as seen in L1505-1515. Otherwise, `ctrl+Z` won't work.

synfig-studio/src/gui/workarea.cpp
```diff
@@ -1031,12 +1031,23 @@ bool
 WorkArea::on_key_release_event(GdkEventKey* event)
 {
 	SYNFIG_EXCEPTION_GUARD_BEGIN()
+	const float radius((std::fabs(pw)+std::fabs(ph))*4);
```
rodolforg, line 1034: Why is this `radius` defined here, so afar from its real scope?

synfig-studio/src/gui/workarea.cpp
```diff
@@ -1787,6 +1799,7 @@ WorkArea::on_vruler_event(GdkEvent *event)
 			set_drag_mode(DRAG_GUIDE);
 			curr_guide = get_guide_list().insert(get_guide_list().begin(),{synfig::Point(0 ,((1.0/2.0)*(drawing_area->get_window()->get_height())*get_ph())+ get_window_tl()[1]),
 																			   synfig::Angle::deg(90)});
+			guide_highlighted = true;
```
rodolforg, line 1802: Why did you write this here?

rodolforg (changes requested):

synfig-studio/src/gui/workarea.cpp
Afterdie: It's being used at 1045 for finding the guide should i move this further down?

synfig-studio/src/gui/workarea.cpp
Afterdie: After dragging a new guide from the rule user wont be able to immediately delete the guide because guide_highlighted is set to false. By adding this line you can drag a new guide and then delete immediately without having to move the mouse. Added to the vruler for the same reason.

synfig-studio/src/gui/workarea.cpp
rodolforg: Item 2 I think I fixed in #3304

PR #3200, opened 2023-09-04
perf: optimized Action_System so that the canvas only renders once during an action

mosasauridae:
The problem was that every time any property node emitted signal_changed() during an action, it would cause the canvas to queue a render. The number of times signal_changed() happens seems to grow non-linearly with the number of objects being modified during the action. This was a killer for bulk update actions, where this could easily trigger thousands or potentially even millions of unnecessary renders. I have seen a single bulk action take almost an hour.

The fix is to temporarily block the Canvas_Interface::signal_dirty_preview() signal while an action is in progress, and then manually trigger it once the action completes. This guarantees the scene will only render once the action is done, and seems to be a dramatic performance improvement in longer animations at least in my testing.

I ran some A-B comparisons on a test file. The file has a spline with 100 vertices linked to a bone, and the bone angle parameter has 600 waypoints.
 - Modify bone angle with Animate Mode off and apply offset:  ~3min 40s before -> ~1.5s after
 - Move a keyframe:  ~4min 25s before -> 0.7s after

[perf-test.zip](https://github.com/synfig/synfig/files/12515338/perf-test.zip)

(Please let me know if there's any other process stuff I should do. As far as I'm aware there's no specific issue for this)

### DIFF
--- a/synfig-studio/src/synfigapp/action_system.cpp
+++ b/synfig-studio/src/synfigapp/action_system.cpp
@@ -106,6 +106,8 @@ Action::System::perform_action(etl::handle<Action::Base> action)
 
 	Action::CanvasSpecific *canvas_specific = dynamic_cast<Action::CanvasSpecific*>(action.get());
 
+	DirtySignalBlocker dirtyBlocker(canvas_specific);
+
 	if (canvas_specific && canvas_specific->get_canvas())
 		uim = static_cast<Instance*>(this)->find_canvas_interface(canvas_specific->get_canvas())->get_ui_interface();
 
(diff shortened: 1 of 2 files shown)


mosasauridae:
This change also has the side effect of buffering `Canvas::signal_changed()` during an action, since `CanvasInterface::signal_dirty_preview()` is actually just a reference to it. I did a review of the places where Canvas's signal_changed is referenced and don't see anything where the extra intermediate calls would matter, so it should be fine either way. But at the same time, there could be edge cases I'm not aware of.

To play it safe, I can split out signal_dirty_preview into its own signal so that it can be buffered without also buffering signal_changed. It's like 3 lines of code and doesn't impact performance either way (I actually had this change implemented originally but it got lost at some point).

ice0 (approved):

ice0:
Merged. Thank you!

PR #3201, opened 2023-09-04
fix: new MotionBLur subsampling None used wrong blend method

### LINKED ISSUE #3199: Incorrect blend mode for "None" Subsampling type in Motion blur layer
**Describe the bug**
Related to this new feature https://github.com/synfig/synfig/pull/3173, when setting Subsampling type to "none", the blending mode is `Add` when I guess it should be `Composite` (or just none as we don't really want the subsamples to blend).

**To Reproduce**
1. create any Geometry layer
2. add Motion blur layer
3. set "subsampling type" to "none"
4. check how they are blended using `Add`

**Expected behavior**
They should not get blended but added one behind the other one

**Screenshots**
Here an example:
example
Another one:
example2

**System information:**
 - OS: macOS (I guess applicable to all)
 - OS version: 12.6.7
 - Commit ID: bb67147

**Additional comments**
Maybe it would be a good idea to "open" the blending mode as a parameter so that users could change it and get different interesting results...

Thanks

rodolforg:
the idea is to have a trail of duplicate canvas.

fix #3199

### DIFF
--- a/synfig-core/src/synfig/layers/layer_motionblur.cpp
+++ b/synfig-core/src/synfig/layers/layer_motionblur.cpp
@@ -163,6 +163,8 @@ Layer_MotionBlur::build_rendering_task_vfunc(Context context) const
 	Real subsample_start = param_subsample_start.get(Real());
 	Real subsample_end = param_subsample_end.get(Real());
 
+	const bool no_blur_effect = subsampling_type == SUBSAMPLING_NONE;
+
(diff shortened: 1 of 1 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

pgilfernandez:
It is working like a charm, I love it.

Thanks!

PR #3207, opened 2023-09-08
fix: selecting a handle while the Graphs window tooltip is open crashes the gui

mosasauridae:
Reproduction steps:
1. Select an animated handle, and open the Graphs window.
2. Right click any waypoint in the Graphs window, and without selecting any menu item (or by selecting an item that's outside of the Graphs area), click in the Canvas work area. Notice the tooltip is still open.
3. Left click any other handle in the work area. Synfig crashes.

The problem was the tooltip draw event was directly referencing an iterator that gets invalidated when the new handle is selected. The next time it tries to draw, it dereferences the bad iterator and crashes.

A nice side effect of this change is the tooltip for the deleted waypoint goes away now after you delete it.

https://github.com/synfig/synfig/assets/143972888/aea7a137-3717-476a-b16a-d1a4419b84a8

The recording didn't capture the context menu. I selected Remove from the menu, and the crash happens after I select the new handle.

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_curves.cpp
+++ b/synfig-studio/src/gui/widgets/widget_curves.cpp
@@ -818,7 +818,7 @@ Widget_Curves::on_draw(const Cairo::RefPtr<Cairo::Context> &cr)
 	}
 
 	// Draw info about hovered item
-	if (channel_point_sd.get_hovered_item().is_valid() || channel_point_sd.get_state() == channel_point_sd.POINTER_DRAGGING) {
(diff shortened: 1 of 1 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

rodolforg:
Do we intend to release 1.4.5? If so, this PR should be backported

ice0:
> Do we intend to release 1.4.5? If so, this PR should be backported

Marked it for backport.

PR #3208, opened 2023-09-08
fix: scale value of transformation on Parameters Panel is dimensionless

rodolforg:
the short textual representation of a parameter that is Transformation value type (e.g. Group Layer's transformation) displayed scale vector as it was a distance (i.e. pixels or Synfig units), but it is just a multiplicative factor.


rodolforg:
I force-pushed to change string concatenation to string format.
And then AppVeyor has a random failure. :P

ice0 (approved):

ice0:
Merged. Thank you!

PR #3214, opened 2023-09-13
fix: issues with importing with absolute paths from different root path

### LINKED ISSUE #3212: Absolute paths don't work anymore
**Describe the bug**
After Preview window got fixed, there's a bigger problem that was lying underneath. Synfig can no longer import external resources (images, sounds) by their absolute paths. Why would you need to use absolute paths instead of relative? Because on Windows, you can have multiple disks: C:, D:, E:, etc. And if your external resources are located on a different disk than your project file, you can no longer use relative paths. So, **this is a Windows-specific bug**.

**To Reproduce**
1. Place an external resource (a sound or a picture, not important) on a different drive than C:. If you have only one drive, plug in a flash drive
2. In empty canvas, Menu->File->Import...
3. Select your external resource and see that its "Filename" parameter is empty

**Expected behavior**
Being able to import resources from different drives.

**System information:**
 - OS: Windows 10, x86-64
 - Commit ID: fb11ad27d80903cf5eaba4698298801512e928b6

rodolforg:
and implements missing lexically_proximate(), that actually do what old ETL relative() did.
relative() returns empty path if path isn't relative to base path.

It also fixes the Widget_Filename error triggered by clicking on its Open button with an absolute path (it just concatenated current path with file path). See https://github.com/synfig/synfig/pull/3214#issuecomment-1718733044 .

fix #3212

### DIFF
--- a/synfig-core/src/synfig/canvas.cpp
+++ b/synfig-core/src/synfig/canvas.cpp
@@ -561,7 +561,7 @@ Canvas::_get_relative_id(etl::loose_handle<const Canvas> x)const
 
 	if(x && get_root()!=x->get_root())
 	{
-		String file_name = filesystem::Path(get_file_name()).relative_to(x->get_file_path()).u8string();
+		String file_name = filesystem::Path(get_file_name()).proximate_to(x->get_file_path()).u8string();
 
 		id=file_name+'#'+id;
 	}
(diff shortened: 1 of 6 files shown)


rodolforg:
@Svarov-RZM if it builds via AppVeyor (I hope 'unit'-tests work on Windows), please test it.

ice0:
@rodolforg 
Any idea how I can reproduce it by myself?

rodolforg:
Maybe importing in a saved file?

ice0:
> Maybe importing in a saved file?

Yes. It looks like this fixes the problem.
But I noticed another issue:

https://github.com/synfig/synfig/assets/5604544/6de06f61-d4c7-4dc8-9070-c83b251431d1

rodolforg:
> But I noticed another issue:

By looking at the code, it seems it exists since ever. I hope I fixed in last commit.

rodolforg:
I accidentally pushed the wrong branch in the before-last push. The last one was to revert my mistake.

rodolforg:
(rebased)

rodolforg:
(rebased)

ice0 (approved):

ice0:
Merged. Thank you!

Issue #3222, opened 2023-09-19
macOS autotools builds miss app icon


pgilfernandez:
**Describe the bug**
Builds produced with autotools under macOS result in apps without icon

**To Reproduce**
1. run `./1-setup-osx-brew.sh`
2. run `./2-build-production.sh`
3. run `./3-package-osx-dmg.sh`
4. check SynfigStudio.app has no icon

**Expected behavior**
It should have Synfig icon

**Screenshots**

<img width="613" alt="screenshot" src="https://github.com/synfig/synfig/assets/5942369/05d139f3-abb8-47f4-8087-a194fa2ee019">

**System information:**
 - OS: macOS
 - OS version: 12.6.7
 - Synfig version: 1.5.1

**Additional comments**
I observed since a few month but I didn't pay attention as I usually run these test builds without packaging.
Now I wanted to have a shareable updated Synfig dev version and I found out this.

I'm experiencing another problems but I still would like to test or try to solve them by myself as I don't know if they depend on my setup or it's a real bug. I'll open another Issue in case I can't solve it.
Thanks

rodolforg:
Maybe this link helps: https://stackoverflow.com/questions/646671/how-do-i-set-the-icon-for-my-applications-mac-os-x-app-bundle

pgilfernandez:
I found the problem, PR https://github.com/synfig/synfig/pull/3223 solves this issue, let me close this.

PS - While looking for the problem I found out `.icns` macOS icon files are pretty outdated as they don't include higher resolutions used in modern OS versions so I'm creating another PR for updating them both. =)

PR #3223, opened 2023-09-20
fix: fixed macOS Info.plist file

pgilfernandez:
This PR closes https://github.com/synfig/synfig/issues/3222.

After a pretty deep analysis I found out the problem was simple: a `plist` file uses XML language and ampersand symbol "&" is an special character that can't be used alone, causing the XML file to be corrupted and so on the .plist file.

It was used for the following text twice in the file:
```
<string>© 2001-2023 Synfig developers & contributors</string>
```
while they should be:
```
<string>© 2001-2023 Synfig developers &#38; contributors</string>
```
This way the .plist file works fine and shows correctly not only the icon but the version and copyright tags:
info_plist

Cheers

### DIFF
--- a/autobuild/osx/app-template/Contents/Info.plist
+++ b/autobuild/osx/app-template/Contents/Info.plist
@@ -18,13 +18,13 @@
 	<string>SynfigStudio</string>
 
 	<key>CFBundleGetInfoString</key>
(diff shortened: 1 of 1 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

Issue #2132, opened 2021-04-30
[Synfig Studio] Alternate row style in Parameters and Timeline windows for easier editing


FirasH:
**Synfig version & platform**:
master 7d4d8b7

**Issue description**:
When there are many close waypoints in the `Timeline` it becomes hard to recognize to which paramater the waypoint belongs to, like in the following case:
![waypoints-timeline](https://user-images.githubusercontent.com/6705690/116667629-b31dd080-a99c-11eb-9dbd-cc9f3a8aa208.PNG)

and that becomes harder with wider screens as the distance from the `Parameters` window on the left is increased.

Sometimes I find myself selecting the parameter that I want to edit to highlight the exact row:
![waypoints-timeline-selected](https://user-images.githubusercontent.com/6705690/116668059-3808ea00-a99d-11eb-9828-ad5aa63e9083.PNG)

What about styling the `Parameters` and `Timeline` windows with an alternate row style, something like this mockup?
![waypoints-timeline-alternate](https://user-images.githubusercontent.com/6705690/116669711-32ac9f00-a99f-11eb-91f7-ed1b35683b7e.png)

rodolforg:
It looked like this way in older versions. I guess it was the current Gtk Theme (Adwaita) that doesn't implement it.
We have to change the requested behavior in our custom CSS.

FirasH:
The style should be applied to `tree_view` somewhere here:
https://github.com/synfig/synfig/blob/7d4d8b79ba5e00e1e28d66fb3b2ebb30f875387e/synfig-studio/src/gui/docks/dock_params.cpp#L93-L113

rodolforg:
I think it should be somewhere else, where the Params treeview is created.

I think it is in layertree.

Keyikedalube:
Over [here](https://stackoverflow.com/questions/9036812/how-do-you-change-alternating-background-row-colors-of-a-gtk-treeview-in-pygtk) this guy did it with TreeView API member function [Gtk::TreeView::set_rules_hint()](https://developer-old.gnome.org/gtkmm/3.24/classGtk_1_1TreeView.html#afc95decfb73c43e44e2fbf1e87818dcf):
```c++
treeview_obj.set_rules_hint() // <-- default true in GTKMM
```

Then I noticed Synfig has that turned on already:
[...]

PR #3230, opened 2023-10-01
feat: add CSS classes and id to widgets/panels to allow visual customization

rodolforg:
- Timetrack panel named as `timetrack_panel`
- Timetrack widgets use class `timetrack`
- Keyframes panel named as `keyframes_panel`
- Keyframes widgets use class `keyframes`
- Layers panel named as `layers_panel`
- LayerTree widgets use class `layers`
- Parameters panel named as `parameters_panel`
- ParameterTree widgets use class `parameters`
- LayerSets panel named as `layersets_panel`
- LayerSetTree widgets use class `layersets`
- Library panel named as `library_panel`
- Library widgets use class `library`

### DIFF
--- a/synfig-studio/src/gui/docks/dock_children.cpp
+++ b/synfig-studio/src/gui/docks/dock_children.cpp
@@ -58,7 +58,9 @@ using namespace studio;
 Dock_Children::Dock_Children():
 	Dock_CanvasSpecific("children",_("Library"),"library_icon")
 {
-    //! \todo the combination of set_use_scrolled(false) and
+	set_name("library_panel");
+
+	//! \todo the combination of set_use_scrolled(false) and
 	//!       add(*tree_view) causes crashes when
 	//!       Dockable::prev_widget_ is left referencing a
 	//!       ChildrenTree that has already been destroyed.
--- a/synfig-studio/src/gui/docks/dock_keyframes.cpp
+++ b/synfig-studio/src/gui/docks/dock_keyframes.cpp
@@ -64,6 +64,7 @@ Dock_Keyframes::Dock_Keyframes():
 	Dock_CanvasSpecific("keyframes", _("Keyframes"),"keyframe_icon"),
 	keyframe_action_manager(new KeyframeActionManager())
 {
+	set_name("keyframes_panel");
 	// Make Keyframes toolbar small for space efficiency
 	get_style_context()->add_class("synfigstudio-efficient-workspace");
 
--- a/synfig-studio/src/gui/docks/dock_layergroups.cpp
+++ b/synfig-studio/src/gui/docks/dock_layergroups.cpp
@@ -64,6 +64,8 @@ Dock_LayerGroups::Dock_LayerGroups():
 	action_group_group_ops(Gtk::ActionGroup::create("action_group_dock_layergroups")),
 	group_action_manager(new GroupActionManager)
 {
+	set_name("layersets_panel");
+
 	// Make Sets toolbar buttons small for space efficiency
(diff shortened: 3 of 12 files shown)


rodolforg:
@pgilfernandez is that what you want?
https://github.com/synfig/synfig/issues/2132#issuecomment-1733415166

I'm trying to allow to style it with alternate row colors (as with tr:even and tr:odd), but no success so far.

pgilfernandez:
Hi @rodolforg,

> @pgilfernandez is that what you want?
> [#2132 (comment)](https://github.com/synfig/synfig/issues/2132#issuecomment-1733415166)

Yes, this is right what I'm looking for. I was able to guess some of them but others no, specially the timetrack ones.
Please, have a look at these 2 problems I'm facing:

- `timetrack` class applies to the whole "box" of waypoints, I would like to be able to style it "per row", with a different class called something like `timetrack_row`
- the selected `timetrack` row gets the style of `parameters`class, something not desiderable... do you know why or how could I override it so that I can style a "timetrack selected row"?:
selected




> I'm trying to allow to style it with alternate row colors (as with tr:even and tr:odd), but no success so far.

That's my target, I'm trying to get that "alternate row colors" with another CSS approaches but I need to be able to style individual rows.
Thanks

pgilfernandez:
> I'm trying to allow to style it with alternate row colors (as with tr:even and tr:odd), but no success so far.

The following test was ran just with Layer panel:
It looks like "even" and "odd" works, but each row is the first one so that's why it looks like it doesn't work. It's like there is no container and each row it's a different "tree"...
Try it with:
```
 .layers:nth-child(odd) {
	background-color: orange;
 }

 .layers:nth-child(even) {
	background-color: green;
 }
```

rodolforg:
Bad news: https://gitlab.gnome.org/GNOME/gtk/-/issues/847
https://gitlab.gnome.org/GNOME/gtk/-/issues/581



> * the selected `timetrack` row gets the style of `parameters`class, something not desiderable... do you know why or how could I override it so that I can style a "timetrack selected row"?:

only for the selected row color. The widget I used for 'new' widget timetrack (Gtk::DrawingArea) doesn't have the `:selected` CSS set by Adwaita and many Gtk themes... so it wouldn't have the color for selected row... :(

pgilfernandez:
Yes, it looks like bringing back the odd/even styling is not going to be possible... I still have some other alternative solutions in the pocket, but I need to "solve" this:


> * `timetrack` class applies to the whole "box" of waypoints, I would like to be able to style it "per row", with a different class called something like `timetrack_row`

It's the same I was asking in the other issue:
rows
Long story short: could you create a class name for unselected timetrack rows? (the ones in red in the previous picture)

Thanks

rodolforg:
But it is possible, isn't it?

```css
.timetrack{
  background-color:red;
}
```
image

pgilfernandez:
> But it is possible, isn't it?

Yes and no:

- Yes: it's possible to style the whole background with colors or images
- No: the workaround I'm working on needs to apply an image to each row, so that with the help of CSS use images and overflown images to create the effect of "alternate row backgrounds". Another experiment follows the idea of very shuttle grey line in the middle of the row to visually help follow the row... but it needs an image per row, it's important this way as these panels (Parameters and Timetrack) have their scrolling synchronized when the information doesn't fit the panel dimensions. This is what I get now with this approach:
 
lines

rodolforg:
How do you do in parameters panel?

ice0 (approved):

ice0:
Merged. Thank you!

PR #3231, opened 2023-10-03
feat: workaround to allow to alternate even/odd row colors for Timetrack

rodolforg:
Odd rows have an additional class "odd".
Similarly, even rows have "even" class.

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_timetrack.cpp
+++ b/synfig-studio/src/gui/widgets/widget_timetrack.cpp
@@ -427,8 +427,10 @@ bool Widget_Timetrack::on_draw(const Cairo::RefPtr<Cairo::Context>& cr)
 	// instead of iterate over entire param list
 	//	Gtk::TreePath start_path, end_path;
 	//	params_treeview->get_visible_range(start_path, end_path);
+	int row_number = 0;
+	params_store->foreach_path([=, &row_number](const Gtk::TreeModel::Path &path) -> bool {
+		++row_number;
 
-	params_store->foreach_path([=](const Gtk::TreeModel::Path &path) -> bool {
 		const RowInfo &row_info = param_info_map[path.to_string()];
 		if (!row_info.is_valid()) {
 			queue_rebuild_param_info_list();
@@ -441,6 +443,15 @@ bool Widget_Timetrack::on_draw(const Cairo::RefPtr<Cairo::Context>& cr)
 
 		if (geometry.y + geometry.h < 0 || geometry.y > get_height())
 			return false;
+
+		// alternate even-odd row colors
+		{
+			const std::string class_name = row_number % 2 == 0 ? "even" : "odd";
+			auto context = get_style_context();
+			context->add_class(class_name);
+			context->render_background(cr, 0, geometry.y, get_width(), geometry.h);
+			context->remove_class(class_name);
+		}
 
 		// is param selected?
 		draw_selected_background(cr, path, row_info);


pgilfernandez:
Nice,
I still was trying things in the other PR but I see this one could make things easier... how could I style "even" and "odd" in this PR?
I tried `.even` and `#timetrack .even` but I couldn't change their background...

rodolforg:
ouch I committed wrong. it shouldn't be `bool`, but `int`.

You should use it like this:
`.timetrack.odd`

But while previous PR isn't merged, use like this:
`.odd`

pgilfernandez:
I still can't make it work... I tried:
```
.timetrack.odd,
.timetrack .odd,
.odd {
   background-color: green;
}
```
without any luck...

rodolforg:
Ugh. I don't know why then :(
image

rodolforg:
You are using Timetrack 2 Panel, right?

pgilfernandez:
> Ugh. I don't know why then :(

Ouch! my fault... I was compiling another repository... sorry about it.
It works like a charm.

Could you extend this workaround to the parameters panel so that we can have the same CSS style for both? I think I could mimic it with background images as I told you but your workaround is better.

light
dark


Here is the style I would propose as is valid for both light and dark theme styles:
```
.timetrack.odd {
   background-color: rgba(255,255,255,0.05);
}

.timetrack.even {
   background-color: rgba(0,0,0,0.05);
}
```

If you agree I could PR it or add it to yours

PS1 - I thik less important are the other treeview panels such as layers, keyframes and so, but it may be nice that they all recover the alternate row styling...
PS2 - could this styling be activated/deactivated by users with a Preferences toggle or it's better to just make it visible to all users? (I guess pro users would always be able to go to the CSS files and make custom changes...)

rodolforg:
> Could you extend this workaround to the parameters panel so that we can have the same CSS style for both? I think I could mimic it with background images as I told you but your workaround is better.

Sadly, it isn't that easy, because they aren't the same UI widget type. Parameters, Layers and Sets Panels are GtkTreeView. Timetrack is a GtkDrawingArea.
I did a quick experiment here and I only a drawing over the widget or below its own backgroun (i.e, completely invisible). Nothing between background and text. The first case maybe can be used to draw lines as you background image proposes.



> Here is the style I would propose as is valid for both light and dark theme styles:

Maybe somehow this color operations help? https://docs.gtk.org/gtk3/css-overview.html#an-example-for-defining-colors



> If you agree I could PR it or you can just add it to this one.

I prefer another commit/PR. This one provide the functionality/possibility only :)

> PS2 - could this styling be activated/deactivated by users with a Preferences toggle or it's better to just make it visible to all users? (I guess pro users would always be able to go to the CSS files and make custom changes...)

I don't think it is needed. But...

pgilfernandez:
> Sadly, it isn't that easy, because they aren't the same UI widget type. Parameters, Layers and Sets Panels are GtkTreeView. Timetrack is a GtkDrawingArea.

Well, I've been playing with it around and I wasn't able to style it ala "alternate rows"... I've tried everything, but the main blocker I find is that the class you called `parameters` doesn't stack itself so that I could use `:nth-child(even)` and `:nth-child(odd)`. I mean, it works but all "parameter rows" are "even" ones... That is:
```
row 1 (usually Z Depth) -------- .parameters even
row 2 (usually Opacity)-------- .parameters even
row 3 (usually Blend Method)-- .parameters even
row 4 ------------------------- .parameters even
row 5------------------------- .parameters even
...
```

What it would be expected is:
```
row 1 (usually Z Depth) -------- .parameters even
row 2 (usually Opacity)-------- .parameters odd
row 3 (usually Blend Method)-- .parameters even
row 4 ------------------------- .parameters odd
row 5------------------------- .parameters even
...
```

without this, the only solution I find is the following (a line at the top/bottom, it's again compatible with light/dark themes):
screenshot

> Maybe somehow this color operations help? https://docs.gtk.org/gtk3/css-overview.html#an-example-for-defining-colors

I explored the color operations but, or they don't work or they partially work... BTW, using RGBA as I did results in the same effect and it fully seems to work in Synfig so I don't find any problem using it.
This being said, the full code I used was:
```
#timetrack {
   border-top: 1px solid rgba(255,255,255,0.05);
   border-bottom: 1px solid rgba(0,0,0,0.05);
}

.odd {
   background-color: rgba(255,255,255,0.05);
}

.even {
   background-color: rgba(0,0,0,0.05);
}
```

So, I think it will need to manually hack it as you did with the Timetrack panel...

> I prefer another commit/PR. This one provide the functionality/possibility only :)

OK, I will when there is something we find complete

pgilfernandez:
I forgot to post about this solution which is not perfect but fully works now: add a background image with a slight line in the middle.
screenshot2

It's not an alternate colors solution with probably is better but it makes "visually following the line" much easier than just a white background...

rodolforg:
> > Sadly, it isn't that easy, because they aren't the same UI widget type. Parameters, Layers and Sets Panels are GtkTreeView. Timetrack is a GtkDrawingArea.
> 
> Well, I've been playing with it around and I wasn't able to style it ala "alternate rows"... I've tried everything, but the main blocker I find is that the class you called `parameters` doesn't stack itself so that I could use `:nth-child(even)` and `:nth-child(odd)`. I mean, it works but all "parameter rows" are "even" ones... That is:
> 
> ```
> row 1 (usually Z Depth) -------- .parameters even
> row 2 (usually Opacity)-------- .parameters even
> row 3 (usually Blend Method)-- .parameters even
> row 4 ------------------------- .parameters even
> row 5------------------------- .parameters even
> ...
> ```
> 
> What it would be expected is:
> 
> ```
> row 1 (usually Z Depth) -------- .parameters even
> row 2 (usually Opacity)-------- .parameters odd
> row 3 (usually Blend Method)-- .parameters even
> row 4 ------------------------- .parameters odd
> row 5------------------------- .parameters even
> ...
> ```
> 
> without this, the only solution I find is the following (a line at the top/bottom, it's again compatible with light/dark themes): screenshot
> 
> > Maybe somehow this color operations help? https://docs.gtk.org/gtk3/css-overview.html#an-example-for-defining-colors
> 
> I explored the color operations but, or they don't work or they partially work... BTW, using RGBA as I did results in the same effect and it fully seems to work in Synfig so I don't find any problem using it. This being said, the full code I used was:
> 
> ```
> #timetrack {
>    border-top: 1px solid rgba(255,255,255,0.05);
>    border-bottom: 1px solid rgba(0,0,0,0.05);
> }
> 
> .odd {
>    background-color: rgba(255,255,255,0.05);
> }
> 
> .even {
>    background-color: rgba(0,0,0,0.05);
> }
> ```
> 
> So, I think it will need to manually hack it as you did with the Timetrack panel...
> 
> > I prefer another commit/PR. This one provide the functionality/possibility only :)
> 
> OK, I will when there is something we find complete

That's a GTK issue, as far as I understood and I 'mentioned' here: https://github.com/synfig/synfig/pull/3230#issuecomment-1744086654



> I explored the color operations but, or they don't work or they partially work...

Ugh. I thought it could help to make easier light/dark themes.



> So, I think it will need to manually hack it as you did with the Timetrack panel...

Sadly it seems to be impossible, unless we hack the Gtk code... and that's really annoying :P They won't fix it upstream, as they are focused on Gtk4, that has so many differences...

ice0 (approved):

ice0:
Merged. Thank you!

rodolforg:
@pgilfernandez guess what? I got a trick to do what you asked for ;)

I have a question for you: among both options below, what do you think is the best one?
1. `selected` color is the same no matter what row it is (so `selected` state is not affected by odd/even selectors)
2. `selected` may depend on even/odd row

pgilfernandez:
Great news! I can't wait to see it.

> I have a question for you: among both options below, what do you think is the best one?
>

Ummm, both options are good:

1) it is the closest approach to how real alternate rows work as the selected row always takes the default selection color. I mean, is enough for our intentions here.

2) it would make styling more configurable in case it is needed in the future and you can always leave both selected rows (even and odd) with the same style as default, that is, looking as if option 1 were...

So, if what you are going to do is as easy/hard for either options I would go for number 2 and we have more flexibility for future.

=)

PR #3232, opened 2023-10-03
feat: widget Timetrack now can have its own color for selected row

rodolforg:
You can provide it with CSS selector:
.timetrack:selected

If it doesn't exist (i.e. provides the same color as its background), it uses the same color of Parameters Panel.


ice0 (approved):

ice0:
Merged. Thank you!

PR #3071, opened 2023-05-07
refactor: move etl::handle to synfig-core

rodolforg:
Thank you for your service, ETL. At least its folder.

Farewell.

PR #1935, opened 2020-12-19
[ETL/Core] Added locks to avoid data-races

ice0:

PR #3233, opened 2023-10-03
refactor: clean etl::handle and alike

rodolforg:
Some commits of #3071 , but the safe ones. Those that do not add move constructors, for example.
I decided to split it. First do this PR, then a PR to move file ETL/_handle.h to synfig/handle.h.
Then the missing changes (move constructors, move assignment operators and #1935 )


ice0 (approved):

ice0:
Merged. Thank you!

rodolforg:
Hey :) Do you know what was the problem with macOS?

ice0:
> Hey :) Do you know what was the problem with macOS?

Hi! 
I didn't notice the error, so it looked like it just aborted after 30 minutes of run.
There must be a 6 hour limit to run, so this looks like some kind of issue on the GitHub side.

ice0:
@rodolforg 

P.S. Hmm, once I noticed a long time generation of the `action_export_icon.png` file. 
I’m not sure that this is where the problem lies, since the build is multi-thread, but we need to keep an eye on it.

rodolforg:
Well, the only commit I see that could be a problem is that about `noexcept`... https://github.com/synfig/synfig/pull/3233/commits/99f12214119ae7d78202c2e4640cf72a511ffb23

ice0:
> Well, the only commit I see that could be a problem is that about `noexcept`... [99f1221](https://github.com/synfig/synfig/commit/99f12214119ae7d78202c2e4640cf72a511ffb23)

Interesting. Anyway, if there is a problem here, we will find and fix it.

PR #3240, opened 2023-10-15
feat: add "invert" parameter to ChromaKey layer

### LINKED ISSUE #3239: Feature request: Chrome Key layer: Make possible to invert color match
**Description**

Okay, so I checked out the "Chroma Key" layer... it's good.
And I know it's a recent addition that didn't get any wide spread testing, but I already have an improvement for it in mind.

Can we add an 'Invert' parameter for this layer? So instead of making key color transparent it will make all other colors transparent. With this, you can isolate parts of you composition for manual tweaking in external editor. Think about outlines. Let's assume they have black color, you set key color in "Chroma Key" to black, hit "Invert" and now only your outlines are visible.

This way you can render them separately from everything else and apply some filters or maybe process them manually with a tablet which maybe desired since vector lines are generated mathematically and have this "uniform" look. Later you just compose everything in a video editor.

So what do you think? Does someone else think it's useful beside me?

rodolforg:
fix #3239

### DIFF
--- a/synfig-core/src/modules/mod_filter/chromakey.cpp
+++ b/synfig-core/src/modules/mod_filter/chromakey.cpp
@@ -71,7 +71,8 @@ TaskChromaKey::is_transparent() const
 TaskChromaKey::TaskChromaKey()
 	: lower_bound(0.1),
 	  upper_bound(0.1),
-	  desaturate(true)
+	  desaturate(true),
+	  invert(false)
 {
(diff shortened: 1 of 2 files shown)


rodolforg:
(rebased)

ice0 (approved):

ice0:
Merged. Thank you!

PR #3243, opened 2023-10-16
fix: column 'name' of Layers panel should be the largest one

rodolforg:
and not the Z Depth...

From:
image
To:
image


ice0 (approved):

ice0:
Merged. Thank you!

PR #3242, opened 2023-10-16
feat: Layer and Parameter treeviews can now be styled for even/odd rows

rodolforg:
Just like Widget_Timetrack (done in #3231 219c2149 )

PR #3245, opened 2023-10-16
feat: add Layers, Parameters and Timetrack treeviews styles for even/odd rows

pgilfernandez:
Following PR https://github.com/synfig/synfig/pull/3242 here are the proposed styles for having alternate even/odd rows in Synfig.

These styles are compatible with both light/dark themes:

parameters_and_timetrack

layers

Cheers


ice0 (approved):

ice0:
Merged. Thank you!

PR #3264, opened 2023-11-04
feat: allow to fix broken links on load

rodolforg:
Currently, Synfig doesn't load a document if something is linked to a valuenode or canvas of a external file, and this file is not found. This usually happens when user renames, moves or deletes those files.

This PR allows Synfig Studio to track those missing files and presents a dialog where user can point the missing files to their new locations or names. If user deleted the file, however, 'nothing' can be done except creating a new file with the required library items...

The Synfig CLI is not affected by this feature/PR.

### DIFF
--- a/synfig-core/src/synfig/canvas.cpp
+++ b/synfig-core/src/synfig/canvas.cpp
@@ -55,7 +55,7 @@
 
 using namespace synfig;
 
-namespace synfig { extern Canvas::Handle open_canvas_as(const FileSystem::Identifier &identifier, const String &as, String &errors, String &warnings); };
+
 
 /* === M A C R O S ========================================================= */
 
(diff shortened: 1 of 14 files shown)


rodolforg:
I believe this is now properly working.

TODO:
* [ ] move new classes in canvas.h to a proper file?
* [ ] add docs to those new classes
* [ ] create test files?

rodolforg:
Well, it seems to be working...
Could you test @Svarov-RZM @morevnaproject ?

rodolforg:
Some example files (open the ones named "circle-..."
[broken-link-tests.zip](https://github.com/user-attachments/files/30625514/broken-link-tests.zip)

Svarov-RZM:
Hey, thanks for tackling some heavy issues!
I promise to test it on next weekends (very busy during work days) but I am not sure I am the one who should handle the code review...

Svarov-RZM:
This works well. I tried specifying files with spaces and non-latin characters in them to try and mess up the algorithm but all's good. One thing I can nitpick on is that file select window doesn't show all broken files immediately, only one at a time which makes the re-linking process a bit tedious if there are a lot of missing files.

rodolforg:
Can you give an example so I can understand the case?
@svarov-rzm

Svarov-RZM:
> Can you give an example so I can understand the case?

Sure, here:
[BrokenLinks.zip](https://github.com/user-attachments/files/30885211/BrokenLinks.zip)

Unpack and open BrokenLinks.sif. It's missing 4 files, 1..4.sif. The dialog window asks you to only pick one missing file at a time (you need to press OK for the next one to show up) instead of forming a list of all missing files.

rodolforg:
Got it.
I tried to check how could I fix it, but it seems to be more complicated that I want to deal with :(

Svarov-RZM:
> I tried to check how could I fix it, but it seems to be more complicated that I want to deal with :(

The issue I described isn't a big deal, so don't worry. It's much better than, pardon my French, users having the need to masturbate the XML structure inside the sif file when there are some missing files. I think this PR is ready for review but personally I don't know if I can handle the code review myself.

rodolforg:
Thank you for testing anyway.
Sadly ice0 is absent for a long time :(

Svarov-RZM:
> Sadly ice0 is absent for a long time :(

Yeah, people have lives I guess... I do too and so do you but look at us, we carved out a bit of time for Synfig.
Ah, sorry, I feel a bit bitter. Synfig has been always struggling with development, it's like a curse.

PR #3265, opened 2023-11-05
fix: unable to open file paths longer than 23 characters on Windows

mosasauridae:
Attempting to open a file on Windows whose full path is longer than 23 characters shows an error message "Unable to load [first 23 characters of filepath]".

Looks like this started happening after fc26742, though I'm not sure why it wasn't happening before. That version used the same `sizeof` logic except on a `char*` (size 8) instead of a `std::vector<wchar_t>` (size 24) so I would have thought the issue would have happened in that version as well. Either way I guess.

`sizeof` is a doozy of an operator.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -3546,7 +3546,7 @@ App::open(filesystem::Path filename, /* std::string as, */ synfig::FileContainer
 	std::vector<wchar_t> long_name;
 	long_name.resize(buf_size);
 	long_name[0] = 0;
-	if (GetLongPathNameW(filename.c_str(), long_name.data(), sizeof(long_name)) == 0)
+	if (GetLongPathNameW(filename.c_str(), long_name.data(), long_name.size()) == 0)
 		;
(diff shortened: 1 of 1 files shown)


synfig-studio/src/gui/app.cpp
```diff
@@ -3546,7 +3546,7 @@ App::open(filesystem::Path filename, /* std::string as, */ synfig::FileContainer
 	std::vector<wchar_t> long_name;
 	long_name.resize(buf_size);
 	long_name[0] = 0;
-	if (GetLongPathNameW(filename.c_str(), long_name.data(), sizeof(long_name)) == 0)
+	if (GetLongPathNameW(filename.c_str(), long_name.data(), long_name.size()) == 0)
```
rodolforg, line 3549: It's correct.
According [docs](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlongpathnamew), the third argument is:

> [in] cchBuffer
> 
> The size of the buffer lpszLongPath points to, in TCHARs.

rodolforg (approved):

rodolforg:
Merged. Thank you!

PR #3268, opened 2023-11-08
fix: mouse clicks while dragging ducks prevent DRAG from resetting

### LINKED ISSUE #1377: Press ESC key or right-click should cancel handle dragging
All

**Issue description**:
If user select start one or more vertex handles and starts to drag them, he can't cancel the dragging.

If user presses the right mouse button, a context menu is shown and after dismiss the menu, the handles float following the mouse pointer and disappear , but do nothing, when user left-click.

Afterdie:
closes #1377

### DIFF
--- a/synfig-studio/src/gui/workarea.cpp
+++ b/synfig-studio/src/gui/workarea.cpp
@@ -1129,8 +1129,7 @@ WorkArea::on_drawing_area_event(GdkEvent *event)
 		}
 
 		assert(curr_input_device);
-			
-
+		
 		if(std::isnan(x) || std::isnan(y) || std::isnan(p))
 			return false;
 
@@ -1181,7 +1180,7 @@ WorkArea::on_drawing_area_event(GdkEvent *event)
 					}
 				}
 			}
-			if(guide_highlighted){	
+			if(guide_highlighted){
 				guide_dialog.show();
 				guide_dialog.set_current_guide(curr_guide);
 				return true;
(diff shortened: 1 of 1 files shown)


Afterdie:
what is the expected behavior when user middle mouse clicks a duck?

synfig-studio/src/gui/workarea.cpp
```diff
@@ -1319,11 +1318,10 @@ WorkArea::on_drawing_area_event(GdkEvent *event)
 			break;
 		}
 		case 2:	{ // Attempt to drag and move the window
-			if (Duck::Handle duck = find_duck(mouse_pos, radius))
-				duck->signal_user_click(1)();
-			else
-			if(Bezier::Handle bezier = find_bezier(mouse_pos, radius, &bezier_click_pos))
-				bezier->signal_user_click(1)(bezier_click_pos);
```
rodolforg, line 1326: Why do you delete those lines?

Afterdie: this code was responsible for middle mouse clicking ducks, but it doesnt really need to interact so i though it should be removed

rodolforg: Oh sorry, I thought you were talking about the middle button effect while dragging ducks...

I think we should keep as it is, but add your code before this one.

Afterdie:
@rodolforg please have a look

ice0:
@Afterdie The code looks good to me, the only suggestion I can make is since the code is identical in both places, maybe extract it into a private method?

Maybe `cancelDragOnMouseButtonPress()` or something like that? This will allow other users to clearly understand what this code does.

Afterdie:
> @Afterdie The code looks good to me, the only suggestion I can make is since the code is identical in both places, maybe extract it into a private method?
> 
> Maybe `cancelDragOnMouseButtonPress()` or something like that? This will allow other users to clearly understand what this code does.

sure, Ive made the changes

ice0 (approved):

PR #3276, opened 2023-11-19
fix: units in Tool Options panel didn't follow app settings (regression)

### LINKED ISSUE #3274: Units in Tool Options Panel Does Not Change with Preferences (It's Stuck on 'px')
**Describe the bug**
Changing the 'Unit System' in Preferences menu does not affect the unit in Tool Options, regardless of which one you pick, it will keep using 'px' as its unit. Oddly enough, the 'Parameter' panel works correctly.

**To Reproduce**
1. Open the menu Carat (arrow at the top left of the workarea) OR right-click on your workarea to access the menu, then go to 'Edit' -> 'Preferences'
2. Open the 'System' menu within 'Preferences', then on 'Unit System' change it to anything Other Than 'Pixels'
3. Click on any tool (for example, 'Spline Tool') and open the 'Tool Options' panel
4. You should be able to notice that it still uses 'px' as its unit

**Expected behavior**
The 'Tool Options' panel should uses the unit that has been chosen from the 'Preferences' menu (or if it was selected while changing the unit, should change immediately after picking another tool). At least that's how it works on the current stable version (1.4.4)

**Screenshots**
unevenunitbug2

**System information:**
 - OS: Windows
 - OS version: Windows 10 Home Edition
 - I couldn't find the commit ID, but the artfact on AppVeyor came with the following name 'SynfigStudio-1.5.2-testing-2023-11-06-win64-62db8.zip'  (the one posted on 'Weekly Reports' topic here: https://forums.synfig.org/t/weekly-reports-2023/13748/48)

**Additional comments**
I found this first while testing a slightly older Master ver. (on AppVeyor's artifact it was called 'SynfigStudio-1.5.2-testing-2023-11-06-win64-1113a.zip'). But I decided to test it again on a more recent testing version (at least the one I could find). And just to make sure it wasn't a fluke, I tested it after temporarily removing my ver 1.4.4 settings folder. The issue still persist ._.v
unevenunitbug

rodolforg:
fix #3274

### DIFF
--- a/synfig-core/src/synfig/distance.cpp
+++ b/synfig-core/src/synfig/distance.cpp
@@ -123,6 +123,15 @@ Distance::convert(Distance::System target, const RendDesc& rend_desc)
 	system_=target;
 }
 
+Distance
+Distance::as(System target, const RendDesc& rend_desc) const
+{
+	Distance d;
+	d.value_ = get(target, rend_desc);
+	d.system_ = target;
+	return d;
+}
+
 Real
 Distance::get(Distance::System target, const RendDesc& rend_desc)const
 {
--- a/synfig-core/src/synfig/distance.h
+++ b/synfig-core/src/synfig/distance.h
@@ -91,6 +91,7 @@ public:
 	Real get(System system, const RendDesc& rend_desc)const;
 
 	void convert(System system, const RendDesc& rend_desc);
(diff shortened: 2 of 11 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

PR #3280, opened 2023-11-25
fix: dropped image file doesn't load (broken conversion URI to file path on Win32)

### LINKED ISSUE #3279: Image drag'n'drop not working on Windows
Hi!
We are testing o Windows 11 latest build made on Appveyor.
It doesn't import image files through drag-n-drop.
Here's an error:
image_2023-11-21_13-08-44

At the same time, importing same image through "File" -> "Import" is working fine.

rodolforg:
It doesn't strip slash character '/' before drive letter

culprit: f31dd3a79b10d8e14d1f09fa85bccfd53928e855 (#3229)

fix #3279

### DIFF
--- a/synfig-core/src/synfig/filesystem_path.cpp
+++ b/synfig-core/src/synfig/filesystem_path.cpp
@@ -40,6 +40,7 @@
 
 #include <glibmm/miscutils.h>
 
+#include <synfig/general.h>
 #include <synfig/os.h>
 
 # ifdef _WIN32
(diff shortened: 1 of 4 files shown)


ice0:
It looks like you could just use `Glib::filename_from_uri` here?

ice0 (approved):

rodolforg:
I avoided t here because [docs say](https://docs.gtk.org/glib/func.filename_from_uri.html):

> Converts an escaped ASCII-encoded URI to a local filename **in the encoding used for filenames**.

morevnaproject:
Thank you!

PR #3287, opened 2023-12-02
fix: use u8_str() instead of c_str() for filesystem::Path

rodolforg:
as c_str() means native encoding, not UTF-8.

### DIFF
--- a/synfig-core/src/synfig/listimporter.cpp
+++ b/synfig-core/src/synfig/listimporter.cpp
@@ -74,7 +74,7 @@ Importer(identifier)
 	// TODO(ice0): There is no way to report about error to GUI. We can't
 	//  throw error from constructor (bad practice), and can't return error code.
 	if(!stream)	{
-		synfig::error("Unable to open " + identifier.filename.u8string());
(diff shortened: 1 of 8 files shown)


synfig-core/src/synfig/listimporter.cpp
```diff
@@ -178,8 +178,9 @@ ListImporter::get_sub_importer(const RendDesc &renddesc, Time time, ProgressCall
 	Importer::Handle importer(Importer::open(FileSystem::Identifier(FileSystemNative::instance(), filename)));
 	if(!importer)
 	{
-		if(cb)cb->error(_("Unable to open ")+filename);
-		else synfig::error(_("Unable to open ")+filename);
+		std::string error_msg = strprintf(_("Unable to open %s"), filename.c_str());
```
ice0, line 181: Is that correct? I mean, maybe there should be `u8_str()`?

rodolforg: `filename` here is a `synfig::String` variable, not `synfig::filesystem::Path`. So, it doesn't have the `u8_str()` method.
I don't know why I submitted this change XD

ice0:
Merged. Thank you!

PR #3290, opened 2023-12-09
build(autotools): suppress warnings from libxml++

ice0:
There are many warnings about `auto_ptr`, something like:

```
/usr/include/libxml++-2.6/libxml++/parsers/saxparser.h:224:8: warning: ‘template<class> class std::auto_ptr’ is deprecated [-Wdeprecated-declarations]
  224 |   std::auto_ptr<_xmlSAXHandler> sax_handler_;
```

### DIFF
--- a/synfig-core/configure.ac
+++ b/synfig-core/configure.ac
@@ -13,6 +13,18 @@ AC_CANONICAL_HOST
 AM_INIT_AUTOMAKE([nostdinc subdir-objects])
 AM_MAINTAINER_MODE
 
+dnl SUPPRESSED_CPPFLAGS=SUPPRESS_WARNINGS([$SOME_CPPFLAGS])
+dnl Replace -I with -isystem in $SOME_CPPFLAGS to suppress warnings from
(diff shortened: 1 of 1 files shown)


ice0:
Output lines reduced from 6235 to 5569 (-660 lines of warnings).

P.S. synfig-studio still has a warning because I haven't found where the incorrect values are coming from.

rodolforg (approved):
I was thinking about to write a simple wrapper to libxml2 directly to avoid those warnings...
libxmlpp depends on Gnome Glib or, in 5.0 API, C++17.
We use so few classes and methods of the full libxmlpp API...

ice0:
> P.S. synfig-studio still has a warning because I haven't found where the incorrect values are coming from.

These values are taken from the `synfig.pc` dependencies (Thanks to @morevnaproject for pointing on this).

> I was thinking about to write a simple wrapper to libxml2 directly to avoid those warnings...

Agree. I also thought about this way.

PR #3295, opened 2023-12-20
fix: Adjust minimum panel size for better positioning

neesaaa:


neesaaa:
I will redo it tomorrow branches bug

PR #2777, opened 2022-08-03
feat: added the ability to rotate guides in the canvas

mohamedAdhamc:
Fixes #2593 

This is the first step 
- now the rulers are able to rotate by holding control while moving the cursor (p.s vertical rulers have the center of rotation being half of their height, and horizontal have it as half their width which I think makes sense here if not though please let me know)

the main things left:
- handle moving the "rotated" rulers. 
- also if it would be required I could make a dialog which is accessed by right clicking  the guide to manually set the rotation.
- clean ups + optimizations

this is how the rotation works as of now:

https://user-images.githubusercontent.com/100296264/182651640-c854d38f-4893-4d37-8d19-9551989f5fa0.mp4

PR #3303, opened 2024-01-15
fix: guide lines ignored on loading from old files

rodolforg:
Due to changes in b172e3771a131dbd26879aae334b84ac26bcb6e6 (#2777)

### DIFF
--- a/synfig-core/src/synfig/loadcanvas.cpp
+++ b/synfig-core/src/synfig/loadcanvas.cpp
@@ -103,6 +103,35 @@ std::set<FileSystem::Identifier> CanvasParser::loading_;
 
 /* === P R O C E D U R E S ================================================= */
 
+// Guide lines storage changed:
+// Before:
+//   guide_x -> list of X positions of vertical guide lines (separated by spaces)
+//   guide_y -> list of Y positions of horizontal guide lines (separated by spaces)
+// Now:
+//   guide -> list of X*Y*radian-angle coordinates of guide lines (data from same guide is separated by star *) (items separated by spaces)
+static void
+upgrade_guide_metadata(std::string& content, const std::string& metadata, bool is_x_guide)
+{
+	std::string new_content;
+	std::istringstream iss(content);
+	std::string item;
+	while (std::getline(iss, item, ' ')) {
+		if (!item.empty()) {
+			if (is_x_guide) {
+				// X * 0 * PI/2 radians -> vertical guide line
+				new_content += item + "*0*1.57079633 ";
+			} else {
+				// 0 * Y * 0 radians -> horizontal guide line
+				new_content += "0*" + item + "*0 ";
+			}
+		}
+	}
+	// new_content already has a blank space appended
+	content = new_content + metadata;
+	if (content.back() == ' ')
+		content.pop_back();
+}
+
 OpenCanvasMap& synfig::get_open_canvas_map()
 {
 	static OpenCanvasMap open_canvas_map_;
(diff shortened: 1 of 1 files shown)


synfig-core/src/synfig/loadcanvas.cpp
```diff
@@ -3359,7 +3360,46 @@ CanvasParser::parse_canvas(xmlpp::Element *element,Canvas::Handle parent,bool in
...
-				canvas->set_meta_data(child->get_attribute("name")->get_value(),content);
+
+				// Guide lines storage changed:
+				// Before:
+				//   guide_x -> list of X positions of vertical guide lines (separated by spaces)
+				//   guide_y -> list of Y positions of horizontal guide lines (separated by spaces)
+				// Now:
+				//   guide -> list of X*Y*angle coordinates of guide lines (data from same guide is separated by star *) (items separated by spaces)
+				if (meta_name == "guide_x") {
+					std::string new_content;
```
ice0, line 3371: I suggest to extract this code to the separate function, something like this:
```
void upgrade_guide_metadata(std::string& content, const std::string& metadata, bool is_x_guide) {
	std::string new_content;
	std::istringstream iss(content);
	std::string item;
	while (std::getline(iss, item, ' ')) {
		if (!item.empty()) {
			if (is_x_guide) {
				new_content += item + "*0*1.57079633 ";
			} else {
				new_content += "0*" + item + "*0 ";
			}
		}
	}

	content = new_content + " " + metadata;
	if (content.back() == ' ')
		content.pop_back();

}
```

synfig-core/src/synfig/loadcanvas.cpp
```diff
@@ -3359,7 +3360,46 @@ CanvasParser::parse_canvas(xmlpp::Element *element,Canvas::Handle parent,bool in
...
+					std::istringstream iss(content);
+					std::string item;
+					while (std::getline(iss, item, ' ')) {
+						if (!item.empty()) {
+							new_content += item + "*0*1.57079633 ";
```
ice0, line 3376: `1.57079633` - magic number.
Please add explanation in comment.

synfig-core/src/synfig/loadcanvas.cpp
```diff
@@ -3359,7 +3360,46 @@ CanvasParser::parse_canvas(xmlpp::Element *element,Canvas::Handle parent,bool in
...
+					std::string new_content;
+					std::istringstream iss(content);
+					std::string item;
+					while (std::getline(iss, item, ' ')) {
+						if (!item.empty()) {
+							new_content += "0*" + item + "*0 ";
+						}
+					}
+
+					content = new_content + " " +canvas->get_meta_data("guide");
```
ice0, line 3396: > content = new_content + " " +canvas->get_meta_data("guide");

Do we really need to add a space here? 
If we made the conversion, then we already have them, if not, then it seems like we don’t need them.

rodolforg: You're right! Done

synfig-core/src/synfig/loadcanvas.cpp
rodolforg: Done!

synfig-core/src/synfig/loadcanvas.cpp
rodolforg: Done!

ice0 (approved):

ice0:
> I now wonder if this code should be moved to `App::open*()`, as this (any?) metadata is only used by GUI.

I think it's better to keep all the conversion logic in one place. But if there will be a lot of conversions just for the GUI, then yes, it makes sense.

ice0:
Merged. Thank you!

PR #3306, opened 2024-01-16
refactor: port layer to Cobra renderer: Noise gradient

rodolforg:


rodolforg:
Does supersample parameter any effect in original implementation? v1.0.2 or 1.4.x or 1.5.1 ...
Here it has none...

ice0:
> Does supersample parameter any effect in original implementation? v1.0.2 or 1.4.x or 1.5.1 ...
> Here it has none...

That's strange. It did some calculations, but I didn't see any difference in the image.
```
if(super_sample&&pixel_size)
{
...
}
```

rodolforg:
Neither did I...

Svarov-RZM:
> Does supersample parameter any effect in original implementation?

It does in all the versions. You just need to create a gradient where such effect would be visible, like in this case:
NG-SuperSample

On the right side the gradient was generated with "Super Sampling" enabled and you can clearly see it's anti-aliased. 

Do I personally think this parameter is needed? No. We already have a dedicated layer for this purpose called "Super Sample" which provides the user with more options for anti-aliasing.

rodolforg:
And does the PR implementation work equally?

Svarov-RZM:
> And does the PR implementation work equally?

No, toggling the parameter has no effect.
Here's a test file: 
[NG-SuperSample.zip](https://github.com/synfig/synfig/files/14000073/NG-SuperSample.zip)

rodolforg:
I believe now I reproduce it. :) @Svarov-RZM @ice0

Svarov-RZM:
> I believe now I reproduce the supersample effect. :)

Yeah, I confirm. It's working properly now.
Thank you for continuing on porting layers to the new engine, really appreciate it.
If you need any testing, just ping me.

rodolforg:
@Svarov-RZM well, I try to port as long as I understand the engine and its code. :)
I don't check if it performed better, though.

Svarov-RZM:
> I don't check if it performed better, though.

Don't worry, it will perform better, since algorithm stays the same but the unnecessary wrapper is removed. Even if, by some convoluted reason, it won't, the performance can be addressed later after porting.

What's more important is stability. Currently layers such as Twirl, Curve Warp and Spherize are close to being unusable. They crash on zoom in (so annoying!) and often hang Synfig if you use it on complex work. But anyway, there's no hurry, when something is ready we'll test it and fix whatever issues come along.

Svarov-RZM:
Hey, I got closer to the root of the problem! It is not in the implementation!
I was testing this "Noise Gradient" for hours mixing different options and frankly, it produces the correct results except for that blasted shift. But guess what, the shift is not present when you render the project to .bmp instead of .png.

So the problem is somewhere in png render. What handles it? ImageMagick? I remember there was a problem with it. Anyway, here's the most simple test case:
[TestNoiseGr.zip](https://github.com/user-attachments/files/18051038/TestNoiseGr.zip)

Open it, there are two pairs: 0TestNoiseGr.png - 0TestNoiseGr102.png and 1TestNoiseGr.bmp - 1TestNoiseGr102.bmp. In image viewer, switch between png pair and you see there's a small shift. Switch between bmp pair and the result is identical.

The only thing I don't understand is why Preview also gives improper results... Does it somehow rely on png/ImageMagick?

rodolforg:
> But guess what, the shift is not present when you render the project to .bmp instead of .png.

@ice0 @morevnaproject  o.O

morevnaproject:
I am afraid that it's bmp export module gives offset on all versions.
Unfortunately, I didn't managed to check this, because my Synfig 1.0.2 produces unradable bmp file...
@Svarov-RZM Which build of Synfig 1.0.2 you used to render BMP file?

Here I am attaching test file by @Svarov-RZM , rendered to PNG, JPEG, BMP. 
PNG has several versions, because it is rendered using different modules (imagemagick and png module).
[TestNoiseGr-tests2.zip](https://github.com/user-attachments/files/18139923/TestNoiseGr-tests2.zip)

morevnaproject:
@Svarov-RZM Can you please render your test file using different PNG modules?

zzz


And please also render one version of BMP. It would be nice to have JPEG version too.

Svarov-RZM:
> I am afraid that it's bmp export module gives offset on all versions.

Yeah, you're right. I checked and bmp export is indeed broken. I got hopeful because bmp render between 1.0.2 and current branch gave identical (but improper) results.

> Which build of Synfig 1.0.2 you used to render BMP file?

It's official version, 1.0.2, 64 bit running on Windows 10. It's not really important now though, the problem is somewhere else.

> Can you please render your test file using different PNG modules?

I did and it gives identical results to the regular 'png' render. The target render is alright. Appears it's just bmp render that's broken, we can safely ignore it.

Uff... Let's start all over. Screw the original example, I made a better one (archive also includes renders to different types of files):
[TestNoiseGr3.zip](https://github.com/user-attachments/files/18161614/TestNoiseGr3.zip)

The "Noise Gradient" blends onto region (named "SURFACE") and behind it there's a region of the same size but different color (cyan). Both regions have anti-aliasing disabled, so the result should be pixel sharp and any artifacts would be immediately noticed.

Here's combined picture of all major versions handling this file (regular png render):
TestNoiseGr3-Combined

Top row from left to right: 0.64.3, 1.0.2, 1.2.2-18.09.14-win64-286f1.
All produce identical results.

Bottom row from left to right: 1.4.4-2023.03.30-win64-b8d62, 1.5.3-2024.08.23, this branch.

You can see that in 1.4.4 and 1.5.2 a bit of cyan shows - that's underlying "BKG" region meaning that not only gradient is shifted but also "SURFACE" region or the whole group layer is shifted. In current branch though, only gradient is shifted since you can't see the cyan border anymore. But that's only because in this branch "Noise Gradient" is ported to Cobra, if you replace it with "Curve Gradient" which isn't ported, you'll get the same cyan border like in 1.4.4-1.5.2 versions, so the same problem introduced before 1.4.4 release is still here, it just shows differently depending on whether gradient is ported to new engine or not.

With all that being said, the problem seems to be introduced somewhere between 1.2.2 and 1.4.4 releases and never got properly fixed. Perhaps if we find the commit that introduced the problem it will give us some clue on what's happening.

rodolforg:
So, is the current PR equal to 1.5.x and 1.4.x, right?

Svarov-RZM:
> So, is the current PR equal to 1.5.x and 1.4.x, right?

In terms of rendering? Almost. In the post above you can see circles besides the shift have cyan border (when rendered in 1.4..4 and 1.5.3), in this PR shift remains but cyan border disappears.

rodolforg:
@Svarov-RZM  Something changed from 1.4.4 to 1.5.3 :(
See the white "donut" near the center (a bit top left) of each circle of second row of the image you posted.
It was shrink in recent versions... o.O 
image

Svarov-RZM:
> See the white "donut" near the center... It was shrink in recent versions...

Oh, I see it! Good catch. I was too concerned about the shift to notice anything else.
It is very small though, like barely noticeable. Seems like the gradient was rendered at the lower resolution than the canvas, for example, instead of 270x270, the gradient was shrunk to 269x269 and then rendered.

I experimented a bit and came up with this example: 
[TestNoiseGr5.zip](https://github.com/user-attachments/files/18331021/TestNoiseGr5.zip)

It is a Noise Gradient that produces a single red square. If the resolution of Noise Gradient is indeed decreased, then current versions will stop rendering it earlier than older versions. Let's see if that's the case:
0TestNoiseGr5-Comb1

Rendered at 25x25 px, then upscaled in GIMP with no interpolation to better see the details. What we have here from left to right:
* 1.0.2 version - Our point of reference. This is the **lowest resolution** at which the red square is still visible, at 24x24 px it no longer renders. 1.2.2 version produces the same result
* 1.4.4 version - Shift is present, but gradient is not squeezed and still visible. You can see the square is bigger and dimmer, this is because anti-aliasing was introduced at some point
* 1.5.2 version - The shift is present and the red square is no longer visible which to me proves that the resolution of the gradient is indeed decreased by 1 px before rendering
* Current branch - Similar to 1.5.2 version with shift and decreased resolution, but defect with lower circle showing is no longer present

So yeah, something's mighty strange here... Perhaps this anti-aliasing is not intentional but just a side effect of some optimization? Maybe there's some kind of rounding error somewhere? I really don't know at this point...

rodolforg:
@Svarov-RZM I fixed the shift! :smiley:

rodolforg:
Please squash commits on merging. ;)

Svarov-RZM:
> I fixed the shift! 😃 

No freaking way! I won't believe this :)

I made some additional tests and yeah, you did fix the shift - well done! Hmm, but I thought shift was a global issue somewhere in the resampling procedure?

Alright, to make sure I am not crazy, I put some rotate-scale-skew transformation on Noise Gradient, some debug printf's in resample procedure and the thing is, resampling doesn't get called with Noise Gradient even with some transformations applied to it, so 1px shift here was unrelated to 1px shift due to resampling in, let's say, TaskBlur. Insanity.

rodolforg:
Sorry I didn't make myself clear.
After your investigation, I did mine and saw that those layers that don't have issues with resampling always _use_/_apply_ the transformation in its layer itself.

And that's what I did here and with XORPattern layer.

rodolforg:
Btw, I suppose the problem could have its roots here:
https://github.com/synfig/synfig/blob/b561d9f1f31bc86fadc4100ce2d9bb6eaec79f47/synfig-core/src/synfig/rendering/common/task/tasktransformation.cpp#L99-L129

(`Transformation::make_discrete_bounds()`)

morevnaproject:
@Zloi-KoteikA Please guide @MetaruHebi to run all tests on this PR.

Svarov-RZM:
> After your investigation, I did mine and saw that those layers that don't have issues with resampling always _use_/_apply_ the transformation in its layer itself.

Thanks for explanation, this does make things a bit clearer.

> I suppose the problem could have its roots here (Transformation::make_discrete_bounds()):

Yes, I thought so too initially. This procedure in case of TaskBlur extends coordinates to accommodate the blur size and then later resampling gets called. So in my opinion, the problem is either here (wrong coordinates calculation) or there's a bug in rendering->software->function->resample.cpp:resample() or resample_with_downscale().

I planned to compare how the same thing is implemented in 1.2.2 version, but gave up. I just wanted to concentrate on making my next animation...

rodolforg:
Strange. It seems supersample on/off does not have any visual effect in version 1.0.2...

Svarov-RZM:
> It seems changing supersample on/off does not have any visual effect in version 1.0.2...

It's only visible in the final result (when you render it). It's controlled by Quality parameter and only visible if you render at Quality <= 7.

rodolforg:
@Zloi-KoteikA @Svarov-RZM I finally fixed it.

And, by the way, I lost a week trying to understand why it became so gray with the port. I even (re)tried to build synfigstudio 1.0.2 to debug, but it was the gamma issue :P

Please use the attached file to test: it is the same file provided by Svarov, but saved in v1.0.2 . This way, current version get the proper gamma correction.
[NG-SuperSample-1.0.2.zip](https://github.com/user-attachments/files/20149261/NG-SuperSample-1.0.2.zip)

Svarov-RZM:
> I finally fixed it

Seems so, I couldn't find any problems.

> And, by the way, I lost a week trying to understand why it became so gray with the port... it was the gamma issue

Oof.  Gamma setting in Synfig was always confusing to me... Like if we set gamma to 2.2 shouldn't it make image darker overall? Instead it makes it lighter by some reason...

rodolforg:
Phew... Good to know.
So, when time comes, please _squash_ and merge ;)

PR #3307, opened 2024-01-16
refactor: port layer to Cobra renderer: Curve Gradient

rodolforg:


Svarov-RZM:
Hey @rodolforg 
I am on the path of testing and approving all of your Cobra ports. Did you see the issue with this port where it inserts a white pixel in the end of the spline (see my post above)? Do you have any idea what causes it?

rodolforg:
I'll check it.

rodolforg:
I just realize that version 1.4.1 (appImage) suffers the same issue. The investigation will take longer then :(

Versions 1.2.1 and 1.2.2 are OK.
image

Version 1.3.11 is completely broken.
image

Svarov-RZM:
> I just realize that version 1.4.1 (appImage) suffers the same issue

Yeah, the Windows version - SynfigStudio-1.4.5-2024.05.19-win64-f4b8d have the same problem also.
But, what is interesting, is that development version - SynfigStudio-1.5.3-2024.08.23-win64-3b7c5 doesn't suffer from this issue...

rodolforg:
> But, what is interesting, is that development version - SynfigStudio-1.5.3-2024.08.23-win64-3b7c5 doesn't suffer from this issue..

Oh, indeed! Thank you. Now I have a better starting point.
I'm pretty sure it is something related to `Gradient`, that changed after 1.2.x to speed up rendering (`CompiledGradient`).
A corner case that is handled differently from `Gradient` to `CompiledGradient`...

rodolforg:
I did a `git bisect` and it seems this could be the guilty PR:
[ee9fa74b56373710ba3e99993ca0b2c969995d3d] fix: border computation for raster image with Transformation task (#3603)

To be clear: I'm not saying this is the reason of the problem, but probably what made it show up.

rodolforg:
That is what I found out after more investigation:
- The issue appears at zoom level of 100%. It disappears at different levels.
- The specific point where the white pixel appears is at (2.0 , -0.25), the same coordinates of vertex 2 of spline.
- The color is white because that is the color of the first color point of the gradient.
- The same happens at the other vertex (just move it into canvas area to check it out)
- In such specific cases, the vector variable `diff` of the computation (in `color_func()` or `get_color()` of CurveGradient layer .cpp file) is exactly (0, 0). Therefore, the `mag` var is an invalid value (`inf`) because it is the inverse of `diff` magnitude/length.
- Consequently, `diff` becomes invalidated too (`NaN`) and so is `dist` (and `supersample`)
https://github.com/synfig/synfig/blob/e02b254ff1c0759383bec7177a83fd03fbc8759f/synfig-core/src/modules/mod_gradient/curvegradient.cpp#L443-L446
- The Gradient/CompiledGradient class has to deal with such invalid values. It is indeed here the difference (1.2.x vs. >= 1.3.x).

Svarov-RZM:
OK, so the issue is not with the port, am I understanding correctly? I just don't quite get why the issue doesn't appear in the development version. Because gradient is rendered through accelerated_render, is that the case?

> * The issue appears at zoom level of 100%. It disappears at different levels.

Yeah, but when you render, you still get white pixel in the output.

> * The specific point where the white pixel appears is at (2.0 , -0.25)

I checked and it appears at other coordinates too. For example, 0.5, -2; -1, 1; 3, 0.5.
But if you alter the last coordinates (X and/or Y) just a little bit to something like 3, 0.49 - the issue disappears.

rodolforg:
> so the issue is not with the port, am I understanding correctly?

Exactly. You can check the current master branch.



> I just don't quite get why the issue doesn't appear in the development version. Because gradient is rendered through accelerated_render, is that the case?

No. Because of this commit ee9fa74b56373710ba3e99993ca0b2c969995d3d that was merged after the release.


> > * The issue appears at zoom level of 100%. It disappears at different levels.
> 
> Yeah, but when you render, you still get white pixel in the output.

Because it is rendered at 100% ;)


> I checked and it appears at other coordinates too. For example, 0.5, -2; -1, 1; 3, 0.5.

While rasterizing, the pixel surface is 'scanned' pixel-by-pixel. When it coincides exactly with the spline vertex coordinates, the white pixel appears. That is the same reason it doesn't appear before of that culprit commit (the unbalanced addition of two pixels on the surface) and when on different zoom levels.

By the way, I forgot to mention it is white because it is the first color point of the gradient.

rodolforg:
@Svarov-RZM I believe I have fixed now :)

Indeed the problem was how differently original Gradient and CompiledGradient work. The second one is faster than first one, but it misses when gradient is used for zigzag gradient - and that make the first gradient color wrong.

Svarov-RZM:
> I believe I have fixed now :)

All right, thank you! I'll give it one more test during the week and then merge.
Jeesh, there are so much regressions after 1.2.x release...

rodolforg:
> Jeesh, there are so much regressions after 1.2.x release...

Synfig is too flexible. Its layers have so many parameters... It is difficult to test everything and find those corner cases - and our small user base to test doesn't help.

Svarov-RZM:
> I believe I have fixed now :)

No, the issue persists, unfortunately.
The gradient doesn't have to be "zigzaged" though, here's another example:
CurveGrTest-PR

Test-case file:
[CurveGrTest2.zip](https://github.com/user-attachments/files/23913825/CurveGrTest2.zip)

"Zigzag" parameter is unchecked and white pixels are still visible in the start and end of outline.

If you're willing to look more into it, please tell, otherwise I'll merge the port and we will debug this case sometime in the future since it is a global issue.

rodolforg:
> No, the issue persists, unfortunately.

Oh, sorry I didn't test the version I committed - I did have some other debugging code changes here I prefer to left them off of this PR, but I didn't realize it make the previous push unfixed :P 



> The gradient doesn't have to be "zigzaged" though, here's another example:

Wow XD It helped me to fix this not-zigzag case then in this last push. :D

rodolforg:
last commit fixes a warning given by Codacy

Svarov-RZM:
> Oh, sorry I didn't test the version I committed - I did have some other debugging code changes here I prefer to left them off of this PR, but I didn't realize it make the previous push unfixed

Yes, this stupid white dot is gone, thank you!

Okay, so in this case we squash and merge because it was meant to be a single feature and all other commits along the way were addressing mistakes/improving the code quality, so better to squash them, right? You have to bear with me a little bit when it comes to git, I am a system administrator, not a developer, so I better double-check before I get the gist of it.

rodolforg:
In this case, it is better to not squash: the intermediate commits do different small steps (with refactoring).
Keep them is useful to a possible case, in the future, that we would need to check what commit broke something related to CurveGradient.

If this PR is (finally XD) ok, approve it and then merge it :)
As it is a PR with multiple commits, merge it before of single-commit PRs (including those with squash/merge) to keep commit history tree straight-forward.
As a side note, ice0 likes to wait the CI finish on the master branch before merging another PR there.

Svarov-RZM:
All right, my first merge, thank you @rodolforg 

I gotta say, I am terrible at C++ so I can't really review something as complex as this but I do promise to test everything thoroughly before merging.

rodolforg:
> I gotta say, I am terrible at C++ so I can't really review something as complex as this but I do promise to test everything thoroughly before merging.

We're glad you can help us this way ;) I, ice0 and others can review the C++ code, so when stuff are complicated like this you can ask/wait for our review too before merging.

About the merging, only one note: the commit message was ugly / non descriptive :( It would be better if it was the PR title.

rodolforg:
Yay! Maybe only 9 layers left! lol

1. ~~Bevel~~
2. InsideOut
3. Julia
4. MandelBrot
5. SphereDistort
6. Twirl
7. RadialBlur
8. Distort
9. Plant

ice0:
@Svarov-RZM 

Next time, please change merge commit message before merging:

Before:
image

After:
image

This merge commit messages is used to automatically generate the changelog, but it must follow the rules.
We use [Conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for git messages, and adding ` (#PR_number)` for merge commits.

Svarov-RZM:
> About the merging, only one note: the commit message was ugly / non descriptive :( It would be better if it was the PR title.

> Next time, please change merge commit message before merging:

Oh no, I goofed! Should have been obvious but I thought that having PR title in the extended description would be enough... Sorry, I got it and will do better next time!

PR #3310, opened 2024-01-17
feat: add missing scroll bar to Paste Layer dialog

### LINKED ISSUE #3309: There's a need for a scroll bar in a copy dialog
**Description**
All right, I am using master branch in full scale and it has this very handy dialog when you're copying layers with exported values from one file to another. The problem though, if there are A LOT of exported values then buttons get blasted off-screen like so:
CantSeeButtons

So I believe we need to wrap the contents of the window in a scroll bar so window's control would still be visible despite the amount of entries in it.

**To Reproduce**
Look, I can make a simple test file with a lot of exported values but I believe it's not really necessary (I am lazy right now, OK?).

Commit: 768188305412e231309cddf467c58bcc304863d9
It's a bit old, but I don't remember anything addressing the described issue.

rodolforg:
fix #3309


ice0 (approved):

ice0:
Merged. Thank you!

PR #3313, opened 2024-01-21
feat: add crash report for MS Windows

rodolforg:


rodolforg:
It works for those who installed [drmingw](https://github.com/jrfonseca/drmingw) (script `1-setup-windows-msys2.sh` should do it).
I didn't force-linked to drmingw, so, if not installed, Synfig Studio runs normally. It is dynamically linked to the library (as synfig modules are).

As I don't know CMake, I don't know how to change the install script to install the required DLLs (exchndl.dll, mgwhelp.dll, dbghelp.dll, dbgcore.dll, symsrv.dll, and symsrv.yes)
I didn't touch autotools either.

rodolforg:
How to get clean binaries from debug build, with a separate file with debug symbols:
```bash
# create a file with debug symbols
objcopy --only-keep-debug "${binary}" "${debugdir}/${debugfile}"
# strip binary and add reference to possible debug file
objcopy  --strip-debug --strip-unneeded --add-gnu-debuglink="${debugdir}/${debugfile}" "${binary}"
```
This way, we can ship the binary files, without any debug info, as it takes so much space...
And then, after receive a crash report file, we can recover the backtrace, by having the :
* debug binary; or
* the strip binary in a folder and the debugfile in the folder (if any) mentioned in `objcopy` from previous shell commands.
This will return the guilty lines:
```bash
# address is written in crash report
addr2line-drmingw -e ${binary} ${address}
```

PR #3314, opened 2024-01-25
fix: add fill-color instead of outline-color to palette

### LINKED ISSUE #3215: Adding colors to color palette
**Describe the bug**
With the current changes in eyedropper tool, it now sets the fill color to the picked color. But when you try to add color to the color palette, it adds the outline color into the palette, which causes inconvenience.

**Expected behavior**
Color palette should add fill color instead of outline color.

**Screenshots**
image

**System information:**
 - OS: Windows 10
 - OS version: 64 bit 21h2

KartikWatts:
This PR fixes the issue mentioned in https://github.com/synfig/synfig/issues/3215

Fill Color will be passed as the argument in the `add_color` method to add the fill color instead of the outline color to the palette.


ice0:
Merged. Thank you!

PR #3320, opened 2024-01-30
feat(SVG): naive support to Text element

rodolforg:
It doesn't handle <textPath>, text span (<tspan>), stroke, multiple lines, and per-character transformations (offset and rotation) autowrap, font selection among options...

But it does some work, I promise ;)

### DIFF
--- a/synfig-core/src/modules/mod_svg/svg_parser.cpp
+++ b/synfig-core/src/modules/mod_svg/svg_parser.cpp
@@ -77,6 +77,7 @@ static double getDimension(const String& ac, bool use_90_ppi = false);
 static float getRadian(float sexa);
 //string functions
 static std::vector<String> tokenize(const String& str,const String& delimiters);
+static String unquote(const String& str);
 
 static float get_inkscape_version(const xmlpp::Element* svgNodeElement);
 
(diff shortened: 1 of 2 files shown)


synfig-core/src/modules/mod_svg/svg_parser.cpp
```diff
@@ -2705,6 +2888,112 @@ get_inkscape_version(const xmlpp::Element* svgNodeElement)
...
+	{
+		return std::to_string(font_size) + "px";
+	}
+};
+
+const std::map<String, float> FontSize::absolute_size_names = {
+	{"xx-small", default_size * 3./5.},
+	{"x-small",  default_size * 3./4.},
+	{"small",    default_size * 8./9.},
+	{"medium",   default_size},
```
ice0, line 2991: Currently build fails on macOS with the error:
```
Undefined symbols for architecture x86_64:
  "FontSize::default_size", referenced from:
```

This is a little counter-intuitive, but this can be fixed by this:
```
{"medium",   default_size*1},
```

rodolforg: o.O
Done.

ice0 (approved):

ice0:
Merged. Thank you!

PR #3324, opened 2024-01-31
refactor: Bevel Layer port to Cobra render engine

### LINKED ISSUE #3197: Port Bevel Layer to Cobra render engine
It would be nice to port Bevel Layer to Cobra render engine. We use it a lot in recent work. ^__^

The Shade Layer is already ported - https://github.com/synfig/synfig/blob/master/synfig-core/src/modules/lyr_std/shade.cpp#L235

I think porting Bevel can be achieved in similar way (through `composite_fork_task` approach).

morevnaproject:
My attempts to port Bevel Layer to Cobra render engine.

P.S. This PR should be squashed to avoid junky commits.

fix #3197


morevnaproject:
It doesn't work yet.

PR #3327, opened 2024-02-01
feat :replacing redundant genv()

### LINKED ISSUE #2836: Remove redundant getenv() calls

neesaaa:
completing on pr #2826
fixes #2836

### DIFF
--- a/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
+++ b/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
@@ -27,6 +27,7 @@
 ** ========================================================================= */
 
 /* === H E A D E R S ======================================================= */
+#define LOGGING_ENABLED
 
 #define SYNFIG_LAYER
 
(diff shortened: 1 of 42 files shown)


neesaaa:
i dont know how but my build skipped some syntax errors and build 100% correctly at first ,   i handled it in the 2nd commit

ice0:
Hi!

There are two issues in this PR:
1)
```
std::string localdir = Glib::getenv("LOCALAPPDATA");
```
This environment variable (and some others) is required for correct work. They are not for debugging purposes, so no need to change them.

2)
```
std::string localdir = Glib::DEBUG_GETENV("LOCALAPPDATA");
```
This will not work if you undefine `LOGGING_ENABLED`.

neesaaa:
hi @ice0 
okay so every env variable required by the program i have to leave it as it was getenv()

neesaaa:
if u want  changes to be in  a new pr to see changes better  no problem .

PR #3328, opened 2024-02-04
perf: avoid canvas inspection when DockInfo is not shown

rodolforg:
On every mouse move on CanvasView, it reads what color is underneath the mouse pointer.

We can avoid calls to `Layer::get_color()` (that can parse all canvas layers) and their computation if DockInfo is not mapped.

'signal_map', in GTK words, is emitted 'when the widget is visible and all its parents up to the toplevel widget are also visible.'

DockInfo is not mapped when another tab/dock is being shown in its 'block' (DockBook), as it is in default workspace layout.


ice0 (approved):

ice0:
Merged. Thank you!

PR #3330, opened 2024-02-07
feat: add WebM video format to ffmpeg codecs list

pgilfernandez:
Hi,

I found out that Synfig could render WebM files (even with transparency) through FFMPEG but it wasn't clear enough for normal users so I just added VP8 and VP9 codecs to the FFMPEG "Available Video Codecs" list.

Now it's so easy to render this kind of videos for OBS or any other purpose ;-)


ice0:
Merged. Thank you!

PR #3332, opened 2024-02-10
fix: importing via magick++ for version >= 7.0.1

### LINKED ISSUE #3331: Import of raster work is broken in Master
**Description**
I was preparing to test rodolforg's Curve Warp port and before I got to it I noticed that import of raster images (pngs, jpegs) is broken. For example, an imported png would look like this:
Test

While pngs are a complete mess, jpegs are imported almost correctly but have wrong colors.

**To Reproduce**
On empty canvas, import any png image through "File->Import". It will look similar to the image above with weird lines and wrong colors.

**Expected behavior**
It would have been nice if all supported image formats would display correctly.

**System information:**
Windows 10 Pro, version 2004.
Commit: 7f39534c86faf005dae4ded1ac313d97e353c624
Something broke very recently I believe.

rodolforg:
There was an undefined behavior due to multiple increment of same variable in multiple function arguments. - sorry

fix #3331

### DIFF
--- a/synfig-core/src/modules/mod_magickpp/mptr_magickpp.cpp
+++ b/synfig-core/src/modules/mod_magickpp/mptr_magickpp.cpp
@@ -128,35 +128,45 @@ magickpp_mptr::get_frame(synfig::Surface& surface, const synfig::RendDesc& /*ren
 		const auto height = image.size().height();
 		surface.set_wh(width, height);
 
-		image.type(Magick::TrueColorType);
-// Sadly Magick++ API is a mess with namespaces...
-{
-	using namespace Magick;
-	using namespace MagickCore;
-
 #if MagickLibVersion >= 0x701 // MAGICKCORE_CHECK_VERSION(7,0,0) does not work!
-		bool has_alpha = image.alpha();
+		const bool has_alpha = image.alpha();
+		constexpr MagickCore::ImageType TrueColorAlphaType = Magick::TrueColorAlphaType;
 #else
-		bool has_alpha = image.matte();
+		const bool has_alpha = image.matte();
+		constexpr MagickCore::ImageType TrueColorAlphaType = Magick::TrueColorMatteType;
 #endif
+
+		if (has_alpha)
+			image.type(TrueColorAlphaType);
+		else
+			image.type(Magick::TrueColorType);
+
 		const auto packet = image.getConstPixels(0, 0, width, height);
 
 		if (!packet) {
 			synfig::error(_("Magick++ importer: couldn't get pixel packet"));
 			return false;
 		}
+// Sadly Magick++ API is a mess with namespaces...
+{
+	using namespace Magick;
+	using namespace MagickCore;
+
 
 		constexpr synfig::Color::value_type factor = QuantumRange;
 
 		for (size_t y = 0, i = 0; y < height; ++y) {
-			for (size_t x = 0; x < width; ++x, ++i) {
+			for (size_t x = 0; x < width; ++x) {
 #if MagickLibVersion >= 0x701 // MAGICKCORE_CHECK_VERSION(7,0,0) does not work!
-				if (has_alpha)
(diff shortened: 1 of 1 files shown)


synfig-core/src/modules/mod_magickpp/mptr_magickpp.cpp
```diff
@@ -128,33 +128,43 @@ magickpp_mptr::get_frame(synfig::Surface& surface, const synfig::RendDesc& /*ren
 		const auto height = image.size().height();
 		surface.set_wh(width, height);
 
-		image.type(Magick::TrueColorType);
-// Sadly Magick++ API is a mess with namespaces...
-{
-	using namespace Magick;
-	using namespace MagickCore;
+#if MagickLibVersion >= 0x701 // MAGICKCORE_CHECK_VERSION(7,0,0) does not work!
+		const bool has_alpha = image.alpha();
```
ice0, line 132: I suggest moving the ImageType here as well. In this case, we group all the overrides together:
```
#if MagickLibVersion >= 0x701 // MAGICKCORE_CHECK_VERSION(7,0,0) does not work!
		const bool has_alpha = image.alpha();
		constexpr MagickCore::ImageType TrueColorAlphaType = Magick::TrueColorAlphaType;
#else
		const bool has_alpha = image.matte();
		constexpr MagickCore::ImageType TrueColorAlphaType = Magick::TrueColorMatteType;
#endif
```

synfig-core/src/modules/mod_magickpp/mptr_magickpp.cpp
```diff
@@ -128,33 +128,43 @@ magickpp_mptr::get_frame(synfig::Surface& surface, const synfig::RendDesc& /*ren
...
+// Sadly Magick++ API is a mess with namespaces...
+{
+	using namespace Magick;
+	using namespace MagickCore;
+
 
 		constexpr synfig::Color::value_type factor = QuantumRange;
 
 		for (size_t y = 0, i = 0; y < height; ++y) {
 			for (size_t x = 0; x < width; ++x, ++i) {
```
ice0, line 161: It looks like we can make this code clearer by taking `++i` out of the loop:
3 and 4 constants should be clearer than 2 and 3.
```
		for (size_t y = 0, i = 0; y < height; ++y) {
			for (size_t x = 0; x < width; ++x) {
#if MagickLibVersion >= 0x701 // MAGICKCORE_CHECK_VERSION(7,0,0) does not work!
				if (has_alpha) {
					surface[y][x] = synfig::Color(packet[i] / factor, packet[i+1] / factor, packet[i+2] / factor, packet[i+3] / factor);
					i += 4;
				} else {
					surface[y][x] = synfig::Color(packet[i] / factor, packet[i+1] / factor, packet[i+2] / factor, 1.);
					i += 3;
				}
#else
				const auto& color = packet[++i];
				if (has_alpha)
					surface[y][x] = synfig::Color(color.red / factor, color.green / factor, color.blue / factor, 1. - color.opacity / factor);
				else
					surface[y][x] = synfig::Color(color.red / factor, color.green / factor, color.blue / factor, 1.);
#endif
```

synfig-core/src/modules/mod_magickpp/mptr_magickpp.cpp
rodolforg: Done

synfig-core/src/modules/mod_magickpp/mptr_magickpp.cpp
rodolforg: Indeed, it's way better!
I just changed the pre-increment to post-increment on `else` branch, as it would skip `packet[0]` otherwise

ice0: > as it would skip packet[0] otherwise

Good catch!

ice0 (approved):

ice0:
Merged. Thank you!

PR #3334, opened 2024-02-20
feat: confirmation dialog for duplicate palette color addition

### LINKED ISSUE #3215: Adding colors to color palette

KartikWatts:
This PR contains the enhancement discussed in #3215 

A confirmation dialog will be displayed to the User when they attempt to add a duplicate color again into the color palette.

A new public method `Palette::is_color_present` is implemented to check the duplicate color presence
Additional private method `Dock_PalEdit::confirm_duplicate_color_addition()` is implemented to be called within existing `Dock_PalEdit::add_color` method for modularity to trigger `UIInterface::confirmation`

closes #3215

### DIFF
--- a/synfig-core/src/synfig/palette.cpp
+++ b/synfig-core/src/synfig/palette.cpp
@@ -282,6 +282,17 @@ Palette::find_light()
 	return best_match;
 }
 
+bool
+Palette::is_color_present(const Color& color) const
+{
+	for (const_iterator iter = cbegin(); iter != cend(); ++iter) {
+		if (iter->color == color)
+			return true;
+	}
+
+	return false;
+}
+
 Palette
 Palette::grayscale(int steps, ColorReal gamma)
 {
--- a/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
+++ b/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
@@ -397,6 +397,10 @@ Dock_PalEdit::show_menu(int i)
 int
 Dock_PalEdit::add_color(const synfig::Color& x)
 {
+	// Check for duplicate color addition to trigger confirmation dialog
+	if (palette_.is_color_present(x) && !confirm_duplicate_color_addition())
+		return -1;
+
 	palette_.push_back(x);
 	signal_changed()();
 	refresh();
@@ -439,8 +443,11 @@ Dock_PalEdit::add_from_clipboard()
 		col.set_hex(hexcolor);
 		col.set_a(1.0f);
 
-		palette_.push_back(col);
+		// Check for duplicate color addition to trigger confirmation dialog
+		if (palette_.is_color_present(col) && !confirm_duplicate_color_addition())
+			return;
 
+		palette_.push_back(col);
 		signal_changed()();
 		refresh();
 	}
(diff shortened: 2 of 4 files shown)


synfig-core/src/synfig/palette.cpp
```diff
@@ -431,3 +431,17 @@ Palette::load_from_file(const synfig::filesystem::Path& filename)
 
 	return ret;
 }
+
+bool
+Palette::is_color_present(const Color& color)
+{
+        iterator iter;
+
+	for(iter=begin();iter!=end();++iter)
```
KartikWatts, line 440: The current approach requires traversing through the vector for every color addition.
Seems fine to me though, but we may maintain a hashmap if we need further optimization. Thoughts are welcome.

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -537,3 +546,14 @@ Dock_PalEdit::size()const
 {
 	return palette_.size();
 }
+
+bool
+Dock_PalEdit::confirm_duplicate_color_addition()
+{
+        return App::get_ui_interface()->confirmation(
```
KartikWatts, line 553: I have created a separate method to be called within `add_color`. Also, does this behavior require to be controlled with User preferences or a checkbox something like: `Don't ask again`.
However, I felt it should be okay, as this won't happen regularly anyway.

KartikWatts (commented):
During implementation, I had some thoughts, and wanted to point them out. Regards.

Abhishek-kroy:
Hi @KartikWatts,

I’m new to open source and eager to contribute to the implementation of the confirmation dialog for adding duplicate colors to the palette.

Although I’m new to open source contributions, I have a solid understanding of C++ and am keen to start contributing. Could you please provide guidance on how to begin working on this feature? I would greatly appreciate any direction on where to start and how I can assist with the dialog implementation or any improvements related to it.

Looking forward to your feedback and collaborating on this project!

synfig-core/src/synfig/palette.cpp
Abhishek-kroy: I agree that traversing the vector for each color addition works fine for now. However, I think maintaining a hashmap could be a more efficient solution in the long term, especially if the number of colors grows significantly. It would allow for quicker lookups to check for duplicates.

rodolforg:
Hi, @KartikWatts !

Sorry I missed this PR. Are you still around and willing to contribute here?

There are a few code style issues here:
- Indentation must be with tabs
- For new code, we prefer the opening curly bracket `{` in the same line as the `if`, `for`, `while`, `do` (but in the next line for function/method or class/struct declarations).

Regarding the code logic:
- The return value should be `-1` in case of not adding the color, right? It seems the intention is to return the index of added color.
- You missed the `const` declaration in the `Palette::is_color_present()` signature
- The `Palette::is_color_present()` declaration should be right after the find methods in the palette header.

rodolforg:
Rebased and modifications done.
Merging when CI is completed.

rodolforg (approved):

rodolforg:
Merged. Thank you.

Issue #3186, opened 2023-08-21
Feature request: "Do not ask again" checkbox when running a plugin


pgilfernandez:
**Is your feature request related to a problem? Please describe.**
While developing the "animated text" plugin, I found pretty annoying to always have to push "Proceed" button when running a plugin:
<img width="447" alt="Screenshot 2023-08-21 at 17 29 43" src="https://github.com/synfig/synfig/assets/5942369/889a4b90-0a11-4c05-b6a2-db4b33a01560">

**Describe the solution you'd like**
I would like to suggest to add a simple checkbox saying something like "do not ask again" to that popup to let users disable it.

**Describe alternatives you've considered**
none

**Additional comments**
Thanks!

BobSynfig:
As it is written in the warning, it can't be undone and all undo history is lost...
What if you accidentally select the wrong plugin?

pgilfernandez:
Well, if the "do not ask again" checkbox is checked that means the user knows about the warning and takes all responsibility... they couldn't blame Synfig.

BTW, it would be nice and now that I think not difficult to implement as Synfig has an "autosave" feature, to add a backup option near the autosave checkbox. With a simple field you could set backups from 0 (disabled) to any number you want. When enabled, the autosave (or the user when hits Save) creates the backup cascade with files named something like my_animation.sif1, my_animation.sif2,...

If this feature exists, even if you mark "do not ask again", accidentally select the wrong plugin and it totally break your project, you could always go back to your backup files (unless you set the backup to 0... hehehe).

veermetri05:
I agree with @BobSynfig, it is so easy to accidentally select a wrong a plugin from the dropdown and mess your file. But whenever you run a plugin the new file isn't saved automatically. 

BTW we can press "Enter" on keyboard when we get the warning, so I don't think it impacts anybody's productivity.

If this feature is required then I recommend to put the "Do not ask again" checkbox in the preferences dialog as "Do not ask for confirmation while running plugin". So that it is hard to access and user's don't tick the "Do not ask again" checkbox accidentally, and only those aware of the risks will enable it. It increases friction, requiring the user's to locate where the option is.

pgilfernandez:
It's not a bad idea to place it just in Preferences panel. That way probably only pro users would activate it.

BTW, if we combine this feature with the "backups" one then it would be that critical as you could always go to your backups in case of an accident.

BobSynfig:
> That way probably only pro users would activate it.
Even pros can do mistake
[...]

PR #3340, opened 2024-03-18
feat: added toggle setting for plugin warnings

Funinja:
Added functional disable plugin warning setting as requested from issue #3186. 

New setting in preferences:
image

New plugin warning:
image

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -300,6 +300,10 @@ static int max_recent_files_=25;
 int    studio::App::get_max_recent_files()      { return max_recent_files_; }
 void   studio::App::set_max_recent_files(int x) {        max_recent_files_ = x; }
 
+static bool plugin_warn_enabled_=true;
+bool   studio::App::get_plugin_warning_enabled()      { return plugin_warn_enabled_; }
+void   studio::App::set_plugin_warning_enabled(bool x) {        plugin_warn_enabled_ = x; }
+
 static synfig::String app_base_path_;
 
 SoundProcessor *App::sound_render_done = nullptr;
@@ -452,6 +456,10 @@ public:
 			if(key=="file_history.size")
 			{
 				value=strprintf("%i",App::get_max_recent_files());
+				return true;
+			}
+			if(key=="plugin_warning"){
+				value=strprintf("%i",App::get_plugin_warning_enabled());
 				return true;
 			}
 			if(key=="distance_system")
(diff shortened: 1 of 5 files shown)


Funinja:
Force-pushed was for the commit messages only.

ice0:
@mohamedAdhamc @morevnaproject @pgilfernandez @rodolforg 

I'm not sure, but do we need a global option, or should it be for each plugin?

pgilfernandez:
Maybe it's a good idea to have the option per plugin. That way once you fully trust a plugin you can disable warning just for it...

mohamedAdhamc:
> Maybe it's a good idea to have the option per plugin. That way once you fully trust a plugin you can disable warning just for it...

I agree, having it be per plugin does make more sense or at least I see it that way. So I use a plugin a lot and I trust it doesn't mess my work and then I just want to increase the speed of my workflow. That doesn't necessarily mean I would feel the same about the other plugins i have now or in the future.

Although I think we can also possibly have a global one as well, but it'd have to be done in a way that would be easier for the user to understand and not confuse them both.

Funinja:
I agree with everyone here, having the option to toggle warnings per plugin makes sense.

 Unfortunately, I believe the best solution to this feature would be to build a Plugin Dialogue Manager to interface with plugin toggle warning settings individually and make them persistent (after closing and reopening the application). Funny enough, the plugin dialogue manager is one of the ideas in this year's GSOC proposal and I don't know if it is fair for me to start on this contribution, being currently one of GSOC's applicants for the plugin manager for Synfig.

 If it is okay with everyone, I can make a simple plugin dialogue manager with the only feature being to toggle the individual plugins, and the GSOC applicant who is picked to develop the plugin dialogue manager can use it as a base code for their development.

mohamedAdhamc:
@Funinja how about just changing "do not show warning" to something like "do not show warning again for this plugin". And you can then save which plugins have a "do not show warning" preference set to them. If they do then you wouldn't show the warning and if they don't then you can show it normally.

or possibly adding "do not show warning again for this plugin" underneath the current "global" one.

@ice0 any thoughts on this ?

Funinja:
> @Funinja how about just changing "do not show warning" to something like "do not show warning again for this plugin". And you can then save which plugins have a "do not show warning" preference set to them. If they do then you wouldn't show the warning and if they don't then you can show it normally.
> 
> or possibly adding "do not show warning again for this plugin" underneath the current "global" one.
Yes, I can do this, but my main concern is if someone accidentally disables the warning for a certain plugin and wants to turn the warning back on. I'm not sure where would they go if they want to turn it back on.

mohamedAdhamc:
> > @Funinja how about just changing "do not show warning" to something like "do not show warning again for this plugin". And you can then save which plugins have a "do not show warning" preference set to them. If they do then you wouldn't show the warning and if they don't then you can show it normally.
> > or possibly adding "do not show warning again for this plugin" underneath the current "global" one.
> 
> Yes, I can do this, but my main concern is if someone accidentally disables the warning for a certain plugin and wants to turn the warning back on. I'm not sure where would they go if they want to turn it back on.

Ok, yes now I see what you mean.

Issue #3189, opened 2023-08-22
Feature request: create .sif/.sifz backup files when saving and/or autosaving


pgilfernandez:
**Is your feature request related to a problem? Please describe.**
As an example when you may need this feature: https://github.com/synfig/synfig/issues/3186#issuecomment-1688286838

**Describe the solution you'd like**
It would be nice and probably not so difficult to implement as Synfig has an "autosave" feature, to add a backup option near the autosave checkbox. With a simple field you could set backups from 0 (disabled) to any number you want. When enabled, the autosave (or the user when hits Save) creates the backup cascade with files named something like my_animation.sif1, my_animation.sif2,...

![backup_files](https://github.com/synfig/synfig/assets/5942369/e309f026-b80d-487b-9137-77076b32827a)

**Describe alternatives you've considered**
none

**Additional comments**
Thanks

veermetri05:
If I understand this issue properly, the "Backup files" parameter will define the number of previous snapshots of file to be saved based on time interval.

Considering interval is 60s and backup set to 2, we would have following files.
animation1.sif (saved 60s before)
animatino2.sif (saved 120s before)

pgilfernandez:
Not exactly, with that configuration it would be:
my_animation.sif
my_animation.sif1
my_animation.sif2

Notes:
- every 60 seconds the .sif1 is renamed as .sif2, the .sif is renamed to .sif1 and the project is saved in the actual situation as .sif. This way we keep the normal working file (and newest) as .sif and the backups are always numbers taking in count that lower are newer.
- numbers are placed in the file extension as they are not part of the filenames but backup files. Other FOSS do the same way, read Blender (.blend1) or FreeCAD (.FCstd1)
- so at autosave or manual save only one file is really saved, the other ones are just renamed to keep them organized chronologically
- same thing would apply to .sifz files: .sifz1, .sifz2, ...
- if this procedure is very computer consuming (which I don thing so...) it could just be applied when user manually saves the file. But, as said, it's better to make ir work with both save and autosave

PR #3342, opened 2024-03-25
feat: add auto-backup files and backup files settings

Funinja:
This PR implements a feature requested in issue #3189. 

Specifically, it is an autosave feature that saves the current temp file of the current sif file to a sif file in a backup folder created in the directory of the original sif file. Before saving the current temp file, it renames all other backup files by increasing their version name by 1.

For example, for mydir/project.sif:
- creates directory: mydir/backup/ 
- creates backup file from current temp file: mydir/backup/project_1.sif

Say there exists a project_1.sif and project_2.sif in the backup directory. Then, the files are renamed project_2.sif and project_3.sif respectively, before creating the new backup file project_1.sif. Leaving the backup file with project_1.sif, project_2.sif, and project_3.sif. Notably, I have also created a max backup file setting that limits the number of auto-backed up files. Mainly, if the limit was set to 2, instead of having 3 backup files in the previous example we would have 2 most current backup files with the third file being deleted.

I followed the backup file format [project_name]_[version].sif instead of the original issue requested format [project_name].sif[version] since synfig doesn't explicitly read files in the issue requested format.

Backup files setting in preferences:
Untitled
 
Files are backed up in time intervals defined by the interval autosave setting (in the image above it would save every 10 seconds) and if the backup files setting is set to 0 no files will be saved. Additionally, if autosave is turned off, no files will be backed up.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -1977,6 +1977,43 @@ App::set_time_format(synfig::Time::Format x)
 	_App_time_format=x;
 }
 
+/*
+
+Save backup file without altering the current file
+
+*/
+
+bool
+App::save_backup()
+{
+	FileSystemTemporary::Handle temporary_filesystem = FileSystemTemporary::Handle::cast_dynamic(App::get_selected_canvas_view()->get_canvas()->get_file_system());
+
+	// get original filename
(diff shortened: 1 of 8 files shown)


Funinja:
closing to fit git commits more closesly to git commit convention

PR #3343, opened 2024-03-25
feat: add autobackup system and backup file settings

Funinja:
This PR implements a feature requested in issue #3189. 

Specifically, it is an autosave feature that saves the current temp file of the current sif file to a sif file in a backup folder created in the directory of the original sif file. Before saving the current temp file, it renames all other backup files by increasing their version name by 1.

For example, for mydir/project.sif:
- creates directory: mydir/backup/ 
- creates backup file from current temp file: mydir/backup/project_1.sif

Say there exists a project_1.sif and project_2.sif in the backup directory. Then, the files are renamed project_2.sif and project_3.sif respectively, before creating the new backup file project_1.sif. Leaving the backup file with project_1.sif, project_2.sif, and project_3.sif. Notably, I have also created a max backup file setting that limits the number of auto-backed up files. Mainly, if the limit was set to 2, instead of having 3 backup files in the previous example we would have 2 most current backup files with the third file being deleted.

I followed the backup file format [project_name]_[version].sif instead of the original issue requested format [project_name].sif[version] since synfig doesn't explicitly read files in the issue requested format.

Backup files setting in preferences:
Untitled
 
Files are backed up in time intervals defined by the interval autosave setting (in the image above it would save every 10 seconds) and if the backup files setting is set to 0 no files will be saved. Additionally, if autosave is turned off, no files will be backed up.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -1992,7 +1992,14 @@ App::save_backup()
 		return true;
 	}
 
-	FileSystemTemporary::Handle temporary_filesystem = FileSystemTemporary::Handle::cast_dynamic(App::get_selected_canvas_view()->get_canvas()->get_file_system());
+	CanvasView::LooseHandle canvas_view = App::get_selected_canvas_view();
+
+	// no canvas, return immediately
+	if (canvas_view.empty()){
+		return true;
+	}
+
+	FileSystemTemporary::Handle temporary_filesystem = FileSystemTemporary::Handle::cast_dynamic(canvas_view->get_canvas()->get_file_system());
 	// get original filename
 	String filename = temporary_filesystem->get_meta("filename");
 	String as = temporary_filesystem->get_meta("as");
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -300,6 +300,10 @@ static int max_recent_files_=25;
 int    studio::App::get_max_recent_files()      { return max_recent_files_; }
 void   studio::App::set_max_recent_files(int x) {        max_recent_files_ = x; }
 
+static int backup_files_ = 0;
+int    studio::App::get_num_backup_files()      { return backup_files_; }
+void   studio::App::set_num_backup_files(int x) {        backup_files_ = x; }
+
 static synfig::String app_base_path_;
 
 SoundProcessor *App::sound_render_done = nullptr;
@@ -452,6 +456,11 @@ public:
 			if(key=="file_history.size")
 			{
 				value=strprintf("%i",App::get_max_recent_files());
+				return true;
+			}
+			if(key=="backup.num")
+			{
+				value=strprintf("%i",App::get_num_backup_files());
 				return true;
 			}
 			if(key=="distance_system")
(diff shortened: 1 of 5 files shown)


synfig-studio/src/gui/app.cpp
```diff
@@ -1961,6 +1977,105 @@ App::set_time_format(synfig::Time::Format x)
...
+	}
+
+	filesystem::Path backup_dir = proj_dir / filesystem::Path("backups");
+	filesystem::Path base_name = filesystem::Path::basename(filesystem::Path::filename_sans_extension(filename));
+	filesystem::Path file_ext = filesystem::Path::filename_extension(filename);
+	int file_ext_len = file_ext.u8string().length();
+
+	// make backup directory if not already created
+	if (!FileSystemNative::instance()->directory_create(backup_dir.u8string()))
+		return false;
```
ice0, line 2024: I suggest to log the error somehow, so user may see he has some issues here.

synfig-studio/src/gui/app.cpp
```diff
@@ -1961,6 +1977,105 @@ App::set_time_format(synfig::Time::Format x)
...
+
+	int found_backups = 0;
+	std::vector<String> files;
+
+	if (!FileSystemNative::instance()->directory_scan(backup_dir.u8string(), files))
+		return false;
+
+	std::unordered_map<String, String> file_version_map;
```
ice0, line 2032: I suggest to use `std::map` instead `std::unordered_map` here.
I this case you can use the simple logic:
1. `number of backups == map.size()`
2. first element is the earliest which should be removed, `if map.size() > max_backups`, last is the oldest.

synfig-studio/src/gui/app.cpp
Funinja: New fix now shows dialog errors to the user

synfig-studio/src/gui/app.cpp
Funinja: New fix now uses map instead of unordered_map as requested

synfig-studio/src/gui/app.cpp
```diff
@@ -1961,6 +1977,105 @@ App::set_time_format(synfig::Time::Format x)
...
+	String filename = temporary_filesystem->get_meta("filename");
+	String as = temporary_filesystem->get_meta("as");
+	String truncate = temporary_filesystem->get_meta("truncate");
+	if (filename.empty() || as.empty() || truncate.empty())
+		throw (String)strprintf(_("Original filename was not set in temporary container \n\n"));
+	FileContainerZip::file_size_t truncate_storage_size = stoll(truncate);
+	filesystem::Path proj_dir = filesystem::Path::dirname(filename);
+
+	// is new file, do not backup
+	if(proj_dir.u8string() == "."){
```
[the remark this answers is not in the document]
Funinja: That's true, I haven't thought about that. 

Maybe instead of creating a backup directory, the user can designate the backup directory in the preference list. This would allow users to save new files with no directories, however the backed up files would be saved as "Synfig 1_1.sif" and would not be renamed after saving the new file to a different name. 

@pgilfernandez, as the original author of the issue, is this what you envisioned for this feature or did you want something else?

PR #3347, opened 2024-04-28
fix(MotionBlur): undefined behavior when subsample range is 0

### LINKED ISSUE #3345: Bug with linear subsampling type of the motion blur layer
When you set "subsample start amount" & "subsample end amount" to 0, and subsampling type to "linear", the object affected by this layer becomes black. If there are two objects in a group, they both turn into a black square.

1. Create any layer, for example, a star.
2. Add a motion blur layer.
3. Set the following parameters to the motion blur: aperture = 24 f, subsampling factor = 1.0, subsampling type = linear, Subsample Start Amount = 0.0, Subsample End Amount = 0.0.
4. The star disappears.
5. Group the star and motion blur layer.
6. Create a white background under this group.
7. You will see a black star.
8. If you add another shape layer in the group with the star and blur, both objects become one black rectangle.

The other subsampling types work well. In the screenshots, you can see three identical groups, but the only difference is the subsampling type of each motion blur layer.

 - OS: Linux RED OS, 6.1.52-1, el7.3.x86_64 x86_64, MATE 1.26.0
 - Synfig version: 1.5.1

rodolforg:
For linear subsampling type, there is an undefined behavior when both subsample (start and end) amounts are 0.

Synfig version 1.0.2: *none* is drawn!
Synfig version 1.2.1: a pale gray rectangle is drawn
Synfig version 1.4.3: a black gray rectangle is drawn

Anyway, this commit fixes the case when both are 0, by drawing the context without any motion blur effect.

Note:
If they (the subsample amounts) are equal, but both nonzero, the canvas is normally drawn.
Behavior is not the same between 1.0.2 and >=1.2.1 though. The motion trail has a different number of 'echoes': 1.0.2 has less than the next releases.

fix #3345

### DIFF
--- a/synfig-core/src/synfig/layers/layer_motionblur.cpp
+++ b/synfig-core/src/synfig/layers/layer_motionblur.cpp
@@ -204,7 +204,7 @@ Layer_MotionBlur::build_rendering_task_vfunc(Context context) const
 	}
 
 	const Color::BlendMethod blend_method = no_blur_effect ? Color::BLEND_COMPOSITE : Color::BLEND_ADD_COMPOSITE;
-	const Real k = no_blur_effect ? 1.0 : (1.0/sum);
+	const Real k = no_blur_effect ? 1.0 : (approximate_zero(sum) ? 0.0 : (1.0/sum));
 
 	rendering::Task::Handle task;
 	for(int i = 0; i < samples; i++)
@@ -226,5 +226,7 @@ Layer_MotionBlur::build_rendering_task_vfunc(Context context) const
 		task = task_blend;
 	}
 
+	if (!task)
+		return context.build_rendering_task();
 	return task;
 }


rodolforg:
It's possible to port it to 1.4.x

morevnaproject (approved):

morevnaproject:
> It's possible to fix in the 1.4.x branch too. Is it desired, @ice0 @morevnaproject ?

I tried to apply to 1.4.x branch, but it doesn't applies cleanly (conflicts produced). So, let's keep it for 1.5.2.

rodolforg:
I think it would work simply by inserting 
```c++
if (divisor == 0.)
 return context.accelerated_render(surface,quality,renddesc,cb);
```
Before these `for`-loops below:
https://github.com/synfig/synfig/blob/7095a9d82f95af0e08c7e065ee95f3ad3466a518/synfig-core/src/synfig/layers/layer_motionblur.cpp#L259-L261

PR #3355, opened 2024-05-04
build: update script for building with cmake

### LINKED ISSUE #3352: Errors in CMake build script - "2-build-cmake.sh"
In the root of synfig's source dir we have the script to build with CMake - `2-build-cmake.sh` (https://github.com/synfig/synfig/blob/58f522ec25d845ee7734dab0ca385a8ff5ccfd3c/2-build-cmake.sh).

When I try to run the script on Linux it fails with the following error:

```
[konstantin@localhost synfig]$ ./2-build-cmake.sh -r
...
-- OpenEXR include: /usr/include/OpenEXR
--   mod_particle
--   mod_png
--   mod_ppm
--   mod_svg
--   mod_yuv420p
--   mod_mng
--   mod_magickpp
CMake Error at src/modules/CMakeLists.txt:82 (file):
  file failed to open for writing (Permission denied):

    /etc/synfig_modules.cfg


CMake Error at po/CMakeLists.txt:1 (include):
  include could not find requested file:

    SynfigGettext


CMake Error at po/CMakeLists.txt:4 (SYNFIG_PROCESS_PO_FILES):
  Unknown CMake command "SYNFIG_PROCESS_PO_FILES".


-- Configuring incomplete, errors occurred!
See also "/home/konstantin/sources/synfig/_production/synfig-core/CMakeFiles/CMakeOutput.log".
See also "/home/konstantin/sources/synfig/_production/synfig-core/CMakeFiles/CMakeError.log".
Failed to build synfig-core
```

I have examined the script and found that it tries to build ETL, core and studio as separate steps.
In documentation we have simple instructions to build synfig in just one step - https://synfig-docs-dev.readthedocs.io/en/latest/building/Building%20using%20CMake.html#linux

Also there is a file called `./2-build-msys-cmake.sh` which uses much simpler approach.

Is the  script outdated?

morevnaproject:
Closes #3352


1-setup-linux-native.sh
```diff
@@ -245,6 +252,8 @@ elif [ "$ID_LIKE" == "debian" ] || [ "$ID_LIKE" == "ubuntu" ] || [ "$ID_LIKE" ==
             PKG_LIST=" \
                 build-essential \
                 cmake \
+                ninja \
```
ice0, line 255: should be `ninja-build`

rodolforg (approved):

PR #1413, opened 2020-05-16
Added toggleable option for thumbnail preview

Keyikedalube:
This fixes the keyboard input **Ctrl+,** & **Ctrl+.** not responding when
thumbnail is displayed under Wayland session

By default, thumbnail preview is turned on so it doesn't affect X11
users
Wayland users can turn it off in Preferences > Editing > Thumbnail
preview section

Fixes #1144

PR #3356, opened 2024-05-11
perf(GUI): don't render frame thumbnails if disabled

rodolforg:
If thumbnail preview is disabled in Synfig Studio
settings, don't render them, instead of just hide them.

Related to #1413 (6709d28f77f8c3a568e200237221b12b87fd0b12)


ice0 (approved):

ice0:
Merged. Thank you!

PR #3357, opened 2024-05-12
build: added macOS bundle script

ice0:


ice0:
> How complex !!!

You can compare with current implementation:
https://github.com/synfig/synfig/blob/master/autobuild/osx/relocate-binary.sh (149 lines) +
https://github.com/synfig/synfig/blob/master/3-package-osx-dmg.sh (378 lines)

527 lines vs 270 )

Also the `3-package-osx-dmg.sh` script does not sign binaries.

PR #3360, opened 2024-05-19
chore: Push tag to upstream when new version released

morevnaproject:


rodolforg (commented):
# Why `--force`?

morevnaproject:
> Why `--force`?

That's because "testing" is temporary branch, which tells our build system which commit to use for building packages.

"testing" branch can "jump" from 1.4.x to 1.5.x and to master, when needed.

rodolforg:
Use lowercase after the "chore:"

PR #3373, opened 2024-06-09
fix: color/gradient dialog edits the parameter that opened itself

### LINKED ISSUE #3288: Infinite Error Windows from Eyedropper
**Describe the bug**
When attempting to use the eyedropper, I go the error "Unable to set all layer parameters," but the eyedropper is apparently still active. Attempting to close the error window results in the eyedropper trying to pick the color of the error box and spawned new error windows for every click. The escape and enter key all seem to also select colors and spawn new errors instead of letting me exit. I can't close the window from my taskbar, so I either have to use task manager or wait for it to crash. Recovery saved my work, so no big issues there.

Cause of Error: I opened the color picker then realized the color I wanted was on the bottom layer and covered. I moved that layer while the color picker was open, then attempted to sample from it. The correct color was chosen before the error occurred. On my second attempt, the same error did not happen.

My layers were an advanced outline, a region, and a png (in that order from top to bottom). Then I moved the png to the top layer. There is an error that occurs before "Unable to set all layer parameters," but it is covered too fast for me to easily read it.

**To Reproduce**
1. Go to Parameters (for a region)
2. Click on the color box next to the Color parameter to open the selector
3. Move the bitmap layer to the top
4. Attempt to select a color from bitmap image
5. See error

**Expected behavior**
The color is selected, or the resulting error can be exited.

**Screenshots**
image

**System information:**
 - OS: Windows 10
 - OS version: 10.0.19045 Build 19045
 - Synfig version: 1.4.2

**Additional comments**
I was able to reproduce it consistently. This only occurs when I move and sample from a bitmap image, not a region or outline.

rodolforg:
Here we prevent to emit signal_edited_ for Color and Gradient dialog
windows when presented by Parameter Panel.

Those dialogs may be opened by a cell and try to update the parameter
value related to another one, as it just remembered its parameter tree
path.

Reproducible steps:
1. Create a region layer
2. Create an outline layer
3. Click on color parameter to edit it
4. Without closing color dialog, select the region layer instead
5. Try to edit the color in dialog
6. Error Message appears

fix #3288

### DIFF
--- a/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
+++ b/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
@@ -252,6 +252,15 @@ bool get_paragraph(synfig::String& text)
 	return App::dialog_paragraph(_("Text Paragraph"), _("Enter text here:"), text);
 }
 
+static bool
+is_layer_deleted(const synfig::Layer::ConstHandle& layer)
+{
+	if (Canvas::LooseHandle layer_canvas = layer->get_canvas())
+		return std::find(layer_canvas->begin(), layer_canvas->end(), layer) == layer_canvas->end();
+
+	return true;
+}
+
 /* === M E T H O D S ======================================================= */
 
 CellRenderer_ValueBase::CellRenderer_ValueBase():
@@ -566,6 +575,10 @@ CellRenderer_ValueBase::gradient_edited(synfig::Gradient gradient, Glib::ustring
 {
 	synfigapp::ValueDesc vd = App::dialog_gradient->get_value_desc();
 	if (vd.is_valid()) {
+		if (vd.parent_is_layer()) {
+			if (is_layer_deleted(vd.get_layer()))
+				return;
+		}
 		// do not emit signal_edited_ : gradient dialog may be opened by other cell
 		// and would try to update the parameter related to current one
 		// as it just remembered its parameter tree path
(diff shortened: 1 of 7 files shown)


rodolforg:
If a 1.4.6 will ever be released, backport this PR.

ice0:
I do not know if this is important, but it changes the current behavior:

Before:

https://github.com/synfig/synfig/assets/5604544/034d7d05-db4d-4e8f-bfca-43d08824569a


After:

https://github.com/synfig/synfig/assets/5604544/4f82840c-0f60-478d-9c1a-81a6e8c94ff5

rodolforg:
It works like that before just by coincidemce: there is a layer parameter of color type value at same tree path.
If you have a region layer and an outline layer, the steps you did won't work at all.

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
```diff
@@ -252,6 +252,17 @@ bool get_paragraph(synfig::String& text)
 	return App::dialog_paragraph(_("Text Paragraph"), _("Enter text here:"), text);
 }
 
+bool
+is_layer_deleted(const synfig::Layer::ConstHandle& layer)
+{
+	Canvas::LooseHandle layer_canvas = layer->get_canvas();
```
ice0, line 258: Looks good, but I think this part could be clearer:
```
	if (Canvas::LooseHandle layer_canvas = layer->get_canvas())
		return std::find(layer_canvas->begin(), layer_canvas->end(), layer) == layer_canvas->end();

	return true;
```

Anyway you can leave it as-is.
Please rebase.

ice0 (approved):

synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
rodolforg: Refactored and rebased.

ice0 (approved):

ice0:
Merged. Thank you!

PR #3375, opened 2024-06-24
build: remove creation of deprecated synfig-config

rodolforg:
Use pkgconf instead


rodolforg:
@ice0 

I'm unsure what those lines mean and if they can be deleted too:
https://github.com/synfig/synfig/blob/3f0795986a0e0612f86695dade1ebd4f658d3c39/synfig-core/src/synfig/CMakeLists.txt#L153-L165

ice0:
> I'm unsure what those lines mean and if they can be deleted too:

This is a CMake configuration file. Something like `.pc`, but for CMake builds only. (https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#id1)
I suggest leaving it for now, they can be helpful later.

PR #3376, opened 2024-06-24
chore: don't pollute synfig.pc with unneeded flags

rodolforg:
it only should point what is need to link to it

Plus: in synfig-studio, use MLT++ and fontconfig flags for GUI only, not synfigapp


rodolforg:
By the way, AFAIK cmake build doesn't create synfig.pc . Am I right?

ice0:
> By the way, AFAIK cmake build doesn't create synfig.pc . Am I right?

Yes.

PR #3378, opened 2024-07-03
feat: create Plugin Manager dialog

veermetri05:


synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1417,6 +1418,7 @@ CanvasView::init_menus()
 		{"jump-next-keyframe", "animate_seek_next_keyframe_icon", N_("Seek to Next Keyframe"),      "", sigc::mem_fun(*canvas_interface(), &CanvasInterface::jump_to_next_keyframe) },
 		{"jump-prev-keyframe", "animate_seek_prev_keyframe_icon", N_("Seek to Previous Keyframe") , "", sigc::mem_fun(*canvas_interface(), &CanvasInterface::jump_to_prev_keyframe) },
 
+		{"plugin-manager", "", N_("Plugin Manager"), "", sigc::mem_fun0(plugin_manager_dialog, &Dialog_PluginManager::present)},
```
rodolforg, line 1421: I guess you can remove the `0` digit of `sigc::mem_fun0()`.

synfig-studio/src/gui/canvasview.h
```diff
@@ -75,6 +75,7 @@
 #include "dialogs/dialog_keyframe.h"
 #include "dialogs/dialog_preview.h"
 #include "dialogs/dialog_waypoint.h"
+#include "dialogs/dialog_pluginmanager.h"
```
rodolforg, line 78: Move to line 76 to sort it alphabetically.

synfig-studio/src/gui/dialogs/Makefile_insert
```diff
@@ -19,7 +19,8 @@ DIALOGS_HH = \
 	dialogs/dialog_guide.h \
 	dialogs/dialog_workspaces.h \
 	dialogs/dialog_template.h \
-	dialogs/vectorizersettings.h
+	dialogs/vectorizersettings.h \
+	dialogs/dialog_pluginmanager.h
```
rodolforg, line 23: move up to sort it alphabetically

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,126 @@
...
+**	published by the Free Software Foundation; either version 2 of
+**	the License, or (at your option) any later version.
+**
+**	This package is distributed in the hope that it will be useful,
+**	but WITHOUT ANY WARRANTY; without even the implied warranty of
+**	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+**	General Public License for more details.
+**	\endlegal
+*/
+/* ========================================================================= */
```
rodolforg, line 21: Please follow the current template for .cpp

synfig-studio/src/gui/dialogs/dialog_pluginmanager.h
```diff
@@ -0,0 +1,68 @@
...
+
+/* === H E A D E R S ======================================================= */
+
+#include <ETL/handle>
+
+#include <gtkmm/dialog.h>
+#include <gtkmm/notebook.h>
+#include <gui/pluginmanager.h>
+#include <synfig/os.h>
+#include <gui/app.h>
```
rodolforg, line 39: As these are not used by header file itself, move to implementatio file .cpp

rodolforg (changes requested):
Nice public start :)
I'll resume my review later.

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
veermetri05: Where should I find it ?

rodolforg: I just found out where did you get this old template.
synfig-studio/src/gui/dialogs/canvasresize.h / synfig-studio/src/gui/dialogs/canvasresize.cpp

Any other file uses the correct one :)

https://github.com/synfig/synfig/blob/master/synfig-core/src/template.h

https://github.com/synfig/synfig/blob/master/synfig-core/src/template.cpp

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1417,6 +1418,7 @@ CanvasView::init_menus()
 		{"jump-next-keyframe", "animate_seek_next_keyframe_icon", N_("Seek to Next Keyframe"),      "", sigc::mem_fun(*canvas_interface(), &CanvasInterface::jump_to_next_keyframe) },
 		{"jump-prev-keyframe", "animate_seek_prev_keyframe_icon", N_("Seek to Previous Keyframe") , "", sigc::mem_fun(*canvas_interface(), &CanvasInterface::jump_to_prev_keyframe) },
 
+		{"plugin-manager", "", N_("Plugin Manager"), "", sigc::mem_fun0(plugin_manager_dialog, &Dialog_PluginManager::present)}
```
rodolforg, line 1421: 1. I guess you can remove the `0` digit of `sigc::mem_fun0()`, leaving it as `sigc::memfun()` only.
2. I think it's better to name it more descriptive as "open-plugin-manager".

synfig-studio/src/gui/dialogs/Makefile_insert
```diff
@@ -42,7 +44,7 @@ DIALOGS_CC = \
 	dialogs/dialog_guide.cpp \
 	dialogs/dialog_workspaces.cpp \
 	dialogs/dialog_template.cpp \
-	dialogs/vectorizersettings.cpp
+	dialogs/vectorizersettings.cpp
```
rodolforg, line 47: please remove the trailing blank space

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,161 @@
+/* === S Y N F I G ========================================================= */
+/*!	\file template.cpp
+**	\brief Template File
```
rodolforg, line 3: Whoops ;)

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,161 @@
...
+        // pluginListItem->override_background_color(*(new Gdk::RGBA("rgb(242, 242, 242)")));
+        pluginListItem->set_margin_top(10);
+        pluginOptionBox->property_expand().set_value(false);
+
+        restoreSettings->set_image_from_icon_name("view-refresh", Gtk::ICON_SIZE_BUTTON);
+        openFolder->set_image_from_icon_name("document-open", Gtk::ICON_SIZE_BUTTON);
+        deletePlugin->set_image_from_icon_name("user-trash-symbolic", Gtk::ICON_SIZE_BUTTON);
+
+        deletePlugin->signal_clicked().connect([plugin, this](){
+            this->messageDialog.set_message("Do you want to delte the plugin: " + plugin.name.fallback());
```
rodolforg, line 102: 1. You misspelled "delete" here
2. Add internationalization here: surround translatable string with `_()`
3. Maybe you should use `plugin.name.get()` (translated name) instead of `plugin.name.fallback()`

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,161 @@
...
+    installPluginButton->set_image_from_icon_name("document-open", Gtk::ICON_SIZE_BUTTON);
+    installPluginButton->set_always_show_image(true);
+    installPluginButton->signal_clicked().connect(sigc::mem_fun(*this, &Dialog_PluginManager::clicked));
+
+    Gtk::VBox *pluginTab = manage(new Gtk::VBox());
+    pluginTab->property_expand().set_value(false);
+    Gtk::ListBox *pluginListBox = manage(new Gtk::ListBox());
+    pluginListBox->set_selection_mode(Gtk::SELECTION_NONE);
+
+    for (auto plugin : pluginList)
```
rodolforg, line 80: As you won't change `plugin` itself, prefer to use `const auto&` rather plain `auto`. It's more efficient (avoid copies).

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,161 @@
...
+    Gtk::Dialog(_("Plugin Manager"), parent),
+    messageDialog(_("Are you sure you want to delete this plugin?"), false, Gtk::MESSAGE_QUESTION, Gtk::BUTTONS_OK_CANCEL, true),
+    pluginList(App::plugin_manager.plugins())
+{
+    // this->set_resizable(false);
+    messageDialog.set_transient_for(*this);
+    Gtk::Button *installPluginButton = manage(new Gtk::Button(_("Install Plugin")));
+    installPluginButton->set_image_from_icon_name("document-open", Gtk::ICON_SIZE_BUTTON);
+    installPluginButton->set_always_show_image(true);
+    installPluginButton->signal_clicked().connect(sigc::mem_fun(*this, &Dialog_PluginManager::clicked));
```
rodolforg, line 73: I know it's just a test, but it's better to name methods in a more descriptive way.
Maybe `on_test_clicked` or `on_install_plugin_button_clicked`.

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,161 @@
...
+/* === M E T H O D S ======================================================= */
+
+Dialog_PluginManager::Dialog_PluginManager(Gtk::Window& parent):
+    Gtk::Dialog(_("Plugin Manager"), parent),
+    messageDialog(_("Are you sure you want to delete this plugin?"), false, Gtk::MESSAGE_QUESTION, Gtk::BUTTONS_OK_CANCEL, true),
+    pluginList(App::plugin_manager.plugins())
+{
+    // this->set_resizable(false);
+    messageDialog.set_transient_for(*this);
+    Gtk::Button *installPluginButton = manage(new Gtk::Button(_("Install Plugin")));
```
rodolforg, line 70: Our (non-official?) code style now suggests (requires?):
`Gtk::Button* install_plugin_button`  instead of  `Gtk::Button *installPluginButton`.

1. No space before pointer mark `*`, but after
2. snake_case variable/method naming instead of camelCase

(There are other cases below)

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+        restoreSettings->set_image_from_icon_name("view-refresh", Gtk::ICON_SIZE_BUTTON);
+        openFolder->set_image_from_icon_name("document-open", Gtk::ICON_SIZE_BUTTON);
+        deletePlugin->set_image_from_icon_name("user-trash-symbolic", Gtk::ICON_SIZE_BUTTON);
+
+        deletePlugin->signal_clicked().connect([plugin, this](){
+            this->messageDialog.set_message("Do you want to delte the plugin: " + plugin.name.fallback());
+            int response = this->messageDialog.run();
+            if (response == Gtk::RESPONSE_OK)
+            {
```
rodolforg, line 105: In our (official?) code style, for `if`, `for`, `while`, `do-while`, we use initial curly brace `{` on the same line of the statement.

(there are more cases here)

rodolforg (changes requested):
In `synfig-studio/po/POTFILES.in`, add the following line:
src/gui/dialogs/dialog_pluginmanager.cpp

I know it's difficult because the file is a mess, but try to insert it alphabetically (but do not change unrelated lines, just insert yours, to not mix stuff).

I know it seems unnecessary, but, well, Autotools building system requires it, and that's something it disappoints me XD

synfig-studio/src/gui/canvasview.cpp
veermetri05: It's not possible to remove 0, because then I get error, 
```
more than one instance of overloaded function "sigc::mem_fun" matches the argument list
```

synfig-studio/src/gui/pluginmanager.h
```diff
@@ -30,6 +30,8 @@
 
 /* === H E A D E R S ======================================================= */
 
+#include <glibmm/refptr.h>
```
ice0, line 33: `glibmm/refptr.h` - is unused, so better to remove it.
`gtkmm/application.h` - is unused too, you need `#include <sigc++/signal.h>` instead for `sigc::signal`

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1543,6 +1534,35 @@ CanvasView::init_menus()
...
+{
+	auto instance = get_instance().get();
+	if(instance)
+		for ( const auto& plugin : App::plugin_manager.plugins() )
+		{
+			std::string id = plugin.id;
+			if(!action_group->get_action(id))
+				action_group->add(
+					Gtk::Action::create(id, plugin.name.get()),
+					[instance, id](){instance->run_plugin(id, true);}
```
rodolforg, line 1548: The signal slot here should use[ `sigc::track_obj`](https://libsigcplusplus.github.io/libsigcplusplus/reference/html/group__track__obj.html#gad4baa7c8ace9d51344f85d121d08b2cc).
It's important to autodelete signal connection if instance object is deleted first, for example.

image

(the current docs are for sigc++ 3.0 API. Sadly libsigc++ website doesn't host docs for API 2.0 anymore)

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -1543,6 +1534,35 @@ CanvasView::init_menus()
...
+				);
+		}
+
+	// remove group if exists
+	typedef std::vector< Glib::RefPtr<Gtk::ActionGroup> > ActionGroupList;
+	ActionGroupList groups = App::ui_manager()->get_action_groups();
+	for(ActionGroupList::const_iterator it = groups.begin(); it != groups.end(); ++it)
```
rodolforg, line 1555: As you  don't need the iterator `it` itself, you can use range-based `for` here too:

```c++
for (const auto& group : groups) {
```

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,252 @@
...
+    notebook.append_page(*plugin_tab);
+    notebook.set_tab_label(*plugin_tab, *custom_label);
+    
+    notebook.set_tab_pos(Gtk::POS_LEFT);
+    get_content_area()->pack_start(notebook);
+    set_default_size(1200, 600);
+    show_all_children();
+}
+
+void Dialog_PluginManager::refresh()
```
rodolforg, line 114: style:
4. return type should be in previous line:
```c++
void
Dialog_PluginManager::refresh()
{ 
```

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,252 @@
...
+{
+    plugin_list = App::plugin_manager.plugins();
+    auto children = plugin_list_box.get_children();
+    for(auto child : children) {
```
rodolforg, line 118: style: 1. spaces after `for`

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,252 @@
...
+        plugin_list_box.add(*plugin_list_item);
+
+    }
+    plugin_list_box.show_all_children();
+}
+
+Dialog_PluginManager::~Dialog_PluginManager()
+{
+    
+}
```
rodolforg, line 178: style: move this method to be right below the constructor

synfig-studio/src/gui/dialogs/dialog_pluginmanager.h
```diff
@@ -0,0 +1,73 @@
...
+**	You should have received a copy of the GNU General Public License
+**	along with Synfig.  If not, see <https://www.gnu.org/licenses/>.
+**	\endlegal
+*/
+/* ========================================================================= */
+
+/* === S T A R T =========================================================== */
+
+#ifndef __SYNFIG_GTKMM_DIALOG_PLUGINMANAGER_H
+#define __SYNFIG_GTKMM_DIALOG_PLUGINMANAGER_H
```
rodolforg, line 29: style:
5. remove the prepended double underscore in the #define:

synfig-studio/src/gui/pluginmanager.h
```diff
@@ -127,6 +129,7 @@ class Plugin
 	PluginString description;
 
 	bool is_valid() const;
+	void launch_dir() const;
```
rodolforg, line 132: I don't think that PluginManager class should have this method. IMO it should be done in Plugin Manager Dialog

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,252 @@
...
+                                pluginStatus = PluginZipStatus::PLUGIN_ZIP_AT_CHILD_DIR;
+                                break;
+                            }
+                    }
+                if(pluginStatus == PluginZipStatus::PLUGIN_ZIP_AT_CHILD_DIR) {
+                    output_path = path_to_user_plugins.add_suffix("/").u8string();
+                    for(auto &file : files)
+                        zip_fs->copy_recursive(zip_fs, file, native_fs, output_path + file);
+                }
+                App::plugin_manager.load_plugin(output_path + files[0] + '/' + "plugin.xml", output_path +  files[0] , true);
```
rodolforg, line 237: You could create a method or function just for searching the plugin.xml file and the optional directory it could be in.
Then, you would do the extraction.


I think the installed plugin folder could be named after the plugin metadata info, instead of zip filename or zip internal folder. But it can be done later.

rodolforg (changes requested):

synfig-studio/src/gui/canvasview.cpp
rodolforg: To make myself clear: My comment above was related to lambda function as a signal slot.

I built and uploaded the docs for libsigc++ 2.12 and gtkmm 3.24 :
http://floss.rodolfo.eng.br/libsigc++/2.12.1/
http://floss.rodolfo.eng.br/gtkmm/3.24.4/

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,252 @@
...
+/* === M E T H O D S ======================================================= */
+
+Dialog_PluginManager::Dialog_PluginManager(Gtk::Window& parent):
+    Gtk::Dialog(_("Plugin Manager"), parent),
+    plugin_file_dialog("Select plugin zip file", Gtk::FILE_CHOOSER_ACTION_OPEN),
+    message_dialog(_("Are you sure you want to delete this plugin?"), false, Gtk::MESSAGE_QUESTION, Gtk::BUTTONS_OK_CANCEL, true),
+    plugin_list(App::plugin_manager.plugins())
+{
+    plugin_file_dialog.add_button("_Cancel", Gtk::RESPONSE_CANCEL);
+    plugin_file_dialog.add_button("_Open", Gtk::RESPONSE_OK);
```
rodolforg, line 74: Allow localization: surround translatable strings with `_()` (defined in gui/localization.h)

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,252 @@
...
+    plugin_tab->property_expand().set_value(false);
+    plugin_list_box.set_selection_mode(Gtk::SELECTION_NONE);
+
+    build_listbox();
+
+    plugin_tab->pack_start(plugin_list_box, Gtk::PACK_SHRINK);
+    plugin_tab->pack_end(*install_plugin_button, Gtk::PACK_SHRINK);
+
+    Gtk::Label* custom_label = manage(new Gtk::Label());
+    custom_label->set_text("Installed Plugins");
```
rodolforg, line 100: Allow localization: surround translatable strings with `_()` (defined in gui/localization.h)

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,252 @@
...
+            zip_fs->directory_scan("", files );
+            // check if the archive has a plugin.xml file at root directory
+            enum PluginZipStatus
+            {
+                PLUGIN_ZIP_NO_PLUGIN_XML = 1 << 1,
+                PLUGIN_ZIP_AT_ROOT_DIR = 1 << 2,
+                PLUGIN_ZIP_AT_CHILD_DIR = 1 << 3
+            } pluginStatus = PluginZipStatus::PLUGIN_ZIP_NO_PLUGIN_XML;
+            
+            for(auto& file : files) {
```
rodolforg, line 207: `const auto&`

rodolforg: style: space

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,252 @@
...
+                if(file == "plugin.xml") {
```
rodolforg, line 208: style: space

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -672,6 +681,44 @@ MainWindow::edit_custom_workspace_list()
...
+			"  <menubar name='menubar-main' action='menubar-main'>" + ui_info_menu + "</menubar>"
+			"</ui>";
+	plugins_action_group->set_sensitive(false);
+
+	if(!save_plugins_merge_id)
+		App::ui_manager()->insert_action_group(plugins_action_group);	
+	else {
+		typedef std::vector< Glib::RefPtr<Gtk::ActionGroup> > ActionGroupList;
+		ActionGroupList groups = App::ui_manager()->get_action_groups();
+		for(ActionGroupList::const_iterator it = groups.begin(); it != groups.end(); ++it)
```
rodolforg, line 711: here you can do that range-based `for` loop again :)

rodolforg (changes requested):

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -188,90 +247,51 @@ Dialog_PluginManager::on_install_plugin_button_clicked()
...
-                                    for (const xmlpp::Node* node : pNode->find("./name") ) {
-                                        const xmlpp::Element* element = dynamic_cast<const xmlpp::Element*>(node);
-                                        std::string lang = element->get_attribute_value("lang");
-                                        if (lang.empty()) {
-                                            dir_name = element->get_child_text()->get_content();
-                                            break;
-                                        }
-                                    }
-                                }
+                                pluginStatus = PluginZipStatus::PLUGIN_ZIP_AT_CHILD_DIR;
```
rodolforg, line 271: Instead of saving this flag `PluginZipStatus`, create a method/procedure that returns the filepath (folder + "/plugin.xml") to be used as argument to `get_plugin_name()`. This new method may be called `find_plugin_xml()` or `find_plugin_metafile()` or anything better XD.

If the file is in in the zip root folder, the filepath would be "./plugin.xml" or simply "plugin.xml".

As you remembered, user may provide an invalid zip file (without plugin.xml). In this case, the returned filepath could be empty to signalize it.

The way you refactored is more organized than before, but it scan the zip twice searching for the same file (as you did before). :) As I propose to you, `get_plugin_name()` will be reduced a lot ;)

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -179,6 +186,58 @@ Dialog_PluginManager::build_listbox()
...
+            }
+        }
+    }
+}
+std::string get_plugin_name(etl::handle<synfig::FileContainerZip> &zip_fs, const PluginZipStatus &pluginStatus, const std::vector<std::string> &files) {
+    switch (pluginStatus) {
+        case PLUGIN_ZIP_AT_ROOT_DIR:
+            for (const auto& file : files) {
+                if (file == "plugin.xml") {
+                        return extract_plugin_name(*zip_fs->get_read_stream(file));
```
rodolforg, line 218: Every time you deference a (smart or not) pointer, check if it isn't null before you do it :)
Otherwise, ... crash is possible ;)

rodolforg (changes requested):

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,290 @@
...
+                            }
+                        }
+                    if (!plugin_metadata_file.empty())
+                        break;
+                    }
+                }
+            }
+            plugin_name = extract_plugin_name(*zip_fs->get_read_stream(plugin_metadata_file));
+            output_path = ((path_to_user_plugins / plugin_name).add_suffix("/")).u8string();
+            if (native_fs->is_exists(output_path)) {
```
rodolforg, line 256: What if it is a file instead of a directory? You have to check if it is a file too.

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,290 @@
...
+                                plugin_metadata_file = file + "/" + innerFile;
+                                break;
+                            }
+                        }
+                    if (!plugin_metadata_file.empty())
+                        break;
+                    }
+                }
+            }
+            plugin_name = extract_plugin_name(*zip_fs->get_read_stream(plugin_metadata_file));
```
rodolforg, line 254: You forgot to check if `plugin_metadata_file` is empty and, if so, tell user about invalid plugin zip file.

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,290 @@
...
+                    break;
+                } else if(zip_fs->is_directory(file)) {
+                    std::vector<std::string> child_files;
+                    if (zip_fs->directory_scan(file, child_files)) {
+                        for (auto& innerFile : child_files) {
+                            if (innerFile == "plugin.xml") {
+                                plugin_metadata_file = file + "/" + innerFile;
+                                break;
+                            }
+                        }
```
rodolforg, line 248: I don't know if `std::find(child_files.begin(), child_files.end(), "plugin.xml")` would be better.
What do you think?

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,290 @@
...
+            }
+            plugin_name = extract_plugin_name(*zip_fs->get_read_stream(plugin_metadata_file));
+            output_path = ((path_to_user_plugins / plugin_name).add_suffix("/")).u8string();
+            if (native_fs->is_exists(output_path)) {
+                message_dialog.set_message(_("Plugin already exists. Do you want to overwrite it?"));
+                int response = message_dialog.run();
+                if (response == Gtk::RESPONSE_CANCEL) {
+                    message_dialog.close();
+                    return;
+                }
```
rodolforg, line 262: I recommend you to check if it is different of OK, instead of equals to Cancel.
If user closes the dialog instead of pressing Cancel, the dialog sets a different response.

rodolforg (changes requested):
Good progress! :)

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
veermetri05: You are referring to the possibility that there could a "plugin.xml" named directory (for whatever reason). Am I correct ?

rodolforg: No. I meant a file named as the same output folder name.
If plugin name is "Skel" the plugin will be installed in "Skel" folder, right?
There could be a file named "Skel" instead of a directory named so.

veermetri05: Ok, let me explain the process here so you can pinpoint where this issue could occur.
Plugins exist in C:/Users/username/Synfig/Plugins/
(This directory by default is empty.)

When a user selects a zip file, we first find the "plugin.xml" file within the folder, it could be at the root level or a child directory. Then we create a directory at the plugin installation path (in our case: C:/Users/username/Synfig/Plugins/)
So, if a plugin name is "Skel" (as per its plugin.xml) we create a directory at the installation path with the name of the plugin (so the path where the plugin exists would be  C:/Users/username/Synfig/Plugins/Skel/ ). 

(Maybe I understand it now), you are referring to the possibility that for whatever reasons, we may encounter a "Skel" named file in "C:/Users/username/Synfig/Plugins/".

rodolforg: Exactly. Sorry I couldn't express me well before :)



> You are referring to the possibility that there could a "plugin.xml" named directory

By the way, this is important too :) Nice catch

veermetri05: What should be the behavior if we encounter a file with the to-be-created directory name ? Probably just delete it. The purpose of having a plugin manager interface is also to remove the need for manual intervention of the user in the directory, so if we encounter something unexpected we should deal it through the code itself.

rodolforg: Probably ask user about it

veermetri05: It missed over my mind, we do not have any issue if there exists a file with same name as the directory we are checking, so we will leave it as is.

Only thing that is needed to do is make sure that we are checking if it is directory. 

If there exists a file and a folder with the same name (we have a file "Skel" as well as a directory named "Skel"), we should check which is a directory. I will take a look at it how should it be implemented.

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,298 @@
...
+    {
+        for (const xmlpp::Node* node : pNode->find("./name") ) {
+            const xmlpp::Element* element = dynamic_cast<const xmlpp::Element*>(node);
+            std::string lang = element->get_attribute_value("lang");
+            if (lang.empty()) {
+                return element->get_child_text()->get_content();
+            }
+        }
+    }
+}
```
rodolforg, line 212: It misses a `return` statement in case of an empty `execlist`.

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,298 @@
...
+void
+Dialog_PluginManager::on_install_plugin_button_clicked()
+{
+    // Show the dialog and wait for a user response
+    int result = plugin_file_dialog.run();
+    switch (result) {
+        case (Gtk::RESPONSE_OK):
```
rodolforg, line 219: Why a `switch`?
A simple `if (result != Gtk::RESPONSE_OK) return;` should do it :)

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,298 @@
...
+        { 
+            std::vector<std::string> files;
+            PluginZipStatus pluginStatus = PluginZipStatus::PLUGIN_ZIP_NO_PLUGIN_XML;
+            etl::handle<synfig::FileSystemNative> native_fs = FileSystemNative::instance();
+            etl::handle<synfig::FileContainerZip> zip_fs = new FileContainerZip();
```
rodolforg, line 224: Prefer `synfig::FileSystemNative::Handle` and `synfig::FileContainerZip::Handle`.

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,298 @@
...
+            for (const auto& file : files) {
+                if (file == "plugin.xml") {
+                    plugin_metadata_file = file;
+                    break;
+                } else if(zip_fs->is_directory(file)) {
+                    std::vector<std::string> child_files;
+                    if (zip_fs->directory_scan(file, child_files)) {
+                    auto it = std::find(child_files.begin(), child_files.end(), "plugin.xml");
+                    if (it != child_files.end()) {
+                        plugin_metadata_file = file + "/" + *it;
```
rodolforg, line 245: maybe directly `plugin_metadata_file = file + "/plugin.xml"` ?

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
```diff
@@ -0,0 +1,298 @@
...
+                if (!native_fs->file_remove(output_path)) {
+                    synfig::error("Failed to remove file: " + output_path);
+                    return;
+                }
+            }
+            if (native_fs->directory_create(output_path)) {
+                if (plugin_metadata_file.find("/") == std::string::npos) {
+                    for(const auto& file : files) {
+                        zip_fs->copy_recursive(zip_fs,  file, native_fs, output_path + file);
+                    }
```
rodolforg, line 282: Why do you  want to extract all files from zip here?

rodolforg (changes requested):

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
veermetri05: The zip contains the files for plugin, and when user installs it, those files will be installed in the plugins directory for the user.

(This should be highly obvious, am I missing something from the question?)

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
rodolforg: AFAIK, you can't have a file and a folder with same name at same location:
https://superuser.com/questions/1639147/does-windows-explorer-allow-the-creation-of-duplicate-folder-names-in-the-same-l
https://askubuntu.com/questions/1024781/a-folder-and-a-file-with-the-same-name-in-the-same-location
https://unix.stackexchange.com/questions/22447/why-cant-i-have-a-folder-and-a-file-with-the-same-name

synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp
rodolforg: You plan to support two plugin zip file layouts:
1. A 'plain' zip file, where 
2. A zip file that contains a folder that contains the metadata plugin.xml and all its other files

The 'its other files' may include folders too.

Suppose a badly structured zip file with this layout:
- a folder called `skel` (in the zip root), that contains:
  - plugin.xml
  - skel.py
- another folder (in the zip root) named `trash`
  - trash.txt
  - whyonhell.txt

If I read the code correctly, your code will extract both folders in the plugin folder you created. You should extract the `skel` folder contents only. Or am I wrong?

veermetri05: There are some corrections in your understanding. 
The above pointed code is only executed when if the "plugin.xml" file is at root level. If the "plugin.xml" is at sub directory then the below code is executed:
https://github.com/synfig/synfig/blob/e985c198cc1d53814c62e55e43fe8ef738367a4b/synfig-studio/src/gui/dialogs/dialog_pluginmanager.cpp#L283-L285

(Which requires some correction, because it assumes that the zip file contains only one directory)

synfig-studio/src/gui/pluginmanager.cpp
```diff
@@ -196,14 +315,15 @@ fetch_data_in_widget(const Gtk::Widget* w, std::map<std::string, std::string>& d
 			data[w->get_name()] = static_cast<const Gtk::Entry*>(w)->get_text();
 		}
 	}
-	if (GTK_IS_CONTAINER(w->gobj())) {
+	if (GTK_IS_CONTAINER(w->gobj()) && (w->get_name().find("gtkmm__") == 0)) {
```
rodolforg, line 318: @veermetri05 do you recall why do you filter the container name?

veermetri05: I can't recall exactly why I did it, gtkmm_ is inserted for widgets that are programmatically inserted. I will revert it.

PR #3380, opened 2024-07-15
fix: proper reset of bonesetup size in WorkArea

rodolforg:


rodolforg:
This one is just a single copy'n paste mistake from another file.

ice0 (approved):

PR #3384, opened 2024-07-23
perf: minor optimization of Bevel layer

rodolforg:


morevnaproject:
I have tested with this file (rendering from CLI, 120 frames PNG at 1920x1080). Unfortunately, there are no difference for speed.
[bevel-test-1.sif.zip](https://github.com/user-attachments/files/16421867/bevel-test-1.sif.zip)

ice0:
I have tested it too using profiler, here is the result:

**Before**
before

**After**
after

Wow! Awesome result!

This is the total time part:

**Total time (before):**
sum-before

**Total time (after):**
sum-after

**CPU time reduced from 24.7 secs to 0.273 secs!!! Amazing!**

The bad thing is that the total time is the same :(

The issue as I already told to @morevnaproject is some sort of bad scheduling in threads, they are just waiting each other doing nothing.
Unfortunately I didn't find how to fix this for now, but when we find the issue the result will be bold.

ice0 (approved):

ice0:
Merged. Thank you!

ice0:
Sorry guys, I didn't test twice before posting the results.
For some reason profiler didn't get correct data for the first time :/

Here's an updated comparison of before and after results:

compare-result

Mostly unchanged, but cache misses reduced to 30% so I still think this is a good result!

PR #3394, opened 2024-08-04
refactor: clean up Bevel layer

rodolforg:
Maybe a minor performance improvement because it now does a single memory allocation (instead of 4) per Blur call.

### DIFF
--- a/synfig-core/src/modules/lyr_std/bevel.cpp
+++ b/synfig-core/src/modules/lyr_std/bevel.cpp
@@ -38,17 +38,8 @@
 #include <synfig/localization.h>
 #include <synfig/general.h>
 
-#include <synfig/string.h>
-#include <synfig/time.h>
+#include <synfig/blur.h>
 #include <synfig/context.h>
-#include <synfig/misc.h>
-#include <synfig/paramdesc.h>
-#include <synfig/renddesc.h>
-#include <synfig/surface.h>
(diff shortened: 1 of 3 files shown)


morevnaproject:
I have tested this PR and it looks good - it gives identical results comparing to previous commits.

There is just one test with Disc Blur type that looks slightly different, but the difference is very hard to notice with unarmed eye.
[bevel-depth12-soft6-angle135-luma0-solid0-blur4-onto.sif.zip](https://github.com/user-attachments/files/16581175/bevel-depth12-soft6-angle135-luma0-solid0-blur4-onto.sif.zip)
Before PR:
bevel-depth12-soft6-angle135-luma0-solid0-blur4-onto-1
After PR:
bevel-depth12-soft6-angle135-luma0-solid0-blur4-onto-2

I haven't tested performance change, though.

I think it is good to merge!

morevnaproject (approved):

rodolforg:
> There is just one test with Disc Blur type that looks slightly different, but the difference is very hard to notice with unarmed eye.

fixed with [6f26fba](https://github.com/synfig/synfig/pull/3394/commits/6f26fbaaac87632e73c1948718621e6c1c85b0dd)

morevnaproject (approved):
I have tested - the results now identical with 1.5.1! Ready to merge. Thank you!

PR #3399, opened 2024-08-10
chore: Fix generation of source packages using Autotools

morevnaproject:
The "./autobuild/synfigstudio-release.sh" command was failing without this fix.


morevnaproject:
#3395 is failing without this fix

ice0:
Automake did not install `depcomp` because there are no more compilation dependencies in ETL

ice0 (approved):

morevnaproject:
> Automake did not install `depcomp` because there are no more compilation dependencies in ETL

Yes, this started from a62438c4dc9f14506e6e3587ff0fc78be215bf02 (part of #3233)

PR #3412, opened 2024-08-23
fix: drag-n-drop import when filename contains non-latin chars

### LINKED ISSUE #3409: Unable to import image via drag-n-drop
If image filename contains non-latin characters, then it is unable to import it via drag-n-drop.

Demo:
Peek 2024-08-22 16-19

Here is a sample file with non-latin characters -
[006_уличный двор_02.png.tar.gz](https://github.com/user-attachments/files/16708331/006_._02.png.tar.gz)

Tested on master branch - 78e866cde56d844b1d2b66d83b5de7249788d0ab

rodolforg:
wrong hexadecimal to decimal conversion

fix #3409


morevnaproject (approved):
Works perfectly! Thank you!

PR #3416, opened 2024-09-06
fix: change convert.exe syntax to avoid clash in Windows

rodolforg:
On Windows (I built with CMake) for some weird reason calling ImageMagick's convert calls Windows/System32/convert.exe . It's odd because my Path env var seems to list the mingw64 directory first, but a solution is using the 'magick' syntax, where you just call 'magick' to convert as per https://imagemagick.org/script/command-line-tools.php

The problem is widely known as per https://stackoverflow.com/questions/3060205/error-invalid-parameter-fom-imagemagick-convert-on-windows , I was getting the same 'Invalid Parameter' error

The patch uses 'magick convert' (legacy syntax), instead of 'magick' for compatibility with ImageMagick 6, since many distros don't have official packages for ImageMagick 7, despite the release being over 5 years old:

    https://packages.debian.org/search?searchon=sourcenames&keywords=imagemagick
    https://packages.fedoraproject.org/pkgs/ImageMagick/ImageMagick/
    https://packages.ubuntu.com/search?keywords=imagemagick

The reason is that v7's syntax is not fully compatible with v6 but it our syntax seems safe

TL;DR:
Instead of using ImageMagick with 'convert' use 'magick convert' since Win has another 'convert' program

Supersedes: #2951

### DIFF
--- a/synfig-core/src/modules/mod_imagemagick/mptr_imagemagick.cpp
+++ b/synfig-core/src/modules/mod_imagemagick/mptr_imagemagick.cpp
@@ -99,17 +99,35 @@ imagemagick_mptr::get_frame(synfig::Surface &surface, const synfig::RendDesc &re
 		}
 	}
 
-	OS::RunArgs args;
-	args.push_back(filename);
+	bool success = false;
+	// pair (binary name, first argument)
+	std::vector<std::pair<std::string, std::string>> binaries {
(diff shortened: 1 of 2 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

PR #3417, opened 2024-09-08
refactor: add move constructor to handle/loose_handle/rhandle

rodolforg:
Split from PR #3071 . That one will be only about moving last ETL files to synfig-core.

### DIFF
--- a/ETL/ETL/_handle.h
+++ b/ETL/ETL/_handle.h
@@ -53,13 +53,24 @@ template <class T> class handle;
 template <class T> class loose_handle;
 template <class T> class rhandle;
 
-
-// ========================================================================
-/*!	\class	shared_object _handle.h	ETL/handle
-**	\brief	Shared Object Base Class
-**	\see handle, loose_handle
-**	\writeme
-*/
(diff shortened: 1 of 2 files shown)


ETL/ETL/_handle.h
```diff
@@ -385,16 +496,34 @@ class rhandle : public handle<T>
 		if(obj)add_to_rlist();
 	}
 
-	//! Default copy constructor
+	/** Copy constructor */
 	rhandle(const rhandle<value_type> &x):handle<T>(x)
 	{
 		if(obj)add_to_rlist();
```
rodolforg, line 502: I wonder if the `rhandle::next_`, `rhandle::prev_`, `rshared_obj::front_` and `rshared_obj::back_` should be 'protected' by a mutex.

ice0: I think yes. Since it is possible to access the same object from multiple threads, it is a good idea.

ice0 (approved):

ice0:
Merged. Thank you!

PR #3428, opened 2024-09-29
fix: handle (Radial)Composite valuenodes via Widget_Curves

### LINKED ISSUE #3426: Allow manipulation of tangent waypoints in Graphs panel
### Description
In "Graphs" panel we can manipulate waypoints of all the default types (vector, real, etc.) but Radial Composite isn't one of them. Therefore, when you try to move tangent's waypoints in the Graphs you get "Add waypoint: Action is not ready" error:
ActionIsntReady

### To Reproduce
1. Create an outline
2. Move one of the tangents in the animated mode - a waypoint is created
3. Select the tangent you animated in the Parameters panel
4. Go to Menu->Window->Graphs
5. Try to move a waypoint, you'll get the above error

### Expected behavior
Being able to manage waypoints that have Radial Composite type in the Graphs panel. I can understand the problem of moving waypoints if valuenode has some user conversions but in case of Radial Composite it's what Synfig itself saves tangents in by default so it's reasonable to expect them to work.

rodolforg:
fix #3426

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_curves.cpp
+++ b/synfig-studio/src/gui/widgets/widget_curves.cpp
@@ -77,6 +77,47 @@ using namespace studio;
 
 /* === P R O C E D U R E S ================================================= */
 
+static bool
+is_a_kind_of_composite_value_node(const ValueNode::Handle& value_node)
+{
+	return value_node && (value_node->get_name() == "composite" || value_node->get_name() == "radial_composite");
+}
+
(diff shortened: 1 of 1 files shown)


rodolforg:
Do you think I have to rebase just to check in macOS13?

ice0:
Yes, please rebase.

ice0:
Merged. Thank you!

PR #3446, opened 2024-10-29
feat: Add gdb as a package to be installed during setup

shreyaspranav:
Description:
This PR enhances the setup script by including `gdb` as a package to be installed during the initial setup. This addition simplifies the installation process for developers by ensuring that `gdb` is readily available for debugging right after the environment is set up.

`gdb` will now be installed while running the 1-setup-* shell scripts in Windows(MSYS2), Linux and OSX.


mohamedAdhamc:
Does this really need to be added though ? The way I see it is the the setup scripts include the required packages for the Synfig to run. Any additional things for debugging, performance analysis, etc... are better left for everyone to download what they need. AFAIK I don't even remember gdb explicitly being mentioned in the dev docs. 

However, that's just my opinion... Let's see what others think.

Issue #3292, opened 2023-12-19
Feature request: Shift + left click on color palette to set outline color


yoSachinkr:
`Left clicking` on any color in the color palette currently changes the fill color. In a similar manner, if `shift+left clicking `on any color in the color palette would change the outline color, it would be more productive.

Afterdie:
Current binding is on middle click. Are you suggesting a shortcut change?

yoSachinkr:
@Afterdie Yes 😊. I think that's more intuitive.

mohamedAdhamc:
> Current binding is on middle click. Are you suggesting a shortcut change?

is it not a customizable shortcut (can't remember)? if it's not I think it may be better to just make it a user customizable shortcut...

yoSachinkr:
> > Current binding is on middle click. Are you suggesting a shortcut change?
> 
> is it not a customizable shortcut (can't remember)? if it's not I think it may be better to just make it a user customizable shortcut...

Yeah, i agree. Making it customizable is a good idea.

shreyaspranav:
Hi, can I work on this issue?

mohamedAdhamc:
> Hi, can I work on this issue?

Sure :).

shreyaspranav:
Dear Maintainers,

I have successfully changed the binding from middle click to shift click. If possible, could someone guide me on how to make the shortcut customizable? Any assistance would be greatly appreciated.

Thank you for your help!

Best regards,
Shreyas.

mohamedAdhamc:
> I have successfully changed the binding from middle click to shift click. If possible, could someone guide me on how to make the shortcut customizable? Any assistance would be greatly appreciated.

Hello sherya,

 I suggest you take a look at `app.cpp`, speciifically the `get_default_accel_map()` method. There you can find all the shortcuts/accelerators. Try to understand how they work by tracing things. You will probably have to know a bit of gtk if you're not already familiar with it. But this should lead you into the right direction. Best of luck!

shreyaspranav:
Dear Maintainers,

I've been working on making the mouse binding customizable for a while now and here's how it went:

as @mohamedAdhamc mentioned to look at `get_default_accel_map()` method, I tried to add the shortcut there but I realized that mouse clicks cannot be used as "accelerators" (correct me if I am wrong). Therefore I made a seperate `std::map` object:

``` cpp
// const char* -> Name of the action
// std::pair<GdkModifierType, int> -> Pair of modifiers and mouse button
[...]

PR #3448, opened 2024-11-02
feat: Color Palette Improvements.

shreyaspranav:
## Description

Earlier, the shortcut for selecting the outline color from the palette was middle click, which is unintuitive. Therefore the binding is changed to shift click instead of middle click. This is also customizable via **Edit->Preferences->Shortcuts** as shown below:

Mouse Shortcuts

This PR addresses the issue #3292 

## Features Implemented in Code
A seperate tree view that displays mouse bindings similar to keyboard shortcuts is implemented with the ability of modifying mouse shortcuts.

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -218,6 +218,9 @@ bool App::shutdown_in_progress;
 
 Glib::RefPtr<studio::UIManager>	App::ui_manager_;
 
+// Mouse Bindings:
+std::map<synfig::String, std::pair<GdkModifierType, int>> App::mouse_bindings_;
+
 int        App::jack_locks_ = 0;
 synfig::Distance::System  App::distance_system;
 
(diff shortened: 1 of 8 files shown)


shreyaspranav:
Dear Maintainers, 

There are a few things to work on this PR:

1. For some reason, this mouse button callback function doesn't register single clicks. It only registers double and triple clicks:
https://github.com/synfig/synfig/blob/75a13c5cf11f73df1c132358230f960774057344/synfig-studio/src/gui/dialogs/dialog_setup.cpp#L672

I attempted to fix the issue, but unfortunately, I was not able to resolve it. Therefore instead of single-clicking, one must double click to edit the mouse binding. Can any one kindly point out what was wrong with my implementation?  

2. Just like keyboard shortcuts, is there a need to add a feature to save mouse binding settings to disk? This feature is currently not implemented, but will be implemented upon request.

If there are any additional changes to be done related to this PR then feel free to ask. I will try to reply ASAP.

I kindly appreciate your feedback.

Best, 
Shreyas.

shreyaspranav:
Dear Maintainers,

I would request you to please provide feedback for the above pull request, it would help me in making sure that everything is in line with the project standards

Best, 
Shreyas.

PR #3452, opened 2024-11-15
refactor: minor cleanup on Plant layer

rodolforg:


rodolforg:
@Ahmed-Khaled-dev  @ice0  @mohamedAdhamc What does this error in CI mean?
`Show Pull Request Change Note / Show Pull Request Change Note (pull_request) `
image

Ahmed-Khaled-dev:
It means that the GitHub workflow doesn't have enough permissions to comment on the pull request. This is my oversight. All the tests I did were with creating pull requests from branches in the main repository, but here, all pull requests are made from forks, which behave differently in terms of permissions.

After further research, I figured out the bug, how to fix it, and created a pull request to add this fix (#3453), it's a simple one-liner fix, more details are in the PR.

As soon as #3453 is merged if you rerun the workflow here it should work as expected :)

I am assuming that the required permissions are already enabled in the Synfig settings. If not, then it still won't work, we can either do the below edit in the Synfig settings, or if you would not prefer to do that I can edit show-pr-change-note to specify the specific permissions that it needs in it.

In Settings -> Actions -> General

image

ice0 (approved):

ice0:
Merged. Thank you!

PR #3454, opened 2024-11-16
feat: allow to switch to previous or favorite workspace

rodolforg:


morevnaproject:
image
I would like to note that. default workspaces are not shown in Workspaces editor dialog.
Thus, it is not possible to set any of them as Favorite.
Is it possible to add them in the list?

morevnaproject:
Issue:
1. Create a custom workspace and mark it as Favorite.
2. Switch to default workspace (or any other than Favorite workspace)
3. Press Shift+F11 to switch into Favorite workspace
4. Press Shift+F11 again to switch into Favorite workspace
5. Press Ctrl+F11 to switch into Previous workspace. -> This wouldn't work, because we pressed Shift+F11 twice on steps 3 and 4.

morevnaproject (changes requested):
Please see my previous two comments

PR #3456, opened 2024-11-22
fix(ci): package installation failure in show-pr-change-note

Ahmed-Khaled-dev:
This fixes the new error below, I did a lot of testing and this never occurred, so I am not exactly sure why this happened now, but generally, this fixes it.

Screenshot from 2024-11-22 11-37-11

It turns out GitHub generally advises always adding `sudo apt-get update`, as mentioned in their [docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners).

This will make the workflow work correctly as expected in new PRs and for the current PRs in which the workflow failed, rebasing the PR's branch and force-pushing will fix the workflow, and if any edits in the title or description in the current PRs were made the workflow should work as expected 👌🏻


Ahmed-Khaled-dev:
show-pr-change-note is currently working great with all new PRs, but still merging this PR should ensure it never fails.

PR #3458, opened 2024-11-28
fix: prevent potential null pointer dereference in ValueDescDisconnect action

rodolforg:
Also this fixes MSYS2 MinGW warning:

```bash
/synfig-studio/src/synfigapp/actions/valuedescdisconnect.cpp:122:104:
C:/msys64/mingw64/include/c++/14.2.0/bits/atomic_base.h:501:31: warning: 'unsigned int __atomic_load_4(const volatile void*, int)' writing 4 bytes into a region of size 0 overflows the destination [-Wstringop-overflow=]
  501 |         return __atomic_load_n(&_M_i, int(__m));
      |                ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
In static member function 'static bool synfigapp::Action::ValueDescDisconnect::is_candidate(const synfigapp::Action::ParamList&)':
cc1plus.exe: note: destination object is likely at address zero
```


ice0 (approved):

ice0:
Merged. Thank you!

PR #3462, opened 2024-11-30
refactor: port InsideOut to Cobra renderer

rodolforg:

PR #3461, opened 2024-11-29
refactor: port Plant layer to Cobra renderer

rodolforg:
Simple port, not trying to optimize anything (if it's possible :P )

PR #3457, opened 2024-11-28
refactor: port Bevel layer to Cobra engine

rodolforg:

PR #3463, opened 2024-11-30
refactor: port Noise Distort layer to Cobra renderer

rodolforg:


rodolforg:
Pending Cobra ports:

1. SphereDistort
2. Twirl
3. Halftone2 (#3300 - it has an issue to be solved yet)
4. Halftone3 (#3300 - it has an issue to be solved yet)
5. RadialBlur (#3464)
6. CurveGradient (#3307)
7. InsideOut (#3462)
8. Noise (#3306)
9. Bevel (#3457)
10. Plant (#3461)
11. Julia
12. Mandelbrot

Svarov-RZM:
Wow, I see someone is busy! Are these ports ready for testing or is still work in progress?

rodolforg:
Ready for testing :) (except Halftone 2 and 3, that Konstantin already catch errors I'll investigate later)

Svarov-RZM:
Alright, I did some heavy testing on "Noise Distort" and got good news and bad news.
Good news are the distortion on regular layers (outlines, regions) works well. Here are the results on my vase example from inside out test (isolated from the gradient):
NDcomparision

It differs from the 1.0.2 version but in a very subtle way. You can see that shading on the flower is preserved in current branch better while in 1.0.2 version it is barely visible. Everything's fine here if you ask me.

The gradients though... that's bad news.
It produces the following results:
Current branch:
2TestND-GR

1.0.2 version:
2TestND-GR-102

You can see the gradient is shifted towards bottom-left of the screen, pay attention to magenta border like it is out of boundaries. And again, it looks fine in the canvas view, but when you render/preview it, the gradient bugs out.

And I picked the "Noise Gradient" because the effect is most noticeable there, but it does happen on all gradients, the shift is just very small (not even sure it can be considered an issue there), you can only see it when you render two files and switch between them in image viewer.

So, shift in "Noise Gradient": definitely a problem, perhaps it's what breaks rendering in "Inside Out" layer
Shift in other gradients: Probably just me nitpicking, it is very small (check the attached zip file for rendered pngs, so you can see for yourself)

For additional example, let's see how "Conical Gradient" looks with the same distortion:
Current branch:
3TestND-GR2

1.0.2 version:
3TestND-GR2-102

The shift is there, but as I said, you can't see it with unarmed eye. But you do see this ugly black corner. Which leads me to believe that there's also not enough space for algorithm to properly account for "limitless" nature of gradients. You mentioned something about "rendered underlayer", maybe it should be rendered at a slightly bigger resolution so the corners of the gradient won't get distorted? Just my two cents, I don't know how 1.0.2 handles this and why it doesn't produce those black corners.

All test case files used in this along with rendered pictures:
[TestND.zip](https://github.com/user-attachments/files/18046390/TestND.zip)

rodolforg:
@Svarov-RZM thank you for testing and pointing me the problems!

> pay attention to magenta border like it is out of boundaries.

Yes, I make painting magenta when this happens :)



> it looks fine in the canvas view, but when you render/preview it, the gradient bugs out.

I can't understand why this happens. Final rendering should be the same in CLI and on canvas view with "Final" render selected.



> maybe it should be rendered at a slightly bigger resolution so the corners of the gradient won't get distorted

The algorithm does so. Maybe I'm computing it wrong... But I was pretty sure about hot to calculate the size of underlayer for Noise Distort......


Could you please test the Noise Gradient also? Because I can see some differences between the two pics you posted besides the image shifting.

Svarov-RZM:
> Could you please test the Noise Gradient also?

Hey, I remember that. I didn't really test it except for that super sample option.
I will take a look tomorrow.

rodolforg:
I fixed the shift of the second image (but I didn't pushed the commit).

However now it gives the same issue as the last image... I'll keep investigating...

rodolforg:
@Svarov-RZM Fixed that issue with black-pixels on the image edges :)

Svarov-RZM:
Works well, couldn't find any issues with various combinations of layers, besides the obvious issue of 1px shift.

Benchmarks (average of 3, -b argument):
1.0.2 ver: 1.26322 seconds
Current: 0.576 seconds

Test case file: 
[TestND-GR.zip](https://github.com/user-attachments/files/18314393/TestND-GR.zip)

rodolforg:
@Svarov-RZM @Zloi-KoteikA @morevnaproject @ice0 fixed :)

The CurveWarp has this issue too; I can fix it once this PR is merged as it is (please do not squash)

--- Edit ---
I pushed the wrong branch! 
Pushed the correct one 😉

Svarov-RZM:
Oh, you fixed the shift? I am on it. Will report back on weekends, most likely...

Svarov-RZM:
Hey, @rodolforg 

As promised, I did some tests, and the good thing is - the shift is gone, yay!
But... there's a problem when size of distort is very small, it produces magenta colored spots. You can grab an example from my latest Sonic animation: https://forums.synfig.org/t/sonic-the-hedgehog-intro-parody/16881

The waterfall example from above link looks like this:
ND-MG

You can see there are a bunch of magenta lines visible in canvas view (they are also present when rendered).

rodolforg:
I can't access a pc right now, but... without the last commit, would it work with your fix?

Svarov-RZM:
> but... without the last commit, would it work with your fix?

Well, sort of... not.
It changes things, but the problem is still there. Without any fixes (head at commit 7ff2ae0cec1b2b5f28867aaa42060715cbebae2c) the shift in Noise Distort is diagonal (1px up and 1px right). With my fix the shift is just 1px up.

PR #3470, opened 2024-12-15
fix: regression: couldn't edit color or gradient of multiple layers

### LINKED ISSUE #3469: Layer color doesn't change simultaneously when multiple layers are selected
**Describe the bug**
When multiple layers are selected and you change color, the color is applied only in the upper layer instead of all of them. 
**To Reproduce**
1. Create 2 or more circles.
2. Select all of them
3. Change layer color
4. See that the color changes only in the upper layer.

**Expected behavior**
The same color should be applied in every selected layer. 

**System information:**
 - OS: Linux 6.1.1.110-1.el7.3 x86_64 x86_64
 - OS version: RED OS, MATE 1.28.1
 - Commit ID: Synfig-Studio Master branch.

**Additional comments**
It does work in the version 1.5.3, but not in the master branch version. The other parameters, such as radius, work as expected.

rodolforg:
Introduced when fixing #3288 ( 483b5ed - PR 3373 ).

By the way, both commits fix another case:

(In Synfig <= 1.5.3)
1. Create a text layer
2. Create a star layer
3. Select the text layer
4. Click on color parameter of text layer to open Color editor
5. Select both layers (text and star)
6. Change the color in the previously opened Color editor
7. It changes the text color, but yells an error to star

The problem does not happen if you open color dialog after select both layers (step #5 above)

fix #3469

### DIFF
--- a/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
+++ b/synfig-studio/src/gui/cellrenderer/cellrenderer_value.cpp
@@ -574,58 +574,95 @@ void
 CellRenderer_ValueBase::gradient_edited(synfig::Gradient gradient, Glib::ustring path)
 {
 	synfigapp::ValueDesc vd = App::dialog_gradient->get_value_desc();
-	if (vd.is_valid()) {
-		if (vd.parent_is_layer()) {
-			if (is_layer_deleted(vd.get_layer()))
-				return;
-		}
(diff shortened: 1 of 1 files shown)


Zloi-KoteikA:
Hello. I checked it. Everything works. Thank you.

morevnaproject (approved):

PR #3476, opened 2025-01-05
refactor: port Julia layer to Cobra renderer

rodolforg:


rodolforg:
@Zloi-KoteikA could you please try it? :)

PR #3477, opened 2025-01-06
fix: inverted interpolation of gradient since 1.3.11

rodolforg:
AFAIK it only affects Plant, Mandelbrot and Metaballs layers.
All the others use CompiledGradient, that doesn't suffer the issue.

Gradient::operator() was getting the values in wrong direction
(except for the gradient start and end positions)
@ 0.0 - first color (OK)
@ 1.0 - last color (OK)
@ 0.3 - closer to last color instead of first one
          (it looks like how it should be if position was 0.7)
@ 0.7 - closer to last color instead of first one
          (it looks like how it should be if position was 0.3)

For 3-or-more-color gradients, each interval is reversed,
but the start and end color are correct...

So the solution was to create an intermediate gradients that
keep the start and end color points, but create swapped color
points for each pair of them.

Consider a 3-color gradient like this:

Color A @0.0 -> Color B @0.3 -> Color C @1.0

, where @ means the color point position.

It is transformed to:

Color A @0.0 -> Color B @0.0 -> Color A @0.3 -> Color C @0.3 -> Color B @1.0 -> Color C @1.0

And it works fine.

Tests were created to check this inversion.

Layer versions were bumped, so it won't depend anymore about the Canvas version on load.

### DIFF
--- a/synfig-core/src/modules/lyr_std/mandelbrot.cpp
+++ b/synfig-core/src/modules/lyr_std/mandelbrot.cpp
@@ -59,7 +59,7 @@ SYNFIG_LAYER_INIT(Mandelbrot);
 SYNFIG_LAYER_SET_NAME(Mandelbrot,"mandelbrot");
 SYNFIG_LAYER_SET_LOCAL_NAME(Mandelbrot,N_("Mandelbrot Set"));
 SYNFIG_LAYER_SET_CATEGORY(Mandelbrot,N_("Fractals"));
-SYNFIG_LAYER_SET_VERSION(Mandelbrot,"0.2");
+SYNFIG_LAYER_SET_VERSION(Mandelbrot,"0.3");
 
 /* === P R O C E D U R E S ================================================= */
 
@@ -103,7 +103,8 @@ Mandelbrot::Mandelbrot():
 	param_gradient_loop_inside(ValueBase(true)),
 	param_gradient_outside(ValueBase(Gradient(Color::alpha(),Color::black()))),
 	param_gradient_offset_outside(ValueBase(Real(0.0))),
-	param_gradient_scale_outside(ValueBase(Real(1.0)))
+	param_gradient_scale_outside(ValueBase(Real(1.0))),
+	param_broken_gradient(ValueBase(bool(false)))
 {
 	param_iterations=ValueBase(int(32));
 
@@ -173,6 +174,7 @@ Mandelbrot::set_param(const String & param, const ValueBase &value)
 	  }
 	  );
 
+	IMPORT_VALUE(param_broken_gradient);
 	return false;
 }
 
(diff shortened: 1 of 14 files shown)


morevnaproject:
Looks good to me! Please resolve merge conflicts.

rodolforg:
Done!
@Svarov-RZM , coudl you test this PR too please?

Svarov-RZM:
> coudl you test this PR too please?

Sure, will be done on weekends.

Svarov-RZM (dismissed):

Svarov-RZM:
Alright, I tested all 3 affected layers, gradient works as expected now (compared to 1.0.2).

@rodolforg A tiny question, is it worth refactoring these layers to use CompiledGradient so we have uniform behavior across all layers that use gradients? Or is there a reason why Plant and friends has to be different? I am going to merge this PR regardless, just curious.

rodolforg:
Well, that's the intent, but I needed to make sure the old and current SIF files load properly as intended by artist when it created the document.
I realize these errors when trying to port those layers to Cobra renderer.

rodolforg:
Rebased.

rodolforg:
One doubt I have is: should I name the newly created parameter (invisible in GUI) as `param_invert_gradient` or change it to `param_broken_gradient`? XD

Svarov-RZM:
> One doubt I have is: should I name the newly created parameter (invisible in GUI) as `param_invert_gradient` or change it to `param_broken_gradient`? XD

It's not a big deal, but yeah, I think `param_broken_gradient` fits better. `param_invert_gradient` sounds like a potential feature where a user could invert the gradient without editing the color marks (not sure if it's useful though).

Actually, I feel like you care a bit too much...
If I were you, I would omit the logic for loading old files created in broken versions completely. There's just not much work that use Plant layer and almost none that use other two.

P.S. Judging by the commit structure you want this PR to be merged as is, not squashed, correct?

rodolforg:
> I think `param_broken_gradient` fits better. 

Done. Updated also the comment I wrote about this being broken until 1.5.3 - current version is now 1.5.5 XD

> Actually, I feel like you care a bit too much... If I were you, I would omit the logic for loading old files created in broken versions completely. There's just not much work that use Plant layer and almost none that use other two.

Yes, maybe it was excessive preciousness. But Synfig is so badly viewed by regular users (oh my, user can't simply drag a 'shape' XD), I didn't want to break any existent file. Plant layer was used by Khemardi in some tutorials like this one: https://www.youtube.com/watch?v=mOt4l-Jb-ss&list=PLjuma0nCUc49ouTXvF9DtqBAOvMKjQLhF&index=119 .

> P.S. Judging by the commit structure you want this PR to be merged as is, not squashed, correct?

Yes, I prefer to be merged as is.

Svarov-RZM (approved):

Svarov-RZM:
All good, merging...
Thanks @rodolforg!

PR #3485, opened 2025-01-30
refactor: port XORPattern layer to Cobra renderer

rodolforg:


rodolforg:
@Zloi-KoteikA could you please test this port to Cobra? :)

Zloi-KoteikA:
@rodolforg 
Conducted tests of the XOR pattern layer. According to the results, there is a displacement of the finished image in the Synfig PR version.
I have not been able to determine exactly which way the shift is going. As a test, I changed the value of the "size" parameter, starting from 0.1. The result will be the same for any value.

I also tried using other file extensions (PNG, JPEG). In version 1.0.2, it was not possible to create the bmp format, so I'll check it later.
V.PR
xor-patern-size-0 1

V.1.0.2
xor-patern-size-0,1(old)



Peek 2025-03-24 17-14

rodolforg:
@Zloi-KoteikA I finally correct it: :smile:

Zloi-KoteikA:
@rodolforg Hello. Sorry for the long wait for the tests. I checked and everything is fine. The results in versions PR and 1.0.2 are the same.

morevnaproject (approved):

Issue #3493, opened 2025-02-13
Random Crashes on Version 1.5.3


Rdc19:
**Describe the bug**
Synfig keeps crashing like 5 times in 30 mins, either when I'm moving way points or moving a group, and at times when I playback my animation. 

**Expected behavior**
Basic tasks should work smoothly 

**System information:**
 - OS: [Windows]
 - OS version: 11
 - Synfig version: 1.5.3

ice0:
Hi!
Can you record the workflow so that we can see at what point the program crashes?
This will help us reproduce your situation.

Rdc19:
https://github.com/user-attachments/assets/5c23fa4e-3195-4606-8c46-33a881f3d93a

It occurs more often during playback

BobSynfig:
Please try to activate pre-rendering before playing.
From the Timeline we can see that not all the frames are generated, maybe your computer is not powerful enough for realtime.
What are its characteristics?

Note: be sure that you activated the usage of the dedicated GFX card in Windows settings for Synfig, it may help using the GPU instead of the CPU

Rdc19:
> Note: be sure that you activated the usage of the dedicated GFX card in Windows settings for Synfig, it may help using the GPU instead of the CPU

How do I activate this, I can't seem to find it

BobSynfig:
Please see here
https://www.xda-developers.com/how-force-app-use-discrete-gpu-windows-11/

Of course, you need to have a dedicated GFX card to modify this setting, maybe your computer doesn't have one

Rdc19:
> Of course, you need to have a dedicated GFX card to modify this setting, maybe your computer doesn't have one
I've set it, my laptop has a dedicated GPU. 
I am using the stable version of Synfig now, but I'll try reverting to the development version now, I hope the crashing stops

Rdc19:
The development version keeps crashing, but not on the stable version. I've switched to the stable version for now, although I prefer the features in the development version

rodolforg:
Could you share your sif file?
This way we could verify if it happens on Linux too and then be avle to investigate the issue.

Rdc19:
Here You go
https://www.mediafire.com/file/i2favmolqusnax0/Chat.sifz/file

rodolforg:
Thank you.
I noticed the file depends on 15 external png files.
Could you send us too? You can zip them and upload it directly here.

If they are private you can send me in a DM/PM via synfig forums.

Rdc19:
[...]

PR #3496, opened 2025-02-22
fix: possible double free due to multithreading with Contour

rodolforg:
Random crashes :'(

fix https://github.com/synfig/synfig/issues/3493

### DIFF
--- a/synfig-core/src/synfig/rendering/common/task/taskcontour.h
+++ b/synfig-core/src/synfig/rendering/common/task/taskcontour.h
@@ -53,7 +53,7 @@ public:
 	virtual Token::Handle get_token() const { return token.handle(); }
 
 public:
-	Contour::Handle contour;
+	std::shared_ptr<Contour> contour;
 	Real detail;
 	bool allow_antialias;
 	Holder<TransformationAffine> transformation;
--- a/ETL/ETL/_handle.h
+++ b/ETL/ETL/_handle.h
@@ -99,8 +99,7 @@ public:
 	 */
 	virtual void unref() const
 	{
-		--refcount;
-		if (refcount == 0)
+		if (--refcount == 0)
 			delete this;
 	}
 
--- a/synfig-core/src/modules/lyr_std/perspective.cpp
+++ b/synfig-core/src/modules/lyr_std/perspective.cpp
@@ -1113,7 +1113,7 @@ Perspective::build_rendering_task_vfunc(Context context) const
 	
 	if (clip) {
 		rendering::TaskContour::Handle task_contour(new rendering::TaskContour());
-		task_contour->contour = new rendering::Contour();
+		task_contour->contour = std::make_shared<rendering::Contour>();
 		task_contour->contour->move_to( Vector(clip_rect.minx, clip_rect.miny) );
 		task_contour->contour->line_to( Vector(clip_rect.minx, clip_rect.maxy) );
 		task_contour->contour->line_to( Vector(clip_rect.maxx, clip_rect.maxy) );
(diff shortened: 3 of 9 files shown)


rodolforg:
Maybe change 
```c++
class Contour: public etl::shared_object
{
public:
	typedef etl::handle<Contour> Handle;
```

to

class Contour
{
public:
	typedef std::shared_ptr<Contour> Handle;
```
?

File: synfig-core/src/synfig/rendering/common/task/taskcontour.h [hunk 1]
```diff
@@ -53,7 +53,7 @@ class TaskContour: public Task, public TaskInterfaceTransformation
 	virtual Token::Handle get_token() const { return token.handle(); }
 
 public:
-	Contour::Handle contour;
+	std::shared_ptr<Contour> contour;
```
rodolforg, line 56: I tried to use `std::unique_ptr`, but couldn't compile because some Task copy calls... I didn't investigate it though

ice0:
> Maybe change
> 
> ```c++
> class Contour: public etl::shared_object
> {
> public:
> 	typedef etl::handle<Contour> Handle;
> ```
> 
> to
> 
> ```c++
> class Contour
> {
> public:
> 	typedef std::shared_ptr<Contour> Handle;
> ```
> 
> ?

I tried, but it affects other parts of the code. You can try too, maybe you'll have a better luck.

synfig-core/src/synfig/rendering/common/task/taskcontour.h
ice0: I didn't remember, but maybe you know - if the Task is always unique, or copies are possible?

rodolforg: Hm... They can be set as sub_task of one or more tasks...

rodolforg: But contour itself I don't know if it is shared to other stuff besides the task.

ice0:
I didn't merge because Synfig is still crashing. I'm currently looking into why this is happening.

rodolforg:
Still double free problem?

rodolforg:
I particularly do not agree with this:
https://github.com/synfig/synfig/blob/dc31e9d183a640ae1025b0586b53d71c844da99a/synfig-core/src/synfig/rendering/primitive/contour.cpp#L476-L479

The code is assigning a Contour to another Contour object, but ties the Intersector, instead of copying it.

ice0:
> Still due to double free problem?

I am not sure.
To reproduce, run looped animation and zoom in/zoom out work area until it crashes.

This is call stack I got:

image

rodolforg:
Hm... I believe it is due to double free too...
Maybe we should std::shared_ptr to tasks too? (btw, the only thing that currently needs this ETL classes is ValueNode - to easily replace a valuenode).

rodolforg:
What if this code of `etl::shared_object::unref()`:
https://github.com/synfig/synfig/blob/518486435c0a5889324b7877312cd223d92c627d/ETL/ETL/_handle.h#L100-L105

was like this:
```c++
	virtual void unref() const
	{
		if (--refcount == 0)
			delete this;
	}
```

ice0:
> What if this code of `etl::shared_object::unref()`:

Nice catch! 
After making the specified changes, the issue no longer occurs.
I spent three minutes trying to reproduce the problem, but everything worked without errors.
Awesome!

ice0:
> Maybe we should std::shared_ptr to tasks too?

I agree :)

rodolforg:
Done the unref() stuff :)

ice0 (approved):

ice0:
Merged. Thank you!

PR #3503, opened 2025-03-04
refactor: make code for importing image sequence more readable

rodolforg:
the main change is how we loop to detect consecutive duplicated images.

However, the idea is still the same:
1. create an animated valuenode to "layer_name" parameter of a (future) layer Switch
2. add each imported image to current canvas
3. check if it is duplicated (equal to previous image) 
3.1. If so, add it to a future remove layer action 
3.2. Otherwise add a new waypoint in the "layer_name" valuenode
4. delete the duplicated layers (if it is supposed to do so)
5. add an action to encapsulate the non-deleted import layers to a new switch layer
6. convert/link the layer_name parameter to the animated valuenode

### DIFF
--- a/synfig-studio/src/synfigapp/canvasinterface.cpp
+++ b/synfig-studio/src/synfigapp/canvasinterface.cpp
@@ -953,11 +953,12 @@ CanvasInterface::import_sequence(
 {
 	Action::PassiveGrouper group(get_instance().get(),_("Import sequence"));
 
-	float fps = get_canvas()->rend_desc().get_frame_rate();
-	
+	const float fps = get_canvas()->rend_desc().get_frame_rate();
+
 	synfig::info("Attempting to import sequence");
 	Layer::Handle layer_switch;
 	ValueNode_Animated::Handle layer_name_animated = ValueNode_Animated::create(type_string);
+
 	try {
 		// add imported layers into switch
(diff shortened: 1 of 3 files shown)


synfig-studio/src/synfigapp/canvasinterface.cpp
```diff
@@ -1002,73 +996,75 @@ CanvasInterface::import_sequence(
...
 					throw int();
 
-				if(remove_dups){
-					curr_layer= Layer_Bitmap::Handle::cast_dynamic(layer);
-					if(!curr_layer){
+				bool is_layer_equal_previous_one = false;
+
+				if (remove_dups) {
+					// Gets the surface of current imported image
+					auto curr_layer = Layer_Bitmap::Handle::cast_dynamic(layer);
```
rodolforg, line 1009: Maybe I should change it to `static_cast`? The layer is Import type as defined in line 999.

ice0: I suggest to leave it as-is.

ice0 (approved):

synfig-studio/src/synfigapp/canvasinterface.cpp
rodolforg: Ok

ice0:
Merged. Thank you!

PR #3504, opened 2025-03-04
refactor: reduce visibility of internal variable members of Layer_Bitmap

rodolforg:
more should be done, but that's better than current situation


ice0 (approved):

ice0:
Merged. Thank you!

Issue #3471, opened 2024-12-16
[Lottie exporter] Cannot export if an imported asset is in the 'images' subdirectory


nullst:
**Describe the bug**

Exporting an animation to lottie format currently copies all imported assets into an 'images' subdirectory (#936). If an imported asset was *already* in the 'images' subdirectory (I happened to name my assets directory 'images'), the export fails with a long and somewhat confusing stacktrace due to the fact that lottie-exporter uses the python function `shutil.copy` to put the asset in the `images` directory, and this function throws an exception when the source and the target file are the same.

**To Reproduce**
1. Create new synfig project.
2. Save it to some directory.
3. Manually create a subdirectory 'images' and put some jpg file in it.
4. Import that image in Synfig as a new layer.
5. Try to export to Lottie, see the break.

**Expected behavior**
If the file is *already* in the 'images' subdirectory, it does not have to be copied, so the export should work as expected.
Workaround: rename the assets directory to some other name.

**Screenshots**
![Screenshot from 2024-12-16 13-58-32](https://github.com/user-attachments/assets/31afcee6-9a75-4a49-b45a-74bf8bc82736)

**System information:**
 - OS: Linux Pop!_OS
 - OS version: 22.04 LTS
 - Synfig version: 1.5.3

PR #3506, opened 2025-03-05
Fix: Prevent SameFileError when exporting assets already in the images directory

Stargazer10101:
Related issue: (#3471)
This PR fixes the issue where exporting animations to Lottie format fails when imported assets are already located in the images subdirectory. The error occurs because shutil.copy throws a SameFileError when attempting to copy a file to itself.
Changes:
    Added a try-except block to handle shutil.SameFileError when copying assets to the images directory.
    Ensured that file metadata is only copied if the file copy operation succeeds.
    Skipped redundant copying when the source and destination files are the same.


rodolforg:
Instead of "Related issue:" at the beginning of the PR description, write it at the last line with the words "Close #", "Closes #", "Fix #" or "Fixes #".

We are using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for a while. The PR and commit 'titles' should not start in upper case.

Stargazer10101:
Hey @rodolforg thanks for suggesting corrections. Will follow the conventional commits hereafter. Apologies for multiple forced pushes.

ice0:
> Apologies for multiple forced pushes.

No issue here, it's ok.

ice0 (approved):

ice0:
Merged. Thank you!

PR #3510, opened 2025-03-10
ci: enable ccache compression for AppVeyor build

ice0:
AppVeyor cache size is only 1Gb:
https://www.appveyor.com/docs/build-cache/#cache-size-beta

### DIFF
--- a/2-build-cmake.sh
+++ b/2-build-cmake.sh
@@ -21,10 +21,19 @@ then
 mode=$1
 fi
 
+# AppVeyor cache size limit:
+# https://www.appveyor.com/docs/build-cache/#cache-size-beta
+ccache --set-config compression_level=5
+ccache --max-size=1G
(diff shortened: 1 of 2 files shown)


rodolforg:
Is it good to be in the `2-build-cmake.sh`?
Or is there a way to enable compressed cache for CI only?

ice0:
> Is it good to be in the `2-build-cmake.sh`? Or is there a way to enable compressed cache for CI only?

Agree. I've been thinking about it, and this solution seemed quite controversial to me.
Since it raises questions not only for me, I'll consider an alternative solution.

rodolforg:
> I'll consider an alternative solution.

Maybe the script accepts an extra argument that allows user to choose whether the ccache should be compressed or not.
In its turn, the CI setup forces this argument to 'true'.

Is it possible?

PR #3512, opened 2025-03-11
fix(audio): add length check for MLT track loading

### LINKED ISSUE #3484: Displaying the sound wave of the attached audio file crashes Synfig Studio
**Describe the bug**
Displaying the sound wave of the attached audio file crashes Synfig Studio.

**To Reproduce**
1. Create a new synfig document
2. Import the attached audio file
3. Switch to Sound panel
4. In the combobox, select the listed audio file
5. It crashes

**Expected behavior**
No crash and the sound wave is shown

**System information:**
 - OS: Linux, Debian
 - OS version: testing (future Trixie)
 - Synfig version: 1.5.3

[crash-test-audio-file.zip](https://github.com/user-attachments/files/18605806/crash-test-audio-file.zip)

Abdelhadi-Wael:
Fixes #3484

### DIFF
--- a/synfig-studio/src/gui/widgets/widget_soundwave.cpp
+++ b/synfig-studio/src/gui/widgets/widget_soundwave.cpp
@@ -253,32 +253,6 @@ void Widget_SoundWave::on_time_model_changed()
 	queue_draw();
 }
 
-bool Widget_SoundWave::is_file_header_valid(const std::string& filename)
-{
-	FILE* file = fopen(filename.c_str(), "rb");
-	if (!file)
-		return false;
-	unsigned char header[4];
-	size_t bytes_read = fread(header, 1, 4, file);
-	fclose(file);
-	// Need 4 bytes for header check
-	if (bytes_read < 4)
-		return false;
-	// Check for WAV format
-	if (header[0]=='R' && header[1]=='I' && header[2]=='F' && header[3]=='F')
-		return true;
-	// Check for OGG format
-	if (header[0]=='O' && header[1]=='g' && header[2]=='g' && header[3]=='S')
-		return true;
-	// Check for MP3 with ID3 tag
-	if (header[0]=='I' && header[1]=='D' && header[2]=='3')
-		return true;
-	// Check for MP3 without ID3 tag
-	if (header[0] == 0xFF && (header[1] & 0xE0) == 0xE0)
-		return true;
-	return false;
-}
-
 void Widget_SoundWave::setup_mouse_handler()
 {
 	mouse_handler.set_pan_enabled(true);
@@ -302,15 +276,14 @@ bool Widget_SoundWave::do_load(const synfig::filesystem::Path& filename)
 {
 #ifndef WITHOUT_MLT	
 	std::string real_filename = Glib::filename_from_utf8(filename.u8string());
-	if (!Widget_SoundWave::is_file_header_valid(real_filename)) {
-		return false;
-	}
 	Mlt::Profile profile;
 	Mlt::Producer *track = new Mlt::Producer(profile, (std::string("avformat:") + real_filename).c_str());
-	if (!track->get_producer() || track->get_length() <= 0) {
+	int sample_rate = track->get_int("audio_sample_rate");
+	if (!track->get_producer() || track->get_length() <= 0 || sample_rate <= 0) {
 		delete track;
 		track = new Mlt::Producer(profile, (std::string("vorbis:") + real_filename).c_str());
-		if (!track->get_producer() || track->get_length() <= 0) {
+		sample_rate = track->get_int("audio_sample_rate");
+		if (!track->get_producer() || track->get_length() <= 0 || sample_rate <= 0) {
 			delete track;
 			return false;
 		}
--- a/synfig-studio/src/gui/widgets/widget_soundwave.cpp
+++ b/synfig-studio/src/gui/widgets/widget_soundwave.cpp
@@ -253,22 +253,27 @@ void Widget_SoundWave::on_time_model_changed()
 	queue_draw();
 }
 
-bool Widget_SoundWave::isFileHeaderValid(const std::string &filename)
+bool Widget_SoundWave::is_file_header_valid(const std::string& filename)
 {
 	FILE* file = fopen(filename.c_str(), "rb");
 	if (!file)
 		return false;
-	unsigned char header[16];
-	size_t bytesRead = fread(header, 1, 16, file);
+	unsigned char header[4];
+	size_t bytes_read = fread(header, 1, 4, file);
 	fclose(file);
-	if (bytesRead < 16)
+	// Need 4 bytes for header check
+	if (bytes_read < 4)
 		return false;
+	// Check for WAV format
 	if (header[0]=='R' && header[1]=='I' && header[2]=='F' && header[3]=='F')
 		return true;
+	// Check for OGG format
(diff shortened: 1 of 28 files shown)


synfig-studio/src/gui/widgets/widget_soundwave.cpp
```diff
@@ -253,6 +253,27 @@ void Widget_SoundWave::on_time_model_changed()
 	queue_draw();
 }
 
+bool Widget_SoundWave::isFileHeaderValid(const std::string &filename)
```
rodolforg, line 256: In our code style, we don't use camelCase or PascalCase, but snake_case.
Please change method name as such.

Besides, prefer `const std::string& filename` rather `const std::string &filename`

synfig-studio/src/gui/widgets/widget_soundwave.cpp
```diff
@@ -253,6 +253,27 @@ void Widget_SoundWave::on_time_model_changed()
...
+{
+	FILE* file = fopen(filename.c_str(), "rb");
+	if (!file)
+		return false;
+	unsigned char header[16];
+	size_t bytesRead = fread(header, 1, 16, file);
```
rodolforg, line 262: snake_case, please.

synfig-studio/src/gui/widgets/widget_soundwave.cpp
```diff
@@ -253,6 +253,27 @@ void Widget_SoundWave::on_time_model_changed()
...
+	fclose(file);
+	if (bytesRead < 16)
+		return false;
```
rodolforg, line 265: Why 16 specifically? Maybe a comment to explain it?

synfig-studio/src/gui/widgets/widget_soundwave.cpp
```diff
@@ -253,6 +253,27 @@ void Widget_SoundWave::on_time_model_changed()
...
+	if (header[0]=='R' && header[1]=='I' && header[2]=='F' && header[3]=='F')
```
rodolforg, line 266: Before each comparison, describe in a comment line what file format(s) are you trying to check.
`// WAV or ... `

synfig-studio/src/gui/widgets/widget_soundwave.cpp
```diff
@@ -276,6 +297,10 @@ bool Widget_SoundWave::do_load(const synfig::filesystem::Path& filename)
 {
 #ifndef WITHOUT_MLT	
 	std::string real_filename = Glib::filename_from_utf8(filename.u8string());
+	Widget_SoundWave obj;
+	if (!obj.isFileHeaderValid(real_filename)){
```
rodolforg, line 301: 1. `isFileHeaderValid()` does not depend at all on a `Widget_SoundWave` object data.
So, please, change the method signature in header file to `static`.

This way, you can call it here without a object like this: `Widget_SoundWave::is_file_header_valid(real_filename)`

2. Please add a single space before `{`

rodolforg (changes requested):

synfig-studio/src/gui/widgets/widget_soundwave.cpp
Abdelhadi-Wael: done

synfig-studio/src/gui/widgets/widget_soundwave.cpp
Abdelhadi-Wael: that was a mistake sorry

synfig-studio/src/gui/widgets/widget_soundwave.cpp
Abdelhadi-Wael: done

ice0:
> the application now properly validates the file format before processing.

Synfig can import a variety of audio formats (depending on the loaded plugins and their capabilities), so checking the file header might not be sufficient. A more accurate approach to solving the problem is to ask the plugin whether it managed to load the file or not.

In this case, we are using the MLT framework to load files.
There is a check to ensure that the file has been loaded successfully, but this check turned out to be insufficient.
```cpp
if (!track->get_producer() || track->get_length() <= 0) {
```

I have output additional details to see which parameters could be used to add another condition. Here is the output for this file (based on the information provided by MLT library):
```bash
Duration (frames): 2147483647
Frame rate: 25
Sample rate: 0
Channels: 0
```
I suggest adding an additional check based on the `Sample rate`.

Abdelhadi-Wael:
Okay, done thanks for the clarification

ice0:
Looks good. Please fix the PR description to better match the changes made.

Abdelhadi-Wael:
done

rodolforg:
Interesting. 
The 'linux' software `file` detects the sample rate, as well as the Gnome `nautilus` when right-clicking to get file properties. Audacity imports fine too, with correct sample rate. Why MLT doesn't detect it? :/

> Here is the output for this file (based on the information provided by MLT library):
> 
> ```shell
> Duration (frames): 2147483647
> Frame rate: 25
> Sample rate: 0
> Channels: 0
> ```

ice0:
> Interesting. The 'linux' software `file` detects the sample rate, as well as the Gnome `nautilus` when right-clicking to get file properties. Audacity imports fine too, with correct sample rate. Why MLT doesn't detect it? :/


Added issue: https://github.com/mltframework/mlt/issues/1073

synfig-studio/src/gui/widgets/widget_soundwave.cpp
```diff
@@ -302,15 +276,14 @@ bool Widget_SoundWave::do_load(const synfig::filesystem::Path& filename)
 {
 #ifndef WITHOUT_MLT	
 	std::string real_filename = Glib::filename_from_utf8(filename.u8string());
-	if (!Widget_SoundWave::is_file_header_valid(real_filename)) {
-		return false;
-	}
 	Mlt::Profile profile;
 	Mlt::Producer *track = new Mlt::Producer(profile, (std::string("avformat:") + real_filename).c_str());
-	if (!track->get_producer() || track->get_length() <= 0) {
+	int sample_rate = track->get_int("audio_sample_rate");
```
ddennedy, line 281: There is no "audio_sample_rate" rate property in MLT unless you have forked and modified it. You need to look in either `meta.media.attr` properties (do not assume index is 0), or you need to get a frame and get "audio_frequency" propery from the frame.

Abdelhadi-Wael:
> I suggest adding an additional check based on the `Sample rate`.
The additional check based on the Sample rate didn't solve the issue, so I added a length check as well. Hopefully, this resolves the problem. please Let me know if any further changes are needed
@ice0

ice0:
> The additional check based on the `Sample rate` didn't solve the issue

I suggest removing this check since it doesn't work, to avoid cluttering the code.

I also suggest getting rid of the magic number (2147483647) and doing something like this (to improve clarity).
```cpp
// That value is INT_MAX, which is our special value that means unknown or possibly a live source.
// https://github.com/mltframework/mlt/issues/1073#issuecomment-2730111264
constexpr int mlt_error_length = INT_MAX;
```

Abdelhadi-Wael:
> I suggest removing this check since it doesn't work, to avoid cluttering the code.

added a commit : )

ice0:
It failed last time because the self-hosted runner was offline.
I re-ran the job. Now should be ok.

ice0 (approved):

rodolforg (approved):

Issue #2188, opened 2021-06-12
Implement toolbars using GTK Box


ice0:
https://developer.gnome.org/gtk4/stable/ch41s02.html#id-1.7.4.4.75

> 
![Screenshot_2021-06-04_10-31-32](https://user-images.githubusercontent.com/24214509/120748267-1fea3480-c520-11eb-92f9-f455b70fc461.png)

However, `GtkToolbar` has been removed in Gtk 4 so we might end up reimplementing this PR solution again in the future when migrating to a series of buttons packed in Box/Grid.

_Originally posted by @Keyikedalube in https://github.com/synfig/synfig/issues/2170#issuecomment-854360300_

Kashyapdevesh:
If no one is working on this issue,can i try it? I'm new to GTK & opensource in general,so it may take a little time

rodolforg:
You can try it :)

rodolforg:
@Kashyapdevesh  Do you need any help?

Kashyapdevesh:
Sorry for the late reply, i was studying about gtk4 & analyzing the codebase, i have started working on designing the toolbar, i will post a pr ASAP

rock16:
I would love to work on this issue, 
below is a list of what I know and intend implementing for now.

1) in `src/gui/docks/dock_layer.cpp`, I have started out by replacing the ui definition for the toolbar 

from this `Glib::ustring ui_info =
        "<ui>"
	"	<toolbar action='toolbar-layer'>"
	"	<toolitem action='popup-layer-new' 
        ........................`

to something like this  `Glib::ustring ui_info =
	"<interface>"
	"	<object class='GtkBox'>"
	"   <property name='action_name'>toolbar-layer</property>"
       ...........................................................`

2) it turns out gtkuimanager is depracated and I would have to replace it with gtkbuilder; I will have to dig a little bit more deeper into the source code for `App::ui_manager` any information or help will be appreciated.
`
3) I will also have to work on the set_toolbar method in `/src/gui/docks/dockable.cpp`

I don't know if I'm headed in the right direction?

ice0:
Hi, Rock!

Assigned this to you.

I suggest adding the `App::get_builder()` method to the `App` class (same way as `App::ui_manager()`) and reimplement the toolbar using it.

rock16:
Hi, @ice0 

[...]

PR #3515, opened 2025-03-13
Draft: feat: replacing toolbar from toolbox and dock history bar (#2188)

neesaaa:
image
* I have converted the two toolbox (right and top) and the dock history toolbar.
* I wonder if this is the best practice or if there's a better approach.
just to continue in the rest of deprecated bars.


mohamedAdhamc:
Hello @neesaaa, I saw you had the title as "draft: feat: ...." so I changed the pr status to "draft" and fixed the title.

PR #3525, opened 2025-03-16
fix(ui): refresh Parameters panel after undo/redo

### LINKED ISSUE #3524: Undo/Redo Does Not Refresh Parameter Value in Converted Layers
**Describe the bug**
When performing an undo or redo operation after modifying a parameter (e.g., angle) in a converted layer (e.g., Radial Composite), the value does not refresh properly in the UI.

**To Reproduce**
1. Create a Circle layer (circle, region, outline, it does not matter)
2. Convert its Origin to Radial Composite (or Composite or Scale, ...)
3. In the Parameter Panel, change a converter parameter. If Radial Composite, its angle to 10, for example
4. Change that converter parameter again. If Radial Composite, its angle to 30, for example
5. Change that converter parameter again. If Radial Composite, its angle to 50, for example
6. undo -> angle is still 50
7. undo -> angle is still 50
8. redo -> angel is still 50

**Expected behavior**
The angle parameter should revert correctly on Undo.
The angle parameter should restore correctly on Redo.
The UI should immediately reflect these changes.

Abdelhadi-Wael:
fixes #3524

### DIFF
--- a/synfig-studio/src/gui/docks/dock_params.cpp
+++ b/synfig-studio/src/gui/docks/dock_params.cpp
@@ -109,9 +109,25 @@ Dock_Params::changed_canvas_view_vfunc(CanvasView::LooseHandle canvas_view)
 			sigc::mem_fun(
 				*this,
 				&Dock_Params::refresh_selected_param ));
+
+		// Connect to the canvas's signal_changed() to detect state changes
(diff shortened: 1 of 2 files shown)


ice0 (approved):

ice0:
Merged. Thank you!

Issue #3387, opened 2024-07-27
Not Support Devanagari script properly.


avnlearn:
### For example:

![Screenshot_20240727_232818](https://github.com/user-attachments/assets/224714d6-fb58-445d-87ac-567f5f9071da)
![Screenshot_20240727_233607](https://github.com/user-attachments/assets/64a48bb6-2989-47e3-92c2-2f2dee9d8398)
 and Inkscape SVG is not import properly 
**Please solve this problem**

**OS :** EndeavourOS  2024.06.25 64bit
**Synfig Studio Version :** v1.4.5

mrraj30:
Same problem....

yoSachinkr:
Also adding multiple layers of text from the same script often results in blocky output.
![image](https://github.com/user-attachments/assets/563386f4-d1a3-40da-8c31-ee45ec384509)

mrraj30:
> Also adding multiple layers of text from the same script often results in blocky output.
> ![image](https://github.com/user-attachments/assets/563386f4-d1a3-40da-8c31-ee45ec384509)
> 

Not working...

avnlearn:
root - WARNING - Layer 'text' is not supported yet. For more information, contact us on Synfig forums or Github page

ice0:
Can you check build from the current master branch?

P.S. You need to compile it yourself.
https://synfig-docs-dev.readthedocs.io/en/latest/building/Building%20using%20CMake.html

rodolforg:
Could you try the development version 1.5.2? I think we fixed it since 1.5.1.

Please keep in mind it is a development version, and opening/saving files with it can block to use it 1.4.x versions. Make a backup before using your files.

rodolforg:
By the way, please try to provide as much info as possible.
The current core developers speak English (not natively), Russian and Portuguese... So we don't know Indian scripts and don't know how to  type them. The screenshots help, but it would be good to provide here the text you typed in Synfig, so we can copy from here and paste there to test it. Or the .sif/.sifz file.

BobSynfig:
Did some test on Ubuntu 22.04
Using copy-paste from web browser for the word देवनागरी (Devanagari).
The UI (font) of Synfig recognizes natively when you enter the word, but only display blocks on the layer.
I had to select manually in properties a font present in the font cache of my system to display it properly (Lohit Devanagari, Nirmala UI is not installed).

Also it seems that Devanagari is not interpreted properly with some fonts, and some combined characters can lead to reverse order.
[...]

PR #3526, opened 2025-03-18
Fixed spacing between tool buttons inside the toolbox

saumi10:
Tool buttons inside the toolbox were too closely packed , hence changed **toolbox orientation to horizontal** and **added spacing between tool buttons**. As a result it gives better view of toolbox to the user.
@ice0 
@mohamedAdhamc 
synfig default toolbax (updated)


mohamedAdhamc:
Hello @saumi10 

1- I don't see why this is needed. The toolbox looked fine the way it was and that's what Synfig users are used to. Also I'm not sure why in the photo you put all the tool buttons are in one column (maybe it's because of the width of the box in the photo), but I think it looks quite awkward. Maybe also @BobSynfig can have more to say regarding this.

2- When you make a pr make sure that the only changes showing are actual changes, currently when I look at the changes about 80% or something are not actual changes but various formatting changes unrelated to the PR itself.

mohamedAdhamc (changes requested):
Please remove unrelated formatting changes.

BobSynfig:
I gave a like on the forum mostly for the initiative and interest in the software.
Lots of graphic software have a vertical presentation for their tools but it doesn"t that Synfig has to do the same.
If it could be given the choice as a setting, why not, but not to make it as a default ...

saumi10:
@BobSynfig Thankyou for the feedback , I completely understand . Also , do you have any issues I can work on currently?

ice0:
> Also , can you suggest some issues I can work on currently?

Hi, Saumitra!

Can you please check this issue? (https://github.com/synfig/synfig/issues/3387)
Does it reproduced on the current master branch?

Issue #3472, opened 2024-12-22
Shift... Shift everywhere...


Svarov-RZM:
### Description
![ShiftEverywhere](https://github.com/user-attachments/assets/d44c1a8d-6320-4580-8dec-90aace3586a5)

Okay, there's a serious problem in Synfig where a considerable amount of layers has a slight shift in them (like 1-2 px) unlike versions <= 1.2.2 that don't. Shift is not only present in the gradients/distortions but even in some composite layers like "Shade" layer for example.

### To Reproduce
Here's a test-case file with this layer:
[ShadeTest.zip](https://github.com/user-attachments/files/18220947/ShadeTest.zip)

Unpack, switch in image viewer between rendered pngs and you will notice that version 1.4.4 (and newer) has a slight shift toward bottom-right corner.

### Additional
This problem starts to drive me crazy. I wanted to test if I can use "Shade" layer to make linework based on a complex shape and I can't because of that damned shift that messes up the result. 

In this thread I discovered that the problem was introduced after 1.2.2 release: https://github.com/synfig/synfig/pull/3306 
Please read it if you haven't.

I am going to head-dive in the commits trying to find the culprit.
I'll report it here if I do and maybe someone can hopefully fix this annoying issue.

rodolforg:
I accidently found this code slice:

https://github.com/synfig/synfig/blob/9f576ad5b30a427fcd901c4ebe00f0c15a89c636/synfig-core/src/synfig/target_scanline.cpp#L101-L109

If I comment it out, the shift doesn't happen.
But why is this a problem to some top layers only?

Svarov-RZM:
Alright, just like I said I tried to build various versions of Synfig to see where exactly the error was introduced and failed :( I was able to go back to version < 1.3.11, exact commit: 2c0523022e2c5d828ca01726fd394d3fdb706b1d and the shift was already there.
I couldn't go past that because Synfig would no longer build on my Linux box, perhaps I needed the older version of Debian... Anyway, judging by what rodolforg said and the commit I stopped at, the shift was probably introduced somewhere along these two commits: da43984f62e7c453e1caab279e5bb11517883d90 and 5e27542ae28cd7fde5ec4995d8b7ca53a03daeb3

> But why is this a problem to some top layers only?

[...]

PR #3530, opened 2025-03-22
fix: shift in TaskBlur due to resampling

Svarov-RZM:
Fixes https://github.com/synfig/synfig/issues/3472
The above link also contains a detailed explanation of the problem.

This commit affects rendering of all layers that utilize blur, so a lot (almost everything in "geometry" category, Shade, etc.). I did some tests on my own and everything seems fine. but if you have some heavy works with a lot of blur/feathering I urge you to test this branch yourself before merging.

### DIFF
--- a/synfig-core/src/synfig/rendering/common/task/taskblur.h
+++ b/synfig-core/src/synfig/rendering/common/task/taskblur.h
@@ -32,6 +32,7 @@
 
 #include "../../task.h"
 #include "../../primitive/blur.h"
+#include "./tasktransformation.h"
 
 /* === M A C R O S ========================================================= */
 
@@ -44,7 +45,9 @@ namespace synfig
 namespace rendering
 {
 
-class TaskBlur: public Task
+// Blur works identically after or before transformation, so we make it a
+// member of TaskInterfaceTransformationPass to avoid unnecessarily resampling
+class TaskBlur: public Task, public TaskInterfaceTransformationPass
 {
 public:
 	typedef etl::handle<TaskBlur> Handle;


ice0:
Thank you for the contribution.
We plan to test this fix next week.

MetaruHebi:
Hi. The fix has been tested on version 1.0.2. and 1.5.3.

V1.0.2:
test

V1.5.3:
Screenshot 2025-04-04 134847

current PR (V1.5.3):
Screenshot 2025-04-04 134834

There was indeed an offset in version 1.5.3.
Your fix is working.

morevnaproject:
Okay, in this case I need to make one more test for the fix made in this PR. 

Let's make this test:
1. Create rectangle
2. Add blur layer on top off it (with big blur, say 60px) 
3. Add rotate layer on top and rotate the rectangle by 30 degrees
4. Put everything into group and scale by 50% on Y axis

After that let's render the result with fixes from  this PR and compare it with render result of Synfig version 1.5.3 (or any recent version without fixed from this pr). 

@MetaruHebi Please check this.

Zloi-KoteikA:
I ran tests with this data in version 1.0.2; 1.5.3; PR...

>     1. Create rectangle
> 
>     2. Add blur layer on top off it (with big blur, say 60px)
> 
>     3. Add rotate layer on top and rotate the rectangle by 30 degrees
> 
>     4. Put everything into group and scale by 50% on Y axis

...and the result was different in all three cases.

current PR:
Synfig Animation pr

V 1.0.2:
Synfig Animation 1 0 2

V 1.5.3
Synfig Animation 1 5 3

Here is the source file of the tests:
[Synfig Animation 1.sifz.zip](https://github.com/user-attachments/files/19620289/Synfig.Animation.1.sifz.zip)

Svarov-RZM:
> I ran tests with this data in version 1.0.2; 1.5.3; PR...

Hey, thanks for testing. Good thing we decided to do that instead of just merging (I had a nasty feeling that something would turn up).

Okay, first thing, fast Gaussian blur seems to be somewhat broken since 1.2.x, that something to keep in mind. Second, the blur in this PR, as this commit suggests, doesn't do resampling and this is not a proper behaviour. I thought it's not needed for blur, but I was wrong, it is needed, the 1.0.2 version does it as this example shows. If you place "Blur" layer out of the group like so:
Layers

You will get a different result for Rectangle005, so blur does get resampled even back in 1.0.2.

What I noticed, is if you disable blur in the example above, this PR (and obviously master and 1.4.x versions) will still produce 1x shift compared to 1.0.2 version. So the issue is global and lurks I believe somewhere in "resample" procedure. At least we continue to narrow down the problem.

morevnaproject:
Just a quick note: the blur algorithm was rewritten in 1.2.0, this is why it gives different results when comparing 1.5.3 to 1.0.2

rodolforg:
> Just a quick note: the blur algorithm was rewritten in 1.2.0, this is why it gives different results when comparing 1.5.3 to 1.0.2

And that visual change was intentional?

Svarov-RZM:
Okay, so my take to fix the problem turned out wrong, so closing this PR.

PR #3535, opened 2025-03-27
feat: add dependency collector file

Stargazer10101:
This is a part of the project: macOS app bundle for GSOC 2025. I have created a prototype script to collect executable/library dependencies and copy them to appropriate locations in the .app bundle. In the overall workflow of the project, this script gets used after the modified CMakeLists file builds a .app structure.

### DIFF
--- a/autobuild/osx/dependency_collector.py
+++ b/autobuild/osx/dependency_collector.py
@@ -1,0 +1,435 @@
+import logging
+import os
+import subprocess
+import shutil
+import sys
+import argparse
+import re
+
+# Logging to console output as well as a file
+def setup_logging():
+    logging.basicConfig(
+        level=logging.INFO,
+        format='%(asctime)s - %(levelname)s - %(message)s',
+        handlers=[
+            logging.FileHandler("dependency_collection.log"),
+            logging.StreamHandler()
+        ]
+    )
+
+# Recursively find all executable files in the app bundle.
+def find_binaries(app_folder):
+    binaries = []
+    for root, _, files in os.walk(app_folder):
+        for file in files:
+            file_path = os.path.join(root, file)
+            if os.access(file_path, os.X_OK) and not os.path.isdir(file_path):
+                binaries.append(file_path)
+    return binaries
+
+def is_binary_file(file_path):
+    try:
+        result = subprocess.run(
+            ["file", "-b", file_path], 
+            capture_output=True, 
+            text=True, 
+            check=True
+        )
+        return any(x in result.stdout for x in ["Mach-O", "shared library"])
+    except:
+        return False
+
+# List dependencies of a binary using otool -L
+def get_dependencies(binary_path):
+    try:
+        output = subprocess.check_output(["otool", "-L", binary_path], text=True)
+        deps = []
+        for line in output.strip().split('\n')[1:]:  # Skip first line because it contains the binary itself.
+            match = re.match(r'^\s*(@?[^\s]+)', line) # Extract only the path from the output, ignoring version details and leading whitespaces.
+            if match:
+                deps.append(match.group(1))
+        return deps
+    except subprocess.CalledProcessError as e:
+        logging.error(f"Error running otool: {e}")
+        return []
+
+# Function for @rpath references in libraries
+def resolve_rpath(binary_path, rpath_lib):
+    try:
+        # Extracting load commands (including LC_RPATH entries).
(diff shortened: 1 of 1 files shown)


autobuild/osx/dependency_collector.py
```diff
@@ -0,0 +1,435 @@
...
+        level=logging.INFO,
+        format='%(asctime)s - %(levelname)s - %(message)s',
+        handlers=[
+            logging.FileHandler("dependency_collection.log"),
+            logging.StreamHandler()
+        ]
+    )
+
+# Recursively find all executable files in the app bundle.
+def find_binaries(app_folder):
```
ice0, line 21: 1. I suggest to rename it to `find_executable_files` to reflect function purpose more correctly.
2. Also it's better to extract executable check to separate function. In this case your logic became more clear in both cases.
3. Better to use python docstring to describe function purpose.

This is suggested change:
```python
def is_executable_file(file_path):
    """Check if the given path is an executable file."""
    return os.path.isfile(file_path) and os.access(file_path, os.X_OK)

def find_executable_files(directory):
    """Find all executable files in the specified directory and its subdirectories."""
    executable_files = []
    
    for root, _, files in os.walk(directory):
        for filename in files:
            file_path = os.path.join(root, filename)
            if is_executable_file(file_path):
                executable_files.append(file_path)
    
    return executable_files
```

autobuild/osx/dependency_collector.py
```diff
@@ -0,0 +1,435 @@
...
+# Function for @rpath references in libraries
+def resolve_rpath(binary_path, rpath_lib):
+    try:
+        # Extracting load commands (including LC_RPATH entries).
+        output = subprocess.check_output(["otool", "-l", binary_path], text=True)
+        rpaths = []
+        binary_dir = os.path.dirname(binary_path)
+        
+    
+        for i, line in enumerate(output.splitlines()):
```
ice0, line 65: I suggest to extract `LC_RPATH` logic to separate function, so we can optimize it and make more clear.

autobuild/osx/dependency_collector.py
```diff
@@ -0,0 +1,435 @@
...
+    logging.info(f"Found {len(binaries)} binaries to process")
+    
+    for binary in binaries:
+        process_binary(binary, app_bundle_path)
+    
+    logging.info(f"Finished processing app bundle: {app_bundle_path}")
+    return True
+
+# Recursively find all executable files in the app bundle.
+def find_binaries(app_folder):
```
ice0, line 398: `find_binaries` and `is_binary_file` is duplicated here. Please remove duplicates.

ice0:
I also suggest avoiding automatic binary detection.
Instead, it's better to let users provide a list of binaries they want to include in the `.app` bundle.
For example, Synfig uses `ffmpeg` and ImageMagick (`magick`) binaries for file reading/writing.

Additionally, ImageMagick and GTK/FontConfig libraries depend on configuration files that can't be automatically detected.
We just need to know their locations and how to copy them.

With this in mind, I suggest splitting the script into the following steps:

1. **Set `BUNDLE_ROOT` folder**:  
   Create and clean the bundle directory (e.g., `Synfig-1.5.3.app`) if it's not empty.

2. **Get Homebrew path**:  
   Use `brew --prefix` to determine the Homebrew installation path, which we'll need for collecting binary dependencies and configuration files.

3. **Populate `BUNDLE_ROOT` from skeleton folder**:  
   Copy initial content from a skeleton folder (containing `readme.txt` and other static files). This approach:
   - Avoids programmatic folder creation
   - Allows easy modifications without changing the script

4. **Relocate required binaries**:  
   Move binaries to their corresponding folders. For the first iteration, focus on `synfig` and `synfigstudio` binaries.  
   - Use a `BUILD_ROOT` variable to specify the Synfig installation path (e.g., `${BUILD_ROOT}/bin/synfig`)  
   - Relocation should happen at this stage because binaries may contain relative paths  

   The relocation function signature should be:  
   ```python
   def relocate_binary(src_file_full_path, dst_folder_path, lib_folder_path)
   ```  
   Where:
   - `lib_folder_path` is the destination for all dependency libraries

Stargazer10101:
Thanks for the review. I will incorporate all of these suggestions.

Stargazer10101:
Hi @ice0 I have worked on your suggestions. Please review it here: https://github.com/synfig/synfig/pull/3608

PR #3540, opened 2025-04-01
fix: Fix bug displaying "Unable to resolve <val>" on project load

### LINKED ISSUE #3538: Synfig may damage the project with exported values on save
### Description

When we have a bunch of exported values in a project, Synfig may damage it when saving the project to disk. The reason for this is that the order of the nodes gets messed up and we get a bunch of "Unable to resolve" messages on loading the project.

### How to reproduce
* Open the test-case file:

[WrongCopyOrder-OK.zip](https://github.com/user-attachments/files/19524642/WrongCopyOrder-OK.zip)

* Click on a black outline
* In layers panel->Vertices (PATH)->Vertex 001->Vertex->Right click->Export Value...->Name: TEST->Export

Image

* Save the project under a different name
* Menu->File->Revert and you get a bunch of "Unable to resolve TEST" messages

Reproducible since version 1.0.2.

### My thoughts
This is a **very serious issue** because it leaves the project damaged and non-tech savvy users have no choice to seek help on the forum. Good thing that I finally was able to create a reproducible example but how did I get there? I honestly don't know.

I did the usual thing... You know, exported a bunch of nodes, made some converters, exported some more stuff in a nested manner to make Synfig fail but it all worked fine. I think the problem arrived after I unexported some nodes and then exported them again but I don't know if all that is relevant.

shreyaspranav:
This PR fixes `Unable to resolve <val>` on project load by changing save file format to write the constant exported values first and then writing the non-constant ones.

By writing the constant value nodes first makes sure that the non-constant value nodes that depend on constant value nodes get resolved and hence the project file can be loaded without errors due to unresolved exported values.  

Fix #3538

### DIFF
--- a/synfig-core/src/synfig/savecanvas.cpp
+++ b/synfig-core/src/synfig/savecanvas.cpp
@@ -985,16 +985,23 @@ xmlpp::Element* encode_canvas(xmlpp::Element* root,Canvas::ConstHandle canvas)
 		xmlpp::Element *node=root->add_child("defs");
 		const ValueNodeList &value_node_list(canvas->value_node_list());
 
+		// Save all the constant 'value_node' nodes first.
 		for(ValueNodeList::const_iterator iter=value_node_list.begin();iter!=value_node_list.end();++iter)
 		{
 			// If the value_node is a constant, then use the shorthand
 			if (ValueNode_Const::Handle value_node = ValueNode_Const::Handle::cast_dynamic(*iter))
 			{
 				reinterpret_cast<xmlpp::Element*>(encode_value(node->add_child("value"),value_node->get_value(),canvas))->set_attribute("id",value_node->get_id());
-				continue;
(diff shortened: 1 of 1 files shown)


shreyaspranav:
I apologize for the forced push, didn't realize that the commit message shouldn't end with a full-stop.

Svarov-RZM:
I ran some tests on this PR and everything seems fine.
Tested on particle template (it got some pretty complex conversions) along with my own work. Tried to rename exported nodes, unexport them, save/reload and no problems.

So this fixes the linked issue and doesn't seem to introduce new problems.

shreyaspranav:
> So this fixes the linked issue and doesn't seem to introduce new problems.

@Svarov-RZM Thank you for your valuable time testing my PR.

rodolforg (approved):

PR #3542, opened 2025-04-02
feat: Add atan2 support for value type Real

### LINKED ISSUE #3528: Converter real to angle
When my calculation requires a real value, I normally connect it by converting an angle to a real number. However, the problem is that the angle comes from a formula—for example, the result of `atan2`, which is an angle. I hope the value can be converted into a real number so that my calculation works properly.

Image

Stargazer10101:
The value type "Real" doesn't have atan2 as a convert type. This PR introduces atan2 conversion of "Real" values, so that the Bone Angle → Real → Angle conversion can be achieved, as requested in the issue #3528 .
Fixes #3528

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_atan2.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_atan2.cpp
@@ -65,6 +65,14 @@ ValueNode_Atan2::ValueNode_Atan2(const ValueBase &value):
 		set_link("x",ValueNode_Const::create(Angle::cos(value.get(Angle())).get()));
 		set_link("y",ValueNode_Const::create(Angle::sin(value.get(Angle())).get()));
 	}
+	else if (value.get_type() == type_real)
+	{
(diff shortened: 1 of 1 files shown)


rodolforg:
The reason for real number can't be converted via `atan2()` is that `atan2()` is supposed to return an angle, as it means "arctangent" (i.e. arc/angle that gives such tangent).

You should do the Real -> Angle converter, but I still don't know what it is supposed to do with this conversion.

Stargazer10101:
I agree, Real → Angle doesn't make sense. But the user is trying to achieve this: Bone Angle → Real → Angle. 
Right now, the converter atan2() can be connected to angle-valued parameters. But the user wants to convert "angle" to "real" first and then convert to atan2() .
When atan() is used on "Real" value type, it is treated as if it is a value of an angle in degrees.

rodolforg:
Original request/issue was solved via other PR https://github.com/synfig/synfig/pull/3560.
Thank you for your contribution ;)

PR #3545, opened 2025-04-07
fix: naming system for duplicated layers

neesaaa:
# Improve Layer Duplication Naming

## Background

Currently in Synfig:
default

- **Duplicate Layer**  
  - You duplicate “circle003” → you get another “circle003” (no suffix)  
  - Duplicate again → still “circle003”  
- **Copy‑Paste Layer**  
  - You copy “circle003” and paste → you get “circle003”
  - Paste again → “circle003”  

This differs from common tools (e.g. After Effects), which append increasing numbers:

- **After Effects**  
  - “Layer1” → “Layer1 2” → “Layer1 3” → …  

## What’s Wrong

1. **Duplicate adds no suffix**  
   - Makes it impossible to distinguish original vs. duplicate at a glance, plus designer will need to rename which maybe slows him.
2. **Copy‑Paste adds double `##`**  
   - Because `clone()` (which appends `#`) is called twice in the copy‑paste flow.  
   - Results in “Layer##” instead of “Layer#”.

## Changes in This PR
image

- **Duplicate Layer** now appends a single `#` on each duplicate, e.g.:  
  - First duplicate → `Layer#`  
  - Second duplicate → `Layer##`  
- **Copy‑Paste Layer** behavior is unchanged for now, but i have identified why it adds `##` twice, as Layer::clone() indirectly is called twice.

## Remaining Issues & Next Steps

1. **Copy‑Paste still double‑hashes**    
2. **Adopt numbered suffixes (optional)**  
   - Consider switching from `#` to numbered suffixes (`Layer1 2`, `Layer1 3`, …) to match others.  

## Call for Feedback

- do you agree naming needs refine just wanted to check before working?

Thank you!

### DIFF
--- a/synfig-studio/src/gui/actionmanagers/layeractionmanager.cpp
+++ b/synfig-studio/src/gui/actionmanagers/layeractionmanager.cpp
@@ -536,7 +536,7 @@ LayerActionManager::paste()
 
 	for(std::list<synfig::Layer::Handle>::iterator iter=clipboard_.begin();iter!=clipboard_.end();++iter)
 	{
-		layer=(*iter)->clone(canvas, guid);
+		layer=(*iter)->clone(canvas, guid,false);
 		layer_selection.push_back(layer);
 
 		replace_exported_value_nodes(layer, valuenode_replacements);
--- a/synfig-core/src/synfig/layer.cpp
+++ b/synfig-core/src/synfig/layer.cpp
@@ -462,7 +462,7 @@ Layer::clone(Canvas::LooseHandle canvas, const GUID& deriv_guid) const
 
 	ret->group_=group_;
 	//ret->set_canvas(get_canvas());
-	ret->set_description(get_description());
+	ret->set_description(get_description()+"#");
 	ret->set_active(active());
 	ret->set_optimized(optimized());
 	ret->set_exclude_from_rendering(get_exclude_from_rendering());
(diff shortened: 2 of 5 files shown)


neesaaa:
now all are okay copy paste adds one # to name and duplicate too.

synfig-studio/src/gui/actionmanagers/layeractionmanager.cpp
```diff
@@ -536,7 +536,7 @@ LayerActionManager::paste()
 
 	for(std::list<synfig::Layer::Handle>::iterator iter=clipboard_.begin();iter!=clipboard_.end();++iter)
 	{
-		layer=(*iter)->clone(canvas, guid);
+		layer=(*iter)->clone(canvas, guid,false);
 		layer_selection.push_back(layer);
```
morevnaproject, line 540: Wouldn't  it will be easier to modify layer name right here?

morevnaproject (changes requested):

synfig-studio/src/gui/actionmanagers/layeractionmanager.cpp
neesaaa: of course, i just thought the problem was in many areas  so i made that as safety precaution, i can edit it if you want no problem but i will reset branch or make new pr

PR #3552, opened 2025-04-11
fix: bypass font caching for Devanagari MT and Devanagari Sangam MN

Stargazer10101:
A caching mechanism is in place where FaceCache stores FT_Face instances. When multiple layers use the same font (same FontMeta), they share the same FT_Face.
This works well for most fonts (like Arial), however, some complex fonts (Devanagari MT and Devanagari Sangam MN) don't render correctly when cached - multiple layers using these fonts show corruption. 
I have bypassed caching for these fonts.
Fixes: #3387

### DIFF
--- a/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
+++ b/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
@@ -578,6 +578,14 @@ Layer_Freetype::new_font(const synfig::String &family, int style, int weight)
 		new_font_("sans serif",TEXT_STYLE_NORMAL,TEXT_WEIGHT_NORMAL);
 }
 
+bool should_bypass_cache(const synfig::String& font_family) {
+    const std::vector<synfig::String> bypass_fonts = {
+        "Devanagari MT",
+        "Devanagari Sangam MN"
+    };
+    return std::find(bypass_fonts.begin(), bypass_fonts.end(), font_family) != bypass_fonts.end();
+}
+
 bool
 Layer_Freetype::new_font_(const synfig::String &font_fam_, int style, int weight)
 {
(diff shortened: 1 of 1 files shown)


rodolforg:
Interesting. So it is probably a multithread issue.
So, for performance and memory-saving issues, maybe the solution should be different.

Perhaps `Layer_Freetype::sync_vfunc()` should have a "global" mutex to avoid use of the same `FT_Face face` and/or `hb_font font` in multiple threads. Better yet, a mutex for each cached face/font.

Could you test it, @Stargazer10101 ? Instead of your current proposal, create a global mutex in `Layer_Freetype` and lock/unlock it in `Layer_Freetype::sync_vfunc()` .

Stargazer10101:
> Could you test it, @Stargazer10101 ? Instead of your current proposal, create a global mutex in `Layer_Freetype` and lock/unlock it in `Layer_Freetype::sync_vfunc()` and test if this would work, please.

This did not work. The problem is still there. I've added the commit.

rodolforg:
Sadly I can't reproduce the issue :(

Stargazer10101:
The test file can be helpful.
https://github.com/synfig/synfig/issues/3387#issuecomment-2746256143

Stargazer10101:
> Perhaps `Layer_Freetype::sync_vfunc()` should have a "global" mutex to avoid use of the same `FT_Face face` and/or `hb_font font` in multiple threads. Better yet, a mutex for each cached face/font.

I tried the "a mutex for each cached face/font" approach. I've added the commit, but it didn't solve the issue. This PR is still a work in progress.

rodolforg:
Thank you  for your investigation. #3611 is a better solution because it applies to other fonts as well.
Closing this PR then.

PR #3555, opened 2025-04-13
fix(advanced outline): fix rendering of width point in non-homogeneous position

rodolforg:
Advanced outline allows user to set width along its path, not only in its vertices.
If a path has zero-length segments (possibly due to two consecutive vertices at same coordinates), the width points were placed in a wrong position.

Before:
image

Now:
image

This PR does other stuff for improving testing, and fixing some cases of single-vertex blines I found while investigating this issue.

Reported-by: Svarov in https://forums.synfig.org/t/sonic-model-wip/16286/4

### DIFF
--- a/synfig-core/src/modules/mod_geometry/advanced_outline.cpp
+++ b/synfig-core/src/modules/mod_geometry/advanced_outline.cpp
@@ -94,7 +94,7 @@ namespace {
 	Real calc_position(Real p, const rendering::Bend &bend, bool homogeneous) {
 		return homogeneous
 			 ? p*bend.length1()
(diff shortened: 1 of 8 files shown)


Svarov-RZM:
I tested this PR and width position calculates properly now even with multiple vertexes at the same position.
Thank you very much for looking into this very specific issue.

morevnaproject (approved):

ice0:
@rodolforg Please rebase

rodolforg:
@ice0 rebased! ;)

ice0:
Merged. Thank you!

PR #3559, opened 2025-04-15
fix: missing template argument : compiler in FreeBSD complained

### LINKED ISSUE #3557: error: a template argument list is expected after a name prefixed by the template keyword [-Wmissing-template-arg-list-after-template-kw]
**Describe the bug**
I maintain this port on FreeBSD. It no longer builds as a result of this error.
**To Reproduce**
1. cd /usr/ports/graphics/synfig
2. execute make
3. build fails with the error indicated in the subject of this issue

**Expected behavior**
It has always built correctly in the past. It should continue to build.
**Screenshots**

**System information:**
 - OS: FreeBSD
 - OS version: 13,14,15
 - Synfig version: 1.4.5

**Additional comments**
This appears to be induced by a boost version change. eg;
this problem only started with the introduction of 1.86. 1.85
didn't exhibit this problem.

[GRAPHICS_SYNFIG_BUILD_SESSION.log](https://github.com/user-attachments/files/19742228/GRAPHICS_SYNFIG_BUILD_SESSION.log)

rodolforg:
Issued error message:
> error: a template argument list is expected after a name prefixed by the template keyword [-Wmissing-template-arg-list-after-template-kw]

fix #3557

### DIFF
--- a/synfig-core/src/synfig/rendering/software/function/array.h
+++ b/synfig-core/src/synfig/rendering/software/function/array.h
@@ -429,7 +429,7 @@ public:
 	};
 
 	void fill(const Type &x) const
-		{ for(Iterator i(*this); i; ++i) i.get_array().template fill(x); }
+		{ for(Iterator i(*this); i; ++i) i.get_array().template fill<Type>(x); }
 
 	template<typename TT>
 	void assign(const Array<TT, Rank> &x) const


morevnaproject (approved):

ice0:
Merged. Thank you!

PR #3560, opened 2025-04-15
feat: "Angle" converter for real -> angle conversions

### LINKED ISSUE #3528: Converter real to angle

Stargazer10101:
A new converter "Angle" that can be connected to "real" value types. Adds a sub parameter "Angle" of type "angle", so that the calculations mentioned here : https://github.com/synfig/synfig/issues/3528#issuecomment-2761450106 can be achieved.
Fix #3528

### DIFF
--- a/synfig-core/po/POTFILES.in
+++ b/synfig-core/po/POTFILES.in
@@ -326,6 +326,8 @@ src/synfig/valuenodes/valuenode_add.cpp
 src/synfig/valuenodes/valuenode_add.h
 src/synfig/valuenodes/valuenode_and.cpp
 src/synfig/valuenodes/valuenode_and.h
+src/synfig/valuenodes/valuenode_angle.cpp
+src/synfig/valuenodes/valuenode_angle.h
 src/synfig/valuenodes/valuenode_anglestring.cpp
 src/synfig/valuenodes/valuenode_anglestring.h
 src/synfig/valuenodes/valuenode_animated.cpp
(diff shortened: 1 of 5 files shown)


zaynopenapp:
I want to try Synfig on Artifact. Besides the Windows version, is there a Linux/AppImage version available?

rodolforg:
I can't see why it is useful yet. :'(

rodolforg (approved):
I now understand it is useful for stuff like arc length computation. So I'm okay about merging this PR.

morevnaproject:
Quick example how this converter will be useful:

image
I need to rotate this wheel and make the red deck to move according to rotation.

[angle-to-real-example.sifz.zip](https://github.com/user-attachments/files/20219296/angle-to-real-example.sifz.zip)

zaynopenapp:
I want to create a sample angle converter. After I saved the file and opened it again, the following error appeared.

calculation:
engleconverter_synfig

error
converter_angle_error

file:
[file_smartIK_angleconverter_error.sif.zip](https://github.com/user-attachments/files/20319739/file_smartIK_angleconverter_error.sif.zip)

morevnaproject:
Fixed by #3589

Issue #3427, opened 2024-09-29
Vertex Point's 'Width' Parameter in Advanced Outlines Doesn't do Anything (Potential Bug)


FadhilKwan:
**Describe the bug**
On 'Normal Outline', there's a parameter called 'Width' that you can use to change the width of the width points within their vertex points.

This parameter also available on 'Advanced Outlines'...... yet it doesn't seem to do anything....

**To Test**
Simply create an 'Outline' layer, just "Outline" (probably with 3 vertex points) and look at one of the Vertex Point's parameter called 'Width'. Changing the value of that point will change the line width of the vertex point it sits on.

Now replicate this but with 'Advanced Outlines' instead of normal 'Outlines'. And... it just doesn't work.

**Expected behavior**
I'm not sure what supposed to happen here since I just found out this is a thing. But maybe based on what it does to normal 'Outlines', maybe it supposed to do the same thing?
Or maybe just remove this parameter (or keep it as it is), as it is intended to do exactly nothing to 'Advanced Outlines'?

**Screenshots**
There's only one screenshot here, but this is applicable to both Outline layers (Normal and Advanced)
![advWidthNotBugMaybe](https://github.com/user-attachments/assets/8e5f8b09-c5ec-4de6-ac92-9c5f54a5b2a9)

**System information:**
 - OS: Windows 10 64bit, Home Edition (though I might eventually have to move to Mint if things go awry for Win10)
 - Synfig version: 1.4.5 (probably occurred in 1.4.2 as well but can't tell for sure without actually testing it)

**Additional comments**
Now, this is a bit to off topic to the issue ticket, but let me know if I should make another one for this matter.
I noticed that normal Outlines' width points are embedded into their respective vertex points. While this seem a bit restrictive, this does introduce a nice aspect of simplicity of not having to move the width points around AND having to do it again whenever another vertex point is added or removed; which is, as much as I love 'Advanced Outlines', is one of my biggest gripes with it.

So I wonder if it is possible to add this option into the 'Advanced Outlines', where you can choose whether to use a width point that specifically has its position Permanently Affixed/Locked into its respective vertex point. Like... maybe create a way (or a switch? or a checkmark?) to enable this option within Vertex Point's width parameter?
(yes, the exact same thing that I pointed in the screenshot. So this is related to this bug in some way, albeit barely)

This seemed a bit out of pocket, but I think it has a potential to make working with 'Advanced Outlines' feel more consistent.

[...]

Issue #3433, opened 2024-10-09
Add option to set up default interpolation


Svarov-RZM:
### Description
Synfig has three parameters that affect newly created waypoints: Default interpolation, Lock past keyframes and Lock future keyframes. Here they are:
![DefInt](https://github.com/user-attachments/assets/05d4646d-6043-4bb2-b8ce-21f70a63669f)

These parameters are hard-coded, they are always: Clamped, Lock, Lock. The problem with this is if you're making animation using different kind of interpolation, let's say, constant (you're making a frame-by-frame segment or the whole animation) **this will have to be changed on every start/revert/crash of the program**.

My last two animations were almost exclusively frame-by-frame and changing those settings whenever I opened the project became quite a nuisance not only because it's repetitive but also because sometimes I forgot to change it and ended up with some 'clamped' waypoints in my project that would mess up the flow of animation that I had to spend time fixing manually.

Also, 'Lock future keyframes' is rarely useful, so I would like the ability to change those past/future locks too, so it would be: lock past, unlock future by default.

### Mock-up
We can put these settings in Menu->Edit->Preferences->Documents. Here's a quick mock-up:
![PrefDefInt](https://github.com/user-attachments/assets/d6a5e5ab-cb53-4128-8a29-559792c928c2)

Ayush-5k:
Hey! Considering what Bob mentioned in the other thread about weighing feature value and side effects, I wanted to ask if this idea would still be worth pursuing.

While we already have controls for default interpolation and keyframe locks in the timeline bar, they reset on each launch(even after a crash). Adding these as persistent defaults in Preferences > Document might help streamline workflows—especially for those relying heavily on constant interpolation or unlocked future keyframes.

Since this relates directly to how animations are initially created (not just a UI tweak), I thought it might still be useful for some users. Would love to hear your thoughts before thinking further about it. @BobSynfig @ice0 @rodolforg

rodolforg:
@Ayush-5k feel free to do it :)

Issue #3247, opened 2023-10-22
Bring back icons in menu option


yoSachinkr:
**Describe the bug**
Radio buttons in toolbox menu option look weird and is totally unnecessary.
In earlier versions of Synfig there were icons and they looked good.
**To Reproduce**

![image](https://github.com/synfig/synfig/assets/50204208/6b624130-33c2-416e-9d8c-8f18aba47aa7)

**Expected behavior**
![image](https://github.com/synfig/synfig/assets/50204208/cf4f268f-5733-4949-ade3-57c26443b6f6)

**System information:**
 - OS: Windows 10
 - OS version: 64 bit 22h2
 - Synfig version: 1.5.1

PR #3562, opened 2025-04-15
fix: enable 'Width' Parameter in Advanced Outlines to Function as Expected

Ayush-5k:
This PR addresses an issue (#3427) with the "Width" parameter in the Advanced Outline layer, which was previously non-functional. In the Normal Outline layer, the "Width" parameter adjusts the width of the outline based on vertex point widths, and this behavior was expected to carry over to Advanced Outlines. However, in Advanced Outlines, the parameter did not have any effect, even though it was visible.

How the Bug is Being Rectified:
The "Use Vertex Widths" parameter has been introduced for Advanced Outlines, enabling the use of per-vertex width values in the outline layer. This change ensures that the width of individual vertex points will now properly affect the overall width of the outline, similar to the behavior seen in Normal Outlines. The "Width" parameter is now functional within the Advanced Outline layer, allowing users to adjust the vertex width and see the corresponding changes in the outline width, just like in the Normal Outline.
ss1
ss2

In the first image, we can see that in the Advanced Outline layer, setting the vertex width to 1000 has no effect on the outline, even though the "Width" parameter is present and adjustable. This is the bug.

Fix:
In the second image, after enabling the new "Use Vertex Widths" toggle button, the Advanced Outline layer now responds as expected. The vertex width is applied correctly, without changing the behavior of the Width Points List (or wplist), ensuring that previous functionality remains intact.

### DIFF
--- a/synfig-core/src/modules/mod_geometry/advanced_outline.cpp
+++ b/synfig-core/src/modules/mod_geometry/advanced_outline.cpp
@@ -492,7 +492,8 @@ Advanced_Outline::Advanced_Outline():
 	param_smoothness(ValueBase(Real(1))),
 	param_homogeneous(ValueBase(false)),
 	param_dash_offset(ValueBase(Real(0))),
-	param_dash_enabled(ValueBase(false))
+	param_dash_enabled(ValueBase(false)),
+	param_use_vertex_widths(ValueBase(false))
 {
 	clear();
 
@@ -569,7 +570,7 @@ Advanced_Outline::sync_vfunc()
 		const Real gv = exp(get_outline_grow_mark());
 		const Real wk = 0.5*gv*width;
 		const Real we = gv*expand;
-		const bool use_bline_width = wplist.empty();
+		const bool use_bline_width = wplist.empty() || param_use_vertex_widths.get(bool());
 		
 		// build bend
 		rendering::Bend bend;
@@ -712,6 +713,7 @@ Advanced_Outline::set_shape_param(const String & param, const ValueBase &value)
 	IMPORT_VALUE(param_homogeneous);
 	IMPORT_VALUE(param_dash_offset);
 	IMPORT_VALUE(param_dash_enabled);
+	IMPORT_VALUE(param_use_vertex_widths); 
 
 	// Skip polygon parameters
 	return Layer_Shape::set_shape_param(param,value);
@@ -732,6 +734,7 @@ Advanced_Outline::get_param(const String& param)const
 	EXPORT_VALUE(param_homogeneous);
 	EXPORT_VALUE(param_dash_offset);
 	EXPORT_VALUE(param_dash_enabled);
+	EXPORT_VALUE(param_use_vertex_widths);
 
 	EXPORT_NAME();
 	EXPORT_VERSION();
@@ -823,6 +826,11 @@ Advanced_Outline::get_param_vocab()const
 		.set_is_distance()
 		.set_description(_("Distance to Offset the Dash Items"))
 	);
+	ret.push_back(ParamDesc("use_vertex_widths")
+		.set_local_name(_("Use Vertex Widths"))
+		.set_description(_("When checked, uses the Width value of each vertex point (from the Spline) instead of the Width Points list"))
+		.set_hint("bool")
+	);
 	return ret;
 }
 
@@ -876,4 +884,9 @@ Advanced_Outline::connect_bline_to_wplist(ValueNode::LooseHandle x)
 		return false;
 	wplist->set_bline(ValueNode::Handle(x));
 	return true;
+}
+
+bool 
+Advanced_Outline::get_use_vertex_widths() const {
+    return param_use_vertex_widths.get(bool());
 }
--- a/synfig-core/src/modules/mod_geometry/advanced_outline.h
+++ b/synfig-core/src/modules/mod_geometry/advanced_outline.h
@@ -67,7 +67,8 @@ private:
 	synfig::ValueBase param_dash_offset;
 	//! Parameter: (bool)
 	synfig::ValueBase param_dash_enabled;
-
+	//! Parameter: (bool)
+	synfig::ValueBase param_use_vertex_widths;
 public:
 	enum CuspType
 	{
@@ -85,6 +86,7 @@ public:
 
 	//! Connects the parameter to another Value Node. Implementation for this layer
 	virtual bool connect_dynamic_param(const synfig::String& param, synfig::ValueNode::LooseHandle x );
+	bool get_use_vertex_widths() const;
 
 private:
 	bool connect_bline_to_wplist(synfig::ValueNode::LooseHandle x);


Svarov-RZM:
Hey, I appreciate the effort, but what exactly do we achieve with this?
So now we can downgrade Advanced Outline to a normal one, what's the point if we can just use the normal outline in the first place?

> In the Normal Outline layer, the "Width" parameter adjusts the width of the outline based on vertex point widths, and this behavior was expected to carry over to Advanced Outlines

Where are you getting this information from? As far as I know it was not expected because Adv. Outline uses entirely different width system.

Ayush-5k:
Thanks for the feedback @Svarov-RZM ! 

My goal wasn’t to downgrade the Advanced Outline, but to make it more flexible by supporting both systems—vertex widths and the existing wplist. This change doesn’t interfere with the current behavior—users can still rely on wplist as before.

I added this option so that users who prefer direct per-vertex width control (like in the Normal Outline layer) won’t need to switch layer types or recreate their outlines from scratch. It just makes the Advanced Outline layer more adaptable to different workflows and preferences.

Also, this behavior was mentioned in issue [#3247](https://github.com/synfig/synfig/issues/3247), where it was noted that the "Width" parameter was expected to affect the Advanced Outline similarly to how it does in the Normal Outline. My intention was to meet that expectation while keeping the advanced control intact for users who need it.

Svarov-RZM:
Okay, I see your point about flexibility but I still disagree. Separate width list is almost a sole point of Advanced Outline, it just no longer advanced without it. Why would a user want to convert an Adv. Outline to standard one? I would like to know if there are any people that actually want this and their reasoning.

> where it was noted that the "Width" parameter was expected to affect the Advanced Outline similarly to how it does in the Normal Outline

The OP never stated what was expected because he simply didn't know. Here's a quote from the linked issue:

`I'm not sure what supposed to happen here since I just found out this is a thing. But maybe based on what it does to normal 'Outlines', maybe it supposed to do the same thing?
Or maybe just remove this parameter (or keep it as it is), as it is intended to do exactly nothing to 'Advanced Outlines'?`

I am pretty sure the original intention was to keep "Width" parameter non-functional in Adv. Outline. The reason why it wasn't removed is because it would force us to have two different lists for splines one with "Width" parameter and another one without it which would complicate exporting/linking.

> i was wondering if it was still a good idea to still implement the issue you raised a while earlier: #3433

Umm... Yeah. How can I say no to my own issue? :)
Look, I don't think you realize, but I am just a pitiful user. I don't have a saying in what will get merged and what won't. I just dropped an issue that would benefit me (egoistic, I know). So if you think the linked issue is worth your time then I won't stop you (plus the issue was labeled, that's a good thing).

Ayush-5k:
>Okay, I see your point about flexibility but I still disagree. Separate width list is almost a sole point of Advanced Outline, it just no longer advanced without it. Why would a user want to convert an Adv. Outline to standard one? I would like to know if there are any people that actually want this and their reasoning.

Just wanted to share where I'm coming from—but I totally see your point too. I understand that the wplist is a core part of what makes Advanced Outline “advanced,” and I definitely don’t want to undermine that. Just to clarify, the toggle I added is fully optional and doesn’t interfere with wplist behavior at all—it simply allows for an alternate mode if the user wants it.

You're absolutely right that the original issue was more exploratory than a confirmed bug report, and I can see how the ambiguity there may have led to different interpretations. That said, I agree it would be great to hear from users on whether this is something they would actually use in real projects. 

>Umm... Yeah. How can I say no to my own issue? :)
Look, I don't think you realize, but I am just a pitiful user. I don't have a saying in what will get merged and what won't. I just dropped an issue that would benefit me (egoistic, I know). So if you think the linked issue is worth your time then I won't stop you (plus the issue was labeled, that's a good thing).

Haha, come on, give yourself more credit! We’re all just users trying to make things better 😄
Whether it's reporting an issue or suggesting an idea, it's all part of contributing. Your input helps guide things in the right direction and it really helps new contributors like me understand what’s useful or needed.

FadhilKwan:
Whoaaaa, okay! Hold on~ (O-O;
I did not expect this ticket of mine to spark a little bit of debate (let alone getting a 'fix', but thanks alot @Ayush-5k , I really appreciate it! :D ).
So let me explain a few things (which many were already written in my issue ticket, but I still want to get my point across.)

_(Also deep apologies for only noticing this right now. I'm getting a quite busy with real life stuff I kept forgetting to check in here from time to time .-.v )_

1. Yeah, as Svarov mentioned, my issue ticket was not meant to be taken as a feature request or bugfix. It is really just a question to clarify what the particular option (which in this case, the vertex's 'Width' on advanced outlines) meant to do. Whether it is supposed to work properly, or is it just a dud parameter that was meant to be removed.

**_(Though I do have a bias towards the former...., I'll try to explain~)_**

2. I agree with Svarov's reasoning, that the movable width points are the ones that define normal outlines and advanced outlines apart, making this addition seem very trivial (or even pointless to an extent). But I do have a few reasons for this "feature request" (and; I'm with Svarov on this, it might be a little bit selfish >-<; ).
I have been using Adv. Outlines almost all the time. And one thing that keeps bugging me endlessly is that I cannot lock a specific width point(s) in place (or in this case, any vertex point(s)). This is not a problem at all, except that my workflow demands me to reuse the same line over and over again (basically duplicating them and changing their shapes for every stroke/outline). This includes adding/removing width points that are only meant to be stylizing those strokes. And altering the number of width points can screw up the width points' structures. This can get very annoying when I'm trying to stylize lines with already well-defined width points, like when I'm drawing something spiky (as in lines with sharp edges), or any other lines with defined edges.
So, a width point(s) that'll remain stationery regardless of any changes to the width point count could help speed up my workflow process, as I don't have to rearrange them too often. And this addition, could help solve that problem.

In case that was a bit convoluted, the simplest example of this are hair strands.
I'd start by making a hair strand that is almost looped (but they're not), made of a region layer and an advance outline layer, but the line for it is cut off a bit short to make a hanging region that I can use to mask other future hair strands below it. After that, I'd just duplicate the line and its region over and over to shape up the hair.
This often calls for more or less width points, to give them some variety in thickness. Which; whenever it happens, will offset the other width points, including the point that I cut off. So I have to rearrange the remaining width points again to their original positions.
This example could apply for clothes, drapes, and folds as well (and maybe more).


With that said, this still hasn't answered the other question;
**"How useful is this feature would be? Is it even worth it to implement in the first place?"**

....I dunno, I haven't tested it yet...
Or more like, I couldn't. I came here way too late, so the AppVeyor's artifact for this has expired. And this option won't show up on the newer artifacts for some reason.... ._.?

But if I have to predict, I might occasionally use this feature. (heck, maybe more often than that!).
After all, I kind of yearn for it....?

When I found out that the normal Outline can do that, I actually got excited and wanted to try it on Adv. Outlines. only to find out that it didn't work for some reason. So I created this ticket.
At first I would just accept the final outcome, even if it means that this parameter would be removed forever. Sure, I'd be very disappointed, but I don't think a small setback like the one I mentioned above should be able to stop me whatsoever.
But knowing that this has a potential to be kept in, it actually makes me very happy~
Moreover, you made this feature optional, and it is set to 'Off' by default, which is a really nice touch! It helps to avoid throwing long time users out of the loop, while still only giving the option to the ones that need it.

So if anything, even though this feature might be the least desirable one, for me this is a huge QoL, honestly!

What it only lacks is that the vertex widths are missing their own width points, and maybe being able to only enable this option to the chosen _(edit: vertex)_ points would also be nice. But I wouldn't mind them too much...
**.....actually, I'm just worried that adding those would require a major code overhaul, which if so, I'd say please, don't bother with those small details. This alone is already more than enough of an addition to me~ ._.v**

Thank you so much again for this addition, Ayush!! Hopefully it'll get polished up and implemented in the next version~ 🙏 
(or heck, maybe even in the next AppVeyor's Artifact, this might finally convince me to move to the newer Synfig, even though they still have some issues on their canvas rendering part xD )

Alright, that is my take. Hope someone else can give their take on this addition, whether this one is worth it or not. It might mean that I wont be getting this feature. But whichever works best, I guess~ :)

_p.s. Idk why, but everytime I try to comment on something I kept making an essay by accident....... Oh well..... .-.;_

Svarov-RZM:
> 2\. And altering the number of width points can screw up the width points' structures

It can't... You just right-click on a width point and select "Insert item and Keep Shape". If it does alter the width structure, then it's a bug.

I am gonna be honest I still don't understand what exactly is the problem. Can you provide some screenshots or a simple test-case file on how to reproduce your scenario? What I have to do to get to the point where you see a problem?

FadhilKwan:
> > 2. And altering the number of width points can screw up the width points' structures
> 
> It can't... You just right-click on a width point and select "Insert item and Keep Shape". If it does alter the width structure, then it's a bug.
> 
> I am gonna be honest I still don't understand what exactly is the problem. Can you provide some screenshots or a simple test-case file on how to reproduce your scenario? What I have to do to get to the point where you see a problem?

Oh.... yeah, about that....

I.... misremembered my own issue..... Sorry about that.... ><

I read your reply and thought, "yeah, I've been doing that. So..... what was the issue again?"
And then I finally re-remembered it all again.

Turns out my point of issue was not centered on width points at all. Sure they can contribute in a few ways, but that is definitely not the main point (pun intended).
My gripes was with altering the vertex point count all along (and with line length alteration on the side).

So what I mean is, when I want to use the same line, most of the time I'd have to slightly tweak it, which; almost all the time, involves adding or removing vertex point(s). This will cause all of the width points to shift slightly to a side. (unless it is located at the start and end of the line a.k.a. position 0 and 1, which will not get altered at all or barely noticeable.)
The line length (stretching and shrinking) can contribute to messing up width points as well, but there's already a quite a viable fix for it; simply disable the 'Homogeneous' option. should do the trick.

Just replace some of the words in my previous comment from "width point(s)" to "vertex point(s)", and that should clear it up. (no worries, I'll correct them up later ._.v )

As for a simple scenario for it, here's one that I just created. It is hopefully as simple as it can get:

So here, I made a shape of a hair strand. Starting off with simple three-point shape, made of a region and an adv. outlines. Sharp line at the spike, and widening strokes before and after it,
As for the 'hanging region' bit, I'd tweak it to look like the shape on the right. A length of the line has been cut off by changing that width point's cusp type.
I'll be using the shape on the right, for the sake of clarity.
wplockingproposalb-1

As you can see, I've circled some of the vertex points. These are the spots where I'd put my width points, with each having their own widths and setups, especially the cutoff point. And I'd like to keep these points in place, right on their respective vertex points' positions, not moving at all.
wplockingproposalb-2

Alright, now, as part of my workflow, I would duplicate this shape (the entire advanced outline+region pair). And after that, I'd alter it slightly to make a longer, wavy strand, while still using the points that I've established earlier as the guides.
But as soon as I add my vertex points to start with, this happened:
wplockingproposalb-3

A slight mess-up has occurred, nudging almost all of my width points all over the place.
wplockingproposalb-4

Thus, I have to nudge them back tho where they where, back on their vertex points' position.

Now imagine this, but about 100+ times. And with increasingly complex lines and more vertex points to alter, there will be more nudging width points to deal with...
And that's where these inconveniences will start to get me .-. .

_**So, the main thing is, what I really (Really!) want, is a way to lock some width points of my choice to a vertex.**_ And so far, I haven't found any solution for it.
(I might have mentioned this on my issue ticket as well, but I'd like to put this in again, just to make it neat and clear~)

_edit:
I also would like to add, that **I'm not intending to lock ALL width points that I made. If so I'd just use normal outlines instead.**
I just want to lock few width points that feels important to me, in order to keep the line structure consistent, such as start and endpoint, any lines with sharp changes in width (like spiky edges and such), or any established points that I have no intention to move around whatsoever. As for the rest, I'd just freestyle them as much as I like~ :)
Also, using normal outlines as the sole solution for this issue can cause me quite a pain on the long run, as the vertex count for each subsequent lines can and will grow higher and higher, and they would become much harder to manage x!_

One way would be to make it as an option, like link to spline, but it'd be link to vertex instead. Which, while I think sounds good, it might cause some issues when someone remove a vertex point that currently has a width point(s) connected onto it.
And, as of now (on 1.4.4), trying to link width point's position onto a vertex point directly will result on this error:
wplockingproposalb-5
_(btw I can't convert width point position into vector, I just tried it now...)_

Another way would be what this fix is intending to do. Adding an option to alter the line width within the vertex points' parameter, just like what it does with the normal outline. While it doesn't come with cusp altering options (which would hopefully be the next step, but I won't force y'all to do it~ .-.v ), this already a really good addition to me.

Once again, I'm very sorry for the confusion. I'll try fix up my last comment just to clear things up and avoid some misinform. And hopefully that clears up my reasoning.

Or. maybe, do you have a better plan in mind? I'm kinda hoping so? I genuinely would like to know if there's a better way to implement this~
Also thanks alot for the reply~  ^^v

Svarov-RZM:
> And hopefully that clears up my reasoning

Thank you for detailed explanation, I read it carefully and believe I understand your struggle now.

It seems to me that a proper solution would be to implement a new context menu with the option "Insert Item & Keep Width". After hitting it, a new vertex will be inserted in the desired position and all width points of the spline will get recalculated to stay in the same place. 

Locking specific width points to a vertex is basically the same thing but more awkward since instead of clicking one option user will have to go over all needed width points and do the same thing on them.

> Another way would be what this fix is intending to do

This fix just removes "advanced" from Advanced Outline. As you said, there's no cusp functionality and separate width points are just ignored. In theory, we could make another outline layer - Super Advanced Outline layer where each vertex point has embedded width and dash points where their positions are relative to the next/previous vertex. I would totally use it.

FadhilKwan:
> This fix just removes "advanced" from Advanced Outline. As you said, there's no cusp functionality and separate width points are just ignored.

Oh.... 
I thought the option will make it possible to work with both Advanced Outlines' WP  and normal Outlines WP in tandem. So.... they cant?
Awwww man, I'm quite disappointed now... ;-;
I... didn't read the pull request carefully, dang silly me... 🙃

> It seems to me that a proper solution would be to implement a new context menu with the option "Insert Item & Keep Width". After hitting it, a new vertex will be inserted in the desired position and all width points of the spline will get recalculated to stay in the same place.
>
>Locking specific width points to a vertex is basically the same thing but more awkward since instead of clicking one option user will have to go over all needed width points and do the same thing on them.

Yeah, actually, that could work as a possible alternative, I think I really like that Idea! We already have insert Item and keep shape option. Maybe, just maybe, the same thing could be done with this one?
The locking specific WP can get awkward especially when their count starts growing big. Though it might be able to give some advantage if you're drawing the lines from scratch? Also less prone to line stretching and shrinking if you much prefer using the 'Homogenous' line option? Idk... ._.?

If we follow your theory for this context menu, it might call for the 'Homogenous' option on Advanced Outlines to be on, since that'll keep the width points from being affected by additional vertex points.

**And I just checked now, I _Can_ just keep the advanced outline's 'Homogenous' option on. I can add a new vertex point without it nudging the WPs, and most of my problems would be gone in an instant.**
_**But. Not Quite.**_
I'm not exactly a fan of the 'Homogenous' option. As I mentioned, that option will make the WPs alot more prone to line stretching/shrinking. Like, they could just get moved to an entirely different line section by doing so. This option could be useful for distance-based stuff, But to me, most of the time; especially with my drawing/animation workflow, this is quite a big hinderance.
In other words, I much prefer to have it turned 'Off' on most cases.

_Also turning 'Homogenous' off or on **Will** change the WPs' positions, as it does not convert the current WPs position value to their Homogenous equivalents. It just takes the values on that parameter and uses those instead._


Soooo.... this would probably involve converting the current WPs' positions to their Homogenous equivalents, then add the new vertex, and finish it off by converting the WPs' positions back to their Non-Homogenous value (or whatever it is called).

In any case, I think I like your solution better. Seems easier to implement (at least on paper), and it is a quick; almost one-click solution to my problems. But the only way to find out is by test it~ (if we do get it :) )

> In theory, we could make another outline layer - Super Advanced Outline layer where each vertex point has embedded width and dash points where their positions are relative to the next/previous vertex. I would totally use it.

Oh dang, Super Advanced Outline!! I like that name~ xD
Yeah, I'd unironically gonna use this feature in a heartbeat!
But how'd the value for the WPs' positions be calculated tho? In decimals?
Or maybe just by counting? i.e. 1 for WP 1 on Vertex Point 1, 2 for WP 2 Vertex Point 2, etc.

rodolforg:
Well, my two cents here:

1. `Insert and keep shape` should behave with special care if it is an Advanced Outline and let the Width points where they visually were (not numerically).
2. @FirasH , If you don't need dashes, regular Outline layer wouldn't be enough? That's where the vertex width works.

FadhilKwan:
> Well, my two cents here:
> 
>     1. `Insert and keep shape` should behave with special care if it is an Advanced Outline and let the Width points where they visually were (not numerically).
Oooohhh, I see both you and Svarov are on the same page regarding this matter. And you know what? I fully support that idea! Probably one of the best alternatives to this pull request~

>     2. @FirasH , If you don't need dashes, regular Outline layer wouldn't be enough? That's where the vertex width works.
I'm guessing this is supposed to be directed to me? If not, I'm still answering it anyways :)
Sure, normal Outline layer can solve this matter. But I want to be able to stylize my lines while keeping the vertex count as low as possible, as it will become very hard to manage once it reaches like 6+ or even somewhere on double digits; as low as 10. And sometimes I might just need more dynamic, freely movable WPs and nothing else.
Also I still need those Advanced Outlines' cusps. They're very useful to cut-off my lines as much as I need. ._.;

Svarov-RZM:
> If we follow your theory for this context menu, it might call for the 'Homogenous' option on Advanced Outlines to be on

It's not important. Position of width points will always be calculated correctly whether Homogeneous parameter is on or not. It's just a small check for the implementation algorithm to be aware of.

> But how'd the value for the WPs' positions be calculated tho? In decimals?

Well, if we wanna talk about something that's has almost zero chances to be implemented (having 3 different outline layers in one program, we may be the only ones into that) then here's a quick mock-up:

SuperAdvOutline

In this new layer, each vertex has a width point, just like the normal outline but more complex one. By default, it has position "0.5" meaning at current vertex, "0.0" means at previous vertex, "1.0" means at next vertex and something like "0.6" would get you slightly towards the next vertex, you get the idea.

Pros of this type of outline:
* Everything is in one structure, more intuitive for the user
* If you need to lock specific width point to a vertex position now you can
* When inserting a new vertex all width point will remain in place except for neighbours

Cons:
* You can't have more or less width points than there are vertexes (like in Adv. Outline)

Is it really worth having this kind of layer? Putting my personal bias aside, I'd say no. The gains are just very specific to justify a whole new layer.

FadhilKwan:
> It's not important. Position of width points will always be calculated correctly whether Homogeneous parameter is on or not. It's just a small check for the implementation algorithm to be aware of.

Yeah, that should be how it supposed to work. Right now, adding more vertex points doesn't readjust the WPs' position's value based on their actual positions within that line, before and/or after adding the new VP. The positions value of all WPs would stay the same, and that's what messes the whole WPs up.

> If we follow your theory for this context menu, it might call for the 'Homogenous' option on Advanced Outlines to be on

At that moment I was trying to make an assumption/theory on how the 'add vertex and keep shape (including width)' fix would work in Synfig.
Of course, the 'Homogenous' option shouldn't be necessary to be turned 'on' at all to make this possible. But the whole process of adding vertex and keeping width points in place might involves converting the WPs' position into another unit in order to keep the WPs consistencies after adding the vertex, before converting them back to the original form(?).

And the thing with Homogenous, is that it calculates the WP's positions' value based on the line's total length (i.e. start at 0, 0.5 at halfway, and 1 at the end). Which is why it can be helpful, compared to its non-Homogenous...... thing..... (counterpart?).
_It is also why WPs get all messed up upon stretching or shrinking the line when this option is on._

If Homogenous is UNchecked, the WP's positions are calculated based on how many vertex points that are currently present. So if there are 5 vertex points, WP on vertex 1 has a position value of 0, and for the remaining vertex points, they are calculated in percentages (converted to their decimal forms). So back to the 5 vertex points example, WPs on 2nd VP would be position 0.25, on 3rd would be 0.5, 4th is 0.75, and 5th is 1. 
_Which is why adding more vertex points shuffles these WPs around, as their positions' values remain unchanged after adding the new VP, and they'd get **"repositioned accordingly"** as the side effect._

But there might be a better way to calculate this bit. After all, I'm just assuming on a theory that Synfig cannot see the WPs positions and only takes all the values in the parameter as they are, before positioning the WPs based on the Homogenous option... thing....

Also, I'm kinda considering that turning Homogenous on/off Should Not Change The WPs Positions At All.
Instead, it might be better to have those positions' values converted to the desired form, regardless if you use Homogenous of not.
There might be a few downsides like Synfig lagging out from calculating these WP positions, tho. So I'd think twice before asking for this one. 

>Cons:
>
>    You can't have more or less width points than there are vertexes (like in Adv. Outline)

Yeaaaahhh... That would be a deal-breaker for me sadly. :(
I don't know why I kept thinking that normal and advanced WPs can work together, even in this context. Maybe a bit too much of wishful thinking~

Issue #3389, opened 2024-07-29
get_color() not implemented correctly in Blur Layer


morevnaproject:
This function -
https://github.com/synfig/synfig/blob/d2816730fd6e0a58785fd39571ec3f728e99c616/synfig-core/src/modules/mod_filter/blur.cpp#L118-L134

As result we get bad behavior of Eyedropper tool on Blur Layer.

PR #3563, opened 2025-04-16
fix: Implemented get_color() correctly in Blur layer

Ayush-5k:
Problem:
The get_color() function in the Blur_Layer was not correctly implemented (as mentioned in issue #3389), leading to unexpected behavior when using the Eyedropper tool on blurred areas. Specifically, the color picked by the Eyedropper would not accurately reflect the blurred area, resulting in inconsistencies in the preview.

Solution:
This update improves the get_color() function by introducing an averaging technique. Instead of sampling a single point, the function now samples multiple points around the target position, based on the blur size. This ensures that the color returned is a more accurate representation of the blurred area.

The following changes were made:
- The color is now averaged over a small sample grid (steps = 3), with each point within the grid contributing to the final color.
- This approach simulates a more accurate blurred color, reflecting the visual appearance of the blur effect.

BEFORE:
Screenshot 2025-04-17 025956

AFTER:
Screenshot 2025-04-17 025933

### DIFF
--- a/synfig-core/src/modules/mod_filter/blur.cpp
+++ b/synfig-core/src/modules/mod_filter/blur.cpp
@@ -117,20 +117,35 @@ Blur_Layer::get_param(const String &param)const
 Color
 Blur_Layer::get_color(Context context, const Point &pos)const
 {
-	synfig::Point size=param_size.get(Point());
-	int type=param_type.get(int());
-  	size *= rendering::software::Blur::get_size_amplifier((rendering::Blur::Type)type)
-  	      * ::Blur::get_size_amplifier(type);
-	
-	Point blurpos = Blur(size,type)(pos);
+	synfig::Point size = param_size.get(Point());
+	int type = param_type.get(int());
+	size *= rendering::software::Blur::get_size_amplifier((rendering::Blur::Type)type)
+	      * ::Blur::get_size_amplifier(type);
 
-	if(get_amount()==1.0 && get_blend_method()==Color::BLEND_STRAIGHT)
-		return context.get_color(blurpos);
+	const int steps = 3;
+	Color accum;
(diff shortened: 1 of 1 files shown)


BobSynfig:
```
	const int steps = 3;
	Color accum;
	int count = 0;

	for (int y = 0; y < steps; ++y)
	{
		for (int x = 0; x < steps; ++x)
		{
			Real dx = size[0] * ((Real)x / (steps - 1) - 0.5);
			Real dy = size[1] * ((Real)y / (steps - 1) - 0.5);
			accum += context.get_color(pos + Point(dx, dy));
			++count;
		}
	}
```

As a general tip (for speed/efficiency), always try to precalculate expensive redundant parts (especially with divisions and reals) **outside** of loops.
You can keep you current code as illustration in comments but optimization of loops are one of the most common places where you can make the difference in speed.
Keep them as small as possible (in term of executed code, not in term of source)
Synfig source if full of such slowing-down code :/

One possibility is:

```
const int steps = 3;
// count is always 9 (steps * steps), no need to increment in loop.
const int count = 9;
Color accum;

// Pre-calculate step sizes and half-extents
const Real stepScale = 1.0 / (steps - 1); // 0.5
const Real offsetX = size[0] * 0.5;
const Real offsetY = size[1] * 0.5;

for (int y = 0; y < steps; ++y) {
    // Calculate dy once per row
    Real dy = size[1] * (y * stepScale - 0.5);
    
    for (int x = 0; x < steps; ++x) {
        // Calculate dx once per column
        Real dx = size[0] * (x * stepScale - 0.5);
        accum += context.get_color(pos + Point(dx, dy));
    }
}
```
(Note: it's an illustration, generated by Gemini)

PR #3564, opened 2025-04-18
fix: avoid buffer overflow in Widget_SoundWave

rodolforg:


ice0:
Looks good, please rebase.

rodolforg:
Rebased.

synfig-studio/src/gui/widgets/widget_soundwave.cpp
```diff
@@ -333,11 +335,18 @@ bool Widget_SoundWave::do_load(const synfig::filesystem::Path& filename)
...
-			int buffer_length = (end_frame - start_frame + 1) * _channels * bytes_per_sample * std::round(_frequency/fps);
+			int buffer_length = (end_frame - start_frame) * _channels * bytes_per_sample * std::round(_frequency/fps);
 			buffer.resize(buffer_length);
 		}
 		int _n_bytes = _n_samples * _channels * bytes_per_sample;
-		std::copy(static_cast<unsigned char*>(_buffer), static_cast<unsigned char*>(_buffer) + _n_bytes, buffer.begin()+bytes_written);
+		if (bytes_written + _n_bytes > buffer.size()) {
+			if (buffer.size() <= bytes_written) {
+				synfig::error(_("Internal error: Widget_SoundWave: trying to read more bytes than buffer size: %i x %zu"), _n_bytes + bytes_written, buffer.size());
+				break;
```
ice0, line 345: `delete frame;`?

rodolforg: Yes!
But then I use unique_ptr instead just in case.
Or do you think it would be better the delete (I don't know how it affects performance)

ice0: For unique_ptr, there is no difference, so it looks better in my opinion

ice0 (approved):

ice0:
Merged. Thank you!

PR #3572, opened 2025-04-21
fix: restore Quality option to synfig command line tool

### LINKED ISSUE #2872: Impossible to set 'Quality' in command line application
**Description**
By some reason, the '-Q' switch was removed from the command line application (synfig), so it's no longer possible to set quality of the animation. You can restore it easily, since all the functionality is still in place.

But I was wondering, does setting 'Quality' affects anything after engine switch? I remember it did back in the day, so I am not sure if we should restore in back for command line application, or remove it from GUI for consistency.

**To Reproduce**
Just try to render any animation with '-Q 6' option, like so: `synfig.exe -Q 6 --time 0f C:\MyCoolAnimation.sif`, synfig will bail out with: "Exception: Unknown option -Q"

**Expected behavior**
To render animation with specified quality or at least silently ignore it.

rodolforg:
--quality or -Q

fix #2872

### DIFF
--- a/synfig-core/src/tool/optionsprocessor.cpp
+++ b/synfig-core/src/tool/optionsprocessor.cpp
@@ -154,7 +154,7 @@ SynfigCommandLineParser::SynfigCommandLineParser() :
 	add_option(og_set, "height",      'h', set_height,		_("Set the image height in pixels (Use zero for file default)"), "NUM");
 	add_option(og_set, "span",        's', set_span,		_("Set the diagonal size of image window (Span)"), "NUM");
 	add_option(og_set, "antialias",   'a', set_antialias,	_("Set antialias amount for parametric renderer."), "1..30");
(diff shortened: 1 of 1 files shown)


rodolforg:
@Svarov-RZM could you please test it? :P

Svarov-RZM:
> could you please test it? :P

There isn't much to test though, is it? :)
I confirm that synfig tool no longer throws error on specifying quality. 
Don't know what else to test since layers that were ported to Cobra don't support quality setting.

We need to have some discussion in the future about what to with quality parameter overall.
Maybe we should re-introduce it to new engine somehow? Or perhaps throw it away completely, I don't know...

rodolforg:
Hmmm. I was looking for old issues I commented XD
Indeed, I believe quality is fixed to 4 in both GUI and CLI.

And it affects only 4 layers now.

PR #3573, opened 2025-04-22
feat : Default Interpolation

Ayush-5k:
This PR introduces the feature of default interpolation added in the Edit > Preferences > Document section as mentioned in the issue(fixes : #3433).
image

### DIFF
--- a/synfig-core/src/synfig/canvas.cpp
+++ b/synfig-core/src/synfig/canvas.cpp
@@ -87,7 +87,8 @@ Canvas::Canvas(const String &id):
 	cur_time_	(0),
 	is_inline_	(false),
 	is_dirty_	(true),
-	outline_grow(0.0)
+	outline_grow(0.0),
+	interpolation_mode_(INTERPOLATION_CLAMPED)
 {
 	identifier_.file_system = FileSystemNative::instance();
 	_CanvasCounter::counter++;
--- a/synfig-core/src/synfig/canvas.h
+++ b/synfig-core/src/synfig/canvas.h
@@ -262,6 +262,7 @@ private:
 	/*! \see get_grow_value set_grow_value */
 	Real outline_grow;
 
+	int interpolation_mode_; 
 
 	/*
  -- ** -- S I G N A L S -------------------------------------------------------
@@ -683,6 +684,9 @@ public:
 	//! Set/Get members for the outline grow value
 	Real get_outline_grow()const;
 	void set_outline_grow(Real x);
+
+	void set_interpolation(int mode) { interpolation_mode_ = mode; }
+    int get_interpolation() const { return interpolation_mode_; }
 
 #if 0
 	void show_canvas_ancestry(String file, int line, String note)const;
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -288,6 +288,9 @@ String        studio::App::default_background_layer_type  = "none";
 synfig::Color studio::App::default_background_layer_color =
 	synfig::Color(1.000000, 1.000000, 1.000000, 1.000000);  //White
 String        studio::App::default_background_layer_image = "undefined";
+bool App::default_keyframedial_past = false;  
+bool App::default_keyframedial_future = false; 
+int App::default_interpolation = INTERPOLATION_CLAMPED; 
 synfig::Color studio::App::preview_background_color =
 	synfig::Color(0.742187, 0.742187, 0.742187, 1.000000);  //X11 Gray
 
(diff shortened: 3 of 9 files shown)


Svarov-RZM:
Thank you for implementing this! There's a bug though.
Set default interpolation like so:
DefInt

Now, restart Synfig and on an empty canvas enter animation mode.
And you can see the locks have reverted back after we entered animation mode:
DefState
AniState

Ayush-5k:
Hey @Svarov-RZM,

I looked into the bug and noticed that in the function
**CanvasView::on_mode_changed(CanvasInterface::Mode mode)** (located in canvasview.cpp),
the line:
**keyframedial->on_mode_changed(mode);**
was being called each time, which seemed to be resetting the keyframe dials to their default values. After commenting out this line, the issue no longer occurred.

Could you please confirm if this is the intended fix and aligns with the behavior you were expecting?

Svarov-RZM:
> Could you please confirm if this is the intended fix and aligns with the behavior you were expecting?

Cannot confirm. PR does fix the initial visual feedback, but ultimately doesn't do anything.
Lock past and future keyframes are still being overridden somewhere...

How to reproduce:
1. Set default interpolation in Edit->Preferences->Document as Constant, unlock, unlock
2. Restart Synfig
3. On an empty canvas add a keyframe on 1s (24f)
4. Create a circle
5. Go to 12f on timeline, enter animated mode and set Circle's opacity parameter to 0.5
6. Observe that 3 waypoints were created even though we said unlock everything
7. Hit CTRL+Z to undo waypoint creation, lock past and future keyframes (in main window, not in the settings)
8. Repeat entry 5.
9. This time only 1 waypoint was created, so initial state on lock past and future keyframes that we set in the settings doesn't do anything

Ayush-5k:
Have some assignments and stuff going on, but will work on it ASAP and try to rectify the things you mentioned above.

Ayush-5k:
Hey @Svarov-RZM,
I tried replicating the steps you mentioned, and after making the changes, it seems to be working correctly now, which appears to contradict what you mentioned earlier (please correct me if I'm wrong). Could you please confirm if the default interpolation is now working properly?

Svarov-RZM:
> which appears to contradict what you mentioned earlier

Where did I contradict myself? I could have I guess, but I honestly don't remember...

> Could you please confirm if the default interpolation is now working properly?

Well, almost. We're back to initial bug where interpolation changes after you enter animation mode but now it's no longer reproducible on empty canvas but on saved work.

Look, maybe I suck at explaining things. Let's dive for a moment into how Synfig works. We've got waypoints - they are values in time. You go somewhere on the timeline, enter animation mode, change some parameter and Synfig adds a waypoint at that time. If there are 2 waypoints, we can now interpolate intermediate values between them using a specified interpolation (we've got a bunch of them: clamped, linear, TCB, etc.). Simple enough, right.

Now, keyframes on the other hand don't do anything on their own, they are simply markers added by user to signify something important, like "start of walk cycle at 10s_12f". When we enter animation mode and add a waypoint, Synfig immediately adds it on the time we asked it to, then if "Lock past KFs" is active Synfig searches first keyframe before current time and adds a waypoint there with initial value that was changed, therefore "locking" it. The same goes for "Lock future KFs" but Synfig looks ahead of current time.

So now, if none of "locks" are active we only get 1 waypoint on editing a parameter in animation mode, if both are active we get 3 waypoints and 2 waypoints if only one of the "locks" is active. Hopefully things are clear now. If I contradicted myself somewhere, I am sorry, I didn't mean to.

Your current logic almost works properly, but there are 3 bugs (1 major and 2 minor). The first major bug is that "locks" reset on entering animation mode. Here's the video:

https://github.com/user-attachments/assets/56d3e743-ef21-4577-b7a0-081f314ed01a

To reproduce:
1. Open this project:
[DefaultInterpolation.zip](https://github.com/user-attachments/files/19935915/DefaultInterpolation.zip)

3. Enter animation mode

Second bug is about default interpolation. It should be CLAMPED-LOCK-LOCK but in this PR it's CLAMPED-UNLOCK-UNLOCK. It should be like so:
HowItShouldBe

Also, pay attention that icon doesn't change when you click on it. We have two different icons for "Lock keyframe" and "Unlock keyframe".

How to reproduce:
1. Remove Synfig settings: ~/synfig/settings-1.4
2. Start Synfig, go to Menu->Edit->Preferences->Document and you will see that default interpolation is wrong

Third bug is state of a new project is changed by default like so:
ChangedByDefault

You see this asterisk before "Synfig Animation 4"? Something triggers the state of the document to be changed even though nothing was changed. How to reproduce:
1. Menu->File->New and you can see the asterisk is right there from the beggining

veermetri05:
@Ayush-5k , could you please add a small summary about your findings, investigations, and point to the code that will help contributors quickly pick up, and continue the work.

PR #3576, opened 2025-05-02
fix: prevent crash due to null devices in DeviceManager from libgtkmm

rodolforg:
Thanks to fcartengie for discovering the issue:
 https://github.com/synfig/synfig/pull/3498


rodolforg:
fcartegnie, could you please test if the problem still happens to you?

ice0 (approved):

PR #3578, opened 2025-05-10
fix(GUI): add missing icon for type transformation

Svarov-RZM:
This PR adds a missing icon for type transformation.
There were several color variants you can check out here: https://forums.synfig.org/t/icon-for-transformation-type/16425
In the end, 3 people voted for colored version, so that's the one I am including here. Hope that's OK.

### DIFF
--- a/synfig-studio/images/CMakeLists.txt
+++ b/synfig-studio/images/CMakeLists.txt
@@ -246,6 +246,7 @@ set(ICONS
 	type_splinepoint_icon
 	type_string_icon
 	type_time_icon
+	type_transformation_icon
 	type_vector_icon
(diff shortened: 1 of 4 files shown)


rodolforg (changes requested):
Thank you, Svarov!
But you forgot to add it to `synfig-studio/images/Makefile.am`

Svarov-RZM:
> But you forgot to add it to `synfig-studio/images/Makefile.am`

Oh, right... I completely forgot that we have two build systems, I only use CMake for my experiments.
Alright, I added support for Autotools.

rodolforg (approved):

PR #3587, opened 2025-05-17
refactor: remove the optional "group" property of layer parameters

rodolforg:
It isn't used - and exists since the first commit.


ice0 (approved):

ice0:
Merged. Thank you!

PR #3589, opened 2025-05-20
fix: new ValueNode_Angle couldn't be loaded because of tag name clash

rodolforg:
ValueNode "Angle" name clashes with Value "Angle" (constant) name.


Stargazer10101:
Does it solve this?  https://github.com/synfig/synfig/pull/3560#issuecomment-2892521258

zaynopenapp (approved):

morevnaproject (approved):

PR #3591, opened 2025-05-21
fix: modifying a bone-linked spline point with split radius / merged angle mutates tangent 2

### LINKED ISSUE #3554: split radius spline
When I create a spline and attach it to a bone, and I enable "Split Radius" is turned on one of the points — why does the spline behave oddly and move incorrectly when I edit it afterward?

However, if "Split Radius" is turned off, the result is correct — as shown in the video below. Thank you.

https://github.com/user-attachments/assets/98cfae6e-9c45-4451-82c0-4e88ee4352e3

mosasauridae:
Fixed a bug where a spline point with Split Radius / Merged Angle that was linked to a skeleton would change Tangent 2 every time the spline point was modified in any way.

Fixes #3554.

### DIFF
--- a/synfig-core/src/synfig/valueoperations.cpp
+++ b/synfig-core/src/synfig/valueoperations.cpp
@@ -79,9 +79,9 @@ ValueTransformation::transform(const Transformation &transformation, const Value
 	if (type == type_bline_point)
 	{
 		BLinePoint bp(value.get(BLinePoint()));
-		bp.set_vertex( transformation.transform(bp.get_vertex()) );
-		bp.set_tangent1( transformation.transform(bp.get_tangent1(), false) );
-		bp.set_tangent2( transformation.transform(bp.get_tangent2(), false) );
+		bp.set_vertex(transformation.transform(bp.get_vertex()) );
+		bp.set_tangent2(transformation.transform(bp.get_tangent2(), false));
+		bp.set_tangent1(transformation.transform(bp.get_tangent1(), false));
 		return bp;
 	}
 	else


mosasauridae:
(Converting my wall-of-text PR description to a comment to clean up the auto-generated release notes)

Steps:
 1. Create a spline and a skeleton
 2. Right click one of the spline points and select Split Radius.
 3. Link the spline point to the skeleton.
 4. Make any change to the spline point (moving the vertex works) - Tangent 2 incorrectly scales. The amount it scales by changes based on the angle.

The problem:
What I think is happening here is when a spline_point is set to Split Radius / Merged Angle is set, transforming tangent1 via set_tangent1() leaves get_tangent2 in an invalid, partially updated state where the angle has been updated but the radius hasn't. The next line uses the invalid get_tangent2 as the base for the update.

This fix avoids the problem by transforming tangent2 first. I think in general it would be best to update tangent2 before tangent1 when applying a transformation that could potentially scale the radius.

I reviewed other usages of set_tangent1/2 to see if this pattern happened anywhere else. It happens in state_lasso.cpp but I don't know where that file gets used. It also happens in ValueAverage::add and ValueAverage::multiple later in this file which are used for the Average converter, though I tested that and it seemed to work.

zaynopenapp (approved):

rodolforg:
> It happens in state_lasso.cpp but I don't know where that file gets used.

@mosasauridae `state_lasso` is the Cut Tool.

rodolforg:
Is it related to this issue? https://github.com/synfig/synfig/issues/3554

rodolforg:
I wonder if creating a new method `set_tangents(const Vector& tangent1, const Vector& tangent2)` would be better, that would set both tangents before calling `update_tangent2()` - which, I suppose, is the real problem.
What do you think, @ice0 , @mosasauridae ?

mosasauridae:
> I wonder if creating a new method `set_tangents(const Vector& tangent1, const Vector& tangent2)` would be better, that would set both tangents before calling `update_tangent2()` - which, I suppose, is the problem root. What do you think, @ice0 , @mosasauridae ?

I was actually thinking something along the same lines, if you added `set_tangents` and removed the ability to set them separately, you'd be able to enforce this ordering to prevent this from happening again. I think you'd want to get rid of the separate `set_tangent1/set_tangent2` functions though, otherwise it'd still be possible to fall for the trap. It's not _that_ much extra effort to call `pt.set_tangents(..., pt.get_tangent2())` if you only wanted to update tangent1.

I'll do a quick refactor to see what it looks like with this change, and if we don't like it I can just nuke the commit.

mosasauridae:
I elected for the less dramatic approach of just adding set_tangents as a new option and refactoring the places where it was safe to do so, and leaving the set_tangent1/2 in place

synfig-core/src/synfig/valueoperations.cpp
```diff
@@ -79,9 +79,9 @@ ValueTransformation::transform(const Transformation &transformation, const Value
 	if (type == type_bline_point)
 	{
 		BLinePoint bp(value.get(BLinePoint()));
-		bp.set_vertex( transformation.transform(bp.get_vertex()) );
-		bp.set_tangent1( transformation.transform(bp.get_tangent1(), false) );
-		bp.set_tangent2( transformation.transform(bp.get_tangent2(), false) );
+		bp.set_vertex(transformation.transform(bp.get_vertex()) );
```
rodolforg, line 82: you forgot to delete the whitespace before the last close-parenthesis `)` ;)

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -515,11 +515,10 @@ Duckmatic::update_ducks()
 							{
 								// Create a new BLinePoint
 								BLinePoint nbp;
-								// Terporary set the flags for the new BLinePoint to all split
+								// Temporarily set the flags for the new BLinePoint to all split
 								nbp.set_split_tangent_both(true);
 								// Now we can set the tangents. Tangent2 won't be modified by tangent1
```
rodolforg, line 520: Part 1: Apparently this trick (temporarily changing the "split tangent" flag was to avoid the problem you solved by creating this new method).

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -515,11 +515,10 @@ Duckmatic::update_ducks()
...
-								nbp.set_tangent1(c1->get_point());
-								nbp.set_tangent2(bp.get_tangent1());
+								nbp.set_tangents(c1->get_point(), c1->get_point());
 								// Now update the flags
 								nbp.set_split_tangent_radius(bp.get_split_tangent_radius());
 								nbp.set_split_tangent_angle(bp.get_split_tangent_angle());
```
rodolforg, line 524: Part 2: restoring the flags here... 
So maybe both temporary-set, restore-set lines should be deleted ;)

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -563,11 +562,10 @@ Duckmatic::update_ducks()
 							{
 								// Create a new BLinePoint
 								BLinePoint nbp;
-								// Terporary set the flags for the new BLinePoint to all split
+								// Temporarily set the flags for the new BLinePoint to all split
 								nbp.set_split_tangent_both(true);
 								// Now we can set the tangents. Tangent2 won't be modified by tangent1
```
rodolforg, line 567: Part 1 again? ;)

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -563,11 +562,10 @@ Duckmatic::update_ducks()
...
-								nbp.set_tangent1(c2->get_point());
-								nbp.set_tangent2(bp.get_tangent1());
+								nbp.set_tangents(c2->get_point(), c2->get_point());
 								// Now update the flags
 								nbp.set_split_tangent_radius(bp.get_split_tangent_radius());
 								nbp.set_split_tangent_angle(bp.get_split_tangent_angle());
```
rodolforg, line 571: Part 2 is back :D

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -710,11 +708,10 @@ Duckmatic::update_ducks()
 										{
 											// Create a new BLinePoint
 											BLinePoint nbp;
-											// Terporary set the flags for the new BLinePoint to all split
+											// Temporarily set the flags for the new BLinePoint to all split
 											nbp.set_split_tangent_both(true);
 											// Now we can set the tangents. Tangent2 won't be modified by tangent1
```
rodolforg, line 713: Part1 once again

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -710,11 +708,10 @@ Duckmatic::update_ducks()
...
-											nbp.set_tangent1(duck->get_point());
-											nbp.set_tangent2(bp.get_tangent1());
+											nbp.set_tangents(duck->get_point(), duck->get_point());
 											// Now update the flags
 											nbp.set_split_tangent_radius(bp.get_split_tangent_radius());
 											nbp.set_split_tangent_angle(bp.get_split_tangent_angle());
```
rodolforg, line 717: And part 2 for the third time

rodolforg (changes requested):
I'm wondering how to check the changes on the cases you replaced ;)

synfig-studio/src/gui/duckmatic.cpp
mosasauridae: This looks like a good change, but I'm having trouble actually testing this case - the parent of the blinepoint is a bline which is not a ValueNode_Composite, so `composite` is null and the condition at line 491 doesn't pass. I played around with a couple converters but couldn't get anything to hit this line of code. Any idea how to test this?

synfig-core/src/synfig/valueoperations.cpp
mosasauridae: fixed

synfig-studio/src/gui/duckmatic.cpp
mosasauridae: I updated these sections to remove the temporary. Never did figure out how to run these sections, but it's a pretty small change

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -513,18 +513,8 @@ Duckmatic::update_ducks()
...
-								// Now we can set the tangents. Tangent2 won't be modified by tangent1
-								nbp.set_tangent1(c1->get_point());
-								nbp.set_tangent2(bp.get_tangent1());
-								// Now update the flags
-								nbp.set_split_tangent_radius(bp.get_split_tangent_radius());
-								nbp.set_split_tangent_angle(bp.get_split_tangent_angle());
-								// Now retrieve the updated tangent2 (which will be stored as t1, see below)
-								Vector t1(nbp.get_tangent2());
+								bp.set_tangents(c1->get_point(), c1->get_point());
+								Vector t1(bp.get_tangent2());
```
rodolforg: Is this really correct?
As far as I understood, the original code make the `tangent2` of `bp` as its original `tangent1`

mosasauridae: So I'm still not sure how to test this section, but I was at least able to test it indirectly by copying the code into a test function I could call manually from main.cpp to make sure that the new code gets the same value for `t1` that it did before.

You were right, I'm not sure where I came up with "both parameters are the same" but it's fixed now. Each of these sections now give the same values for `t1` that they used to before the refactor.

There was a particularly nasty aliasing bug when I originally had it do `bp.set_tangents(..., bp.get_tangent1())` - everything here is references, so the `const Vector& t2` parameter to set_tangents was pointing to `tangent_[0]`, which screwed up the result. I changed set_tangents to take its input by value instead of by reference to prevent that from becoming an issue. Not ideal to have to do a copy, but thankfully Vector's copy is cheap.

rodolforg: > So I'm still not sure how to test this section

Neither do I. I thought I answered this, sorry.
This code is supposed to be when you drag the Bézier curve, but it tests if the parent valuenode is `composite`, but it is not. I don't know if there is actually a way to reach this code... :/

rodolforg: > There was a particularly nasty aliasing bug when I originally had it do `bp.set_tangents(..., bp.get_tangent1())` - everything here is references, so the `const Vector& t2` parameter to set_tangents was pointing to `tangent_[0]`, which screwed up the result.

So it would be good to create a test file with this situation to avoid changing the signature of this method by someone (else?) in the future - at least without knowledge.

mosasauridae: Do we have any unit tests? I could toss something in there.

Also the OSX build keeps failing and I don't know why, doesn't seem to be anything related to these changes though.

rodolforg: Yes, we have a basic unit test 'framework' (header file :P) . It is in `synfig-core/test` and `synfig-studio/test` folders.
They work for CMake and Autotools.

mosasauridae: Finally got back to this. I added some tests for blinepoint tangent handling which are all passing. If the bug is reintroduced then these new tests will fail.

Also I added ENABLE_TESTS as a cache variable so that it shows up in IDEs, defaulted to false to match previous behavior.

mosasauridae:
> I'm wondering how to check the changes on the cases you replaced ;)

I was able to test most of the changes, but not all. I used debugging break points to confirm I hit the line of code.

- duckmatic.cpp:1116 - Modify t2 of split angle, merged radius spline point which has been linked to a bone
- duckmatic.cpp:1120 - Modify t2 of split radius, merged angle spline point which has been linked to a bone
- state_bline.cpp:1554 - Split angle (merge radius) while creating spline, modify t2
- state_bline.cpp:1557 - Split radius (merge angle) while creating spline, modify t2
- state_bline.cpp:750 - Use Spline tool
- state_draw.cpp:1220 - Use Draw tool
- state_lasso.cpp:1136 - Use Cut tool
- valuenode_bline.cpp:541 - Extend Draw tool before finishing spline
- valuenode_bline.cpp:601/717 - Animate a spline point then use "Mark active point as On/Off" on two different frames and go to a frame in between
- valuenode_bline.cpp:992 - Set "SYNFIG_COMPLEX_TANGENT_BONE_INFLUENCE" environment variable, then use Bone Influence converter on a spline point
- valuenode_composite.cpp:265 - Use Draw tool
- valueoperations.cpp:131/180 - Create spline with 3 points, export 2 spline points, then use the Average converter on the 3rd
- valueoperations.cpp:83 - Link bline points to skeleton

These lines I couldn't hit
- duckmatic.cpp:516/554/691 - Not sure how to hit these lines, they require a splint point whose parent is a ValueNode_Composite
- svg_parser.cpp:780 - Couldn't find an svg file that hits this line.

mosasauridae:
I believe I've addressed the changes, let me know if there's anything else I should do here

CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ endif(MSVC)
 
 message("-- CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
 
+set(ENABLE_TESTS OFF CACHE BOOL "Enable testing")
```
rodolforg: What is this? XD

synfig-core/test/bline.cpp
```diff
@@ -233,6 +233,135 @@ void test_calc_vertex() {
 	ASSERT_VECTOR_APPROX_EQUAL_MICRO(Vector(-0.260716, 0.341317), vertex)
 }
 
+#define ASSERT_VECTOR_APPROX_EQUAL(expected, value) {\
+	if (!synfig::approximate_equal_lp(expected[0], value[0]) || \
+		!synfig::approximate_equal_lp(expected[1], value[1])) { \
+		ERROR_MESSAGE_TWO_VALUES(expected, value) \
+	} \
+}
+
```
rodolforg: Maybe move it to `test_base.h`? In both `synfig-core/` and `synfig-studio/`

CMakeLists.txt
mosasauridae: It basically "declares" the variable so that it shows up in IDEs, which lets you toggle it from the IDE without having to manually run cmake. It's really only relevant if you're building directly from an IDE, if you're building from the scripts it doesn't have any effect

It's entirely optional and I can take it out, I figured others might appreciate the minor convenience lol

synfig-core/test/bline.cpp
mosasauridae: Sure I can toss it in there

mosasauridae: all set

CMakeLists.txt
mosasauridae: This pattern also shows up on line 34 in the top level CMakeLists as well as a few times in the gui, so there's some solid precedent for it. Unless there's any concerns I'll leave it in.

mosasauridae:
Rebased branch

Github says there are requested changes, but I'm not familiar enough with the interface to find out what the requested changes actually are.

mosasauridae:
> Rebased branch
> 
> Github says there are requested changes, but I'm not familiar enough with the interface to find out what the requested changes actually are.

@rodolforg I'm officially lost, GitHub is saying that there's 1 unresolved comment from you on this PR, but I can't find it anywhere. I guess it was on a commit, but apparently those comments become inaccessible after rebasing. The red "Show requested changes" button just brings me to an error page since the commit it's trying to get to is inaccessible. I was able to recover the old commits from my local repo and tried force pushing them to this PR. `9f2d92b` at least brought me to a real diff page, but all the comments from that point were addressed.

CMakeLists.txt
rodolforg: It is okay. Leave as it is.

rodolforg (dismissed):

rodolforg:
@mosasauridae are you still around? XD
Could you please split this PR in two?
For a cleaner git history/log ;)

rodolforg:
Rebased and split in to different PRs for a clean git history.

rodolforg (approved):

mosasauridae:
> @mosasauridae are you still around? XD Could you please split this PR in two? For a cleaner git history/log ;)

Sorry yeah I'm still around. Looks like you took care splitting the PR, let me know if you need any help from my end

PR #3594, opened 2025-05-31
fix(bline tool): ignores double-clicking if without any created point

### LINKED ISSUE #3388: Path tool bugs out after double clicking at an empty space in the canvas
**Describe the bug**
Path tool bugs out after double clicking at an empty space in the canvas and starts acting like the select tool.
**To Reproduce**
Just double click on the canvas after selecting the path tool

https://github.com/user-attachments/assets/465b314b-89e5-418d-ad2a-6b132116464c

**System information:**
 - OS: Windows 10 22h2
 - OS version: 64 bit
 - Synfig version: 1.5.1

rodolforg:
On a BLine Tool, double-clicking is used to realize the creation of a new spline.

When no spline being drawn, double-clicking incorrectly makes impossible to draw a new spline - user then would have to press Esc key to be able to draw one more spline (Esc is a shortcut used to cancel an ongoing spline).

Here we make the tool state ignore the double-click in such case, preventing this awkward situation.

Fix #3388

### DIFF
--- a/synfig-studio/src/gui/states/state_bline.cpp
+++ b/synfig-studio/src/gui/states/state_bline.cpp
@@ -728,6 +728,7 @@ StateBLine_Context::run_()
 	}
 	if(bline_point_list.size()<2)
 	{
+		reset();
 		get_canvas_view()->get_ui_interface()->task(_("Information: You need at least two (2) points to create a spline"));
 		return false;
 	}


rodolforg:
@ice0 rebased as requested

ice0 (approved):

PR #3595, opened 2025-05-31
fix(bline tool): couldn't create the spline right after user drags the first vertex

rodolforg:
Steps to reproduce:

1. Select the BLine Tool
2. Click on the Workarea to create the first vertex
3. Click on the Workarea twice to create two other vertices
4. Click on the first vertex (the green handle) and drag it
5. Click on the first vertex again

After step 5, a new (looped) spline is created.
However, user couldn't create a new spline after it without this fix.

Workaround:
* Press Esc after step 5


ice0:
Merged. Thank you!

PR #3597, opened 2025-06-01
feat: add new Map Range valuenode

rodolforg:
Node inspired in the Blender homonym node.

Requested-by: microvswind (https://forums.synfig.org/t/how-to-fit-a-list-range-to-another/12583)


zaynopenapp:
Here is an example of a map range converter.
using timeloop convertion this is similar to finding 'valueAtTime'.

I hope the add and subtract converters can be used in the Spline. So that they can be used for multiple controllers (bones)

[rangemap.zip](https://github.com/user-attachments/files/20589369/rangemap.zip)
you must commit this to use this converter

https://github.com/user-attachments/assets/12615cd9-03fb-474f-a0ee-feb7848bfcae

Issue #3590, opened 2025-05-20
Bug: Build fail with Autotools on fresh clone / Ubuntu 22.04.5


BobSynfig:
**Describe the bug**
Building with Autotools fail of a fresh copy of master branch, both with `2-build-debug.sh` and `2-build-production.sh`
`2-build-cmake.sh` works fine

**To Reproduce**
1. `git clone` ...
2. `2-build-debug.sh` /  `2-build-production.sh`

**System information:**
 - OS: Ubuntu 22.04.5 LTS
 - Commit ID: master
 - `1-setup-linux-native.sh` dependencies installed

**Additional comments**
End of the output:
```
In file included from /mnt/ramdisk/synfig/synfig-core/test/test_base.h:30,
                 from /mnt/ramdisk/synfig/synfig-core/test/surface_etl.cpp:28:
/usr/include/c++/11/sstream: In instantiation of ‘std::__cxx11::basic_ostringstream<_CharT, _Traits, _Alloc>::__string_type std::__cxx11::basic_ostringstream<_CharT, _Traits, _Alloc>::str() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_ostringstream<_CharT, _Traits, _Alloc>::__string_type = std::__cxx11::basic_string<char>]’:
/mnt/ramdisk/synfig/synfig-core/test/surface_etl.cpp:42:2:   required from here
/usr/include/c++/11/sstream:917:7: error: return type ‘std::__cxx11::basic_ostringstream<char>::__string_type’ {aka ‘class std::__cxx11::basic_string<char>’} is incomplete
  917 |       str() const _GLIBCXX_LVAL_REF_QUAL
      |       ^~~
/usr/include/c++/11/sstream:918:32: error: invalid use of incomplete type ‘std::__cxx11::basic_stringbuf<char>::__string_type’ {aka ‘class std::__cxx11::basic_string<char>’}
  918 |       { return _M_stringbuf.str(); }
      |                ~~~~~~~~~~~~~~~~^~
In file included from /usr/include/c++/11/iosfwd:39,
                 from /usr/include/c++/11/ios:38,
                 from /usr/include/c++/11/ostream:38,
                 from /usr/include/c++/11/iostream:39,
                 from /mnt/ramdisk/synfig/synfig-core/test/test_base.h:29,
                 from /mnt/ramdisk/synfig/synfig-core/test/surface_etl.cpp:28:
[...]

PR #3601, opened 2025-06-13
 Issue-#3590 Build fail with Autotools on fresh clone 

nilesh646:


synfig-core/test/test_base.h
```diff
@@ -28,6 +28,10 @@
 #include <cstdlib> // std::abs
 #include <iostream> // std::cerr
 #include <sstream>
+#include <string>    // For std::string
+#include <locale>    // For std::localea
```
ice0, line 32: I don't see usage of `std::locale` here

PR #3611, opened 2025-06-28
fix(text tool): some fonts render corrupted text

### LINKED ISSUE #3387: Not Support Devanagari script properly.

### LINKED ISSUE #3411: Error with the text tool when using some fonts.
**Describe the bug**
When I change to some different fonts it changes what has been written in the text parameter of the text layer.

**To Reproduce**
1. Create text layer.
2. Write something in the text parameter
3. Change the font in the Font Family parameter.
4. What had been written initially is changed by different letters.

**Expected behavior**
What is written in the text parameter should be kept intact.

**Screenshots**
https://youtu.be/kn4WRIuWTzg

**System information:**
 - OS: Archlinux
 - OS version: 64x with Kernel lTS
 - Commit ID: efeeda8c0925ef8e075a454264ae177cb2539bb7

**Additional comments**
Thank you very much for your time and attention. I hope it is useful.

mosasauridae:
Addressed an issue where Text layers would render a corrupted string when using certain fonts (including Calibri and Consolas, among several others).

The corruption happens after the first instance of the font. The error can also be triggered by creating two Text layers with the same font, or by using the File>Revert menu item.

Fixes #3411
Fixes #3387

### DIFF
--- a/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
+++ b/synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
@@ -696,12 +696,11 @@ static std::string fontconfig_get_filename(const std::string& font_fam, int styl
 bool
 Layer_Freetype::new_face(const String &newfont)
 {
-	synfig::String font=param_font.get(synfig::String());
 	int error = 0;
 	FT_Long face_index=0;
 
 	// If we are already loaded, don't bother reloading.
-	if(face && font==newfont)
+	if (face && param_font.get(synfig::String()) == newfont)
 		return true;
 
 	if(face)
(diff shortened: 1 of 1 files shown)


mosasauridae:
Turned out to be a caching error. When the font is first referenced, it generates both a `face` and a `font` and caches them off, but when loading from the cache it only read in the `face`. I guess this impacted some fonts but not others.

While I was in there I took the opportunity to clean up a mostly-unused local `font` variable which seemed to be causing some confusion in this function.

Stargazer10101:
Thanks! This has addressed issue #3387 as well.

synfig-core/src/modules/lyr_freetype/lyr_freetype.cpp
```diff
@@ -696,12 +696,11 @@ static std::string fontconfig_get_filename(const std::string& font_fam, int styl
 bool
 Layer_Freetype::new_face(const String &newfont)
 {
-	synfig::String font=param_font.get(synfig::String());
 	int error = 0;
 	FT_Long face_index=0;
 
 	// If we are already loaded, don't bother reloading.
-	if(face && font==newfont)
+	if(face && param_font.get(synfig::String())==newfont)
```
rodolforg: As you are already changing this line, please 'fix' its code style by adding a single space before the `(` and surrounding the `==` operator.

mosasauridae: Fixed

rodolforg (approved):

ice0:
Merged. Thank you!

Issue #3580, opened 2025-05-11
New convert type: Absolute value


morevnaproject:
We need a new convert type, which returns absolute value for Real and Angle - https://en.wikipedia.org/wiki/Absolute_value

Svarov-RZM:
Yeah, it would be useful to have.
I remember needing this and had to make a workaround using "vector length" converter, it was pretty awkward.

Also, let's not forget about Integer type.

BobSynfig:
Linux64 AppImage for test available here
https://github.com/BobSynfig/synfig/releases/tag/SynfigStudio-UNOFFICIAL-1.5.3-2025.07.23_Converter_Absolute

PR #3607, opened 2025-06-20
feat: Add support for Angle type to ValueNode_Logarithm (Log converter)

nilesh646:
This PR adds support for the `Angle` type in the Logarithm converter (`ValueNode_Logarithm`). Previously, the converter only accepted `Real` values. With this update, it can now also accept `Angle` inputs, which are internally converted to radians (as `Real`) before applying the logarithmic operation.

### Changes Made

- Modified `operator()` to handle `ValueBase::TYPE_ANGLE` safely.
- Updated `check_type()` to accept both `type_real` and `type_angle`.
- Improved the class documentation in `valuenode_log.h` to reflect the new support for `Angle`.

PR #3618, opened 2025-07-23
Absolute Converter for angle/integer/real

BobSynfig:
Requested per issue https://github.com/synfig/synfig/issues/3580

Linux64 AppImage for test available here
https://github.com/BobSynfig/synfig/releases/tag/SynfigStudio-UNOFFICIAL-1.5.3-2025.07.23_Converter_Absolute

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
@@ -1,0 +1,228 @@
+/* === S Y N F I G ========================================================= */
+/*!	\file valuenode_absolute.cpp
+**	\brief Implementation of the "Absolute" valuenode conversion.
+**
+**	\legal
+**	Copyright (c) 2002-2005 Robert B. Quattlebaum Jr., Adrian Bentley
+**	Copyright (c) 2007, 2008 Chris Moore
+**  Copyright (c) 2011 Carlos López
+**  Copyright (c) 2025 BobSynfig
+**
+**	This file is part of Synfig.
+**
+**	Synfig is free software: you can redistribute it and/or modify
+**	it under the terms of the GNU General Public License as published by
+**	the Free Software Foundation, either version 2 of the License, or
+**	(at your option) any later version.
+**
+**	Synfig is distributed in the hope that it will be useful,
+**	but WITHOUT ANY WARRANTY; without even the implied warranty of
+**	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+**	GNU General Public License for more details.
+**
+**	You should have received a copy of the GNU General Public License
+**	along with Synfig.  If not, see <https://www.gnu.org/licenses/>.
+**	\endlegal
+*/
+/* ========================================================================= */
+
+/* === H E A D E R S ======================================================= */
+
+#ifdef USING_PCH
+#	include "pch.h"
+#else
+#ifdef HAVE_CONFIG_H
+#	include <config.h>
+#endif
+
+#include <synfig/general.h>
+#include <synfig/localization.h>
+#include <synfig/valuenode_registry.h>
+#include "valuenode_absolute.h"
+#include "valuenode_const.h"
+#include <stdexcept>
+#include <synfig/misc.h>
+#include <synfig/angle.h>
+#include <synfig/real.h>
+
+#endif
+
+/* === U S I N G =========================================================== */
+
+using namespace synfig;
+
+/* === M A C R O S ========================================================= */
+
+/* === G L O B A L S ======================================================= */
+
+REGISTER_VALUENODE(ValueNode_Absolute, RELEASE_VERSION_1_6_0, "absolute", N_("Absolute"))
+
+/* === P R O C E D U R E S ================================================= */
+
+/* === M E T H O D S ======================================================= */
+
+ValueNode_Absolute::ValueNode_Absolute(const ValueBase &value):
+	LinkableValueNode( value.get_type() )
+{
+	init_children_vocab();
+
+	Type &type( value.get_type() );
+		
+	if (type == type_angle)
+		set_link("link", ValueNode_Const::create( value.get( Angle() ) ) );
+	else
+	if (type == type_integer)
+		set_link("link", ValueNode_Const::create( value.get( int()   ) ) );
+	else
+	if (type == type_real)
+		set_link("link", ValueNode_Const::create( value.get( Real()  ) ) );
+	else
+	{
+		assert(0);
+		throw std::runtime_error( get_local_name() + _(":Bad type ") + type.description.local_name);
+	}
+
+	assert(value_node);
+	assert(value_node->get_type() == type);
+	assert(get_type() == type);
+}
+
+LinkableValueNode*
+ValueNode_Absolute::create_new() const
+{
+	return new ValueNode_Absolute( get_type() );
+}
+
+ValueNode_Absolute*
+ValueNode_Absolute::create(const ValueBase& value, etl::loose_handle<Canvas>)
+{
+	return new ValueNode_Absolute( value );
+}
+
+synfig::ValueNode_Absolute::~ValueNode_Absolute()
+{
+	unlink_all();
+}
+
+synfig::ValueBase
+synfig::ValueNode_Absolute::operator()(Time t)const
+{
+	DEBUG_LOG("SYNFIG_DEBUG_VALUENODE_OPERATORS",
+		"%s:%d operator()\n", __FILE__, __LINE__);
+
+	if (!value_node)
+		throw std::runtime_error( 
+				strprintf( "ValueNode_Absolute: %s",
+				           _("My parameter isn't set!" )
+				)
+			);
+	else
+	if ( get_type() == type_angle  )
+		return 
+			Angle::deg(
+				std::abs(
+					Angle::deg(
+						(*value_node)(t).get( Angle() )
+					).get()
+				)
+			);
+	else
+	if ( get_type() == type_integer)
+	    return std::abs( (*value_node)(t).get( int()   ) );
+	else
+	if ( get_type() == type_real   )
+		return std::abs( (*value_node)(t).get( Real()  ) );
+
+	assert(0);
+	return ValueBase();
+}
+
+synfig::ValueBase
+synfig::ValueNode_Absolute::get_inverse(const Time& t, const synfig::ValueBase &target_value) const
+{
+	const Type& target_type = target_value.get_type();
+	
+	if (target_type == type_angle)
+		return target_value.get( Angle() );
+	
+    if (target_type == type_integer)
+		return target_value.get( int()   );
+	
+	if (target_type == type_real)
+		return target_value.get( Real()  );
+	
+	throw std::runtime_error(
+		strprintf( "ValueNode_%s: %s: %s",
+				   get_name().c_str(),
+				   _("Attempting to get the inverse of a non invertible Valuenode"),
+				   _("Invalid value type")
+	    )
+	);
+}
+
+LinkableValueNode::InvertibleStatus
+synfig::ValueNode_Absolute::is_invertible(const Time& t, const ValueBase& target_value, int* link_index) const
+{
+	if ( !t.is_valid() )
+		return INVERSE_ERROR_BAD_TIME;
+
+	const Type& type = target_value.get_type();
+	
+	if ( type != type_angle   &&
+		 type != type_real    &&
+	     type != type_integer  )
+		return INVERSE_ERROR_BAD_TYPE;
+
+	if (link_index)
+		*link_index = get_link_index_from_name("link");
+
+	return INVERSE_OK;
+}
+
+bool
+ValueNode_Absolute::set_link_vfunc(int i, ValueNode::Handle value)
+{
+	assert( i>=0 && i<link_count() );
+
+	switch(i)
+	{
+		case 0: CHECK_TYPE_AND_SET_VALUE(value_node, get_type() );
+	}
+	return false;
+}
+
+ValueNode::LooseHandle
+ValueNode_Absolute::get_link_vfunc(int i) const
+{
+	assert( i>=0 && i<link_count() );
+
+	if (i == 0) return value_node;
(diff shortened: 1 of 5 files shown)


BobSynfig:
CommitLint was considering the PR cheks as failed because of a missing feat(something): in front ...
Not because the code was bad ...
Other PR suffer the same issue, like https://github.com/synfig/synfig/pull/3607 for example
Do we really need this ~~sh!t~~ feature?

Also Gemini advices could be great if they were applied also to the rest of the source code.
Obviously it doesn't recognize when some code is directly inspired from existing ones and the "intention" that lies behind.
Too much optimization for a specific purpose could lead to other deeper bugs in the future.

mohamedAdhamc:
> CommitLint was considering the PR cheks as failed because of a missing feat(something): in front ... Not because the code was bad ... Other PR suffer the same issue, like #3607 for example Do we really need this ~sh!t~ feature?
> 
> Also Gemini advices could be great if they were applied also to the rest of the source code. Obviously it doesn't recognize when some code is directly inspired from existing ones and the "intention" that lies behind. Too much optimization for a specific purpose could lead to other deeper bugs in the future.

About the gemini part, yes. I made a pr and I felt like I was instantly attacked by Gemini xd.

Svarov-RZM:
@BobSynfig Thanks for implementing this! I remember needing this converter and had to workaround stuff by using "vectorlength" which was quite awkward.  I'll test it during next weekends, as usual, very busy during work weeks.

Svarov-RZM:
Alright, I tested it and all works well. Two things I would point out from the user perspective:

* The 'Absolute' node is invertible but GUI doesn't see it as such. I think there must be some hook in the GUI logic for it to recognize it. I am not currently sure whether this converter should support inverse manipulation or not. If there will be a requirement for it we will add it in the future
*  Perhaps this converter should also support 'vector' type, but again, I don't have a use-case for it now. We can add it later if needed

@rodolforg @ice0 @Abdelhadi-Wael 
Guys, if you don't have any issues with the code, I will squash and merge it closer to the middle of the next week

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+#	include "pch.h"
+#else
+#ifdef HAVE_CONFIG_H
+#	include <config.h>
+#endif
+
+#include <synfig/general.h>
+#include <synfig/localization.h>
+#include <synfig/valuenode_registry.h>
+#include "valuenode_absolute.h"
```
rodolforg: please list this header first :)

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+	return new ValueNode_Absolute( value );
+}
+
+synfig::ValueNode_Absolute::~ValueNode_Absolute()
+{
+	unlink_all();
+}
+
+synfig::ValueBase
+synfig::ValueNode_Absolute::operator()(Time t)const
```
rodolforg: please add space before `const`

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+/* === G L O B A L S ======================================================= */
+
+REGISTER_VALUENODE(ValueNode_Absolute, RELEASE_VERSION_1_6_0, "absolute", N_("Absolute"))
+
+/* === P R O C E D U R E S ================================================= */
+
+/* === M E T H O D S ======================================================= */
+
+ValueNode_Absolute::ValueNode_Absolute(const ValueBase &value):
```
rodolforg: please swap the space surrounding `&`:
`const ValueBase& value`

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+	LinkableValueNode( value.get_type() )
+{
+	init_children_vocab();
+
+	Type &type( value.get_type() );
```
rodolforg: Same here for `Type& `

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+	else
+	if ( get_type() == type_real   )
+		return std::abs( (*value_node)(t).get( Real()  ) );
+
+	assert(0);
+	return ValueBase();
+}
+
+synfig::ValueBase
+synfig::ValueNode_Absolute::get_inverse(const Time& t, const synfig::ValueBase &target_value) const
```
rodolforg: `const synfig::ValueBase& target_value` :)

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+	const Type& target_type = target_value.get_type();
+	
+	if (target_type == type_angle)
+		return target_value.get( Angle() );
+	
+    if (target_type == type_integer)
+		return target_value.get( int()   );
+	
+	if (target_type == type_real)
+		return target_value.get( Real()  );
```
rodolforg, line 153: I know (all?) other valuenodes use this code style, but what about:

```c++
if (check_type(type))
    return target_value;
```
?

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+		"%s:%d operator()\n", __FILE__, __LINE__);
+
+	if (!value_node)
+		throw std::runtime_error( 
+				strprintf( "ValueNode_Absolute: %s",
+				           _("My parameter isn't set!" )
+				)
+			);
+	else
+	if ( get_type() == type_angle  )
```
rodolforg, line 121: Why the space after open parenthesis `(`?

Although I don't like it, I understand the spaces before the closing parenthesis `)` is for alignment with the following `if`. But I don't get why that after `(`.

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+synfig::ValueNode_Absolute::is_invertible(const Time& t, const ValueBase& target_value, int* link_index) const
+{
+	if ( !t.is_valid() )
+		return INVERSE_ERROR_BAD_TIME;
+
+	const Type& type = target_value.get_type();
+	
+	if ( type != type_angle   &&
+		 type != type_real    &&
+	     type != type_integer  )
```
rodolforg, line 174: Again, I wonder if it would be better to call `check_type()` instead.

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+bool
+ValueNode_Absolute::set_link_vfunc(int i, ValueNode::Handle value)
+{
+	assert( i>=0 && i<link_count() );
+
+	switch(i)
+	{
+		case 0: CHECK_TYPE_AND_SET_VALUE(value_node, get_type() );
```
rodolforg: please remove the space before `)`

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+ValueNode::LooseHandle
+ValueNode_Absolute::get_link_vfunc(int i) const
+{
+	assert( i>=0 && i<link_count() );
+
+	if (i == 0) return value_node;
+	
+	return 0;
```
rodolforg: prefer nullptr

synfig-core/src/synfig/valuenodes/valuenode_absolute.h
```diff
@@ -0,0 +1,101 @@
+/* === S Y N F I G ========================================================= */
+/*!	\file valuenode_absolute.h
+**	\brief Header file for implementation of the "Absolute" valuenode conversion.
+**
+**	\legal
+**	Copyright (c) 2002-2005 Robert B. Quattlebaum Jr., Adrian Bentley
+**	Copyright (c) 2007, 2008 Chris Moore
+**      Copyright (c) 2011 Carlos López
+**      Copyright (c) 2025 BobSynfig
```
rodolforg: alignment

synfig-core/src/synfig/valuenodes/valuenode_absolute.h
```diff
@@ -0,0 +1,101 @@
...
+            create_new() const override;
+
+	virtual bool 
+            set_link_vfunc(int i, ValueNode::Handle x) override;
+	
+        virtual ValueNode::LooseHandle
+            get_link_vfunc(int i) const override;
+
+	virtual Vocab
+            get_children_vocab_vfunc() const override;
```
rodolforg: I learned with @ice0 that when we use `override` we don't need `virtual` :)

I think I'll do another massive commit with all valuenodes to fix code style :P

rodolforg (changes requested):
As with Modulo converter, this one has multiple (two) possible inversions.
I wonder if we should implement a `get_inverse()` that would allow to return multiple values and let GUI choose what is the best (eg. closer to mouse pointer).

rodolforg:
> * The 'Absolute' node is invertible but GUI doesn't see it as such. I think there must be some hook in the GUI logic for it to recognize it. I am not currently sure whether this converter should support inverse manipulation or not. If there will be a requirement for it we will add it in the future

We could merge as it is. There is another place/file to change to allow the inversion, ~~but I think I'll remove this requirement (it exists before the implementation of `get_inverse()` and related methods)~~ (forget it, too much work lol).

@BobSynfig , to make it editable via mouse dragging, please add this valuenode type here:
`synfigapp::is_editable(synfig::ValueNode::Handle value_node)` in `synfigapp/instance.cpp`.

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
BobSynfig: Don't worry, I started to work on refactoring, it will be done soon, starting from `app.cpp`.
Someone has to do the job, so next time we will add some feature by copy/paste the skeleton of existing one, you will not have to complain.
BTW, do you prefer
```
try {
} catch {
}

```
or
```
try
{
}
catch
{
}
```

synfig-core/src/synfig/valuenodes/valuenode_absolute.h
BobSynfig: Yep, virtual is only needed in the base class for virtual methods.
In derived classes, it is implicit as a virtual method stays virtual, and redeclaring it was "the way" before c++11.
Override is also optional (as it was introduced in c++11, but virtual were working the same before) but to make the compiler more strict in checking the signature (otherwise it could accidentally overload it with a different signature).
This is the problem with unmaintained legacy code, what looks strange now was the norm in the past.
(I'm 49 soon, it doesn't shock me ^^).

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
BobSynfig: Here it is a matter of "air",
Either there is no space between the parenthesis, or if there are at the end, I prefer to have also in the start for visual balance.
We could also write all the code one one line without spaces, the compiler doesn't care.
But the source is meant to be read and understood by humans.
With some structure makes the logic more obvious (and it is even better with code folding).

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
BobSynfig: To be done during a global refactoring of the different valuenode_*.
If you edit one file here, another there, there will be always a moment where someone (like me) will use an existing one as skeleton.
And there is always a high probability that it will be a wrong one!

BobSynfig:
`synfigapp/instance.cpp` is one of those very ugly files with 6-7 levels of nesting and pre-c++11 crappy code.

We should also document common tasks as recipes in dev docs, what and where to add code!

Svarov-RZM:
> to make it editable via mouse dragging, please add this valuenode type here:
> `synfigapp::is_editable(synfig::ValueNode::Handle value_node)` in `synfigapp/instance.cpp`.

Oof, this seems like unnecessary logic... The correct one would be just to check is_invertible method of a valuenode that user is about to manipulate and then depending on return value is either proceed or ignore. No need for synfigapp::is_editable. But just like rodolforg's said it requires some overhaul and there's already a lot of refactoring efforts that are just hanging in PRs...

@BobSynfig I've tested the recent changes, the manipulation of 'Absolute' valuenode works now through GUI handles so I am merging it. Again, thank you for contribution!

P.S. I will most likely take a look at all those refactoring PRs now, like the one with removal of ETL folder.

Svarov-RZM (approved):

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
```diff
@@ -0,0 +1,228 @@
...
+#	include <config.h>
+#endif
+
+#include <synfig/general.h>
+#include <synfig/localization.h>
+#include <synfig/valuenode_registry.h>
+#include "valuenode_absolute.h"
+#include "valuenode_const.h"
+#include <stdexcept>
+#include <synfig/misc.h>
```
rodolforg, line 44: is this `#include`  misc.h necessary/used?

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
rodolforg: But the line 130 doesn't has spaces before `)` :)

synfig-core/src/synfig/valuenodes/valuenode_absolute.cpp
BobSynfig: It was inside the file I used as base, can't tell

PR #3619, opened 2025-07-25
fix(synfigapp): DynamicListInsertSmart didn't keep the width for non-homogeneous Adv. Outline

### LINKED ISSUE #3606: Feature Request: Advanced Outlines' Shapes (including their Width Points) Should Stay Consistent When Adding a New Vertex Points; Even if the 'Homogenous' Option is NOT Checked
**Important Remark:**
This is a follow-up to the footnote of an issue ticket I submitted earlier (#3427). After several discussions regarding its Pull Request (#3562), I decided to officially make this ticket in favor of the mentioned issue and the pull request associated with it, meaning that the fix/feature implementation associated with **this** ticket will close **both** issue tickets, and the pull request mentioned earlier may be marked as obsolete; if the devs wish to.

While the implementation of this feature has been discussed earlier on PR #3562, I will put some of them here once again, for the sake of clarity.

**The Issue:**
Adding vertex points into an advanced outline (while the 'Homogenous' option is off) can nudge the positions of all width points within it, thus they'd need to be corrected again. This may be not a big deal when working with few lines, but on a bigger scale, with many advanced outlines, and even more width points to manage, this can turn into a really tiring chore.
One of the cases where this become an apparent problem, is when someone has to reuse the same advanced outline over and over, to avoid the hassle of drawing new, nearly the exact same copies of it. See here where I explained my own test case, with my workflow as the example: https://github.com/synfig/synfig/pull/3562#issuecomment-2971889179

**The /not-so/ Workarounds:**
While this can be resolved by enabling 'Homogenous' option, it introduces another problem. Since 'Homogenous' advanced outlines measure their WPs' positions by the total length of the line (as opposed to the number of vertex points present), stretching or shrinking the line can easily screw up the width points' structure, Not exactly ideal when these lines have to be stretched or shrunk many-many times whilst wanting to keep their structures quite clean and consistent. See here for more (bare-bone) explanation of it: https://github.com/synfig/synfig/pull/3562#issuecomment-2975915817

Another workaround that some people may have thought, is to just use normal outlines. I have answered that question here: https://github.com/synfig/synfig/pull/3562#issuecomment-2973757575
The bottom line is that I want to be able to work with advanced outlines more consistently, while keeping the amount of vertex points I have to manage as low as possible, and still being able to access the usual advanced outlines features; cusps, freely movable WPs, and maybe others.

**The Feature Request:**
To clarify, the feature request is:
**To be able to keep the advanced outlines' shape; including all of their width points, consistent and remain the same, when adding new vertex points into them, even if their 'Homogenous' option is UN-checked.**

I have some theories how this might be implemented into the Synfig into the first place, and it is all mentioned here: https://github.com/synfig/synfig/pull/3562#issuecomment-2975915817. But this is just an assumption from a non-programmer that doesn't know the inner workings of Synfig that much. There might be a neater solution to this issue.

_**Additional comments**
Add any other comments or screenshots about the feature request here._

Before adding new vertex points (I will be using the shape on the right for this demonstration):

Image

Image

After adding new vertex points (and some line adjustments):

Image

Image

**_If the 'Homogenous' option was checked before adjusting this advanced outline, this issue would still persist. And it probably could have gone much worse.... .-.;_**

rodolforg:
This problem happens only for non-homogeneous advanced outline layers, i.e. the 'standard' coordinates (how it is called in internal code).

It works only for non-animated mode here, because adding/removing bline points ("on"/"off" status of these points) is complicated enough when animated...

fix #3606

### DIFF
--- a/synfig-studio/src/synfigapp/actions/valuenodedynamiclistinsertsmart.cpp
+++ b/synfig-studio/src/synfigapp/actions/valuenodedynamiclistinsertsmart.cpp
@@ -36,8 +36,10 @@
 #include <synfig/general.h>
 
 #include "valuenodedynamiclistinsertsmart.h"
-#include <synfigapp/canvasinterface.h>
 
+#include <synfig/valuenodes/valuenode_wplist.h>
+
+#include <synfigapp/canvasinterface.h>
 #include <synfigapp/localization.h>
 
 #endif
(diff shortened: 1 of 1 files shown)


rodolforg:
@FadhilKwan , does this PR fix your issue?

FadhilKwan:
Heck yeah! I mean, it would, for the time being :D
Though, I did try some artifacts that has this commit (at least I assume they have it ._.), but it doesn't seem to work as intended just yet...
note: right now I used this artifact from AppVeyor to test it: SynfigStudio-1.5.2-testing-2025-08-13-win64-8c609

I've put in some screenshots, before and after adding new a vertex point. I also have given the designated width points some contrasting widths, so the differences would be much more visible (and yes, I've unchecked 'Homogenous' as per usual~)

Before Adding a new vertex point. I've added 4 width points into this line (one for each start and end of the line, and another two in the middle; one has larger width, and the other has a near-zero width)
fixingHomogenLines1A


After adding a new vertex point. (The new vertex point is marked by a red arrow)
Some discrepancies still present between the before and after lines
fixingHomogenLines1B

rodolforg:
> SynfigStudio-1.5.2-testing-2025-08-13-win64-8c609

Are you getting from AppVeyor, right? It should be this one: https://ci.appveyor.com/api/buildjobs/98nyw4suaguxhbxh/artifacts/SynfigStudio-1.5.2-testing-2025-07-25-win64-bba0e.zip .

Until the PR got merged, only the one provided by the PR has the modifications.

FadhilKwan:
> Are you getting from AppVeyor, right? It should be this one: https://ci.appveyor.com/api/buildjobs/98nyw4suaguxhbxh/artifacts/SynfigStudio-1.5.2-testing-2025-07-25-win64-bba0e.zip .
> 
> Until the PR got merged, only the one provided by the PR has the modifications.

Ah okay, got it! That's probably the case~ :)

Yeah, I just tested it as well, and.... it still has the same issue as the one I mentioned earlier... ._.?)

Before
fixingHomogenLines2A


After (new vertex point indicated)
fixingHomogenLines2B

rodolforg:
Ouch!

Maybe it depends on #3555 . I'll re-push it here. Once the new artifact is generated, please try it again ;)

rodolforg:
@FadhilKwan New artifact is done :)

FadhilKwan:
> @FadhilKwan New artifact is done :)

'aight thanks for the heads-up! ^^

I've tried it and.... still the same?...
hmmmm.... I'm really hoping this is not one of those times where my pc decides that it wants to be quirky, and breaks Synfig only on my end.... .-.

But, yeah, still not quite there... ._.;

SynfigStudio-1.5.2-testing-2025-07-25-win64-2e3cb (which I assume is the new artifact that has this fix; hope I got the right one)

Before
fixingHomogenLines3A

After
fixingHomogenLines3B



the newest Artifact as of now: SynfigStudio-1.5.3-testing-2025-08-17-win64-b915e
(Which I'm fully aware is very likely not necessary for this test, but might as well~)

Before
fixingHomogenLines4A

After
fixingHomogenLines4B

I might try this again on a different PC, but it'll might be a while until then~ ><

rodolforg:
The latest artifact is available here:
image
(You can only see it if logged in)

Then, in AppVeyor site, you can click in the Artifacts tab.
image

Finally, you get the download link:
image

https://ci.appveyor.com/api/buildjobs/8n09l0mikjebjkl2/artifacts/SynfigStudio-1.5.2-testing-2025-07-25-win64-2e3cb.zip

rodolforg:
Anyway, I did something wrong :'( I'll check it again ¬¬

rodolforg:
@FadhilKwan now it's working! The problem was I did some code cleanup before pushing, but I did it wrong...

FadhilKwan:
Yes!! It finally worked on mine as well! Thanks goodness it wasn't my pc this time~

Yeah, I've tested it a bit, and it seems the fix has worked very well, I even added one more vertex point than usual just to doubly make sure that it does the job (and sure it did :D )

AppVeyor Artifact tested (for future reference): SynfigStudio-1.5.2-testing-2025-07-25-win64-47d44

Before
fixingHomogenLines5A

After adding 1 vertex point
fixingHomogenLines5B

After 2 vertex points
fixingHomogenLines5C

(on both vertex point additions, all width points seemed to be locked in place, and that's good~)

I was also wondering what would happen if I add a new vertex point on animation mode (since you said it is not implemented in). So I tried that and found out the line was just "glitching" to its normal expected behavior (as if this fix were not present to begin width). Which made me realize that animating everchanging lines without using Homogenous is not viable (it is a bit unfortunate, though won't affect me by much :) ).
But at the very least it won't crash Synfig, so that's like, still a win to me~

Now, I haven't really tested this fix to its limits yet, I'd have to actually draw something in this version of Synfig if I want to do so. Though I am also currently working on a drawing on a different Synfig version right now, so that'll have to wait.... ._.v

Thanks alot, mate for bringing this fix together! Hopefully it'll get fleshed out (if there's still a way to), and made available on future releases.  b^^
This might actually help convince me to move to the newer Synfig (despite of its jagged canvas issue xD )

Issue #3571, opened 2025-04-21
Native Inverse Kinematics Support for Synfig – Current Limitations and Suggested Approach


zaynopenapp:
Hi all,
I believe that Inverse Kinematics (IK) is a fundamental feature that Synfig should have supported natively by now, especially given its capabilities with bone-based rigging. I previously attempted to create an IK plugin, but due to the reliance on multiple converters, performance degraded significantly.
Below is my analysis and a suggestion on how IK might be implemented more efficiently in Synfig...

Inverse Kinematics (IK) is a crucial feature for any animation software that supports bone-based rigging, and Synfig should have had native IK support a long time ago. While Synfig's current system is powerful for parameter-driven animation, it is not well-suited for real-time IK calculations due to its heavy reliance on value nodes and converters.

I previously attempted to implement IK through a custom plugin. However, the solution required chaining a large number of converters to compute joint rotations based on a target position. This introduced significant overhead and caused performance issues, making the setup impractical for regular use.

To improve this, IK in Synfig could be implemented using a dedicated IK solver module that runs outside the regular parameter/converter system. This solver could calculate joint angles based on target positions, then bake those results as keyframes or feed them into the animation pipeline through a more efficient internal channel.

Alternatively, a hybrid solution might be possible—where the user defines IK targets and pole vectors, and Synfig performs offline calculations (e.g., during preview or keyframe generation), rather than evaluating IK on every frame in real time.

Implementing native IK support would make character animation in Synfig more competitive and user-friendly, especially for those migrating from tools like Moho or Blender.

rodolforg:
> I previously attempted to create an IK plugin, but due to the reliance on multiple converters, performance degraded significantly.

Could you share it with us?

zaynopenapp:
Here’s a simple idea that might help make the IK converter faster. Thank you.

![Image](https://github.com/user-attachments/assets/835ff507-390a-496e-911e-374093ede460)

plugin and code 
https://github.com/zaynopenapp/IK_formula.git

morevnaproject:
Related to https://github.com/synfig/synfig/issues/601

PR #3623, opened 2025-08-09
feat: new 'IKangle' converter , inverse kinematics bone

zaynopenapp:
The new 'IK angle' converter is a simple inverse kinematics solution. In line with the issues discussed in https://github.com/synfig/synfig/issues/3571 or https://forums.synfig.org/t/inverse-kinematic-for-bones/3068, and (https://forums.synfig.org/t/ik-system/4132)it is hoped that inverse kinematics will finally be available for users who have long wanted Synfig to have IK bones.

forum: https://forums.synfig.org/t/bone-like-movement/588, https://forums.synfig.org/t/skeleton-bones-duplication/3650, https://forums.synfig.org/t/inverse-kinematic-for-bones/3068

sample here. https://forums.synfig.org/t/ik-angle-converter-test/16552 and https://youtu.be/kgnKCwXtjIQ

The basic idea behind my **IK Angle** converter is based on the geometric intersection of two circles.

Imagine a simple two-bone chain. The first circle is centered at the root bone with a radius equal to the length of the first bone. The second circle is centered at the target with a radius equal to the length of the second bone.

The intersection of these two circles gives the possible position of the intermediate joint (for example, the elbow or the knee). Once that joint position is known, the rotation angle of each bone can be calculated directly using vector angles.

Because this approach is based on geometry rather than an iterative solver, it produces a stable solution with very little computation. There is no need for repeated iterations such as those used by CCD or FABRIK for a simple two-bone chain.

To support different bending directions, the solver simply chooses one of the two possible intersection points (for example, left/right or up/down), depending on the desired pole direction.

This geometric approach is the core idea behind my IK Angle converter.

The distance between the root and the target determines whether the circles intersect. When they do, there are generally two possible solutions, corresponding to the two possible elbow or knee positions. The converter selects one of these solutions and computes the bone rotations from the resulting vectors.

[interception_cycle_IK.webm](https://github.com/user-attachments/assets/ef66af5e-9b23-43b6-be2d-22224d8e8a5d)

### DIFF
--- a/synfig-core/po/POTFILES.in
+++ b/synfig-core/po/POTFILES.in
@@ -384,6 +384,8 @@ src/synfig/valuenodes/valuenode_gradientrotate.cpp
 src/synfig/valuenodes/valuenode_gradientrotate.h
 src/synfig/valuenodes/valuenode_greyed.cpp
 src/synfig/valuenodes/valuenode_greyed.h
+src/synfig/valuenodes/valuenode_ik_angle.cpp
+src/synfig/valuenodes/valuenode_ik_angle.h
 src/synfig/valuenodes/valuenode_integer.cpp
 src/synfig/valuenodes/valuenode_integer.h
 src/synfig/valuenodes/valuenode_intstring.cpp
--- a/synfig-core/src/synfig/layers/layer_skeleton.cpp
+++ b/synfig-core/src/synfig/layers/layer_skeleton.cpp
@@ -113,6 +113,9 @@ Layer_Skeleton::set_param(const String &param, const ValueBase &value)
 	// Skip shape, polygon and composite parameters
 	if (param == "amount")
 		return Layer_Composite::set_param(param,value);
+
+	if (param == "color")
+		return Layer_Shape::set_param(param,value);
 	return Layer::set_param(param,value);
 }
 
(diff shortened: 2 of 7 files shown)


zaynopenapp (commented):
Why my code error?

rodolforg:
> Why my code error?

image

zaynopenapp:
Ok. Thank you. I will immediately fix the part that has the error.

_**Edit BobSynfig: The image was not displayed in the quote, still taking empty space**_

zaynopenapp (commented):
I’ve fixed it according to Gemini’s suggestions, though not all of them. I’ve also tested the Windows artifact, and it runs just like the Linux version I built myself.
ik angle

zaynopenapp:
Below is a comparison of files from an IK example using a plugin and an IK angle converter:

ik file compare

Although compressing into a .sifz file reduces the file size, the original file still contains a large number of converters, which significantly impacts Synfig’s performance. This is only for a single IK rig object (a dinosaur). One can imagine how heavy it would become with 5 such objects using the IK plugin.

[comparefile.zip](https://github.com/user-attachments/files/26275687/comparefile.zip)

SlamRockwell:
This would be amazing to have in Synfig and I hope the developers can pull it in soon! :)

BobSynfig:
AppImage for test [UNOFFICIAL-1.5.5-2026.05.06-test-pr-3623](https://github.com/BobSynfig/synfig/releases/tag/UNOFFICIAL-1.5.5-2026.05.06-test-pr-3623)

zaynopenapp:
You can find the documentation for the **IK angle** converter here. Thank you.
[link](https://github.com/zaynopenapp/IK-angle-new-converter-Synfig)

ikangle_tuto

zaynopenapp:
> This would be amazing to have in Synfig and I hope the developers can pull it in soon! :)

[@SlamRockwell](https://github.com/SlamRockwell) Thank you for your support. you can review this PR

Svarov-RZM:
I see I've been tagged on this, but it's a bit too much for me to handle (I am a sysadmin, remember).
Also, I don't use bones in Synfig so I won't be able to do a good test.

I can do a surface level test of this PR if you want me to, but the code review **must** be done by someone more competent.

zaynopenapp (commented):
I fixed this issue

PR #3635, opened 2025-10-25
perf(Shade layer): do not insert blur task if none specified

Svarov-RZM:
This PR optimizes the case when the 'Size' parameter of the 'Shade' layer is set to 0.0 (no blur). Previously, it was alway inserting TaskBlur task regardles if there's a need for it. Now it only does that when Size parameter is greater than zero improving perfomance by about 1.78 (Synfig 1.5.3 = 5.446s vs This PR = 3.053s) on this test case file:
[ShadeMemory.zip](https://github.com/user-attachments/files/23141744/ShadeMemory.zip)


ice0 (approved):

ice0:
Merged. Thank you!

PR #3640, opened 2025-11-26
fix: prevent negative bline/brush width in rectangle state and constrain the range

mohamedAdhamc:
Copied from #3622 : 

" I noticed that for all the other states the bline width had it's range constrained and didn't allow for negative values. Specifically through setting it explicitly in the respective state class as follow:

bline_width_dist.set_digits(2);
bline_width_dist.set_range(0,10000000);

The rectangle state however didn't have that and it led to a weird behavior for example when setting the value to be a negative number.

The set_digits part is not absolutely needed but most other tools had it so I kept it for consistency."


mohamedAdhamc:
@ice0 I think this should be good now.

ice0:
Merged. Thank you!

Issue #3634, opened 2025-10-25
Filter layer: Add parameter to control saturation


Svarov-RZM:
### Description
Currently, the 'Filter' layer doesn't have the ability to control saturation. It does have 'Contrast' parameter which is similar, but not quite the same because contrast also changes the value of the color and I need to adjust only intensity of it for layers below.

Basically, like in HSV wheel where we have the ability to control hue, value and saturation independently. The 'Filter' layer has hue and value (called exposure) but lacks saturation. Personally, I am planning to use it to improve my Color Theory template (it has quite a few bugs): https://forums.synfig.org/t/synfig-application-color-theory/13671

aviralgarg05:
I would like to solve this issue

rodolforg:
Feel free to do it :)

Svarov-RZM:
Implemented in https://github.com/synfig/synfig/pull/3641

PR #3641, opened 2025-11-26
feat: add saturation parameter to Color Correct layer

aviralgarg05:
- Added param_saturation member variable to Layer_ColorCorrect class
- Initialized saturation parameter to 1.0 (no change) in constructor
- Implemented saturation import/export in set_param and get_param methods
- Added saturation to parameter vocabulary with user-friendly name
- Integrated saturation adjustment into ColorMatrix transformation chain
- Saturation uses ColorMatrix::set_saturation() via YUV color space conversion

This addresses issue #3634 by allowing independent control of color saturation without affecting brightness or contrast. The parameter works as follows:
- 0.0: Grayscale (no saturation)
- 1.0: Original saturation (default)
- >1.0: Enhanced saturation

The saturation parameter works independently from other color adjustments and is positioned in the transform chain after hue but before exposure, brightness, and contrast for optimal color processing.

### DIFF
--- a/synfig-core/src/modules/mod_filter/colorcorrect.cpp
+++ b/synfig-core/src/modules/mod_filter/colorcorrect.cpp
@@ -97,11 +97,10 @@ apply_saturation_impl(Color& dst, const Color& src, Real saturation, const Gamma
 		// At saturation=0, all components become max_val (grayscale at Value)
 		// At saturation=1, no change
 		// At saturation>1, components move away from max_val (more saturated)
-		// Clamp results to [0,1] range for saturation > 1.0
 		ColorReal sat = static_cast<ColorReal>(saturation);
-		linear.set_r(synfig::clamp(max_val - (max_val - linear.get_r()) * sat, ColorReal(0), ColorReal(1)));
-		linear.set_g(synfig::clamp(max_val - (max_val - linear.get_g()) * sat, ColorReal(0), ColorReal(1)));
-		linear.set_b(synfig::clamp(max_val - (max_val - linear.get_b()) * sat, ColorReal(0), ColorReal(1)));
+		linear.set_r(max_val - (max_val - linear.get_r()) * sat);
+		linear.set_g(max_val - (max_val - linear.get_g()) * sat);
+		linear.set_b(max_val - (max_val - linear.get_b()) * sat);
 
 		// Reapply gamma to return to gamma-corrected space
 		dst = canvas_gamma.apply(linear);
@@ -133,8 +132,8 @@ TaskSaturationSW::run(RunParams&) const
 		if (!lsrc)
 			return false;
 
-		synfig::Surface &dst = ldst->get_surface();
-		const synfig::Surface &src = lsrc->get_surface();
+		synfig::Surface& dst = ldst->get_surface();
+		const synfig::Surface& src = lsrc->get_surface();
 
 		for (int y = rs.miny; y < rs.maxy; ++y) {
 			const Color* src_ptr = &src[y - rd.miny - offset[1]][rs.minx - rd.minx - offset[0]];


Svarov-RZM:
Hey @aviralgarg05 
Thanks for addressing my issue! I'll take it for a test on weekends and report back.

rodolforg:
Thank you for your contribution, @aviralgarg05 !
Your PR does a lot of code style changes mixed with your actual contribution.

Could you please remove this code style changes not directly related to your contribution code to be easier for our code review? You can do it just pushing another commit.

Please pay attention we use tabs for indentation instead of spaces.

Svarov-RZM:
Hi @aviralgarg05,

I've tested this PR and it doesn't quite work as expected. Take a look at this:
SaturationTest

The top row is how it should be (when you decrease saturation in HSV wheel). The bottom row is the result produced by Color Correct layer. You can clearly see that not only saturation, but hue and value also changes. Do you have an idea why?

Here's the source for test-case file:
[SaturationTest.zip](https://github.com/user-attachments/files/23833327/SaturationTest.zip)

aviralgarg05:
The saturation implementation is using YUV color space (via ColorMatrix::set_saturation() and Color::set_s()), which causes unwanted hue and value shifts. To get the expected behavior (matching the HSV wheel), the implementation would need to use HSV color space transformations instead, which properly isolates saturation from hue and value.

The YUV approach works for certain color grading workflows, but it doesn't match the expectation of "saturation" as defined in HSV color models.


I will make the required changes

Svarov-RZM:
> To get the expected behavior (matching the HSV wheel), the implementation would need to use HSV color space transformations instead

Are you absolutely sure about this? I haven't looked into it myself, but it's hard to believe that you need to convert stuff to a different representation just to change intensity of color. Anyway, I have tested your recent commits and now when you change saturation parameter nothing happens.

Have you checked the example I provided in previous message? You need to get the bottom row to be the same as top one using Color Correct layer.

aviralgarg05:
I will test it out once again, although I tested it locally

rodolforg:
Maybe the issue was the order of color matrix transformations in the original PR commits?

aviralgarg05:
> Maybe the issue was the order of color matrix transformations in the original PR commits?

Yup, have done some turn arounds and fixes

Svarov-RZM:
> Yup, have done some turn arounds and fixes

Well, it's close, but there's one more problem. Take a look at the second column in a test-case file:
ForgotAboutGamma

The reason for this is that you forgot about gamma settings set in Menu->Canvas->Properties->Gamma correction.
If gamma is set to 1.0 your logic works as expected, if it is set to anything else, results are no longer proper. Can you account for gamma correction in your logic?

aviralgarg05:
Sure I will do it as well

Svarov-RZM:
> Sure I will do it as well

Wait, wait, I am not talking about gamma parameter of the layer itself, I mean global gamma settings that's set in Menu->Canvas->Properties->Gamma correction. Did you check the test-case file I uploaded earlier in this thread? The gamma setting there is set to 0.45 so the result of saturation change isn't computed properly.

aviralgarg05:
Ok I will check and solve

Svarov-RZM:
@aviralgarg05 I checked the last commit and saturation now works as expected, hooray! But...

The 'Contrast' parameter is now broken. To reproduce:
* Open the example file I attached in this thread
* Zoom in past 200%
* Change 'Contrast' parameter of any 'Color Correct' layer to 0.5

After that Synfig will either crash immediately or produce weird artifacts like this:
ContrastBreaksThings
And then crash on zooming out. Can you please investigate what's causing this?

Oh, and also, you still didn't address the indentation issue, please notice that we use tabs for indentation, not spaces.

Svarov-RZM:
@aviralgarg05 OK, I tested the last commit and everything works as expected now.
One little thing left, please bump the version of the layer to 0.2 - we did introduce new functionality after all. After that, the PR is ready for final review.

Svarov-RZM:
@rodolforg @ice0 @BobSynfig @Abdelhadi-Wael 

Hey guys,

Can some of you review the changes? There were a lot of mistakes and I am not confident reviewing it myself for I am not very good with C++. I can confirm that logic is correct and working properly but maybe you have some code-related feedback. If you approve, I'll merge it.

synfig-core/src/modules/mod_filter/colorcorrect.h
```diff
@@ -46,6 +48,38 @@ namespace synfig {
...
+class TaskSaturationSW: public TaskSaturation, public rendering::TaskSW
+{
+public:
+	typedef etl::handle<TaskSaturationSW> Handle;
+	static Token token;
+	virtual Token::Handle get_token() const { return token.handle(); }
+
+	virtual bool run(RunParams &params) const;
+private:
+	void apply_saturation(Color &dst, const Color &src) const;
```
rodolforg: Code style:
1. Delete the two occurrences of `virtual` keyword here and append `override` keyword.
2. Instead of `Type &var_name`, prefer `Type& var_name` (note the different place of the white space). Three occurrences here.

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -203,6 +319,10 @@ Layer_ColorCorrect::get_param_vocab()const
 		.set_local_name(_("Exposure Adjust"))
 	);
 
+	ret.push_back(ParamDesc("saturation")
+		.set_local_name(_("Saturation"))
```
rodolforg, line 308: please add a `set_description(_("SHORT DESCRIPTION TELLING WHAT THE VALUE MEANS"))` here.
I know other parameters of this layer doesn't have this description, but they should :)
It appears as tooltip for user in the Parameters Panel.
You explained the parameter value in the task code comments.

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,86 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+	{
+		// Move each component toward max_val based on saturation factor
+		// At saturation=0, all components become max_val (grayscale at Value)
+		// At saturation=1, no change
+		// At saturation>1, components move away from max_val (more saturated)
+		// Clamp results to [0,1] range for saturation > 1.0
+		ColorReal sat = static_cast<ColorReal>(saturation);
+		linear.set_r(std::max(ColorReal(0), std::min(ColorReal(1), max_val - (max_val - linear.get_r()) * sat)));
+		linear.set_g(std::max(ColorReal(0), std::min(ColorReal(1), max_val - (max_val - linear.get_g()) * sat)));
+		linear.set_b(std::max(ColorReal(0), std::min(ColorReal(1), max_val - (max_val - linear.get_b()) * sat)));
```
rodolforg: Please use `synfig::clamp(VAL, MIN_VALUE, MAX_VALUE)` for better reading :smile:

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,86 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+	ColorReal min_val = std::min({linear.get_r(), linear.get_g(), linear.get_b()});
+
+	// Only adjust if there's actual saturation (max != min) and max > 0
+	if (max_val > 0 && max_val != min_val)
+	{
+		// Move each component toward max_val based on saturation factor
+		// At saturation=0, all components become max_val (grayscale at Value)
+		// At saturation=1, no change
+		// At saturation>1, components move away from max_val (more saturated)
+		// Clamp results to [0,1] range for saturation > 1.0
```
rodolforg: Maybe "clamping" should be optional?

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,86 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+// Task tokens for saturation
+rendering::Task::Token TaskSaturation::token(
+	DescAbstract<TaskSaturation>("Saturation") );
+rendering::Task::Token TaskSaturationSW::token(
+	DescReal<TaskSaturationSW, TaskSaturation>("SaturationSW") );
 
 /* === P R O C E D U R E S ================================================= */
 
+void
+TaskSaturationSW::apply_saturation(Color &dst, const Color &src) const
```
rodolforg: Code style:
Instead of `Type &var_name`, prefer `Type& var_name` (note the different place of the white space). Two occurrences here.
We are slowly changing for this style in new code hoping someday the whole code naturally changes to it :wink:

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,86 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+	Gamma inv_gamma = canvas_gamma.get_inverted();
+	Color linear = inv_gamma.apply(dst);
+
+	// Find the max (Value) and min of RGB components in linear space
+	ColorReal max_val = std::max({linear.get_r(), linear.get_g(), linear.get_b()});
+	ColorReal min_val = std::min({linear.get_r(), linear.get_g(), linear.get_b()});
+
+	// Only adjust if there's actual saturation (max != min) and max > 0
+	if (max_val > 0 && max_val != min_val)
+	{
```
rodolforg: Code style:
for control flow statements like `if`, `for`, `while`, `do - while`, `switch`, keep the opening curling bracket `{` in the same line of the statement like in `if (condition) {`

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,86 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+	// Validation checks matching TaskPixelGammaSW pattern
+	if (!is_valid() || !sub_task() || !sub_task()->is_valid())
+		return true;
+
+	RectInt rd = target_rect;
+	VectorInt offset = get_offset();
+	RectInt rs = sub_task()->target_rect + rd.get_min() + offset;
+	rect_set_intersect(rs, rs, rd);
+	if (rs.is_valid())
+	{
```
rodolforg: Code style:
opening curling bracket `{`

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,86 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+			return false;
+		LockRead lsrc(sub_task());
+		if (!lsrc)
+			return false;
+
+		synfig::Surface &dst = ldst->get_surface();
+		const synfig::Surface &src = lsrc->get_surface();
+
+		for(int y = rs.miny; y < rs.maxy; ++y)
+		{
```
rodolforg: Code style:
1. opening curling bracket `{`
2. insert a single space between keywords like `for`, `if`, `while`, `switch` and the opening parenthesis `(` like in `for (condition) {`

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,86 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+			const Color *src_ptr = &src[y - rd.miny - offset[1]][rs.minx - rd.minx - offset[0]];
+			Color *dst_ptr = &dst[y][rs.minx];
+			for(int x = rs.minx; x < rs.maxx; ++x, ++src_ptr, ++dst_ptr)
+				apply_saturation(*dst_ptr, *src_ptr);
```
rodolforg: Code style:
1. insert a single space between keywords like `for`, `if`, `while`, `switch` and the opening parenthesis `(` like in `for (condition) {`

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,86 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+		if (!lsrc)
+			return false;
+
+		synfig::Surface &dst = ldst->get_surface();
+		const synfig::Surface &src = lsrc->get_surface();
+
+		for(int y = rs.miny; y < rs.maxy; ++y)
+		{
+			const Color *src_ptr = &src[y - rd.miny - offset[1]][rs.minx - rd.minx - offset[0]];
+			Color *dst_ptr = &dst[y][rs.minx];
```
rodolforg: Hmm... What about using reference instead of pointer?
```cpp
			const Color& src_color = src[y - rd.miny - offset[1]][rs.minx - rd.minx - offset[0]];
			Color& dst_color = dst[y][rs.minx];
```

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -88,7 +168,41 @@ Layer_ColorCorrect::correct_color(const Color &in)const
...
+			// Move each component toward max_val based on saturation factor
+			ColorReal sat = static_cast<ColorReal>(saturation);
+			linear.set_r(std::max(ColorReal(0), std::min(ColorReal(1), max_val - (max_val - linear.get_r()) * sat)));
+			linear.set_g(std::max(ColorReal(0), std::min(ColorReal(1), max_val - (max_val - linear.get_g()) * sat)));
+			linear.set_b(std::max(ColorReal(0), std::min(ColorReal(1), max_val - (max_val - linear.get_b()) * sat)));
+
+			// Reapply gamma to return to gamma-corrected space
+			ret = canvas_gamma.apply(linear);
+		}
+	}
```
rodolforg, line 187: Maybe to avoid code repetition here from the task you created, you could implement a `static inline` procedure with three parameters:
1. the dst_color
2. the src_color
3. the canvas_gamma

It can be noticed it would have only one additional parameter than `TaskSaturationSW::apply_saturation()`.
This way, both methods (`TaskSaturationSW::apply_saturation()` and `Layer_ColorCorrect::correct_color()`) could call this new procedure without code repetition and keep the performance due to to `inline` ;)

rodolforg (changes requested):
@Svarov-RZM The only code issue I could see was outside of scope of this PR.
It seems the order of Hue Adjust, Brightness, Contrast and Exposure is different for the cobra render and the color pick/eyedrop tool.
I don't know if it could give different results though.

aviralgarg05:
Thank you for the review feedback, @rodolforg!

I've addressed both comments in commit eb531cedd:

1. **Use references instead of pointers** - Changed the loop in `TaskSaturationSW::run()` to use `const Color& src_color` and `Color& dst_color` as suggested.

2. **Create static inline helper** - Added `apply_saturation_impl()` as a `static inline` function that both `TaskSaturationSW::apply_saturation()` and `Layer_ColorCorrect::correct_color()` now call. This eliminates the code duplication while maintaining performance.

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,96 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+
+		synfig::Surface &dst = ldst->get_surface();
+		const synfig::Surface &src = lsrc->get_surface();
+
+		for(int y = rs.miny; y < rs.maxy; ++y)
+		{
+			for(int x = rs.minx; x < rs.maxx; ++x)
+			{
+				const Color& src_color = src[y - rd.miny - offset[1]][x - rd.minx - offset[0]];
+				Color& dst_color = dst[y][x];
```
rodolforg: So sorry I missed the point of using pointers instead of reference here!
Could you please revert? The other way would be more efficient indeed, as less address computation would be done at each row loop (only increments instead of the whole address computation, right?)

aviralgarg05: Yup, I will revert that

rodolforg:
> Thank you for the review feedback, @rodolforg!
> 
> I've addressed both comments in commit [eb531ce](https://github.com/synfig/synfig/commit/eb531cedd82f93c31e70e3ba53f0103aad69343c):
> 
>     1. **Use references instead of pointers** - Changed the loop in `TaskSaturationSW::run()` to use `const Color& src_color` and `Color& dst_color` as suggested.
> 
>     2. **Create static inline helper** - Added `apply_saturation_impl()` as a `static inline` function that both `TaskSaturationSW::apply_saturation()` and `Layer_ColorCorrect::correct_color()` now call. This eliminates the code duplication while maintaining performance.

Wow, you were fast XD What about the change requests regarding the code style?

aviralgarg05:
> > Thank you for the review feedback, @rodolforg!
> > I've addressed both comments in commit [eb531ce](https://github.com/synfig/synfig/commit/eb531cedd82f93c31e70e3ba53f0103aad69343c):
> > ```
> > 1. **Use references instead of pointers** - Changed the loop in `TaskSaturationSW::run()` to use `const Color& src_color` and `Color& dst_color` as suggested.
> > 
> > 2. **Create static inline helper** - Added `apply_saturation_impl()` as a `static inline` function that both `TaskSaturationSW::apply_saturation()` and `Layer_ColorCorrect::correct_color()` now call. This eliminates the code duplication while maintaining performance.
> > ```
> 
> Wow, you were fast XD What about the change requests regarding the code style?

Yup I was actually fixing them, was going through them one by one

aviralgarg05:
Thank you for the detailed review, @rodolforg! I've addressed all feedback in commits eb531cedd and d39d34838:

### Changes Made:
1. **virtual -> override** - Removed `virtual` keyword and added `override` for `get_token()` and `run()` methods
2. **Reference spacing** - Changed `Type &var` to `Type& var` throughout
3. **set_description** - Added tooltip for saturation parameter: "Color saturation adjustment. 0.0 = grayscale, 1.0 = no change, >1.0 = more saturated"
4. **synfig::clamp** - Replaced `std::max(min, std::min(max, val))` with `synfig::clamp(val, min, max)`
5. **Brace style** - Changed to `if (condition) {` style
6. **Keyword spacing** - Added space between `for`/`if` and `(`
7. **Pointer efficiency** - Kept original pointer-based loop per your latest feedback
8. **static inline helper** - `apply_saturation_impl()` is used by both `TaskSaturationSW::apply_saturation()` and `Layer_ColorCorrect::correct_color()`

### Regarding optional clamping:
I kept clamping enabled for now because without it, saturation > 1.0 can push color components outside [0,1] range, potentially causing rendering artifacts. However, if you prefer, I can add a boolean parameter to make clamping optional for HDR workflows. Please let me know your preference.

synfig-core/src/modules/mod_filter/colorcorrect.cpp
```diff
@@ -59,10 +62,91 @@ SYNFIG_LAYER_INIT(Layer_ColorCorrect);
...
+	if (rs.is_valid()) {
+		LockWrite ldst(this);
+		if (!ldst)
+			return false;
+		LockRead lsrc(sub_task());
+		if (!lsrc)
+			return false;
+
+		synfig::Surface &dst = ldst->get_surface();
+		const synfig::Surface &src = lsrc->get_surface();
```
rodolforg: code style. I missed this one before :frowning:

rodolforg:
> I kept clamping enabled for now because without it, saturation > 1.0 can push color components outside [0,1] range, potentially causing rendering artifacts. However, if you prefer, I can add a boolean parameter to make clamping optional for HDR workflows. Please let me know your preference.

Oh, I suggested because I believe it would be something @morevnaproject would request lol

Svarov-RZM:
> I kept clamping enabled for now because without it, saturation > 1.0 can push color components outside [0,1] range, potentially causing rendering artifacts

When those artifacts are used to create some cool effects they become artistic vision :)
Basically, in Synfig we don't stop user from pushing things out of range, for example, we can set opacity greater than 1 to make blur thicker. Let's remove clamping, it's not necessary here.

@aviralgarg05 We have two things left to do:
* Remove clamping (I'll give it one final test after that)
* Address remaining rodolforg's code-style request

And I believe we're done.

aviralgarg05:
Done with the feedbacks!! @Svarov-RZM @rodolforg

Svarov-RZM:
Sheesh, the history here is a mess... I am squashing it.

@aviralgarg05 I finished testing it and didn't find any more problems. We struggled a bit with this but in the end we got it working. Thanks for not giving up and seeing it through.

aviralgarg05:
@Svarov-RZM @rodolforg is there a way I could personally connect with you, there is someone I got a message from looking to connect with you urgently regarding a few Synfig-related assets and has questions around licensing and legal responsibility for the community.

BobSynfig:
@aviralgarg05 For legal matters, you should contact our project leader @morevnaproject (@KonstantinDmitriev on our forums)

Svarov-RZM:
@aviralgarg05 Yeah, what Bob said - please use forums for this matter: https://forums.synfig.org/
Post your questions there and hopefully someone will provide an answer.

> urgently

There's no "urgently" in Open Source.

PR #3537, opened 2025-03-30
feat: added a way to easily access the rendered file

002harshit:
Feature : #3533 
above issue contains the description of the feature

let me know if there are any bad practices in my code based on synfig's code style

- i have created "Open Rendered File" button in dock_info which opens rendered file
- file is opened based on output. folder is opened in case of multiple outputs

Screenshot from 2025-03-30 08-58-22

PR #3642, opened 2025-11-28
feat: added a way to easily access the rendered file

### LINKED ISSUE #3533: Add a way to easily access the rendered file
Currently, user has to find its own way to play the rendered animation after the rendering finishes.

I suggest to show a button in the Info panel to let user open the rendered file:
- If it is a movie, open it with OS default movie player;
- If it is a single image¹, open it with OS default image viewer;
- If it is a image sequence², open it the folder where it was saved.

¹ In the render dialog, the user can choose whether to render the whole animation or the current frame only.
Note that Spritesheet is also a single image.
² Multiple images (eg. PNG, BMP or JPG) one per frame.

Optionally, user could define the default movie player and image viewer in the Settings (as we have for image editor).

Raw idea (provided by @ice0) of how to provide this feature:
Image

Coder:
You can launch a file with its default app via `synfig::OS::launch_file_async()`

rodolforg:
Supersedes original PR: #3537 (closed by original author)

Adds two buttons in Info Panel to easier access to rendered file: one to play the animation itself, the other to open its containing folder.

Original description:

> Feature : https://github.com/synfig/synfig/issues/3533
> above issue contains the description of the feature
>
>    i have created "Open Rendered File" button in dock_info which opens rendered file
>    file is opened based on output. folder is opened in case of multiple outputs

fix #3533

### DIFF
--- a/synfig-studio/src/gui/asyncrenderer.cpp
+++ b/synfig-studio/src/gui/asyncrenderer.cpp
@@ -162,6 +162,16 @@ public:
 		return r;
 	}
 
+	bool is_multiple_files() const override
+	{
+		return warm_target->is_multiple_files();
+	}
+
+	filesystem::Path get_filename() const override
+	{
+		return warm_target->get_filename();
+	}
+
 	virtual bool wait_render_tiles(ProgressCallback* cb = nullptr)
 	{
 		if(!alive_flag)
@@ -346,6 +356,16 @@ public:
 		alive_flag=false;
 	}
 
+	bool is_multiple_files() const override
+	{
+		return warm_target->is_multiple_files();
+	}
+
+	filesystem::Path get_filename() const override
+	{
+		return warm_target->get_filename();
+	}
+
 	virtual bool start_frame(synfig::ProgressCallback *cb)
 	{
 		this->cb = cb;
@@ -489,7 +509,7 @@ AsyncRenderer::stop(AsyncRenderer::Interaction interaction)
 			if(interaction == INTERACTION_UNDEFINED)
 			{
 				if(status == RENDERING_SUCCESS)
-					signal_success_();
+					signal_success_(target->get_filename());
 				else
 					error_message = _("Animation couldn't be rendered");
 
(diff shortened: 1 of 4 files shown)


Svarov-RZM:
Yeah, this is a good feature.
I'll give it a test on weekends, have no strength left after work.

rodolforg:
Nice. 
This would be iaiai a case of squash and merge, but there is a commit of somebody else, do maybe we should merge it without squashing it.
I'll rebase it to master branch tomorrow for a cleaner merge

Svarov-RZM:
Hmm, if target has multiple files (like a png sequence), "Open rendered file" button doesn't appear:
NoOpenFileButton

Did we come to an agreement to have this button be always visible where in case of multiple files it opens the first frame of rendered animation?

rodolforg:
> Did we come to an agreement to have this button be always visible where in case of multiple files it opens the first frame of rendered animation?

Oh, the problem here was I couldn't assure (or infer) the exact filename in such cases... Each target (export) implementation can/may set how many digits it could use for frame id in the filename after the separator. For now, though, all of them is set as 4 digits.

rodolforg:
@Svarov-RZM but I did a refactoring to accomplish your desires :wink:

Svarov-RZM:
> but I did a refactoring to accomplish your desires

Thank you very much, it works now and according to my desires!

Damn, I am sorry if I put additional work on your shoulders... I didn't mean to. Know that you don't have to please anyone, you could have just said something like "Hey, Svarov, implementing this is a bit tricky because of that and that" and I would have accepted this PR in its initial state. So yeah, please don't take my opinions as something that has to be addressed, I might be wrong, I might not understand the technical nuances behind my opinions/desires.

rodolforg:
It was a pleasure :+1:

rodolforg:
@Svarov-RZM you forgot to append the PR number on the merge commit title : `(#3642)`

Svarov-RZM:
> you forgot to append the PR number on the merge commit title : `(#3642)`

Damn it! Still haven't fully memorized the workflow...

Issue #1550, opened 2020-07-12
Bad quality of Advance Outline bends (comparing to 1.2.2)


morevnaproject:
**Synfig version & platform**:
Master branch, id f4a59e9

**Issue description**:

Example file - 
[linea2.sifz.zip](https://github.com/synfig/synfig/files/4908181/linea2.sifz.zip)

Here is how the line bend looks in 1.2.2 -
![screenshot_001](https://user-images.githubusercontent.com/332868/87241538-676e5a80-c44e-11ea-9b49-3dffb726b2a2.png)

...and here is in current master -
![screenshot_002](https://user-images.githubusercontent.com/332868/87241541-748b4980-c44e-11ea-8aa0-718236e236d6.png)

(NOTE: This issue appears after refactoring done in 573f832f5c0c2c9eecc74b811ae9e2943dbae3a2).

Svarov-RZM:
Solved by https://github.com/synfig/synfig/pull/3649

Issue #2284, opened 2021-08-25
Weird Line Bending (That Doesn't Match its Region) On a Specific Curve


FadhilKwan:
**Describe the bug**
I found an unusual line curving that happens on a very specfic line setup. It looks like this:
![BCB02](https://user-images.githubusercontent.com/25135159/130847367-10970b49-675c-460e-9510-a8db080d1254.JPG)

The issue can be seen very clear with Adv. Outline. With normal Outline the curving doesn't occur and it looks a little more fine, even though it still looks a little bit off (the curve got a little bit cut off).
If you want to make one for yourself, simply click on the region, right click the selected layer from 'Layers', then select 'Make Outline'. Also make sure you hide the other outline (Adv. Outline) as well; in 'Layers' panel, untick the box to the left of the Adv. Outline layer.
![BCB06](https://user-images.githubusercontent.com/25135159/130853252-af814b7f-53ef-4d65-86d9-dd45bc2d12a7.JPG)

**To Reproduce**
**I can't specifically describe the setup for this issue (it all happened by accident), so I've provided a necessary file for it.
Also please note that the line will look slightly different from the screenshots. This is because I readjusted its width handles so it will look evenly spread, though it still doesn't fix the issue.**
[CurvingBug.zip](https://github.com/synfig/synfig/files/7049418/CurvingBug.zip)

Simply download, extract, and open the file......
![BCB01](https://user-images.githubusercontent.com/25135159/130847993-25968b75-362c-41e9-ac5f-f0f653e701a1.JPG)

.......Then zoom in and take a close look on the section marked by a small rectangle outline.
![BCB02](https://user-images.githubusercontent.com/25135159/130848010-8aa958bd-a505-464f-9241-f5d395286ff5.JPG)

What's really strange about it, is that it can be fixed with simple steps.... partially

1. Right click on the problematic spline and select 'Split Tangents' or 'Split Tangents' Angle', both works
![BCB03](https://user-images.githubusercontent.com/25135159/130848863-cc4c7af6-9c24-44d9-b78d-910e76e56db7.JPG)

2. Click on the indicated point and nudge it ever so slightly
![BCB04B](https://user-images.githubusercontent.com/25135159/130848939-927dcef8-df54-4465-8b8e-6f581bc0be2f.JPG)

3. The curve corrected itself as a result, partially. You can see that the line and region slightly don't match (indicated by the red arrow).
[...]

PR #3649, opened 2025-12-02
fix(rendering): remove artifacts in rendering outlines by increasing sampling resolution

### LINKED ISSUE #3598: Last dash on Adv. Outline renders with artifacts
### Description

While working on my current animation I have found another bug, this time about rendering dashed outline. The last dash renders with artifacts. Check this screenshot from canvas view:

Image

You can see that the last dash looks off, has this weird "wavy" effect on it. Here's the test-case file:

[LastDashItem.zip](https://github.com/user-attachments/files/20623261/LastDashItem.zip)

### Affected versions

Synfig 1.4.0 (tested 1.4.4) and higher. I believe the issue was introduced at the same time the new dash type (Off-Peak Stop) was added.

### Workaround

Extend dash outline to be dashes + 1 from desired amount and then alpha over the last item. Quite awkward workaround since it requires grouping both layers but it's something.

Abdelhadi-Wael:
fixes #3598, #1550, #2284

### DIFF
--- a/synfig-core/src/modules/mod_geometry/advanced_outline.cpp
+++ b/synfig-core/src/modules/mod_geometry/advanced_outline.cpp
@@ -540,8 +540,8 @@ Advanced_Outline::sync_vfunc()
 {
 	clear();
 	
-	const int wire_segments = 32;
-	const int contour_segments = 16;
+	const int wire_segments = 64;
+	const int contour_segments = 32;
 	const BLinePoint bp_blank;
 	const WidthPoint wp_blank;
 	const DashItem di_blank;
--- a/synfig-core/src/modules/mod_geometry/outline.cpp
+++ b/synfig-core/src/modules/mod_geometry/outline.cpp
@@ -115,8 +115,8 @@ Outline::sync_vfunc()
 	if (param_bline.get_list().empty()) return;
 
 	const BLinePoint blank;
-	const int wire_segments = 16;
-	const int contour_segments = 8;
+	const int wire_segments = 64;
+	const int contour_segments = 32;
 	
 	const Real width  = param_width.get(Real());
 	const Real expand = param_expand.get(Real());


Abdelhadi-Wael:
This should close #1550 and #2284.

 **test 1 (from #1550)**

### **1.2.2**


### **Current Master**


### **Current PR**


### **Higher Quality (64, 32)**


---

**test 2 (from #2284)**

### **1.2.2**


### **Current Master**


### **Current PR**


### **Higher Quality (64, 32)**

Svarov-RZM:
Hmm, if we increase number of segments, we decrease performance, right? Is it wise to do so to workaround a bunch of specific use-cases? If the decrease in performance is small then it's fine I guess... I need to test it.

By the way, does someone know how many segments there were in 1.0.2 version? I remember seeing something like SAMPLES = 50, is it the same thing?

Abdelhadi-Wael:
> if we increase number of segments, we decrease performance, right?

Theoretically yes, but I think it would not be noticeable since values are still very small.

> Is it wise to do so to workaround a bunch of specific use-cases?

I think so , there is already 3 reported issues regarding this.

> does someone know how many segments there were in 1.0.2 version? I remember seeing something like SAMPLES = 50, is it the same thing?

The implementations are different, but the old `SAMPLES` (50) is functionally similar to the current `wire_segments`.

Svarov-RZM:
> Theoretically yes, but I think it would not be noticeable since values are still very small.

Yeah, definitely nothing drastic but still it would be interesting to see if there's any difference. OK, so what I am gonna do, I will make a simple test-case file and compare various values in terms of performance and also see how it does against the 1.0.2 version then report back, Hopefully, I will do it during the week, but more likely on next weekends.

Svarov-RZM:
Okay, I made a very simple test-case file that has a bunch of advanced outlines in it like so:
LineTest

The source:
[LineTest.zip](https://github.com/user-attachments/files/24354904/LineTest.zip)

In terms of rendering speed (1 frame, average of 3):
* Master at 16,8 =  0,38302000363667799998s
* Master at 32,16 = 0,38698530197143566668s
* Master at 64,32 = 0,38694437344869s

The difference between 32,16 and 64,32 is ultimately non-existent. The difference between 64,32 and original 16,8 is about 4ms. So maybe in a very complex outline-heavy work we'll get a performance downgrade of about 100ms, but it's still nothing. We do get a much better result at rendering various cusps, so I am in support of merging it.

@Abdelhadi-Wael Please bump the value up to 64,32 and do the same in outline.cpp so we have a uniform result in rendering between simple and advanced outlines.

Svarov-RZM:
PR's merged, thanks!

PR #3650, opened 2025-12-11
feat: angle converter can now be used for integers

rodolforg:


Svarov-RZM:
PR's merged, thank you!
I haven't used Angle converter myself yet, but there's no reason why it shouldn't support integer type.

PR #3652, opened 2025-12-14
ci(Github Actions): bump macOS version (14 Sonoma -> 26 Tahoe)

ice0:
Update to newest macOS image

### DIFF
--- a/.github/workflows/synfig-ci.yml
+++ b/.github/workflows/synfig-ci.yml
@@ -29,8 +29,8 @@ jobs:
           toolchain: autotools
           allow_failures: false
 
-        - os: macos-14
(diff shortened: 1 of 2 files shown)


rodolforg (approved):

rodolforg:
There is a macos-latest. Would it be a  better option?

ice0:
> There is a macos-latest. Would it be a better option?

In this case, there's a chance the build will just break without warning when GitHub will update the OS image.
When we update the image manually, we can fix build failures (if any occur) before switching the OS image.

rodolforg:
Got it. So we can merge this PR, right?

ice0:
> Got it. So we can merge this PR, right?

Yes. Done :)

PR #3646, opened 2025-11-30
fix(advanced_outline): incorrect width rendering on open outlines

Abdelhadi-Wael:
fixes #3645

PR #3653, opened 2025-12-14
refactor: separate AdvancedLine to add tests

rodolforg:


rodolforg:
The intention here is create a way to make easier to test PRs as #3646 , #3647 , #3649 etc. and, with those to-be-created tests, to prevent future regressions.

ice0:
The main issue here is that you can link test binary with module library (compiled with `-module` flag) on Linux, but this is not possible on macOS.
So we should:
1. load this module for testing using `libsynfig`
2. link it statically with test binary
3. build dynamic library specially for test
4. another option

P.S. I can check option 3 on the weekend if options 1 and 2 are not suitable.

rodolforg:
@ice0 thank you for your tips! Please take a look!

rodolforg:
@ice0 if you think the strategy I used is ok, I'll add more basic tests and then change the draft status of this PR.
After merging, @Abdelhadi-Wael could add to his PRs related to advanced outline.

PR #3658, opened 2025-12-21
fix(outline): resolve tip rendering artifacts on sharp tangents

### LINKED ISSUE #951: Outline tip artifact
Master branch, commit - 2b54ed842195016f44c246658f092729dea037bd

**Issue description**:
Test file - 
[outline.zip](https://github.com/synfig/synfig/files/3612623/outline.zip)

In version 1.0.2:
screenshot_002

In current master:
screenshot_003

(note the distorted tip on the right)

Abdelhadi-Wael:
fixes #951

We could have simply increased the number of segments as @blackwarthog suggested, but that would have required a large value (at least 128), and the quality still would not have been as good as this approach.

Inspired by the old implementation.

### DIFF
--- a/synfig-core/src/modules/mod_geometry/outline.cpp
+++ b/synfig-core/src/modules/mod_geometry/outline.cpp
@@ -143,8 +143,8 @@ Outline::sync_vfunc()
 		const ValueBase::List &bline = bline_segment.is_valid() ? bline_segment.get_list() : param_bline.get_list();
 
 		std::vector<BLinePoint> points;
-		for(ValueBase::List::const_iterator i = bline.begin(); i != bline.end(); ++i)
-			points.push_back(i->get(blank));
+		for (const auto & i : bline)
+			points.push_back(i.get(blank));
 
 		for (size_t i = 0; i < points.size(); ++i) {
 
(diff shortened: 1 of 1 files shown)


Svarov-RZM:
@Abdelhadi-Wael Can you please rebase?

Abdelhadi-Wael:
@Svarov-RZM rebased

synfig-core/src/modules/mod_geometry/outline.cpp
```diff
@@ -139,16 +142,40 @@ Outline::sync_vfunc()
 		}
 		const ValueBase::List &bline = bline_segment.is_valid() ? bline_segment.get_list() : param_bline.get_list();
 
+		std::vector<BLinePoint> points;
+		for (const auto & i : bline)
+			points.push_back(i.get(blank));
```
rodolforg, line 147: If you don't need the `bline` variable of (original) line 140, you can fill the `points` variable via `get_list_of()`:
```c++
const std::vector<BLinePoint> points = param_bline.get_list_of(BLinePoint());
```
for example. The bline_segment has the same method.

Svarov-RZM:
Hmm, I did some tests and current implementation has a serious bug somewhere.
Look at this:
TipBug

When I move the tangent on last vertex just ever so slightly, the outline renders incorrectly. Other versions of Synfig, including 1.0.2, handle this properly. Here's the source:
[TipTest.zip](https://github.com/user-attachments/files/24685227/TipTest.zip)

To reproduce select the tangent of the last vertex and hit arrow up key one time or maybe a couple, depending on current canvas zoom.

Svarov-RZM:
Hey @Abdelhadi-Wael 
Did you see a regression in rendering in my previous post?

PR #3660, opened 2025-12-22
fix: correct selection after grouping layers

### LINKED ISSUE #3659: Feature request: Changes to grouping layers behaviour in Synfig
Currently, when I try to group 2 or more layers together, synfig groups them and the selection moves to the the layer below it, which isn't supposed to be the intended behaviour. It adds an extra click to select back the newly created layer so that I can rename the group layer and do transformation in the canvas. It's a bit of an annoying behaviour in synfig and I don't know what purpose it serves and why this behaviour is implemented.

After grouping, the newly created group layer should be highlighted instead of the layer below it.

omda777:
Fixes #3659 

This PR improves the grouping workflow by keeping the selection on the
newly created group layer.

### DIFF
--- a/synfig-studio/src/gui/trees/layertree.cpp
+++ b/synfig-studio/src/gui/trees/layertree.cpp
@@ -426,7 +426,10 @@ LayerTree::select_layer(synfig::Layer::Handle layer)
 			iter=sorted_layer_tree_store_->convert_child_iter_to_iter(iter);
 
 		Gtk::TreePath path(iter);
-		layer_tree_view().expand_to_path(path);
+		// do not auto expand when selecting a group layer to avoid "opening" it on selection
+		if(!(layer->get_name() == "group"))
+			layer_tree_view().expand_to_path(path);
+
 		layer_tree_view().scroll_to_row(path);
 		layer_tree_view().get_selection()->select(iter);
 	}
--- a/synfig-studio/src/gui/trees/layertree.cpp
+++ b/synfig-studio/src/gui/trees/layertree.cpp
@@ -432,6 +432,7 @@ LayerTree::select_layer(synfig::Layer::Handle layer)
 
 		layer_tree_view().scroll_to_row(path);
 		layer_tree_view().get_selection()->select(iter);
+		layer_tree_view().set_cursor(path);
 	}
 }
 
--- a/synfig-studio/src/synfigapp/actions/layerencapsulate.cpp
+++ b/synfig-studio/src/synfigapp/actions/layerencapsulate.cpp
@@ -160,7 +160,7 @@ Action::LayerEncapsulate::prepare()
 	if(!child_canvas)
 		child_canvas=Canvas::create_inline(get_canvas());
 
-	Layer::Handle new_layer(Layer::create("group"));
+	new_layer=Layer::create("group");
 
 	if (!description.empty())
 		new_layer->set_description(description);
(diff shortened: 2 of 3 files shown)


synfig-studio/src/synfigapp/actions/layerencapsulate.cpp
```diff
@@ -246,3 +246,13 @@ Action::LayerEncapsulate::prepare()
...
+
+void
+Action::LayerEncapsulate::perform()
+{
+	Super::perform();
+	if(new_layer && get_canvas_interface()){
+		get_canvas_interface()->get_selection_manager()->clear_selected_layers();
+		get_canvas_interface()->get_selection_manager()->set_selected_layer(new_layer);
+	}
+}
```
rodolforg, line 258: I wonder if we have another cases where an action should select the correct layer. In this case, it would be good to create an (helper) action to select a layer and it would be used in the `prepare()` method.

@morevnaproject  do you know/recall another use case for this select proper layer after an action be performed beside this one of 'grouping layers'?

omda777:
Hi @rodolforg, I hope you're doing well.
I wanted to follow up and ask if there are any updates regarding the merge of this PR.

rodolforg:
I'll think about it this weekend :)

Svarov-RZM:
Hey @omda777 
I wanted to merge this PR but found a bug. If you still interested in working on it, here's what's wrong:
The selection after grouping highlights the proper layer but doesn't actually selects it.

How to reproduce:
* Create 3 Circe layers
*  Group 2 of them
* Press Enter key to rename it
* Type something and you will see that a layer below actually gets renamed
WrongSelection

omda777:
hi @Svarov-RZM 
I synced the tree cursor with the selected layer.

Svarov-RZM:
@omda777 OK, it works now, I didn't find any more issues. Thank you!
I am merging this PR because this behavior should have been there in the first place.

P.S. We can always address @rodolforg's feedback later if it turns out there are indeed cases where this functionality would be beneficial to have as a helper in order to not duplicate the same code everywhere.

Svarov-RZM (approved):

synfig-studio/src/gui/trees/layertree.cpp
```diff
@@ -426,9 +426,13 @@ LayerTree::select_layer(synfig::Layer::Handle layer)
...
 
 		Gtk::TreePath path(iter);
-		layer_tree_view().expand_to_path(path);
+		// do not auto expand when selecting a group layer to avoid "opening" it on selection
+		if(!(layer->get_name() == "group"))
+			layer_tree_view().expand_to_path(path);
+
 		layer_tree_view().scroll_to_row(path);
 		layer_tree_view().get_selection()->select(iter);
+		layer_tree_view().set_cursor(path);
```
rodolforg, line 435: Hey @omda777 !
Any idea why did you need to set cursor here?

morevnaproject: I think he is trying to solve this - https://github.com/synfig/synfig/pull/3660#issuecomment-4631178709

But yes, this seems to be a bad place for this line because if this regression - https://github.com/synfig/synfig/issues/3775

synfig-studio/src/synfigapp/actions/layerencapsulate.cpp
morevnaproject: I remember @ice0 implemented the fix which selects image layer right after it was imported.

Before this fix importing image wasn't selecting imported layer.

morevnaproject: Here - https://github.com/synfig/synfig/commit/69f9a693a1a997c7972e13daf2fa6c733ba5bed6

Issue #3654, opened 2025-12-15
Feature request: Ability to lock the Guides in Synfig


yoSachinkr:
**Is your feature request related to a problem? Please describe.**
Accidental movement of guides in Synfig can be a headache specially while creating artwork and requiring guides for precise placement of vertices.

**Describe the solution you'd like**
A feature to lock the guides effectively disabling the accidental movement of them could solve this problem.

BobSynfig:
Per guide or globally for all the guides?
It could be added in the Guide Editor (on right-click)

yoSachinkr:
@BobSynfig If you look at Inkscape's implementation, it has a simple lock icon at the left top corner to lock all the guides when needed.

<img width="144" height="113" alt="Image" src="https://github.com/user-attachments/assets/7ae7b693-c5ad-49da-af8f-fd90ff193566" />
And I think it's quite simple to use.

BobSynfig:
Currently, here we have the triangle button giving access to the menu as context menu but anyway it is accessible from anywhere with a right-click on the canvas, it is redundant.
This could be used instead.
And using a meta for its status as the other options of guides.

BobSynfig:
https://youtu.be/kL7n7XyqmF8

You can Lock/Unlock the Guides
- using the lock button (ex-menu button)
- using the meta
- using the context menu on the guides

If you delete all the Guides while locked, the lcon is deactivated automatically.

The menu is still displayed but using the right-click now.
It is a security to get back the main menu in case that it would be hidden on top of the app window.

Unofficial AppImage available on my GitHub for test

PR #3668, opened 2026-01-14
feat(guide_lock): Ability to lock the guides + a bit of refactoring

BobSynfig:
https://github.com/synfig/synfig/issues/3654

https://www.youtube.com/watch?v=kL7n7XyqmF8

You can Lock/Unlock the Guides
- using the lock button (ex-menu button)
- using the meta
- using the context menu on the guides

If you delete all the Guides while locked, the lcon is deactivated automatically.

The menu is still displayed but using the right-click now.
It is a security to get back the main menu in case that it would be hidden on top of the app window.

Unofficial AppImage available on my GitHub for test

### DIFF
--- a/synfig-studio/src/gui/workarea.cpp
+++ b/synfig-studio/src/gui/workarea.cpp
@@ -139,6 +139,7 @@ WorkArea::WorkArea(etl::loose_handle<synfigapp::CanvasInterface> canvas_interfac
 	show_rulers(true),
 	show_grid(false),
 	show_guides(true),
+	lock_guides(false),
 	background_size(15,15),
 	background_first_color(0.88, 0.88, 0.88),  /* light gray */
 	background_second_color(0.65, 0.65, 0.65),  /* dark gray */
@@ -221,9 +222,23 @@ WorkArea::WorkArea(etl::loose_handle<synfigapp::CanvasInterface> canvas_interfac
 	// Create the menu button
 
 	menubutton_box = manage(new Gtk::Button());
-	menubutton_box->set_image_from_icon_name("pan-end-symbolic");
+	menubutton_box->set_image_from_icon_name(get_lock_guides()?"changes-prevent-symbolic":"changes-allow-symbolic");
 	menubutton_box->set_relief(Gtk::RELIEF_NONE);
-	menubutton_box->signal_clicked().connect(sigc::mem_fun(*this, &WorkArea::popup_menu));
+	menubutton_box->add_events(Gdk::BUTTON_PRESS_MASK);
+	menubutton_box->signal_clicked().connect(sigc::track_obj([this](){
+							toggle_lock_guides();
+						 }, *this));
+
+	menubutton_box->signal_button_press_event().connect(
+		sigc::track_obj([this](GdkEventButton* event) -> bool {
+			if (event->type == GDK_BUTTON_PRESS && event->button == 3) {
+					popup_menu();
+					return true; 
+				}
+				return false; // Allow propagation
+			},
+			*this));
+
 	menubutton_box->show_all();
 
 	// Create scrollbars
@@ -282,6 +297,7 @@ WorkArea::WorkArea(etl::loose_handle<synfigapp::CanvasInterface> canvas_interfac
 	get_canvas()->signal_meta_data_changed("grid_snap").connect(sigc::mem_fun(*this,&WorkArea::load_meta_data));
 	get_canvas()->signal_meta_data_changed("grid_show").connect(sigc::mem_fun(*this,&WorkArea::load_meta_data));
 	get_canvas()->signal_meta_data_changed("status_ruler").connect(sigc::mem_fun(*this,&WorkArea::load_meta_data));
+	get_canvas()->signal_meta_data_changed("guide_lock").connect(sigc::mem_fun(*this,&WorkArea::load_meta_data));
 	get_canvas()->signal_meta_data_changed("guide_show").connect(sigc::mem_fun(*this,&WorkArea::load_meta_data));
 	get_canvas()->signal_meta_data_changed("guide").connect(sigc::mem_fun(*this,&WorkArea::load_meta_data));
 	get_canvas()->signal_meta_data_changed("background_rendering").connect(sigc::mem_fun(*this,&WorkArea::load_meta_data));
@@ -359,6 +375,7 @@ WorkArea::save_meta_data()
 	canvas_interface->set_meta_data("guide_snap", get_guide_snap() ? "1" : "0");
 	canvas_interface->set_meta_data("guide_show", get_show_guides() ? "1" : "0");
 	canvas_interface->set_meta_data("grid_show", show_grid ? "1" : "0");
+	canvas_interface->set_meta_data("guide_lock", get_lock_guides() ? "1" : "0");
 	canvas_interface->set_meta_data("status_ruler", show_rulers ? "1" : "0");
 	canvas_interface->set_meta_data("jack_offset", strprintf("%f", (double)jack_offset));
 	canvas_interface->set_meta_data("onion_skin", onion_skin ? "1" : "0");
@@ -405,23 +422,40 @@ WorkArea::have_meta_data()
 	data_size=canvas->get_meta_data("grid_size");
 	data_show=canvas->get_meta_data("grid_show");
 
-	if(data_size.empty() && !data_show.size())
+	if(data_size.empty() && data_show.empty())
 		return false;
 
 	return true;
 }
 
-void WorkArea::grab_focus()
+void
+WorkArea::grab_focus()
 {
 	if (drawing_area)
 		drawing_area->grab_focus();
 }
 
+bool
+WorkArea::get_bool_from_meta_data(const std::string &metaname, bool &value)
+{
+    String data = canvas->get_meta_data(metaname);
+    if (data.empty()) return false;
+    if (data == "1" || data[0] == 't' || data[0] == 'T') {
+	value = true;
+	return true;
+    }
+    if (data == "0" || data[0] == 'f' || data[0] == 'F') {
+	value = false;
+	return true;
+    }
+    return false; // Existing data but unknown format
+}
+
 void
 WorkArea::load_meta_data()
 {
-	// we need to set locale careful, without calling functions and signals,
-	// otherwise it can affect strings in GUI
+    // we need to set locale careful, without calling functions and signals,
+    // otherwise it can affect strings in GUI
     // ChangeLocale change_locale(LC_NUMERIC, "C");
 
     if(meta_data_lock)
@@ -429,6 +463,7 @@ WorkArea::load_meta_data()
 	meta_data_lock=true;
 
 	String data;
+        bool   bool_value; //for get_bool_from_meta_data
 
 	data=canvas->get_meta_data("grid_size");
 	if(!data.empty())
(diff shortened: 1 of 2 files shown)


Svarov-RZM:
I'll take a look on weekends if no one else is willing to.

synfig-studio/src/gui/workarea.cpp
```diff
@@ -221,9 +222,23 @@ WorkArea::WorkArea(etl::loose_handle<synfigapp::CanvasInterface> canvas_interfac
...
 
 	menubutton_box = manage(new Gtk::Button());
-	menubutton_box->set_image_from_icon_name("pan-end-symbolic");
+	menubutton_box->set_image_from_icon_name(get_lock_guides()?"changes-prevent-symbolic":"changes-allow-symbolic");
 	menubutton_box->set_relief(Gtk::RELIEF_NONE);
-	menubutton_box->signal_clicked().connect(sigc::mem_fun(*this, &WorkArea::popup_menu));
+	menubutton_box->add_events(Gdk::BUTTON_PRESS_MASK);
+	menubutton_box->signal_clicked().connect(sigc::track_obj([this](){
+							toggle_lock_guides();
+						 }, *this));
```
rodolforg, line 230: What about:
```c++
menubutton_box->signal_clicked().connect(sigc::mem_fun(*this, &WorkArea::toggle_lock_guides));
```

synfig-studio/src/gui/workarea.cpp
```diff
@@ -221,9 +222,23 @@ WorkArea::WorkArea(etl::loose_handle<synfigapp::CanvasInterface> canvas_interfac
 	// Create the menu button
 
 	menubutton_box = manage(new Gtk::Button());
-	menubutton_box->set_image_from_icon_name("pan-end-symbolic");
+	menubutton_box->set_image_from_icon_name(get_lock_guides()?"changes-prevent-symbolic":"changes-allow-symbolic");
```
rodolforg: please surround `?` and `:` with spaces XD

synfig-studio/src/gui/workarea.cpp
```diff
@@ -405,30 +422,48 @@ WorkArea::have_meta_data()
 	data_size=canvas->get_meta_data("grid_size");
 	data_show=canvas->get_meta_data("grid_show");
 
-	if(data_size.empty() && !data_show.size())
+	if(data_size.empty() && data_show.empty())
```
rodolforg, line 425: as you are touching this line, why don't add space after `if`? :D

synfig-studio/src/gui/workarea.cpp
```diff
@@ -405,30 +422,48 @@ WorkArea::have_meta_data()
...
+}
+
 void
 WorkArea::load_meta_data()
 {
-	// we need to set locale careful, without calling functions and signals,
-	// otherwise it can affect strings in GUI
+    // we need to set locale careful, without calling functions and signals,
+    // otherwise it can affect strings in GUI
     // ChangeLocale change_locale(LC_NUMERIC, "C");
```
rodolforg: Tabs

synfig-studio/src/gui/workarea.cpp
```diff
@@ -405,30 +422,48 @@ WorkArea::have_meta_data()
...
 
     if(meta_data_lock)
 		return;
 	meta_data_lock=true;
 
 	String data;
+        bool   bool_value; //for get_bool_from_meta_data
```
rodolforg: tabs

synfig-studio/src/gui/workarea.cpp
```diff
@@ -828,6 +822,15 @@ WorkArea::set_show_guides(bool x)
 	queue_draw();
 }
 
+void
+WorkArea::set_lock_guides(bool locked)
+{
+	lock_guides=locked;
+	menubutton_box->set_image_from_icon_name(locked?"changes-prevent-symbolic":"changes-allow-symbolic");
```
rodolforg: spaces surrounding `=` `?` and `:`

synfig-studio/src/gui/workarea.cpp
```diff
@@ -1356,19 +1359,37 @@ WorkArea::on_drawing_area_event(GdkEvent *event)
...
+				Gtk::MenuItem* item;
+
+				String lock_unlock_msg = get_lock_guides() ? _("_Unlock Guides") : _("_Lock Guides");
+				item = manage(new Gtk::MenuItem(lock_unlock_msg));
+				item->set_use_underline(true);
+				item->show();
+				item->signal_activate().connect(sigc::track_obj([this](){
+					toggle_lock_guides();
+				}, *this));
+				guide_menu->append(*item);
```
rodolforg: I think this item should be the second or the last one in the menu...

synfig-studio/src/gui/workarea.cpp
```diff
@@ -1356,19 +1359,37 @@ WorkArea::on_drawing_area_event(GdkEvent *event)
...
+				
 				item = manage(new Gtk::MenuItem(_("_Delete")));
 				item->set_use_underline(true);
 				item->show();
 				item->signal_activate().connect(sigc::track_obj([this](){
 					get_guide_list().erase(this->curr_guide);
+					//Unlock if there is no guide left
+					if (get_guide_list().empty()) {
+						set_lock_guides(false);
+					}
```
rodolforg, line 1387: As it is getting bigger, maybe create a method `delete_current_guideline()`

synfig-studio/src/gui/workarea.cpp
```diff
@@ -1356,19 +1359,37 @@ WorkArea::on_drawing_area_event(GdkEvent *event)
...
+				
+				Gtk::MenuItem* item;
+
+				String lock_unlock_msg = get_lock_guides() ? _("_Unlock Guides") : _("_Lock Guides");
+				item = manage(new Gtk::MenuItem(lock_unlock_msg));
+				item->set_use_underline(true);
+				item->show();
+				item->signal_activate().connect(sigc::track_obj([this](){
+					toggle_lock_guides();
+				}, *this));
```
rodolforg: What about:
```c++
tem->signal_activate().connect(sigc::mem_fun(*this, &WorkArea::toggle_lock_guides));
```

synfig-studio/src/gui/workarea.cpp
```diff
@@ -405,30 +422,48 @@ WorkArea::have_meta_data()
...
-void WorkArea::grab_focus()
+void
+WorkArea::grab_focus()
 {
 	if (drawing_area)
 		drawing_area->grab_focus();
 }
 
+bool
+WorkArea::get_bool_from_meta_data(const std::string &metaname, bool &value)
```
rodolforg: Prefer `Type& varname` rather than `Type &varname` (space position)

synfig-studio/src/gui/workarea.h
```diff
@@ -483,6 +493,8 @@ class WorkArea : public Gtk::Grid, public Duckmatic
 	void grab_focus();
 
 	Gtk::DrawingArea* get_drawing_area() { return drawing_area; }
+	
+	bool get_bool_from_meta_data(const std::string &metaname, bool &value);
```
rodolforg: Prefer Type& varname rather than Type &varname (space position)

synfig-studio/src/gui/workarea.h
```diff
@@ -393,6 +396,13 @@ class WorkArea : public Gtk::Grid, public Duckmatic
 	void set_show_guides(bool x);
 	//! Toggles the showing of the guides
 	void toggle_show_guides() { set_show_guides(!get_show_guides()); }
+	//! Returns the state of the lock_guides flag
+	bool get_lock_guides()const { return lock_guides; }
```
rodolforg: space before `const`

synfig-studio/src/gui/workarea.cpp
```diff
@@ -521,51 +556,18 @@ WorkArea::load_meta_data()
...
+	if (get_bool_from_meta_data("guide_lock",   bool_value)) set_lock_guides(bool_value);
+	if (get_bool_from_meta_data("grid_snap",    bool_value)) set_grid_snap(bool_value);
+	if (get_bool_from_meta_data("guide_snap",   bool_value)) set_guide_snap(bool_value);
+	if (get_bool_from_meta_data("onion_skin",   bool_value)) set_onion_skin(bool_value);
 
 	bool render_required = false;
 	data=canvas->get_meta_data("onion_skin_past");
-	if(data.size())
+	if(!data.empty())
 	{
```
rodolforg: space after `if` and pull `{` to the same line of `if`
"pull" `{` to the same line of `if`

synfig-studio/src/gui/workarea.cpp
```diff
@@ -578,7 +580,7 @@ WorkArea::load_meta_data()
 		}
 	}
 	data=canvas->get_meta_data("onion_skin_future");
-	if(data.size())
+	if(!data.empty())
 	{
```
rodolforg: space after `if` and pull `{` to the same line of `if`
"pull" `{` to the same line of `if`

rodolforg (changes requested):
I didn't test the code yet. Just read and analyzed the code changes and they look fine. :wink:

@morevnaproject it is changing the "arrow button" that pops up the menu. Now the "main menu" would be accessible by right clicking this "Lock Guide" button instead.

@Svarov-RZM as you'll test it, check if it is possible to create new guides when guidelines are locked. :smile:

Svarov-RZM:
> as you'll test it, check if it is possible to create new guides when guidelines are locked

I tested everything yesterday and all's fine. New guidelines are not possible to add while locked, but it is possible to delete them through right-clicking (not an issue).

It's combining menu and lock-guides into one button is something I am conflicted about. It's not intuitive now at all, who would think that in order to bring menu you need to right-click a lock button? You need to know this prior. Perhaps we should insert it along the grid buttons?
LockButtonTo

Do you think it would be better or am I nitpicking? The previous arrow icon for bringing up menu wasn't exactly user-friendly either...

yoSachinkr:
@Svarov-RZM We can already bring out the menu options by right clicking anywhere at the canvas. Having the guide lock icon at the top-left makes it look consistent with other vector graphics software such as Inkscape.

rodolforg:
> Having the guide lock icon at the top-left makes it look consistent with other vector graphics software such as Inkscape.

Maybe only Inkscape? XD

> New guidelines are not possible to add while locked, but it is possible to delete them through right-clicking (not an issue).

On Inkscape, by the way, it automatically unlocks the guidelines when creating a new one. I don't think it should unlock, though, but it should allow to create a new one IMO.

synfig-studio/src/gui/workarea.cpp
BobSynfig: Why to create a dedicated function for a code of a few lines that occurs only once in the source?
There is no point.
In this part I only added the unlocking on deletion of the last guide.
`delete_current_guideline()` would only replace `get_guide_list().erase(this->curr_guide);`
We have enough redundant code to refactor (like for `get_bool_from_meta_data`) instead of adding another spaghetti on the plate

synfig-studio/src/gui/workarea.cpp
BobSynfig: sigc::track_obj adds a layer of safety and catch better references in lambdas, it is even recommended to nest sigc::mem_fun inside sigc::track_obj.
It was from the beginning in the delete part, maybe we should use it more often.

Svarov-RZM:
> We can already bring out the menu options by right clicking anywhere at the canvas.

Only if there is an empty space available on canvas. 
I personally don't much care about menu button, it's just kind of wrong to combine two absolutely different functions in one button. Oh, and by the way, just because Inkscape and everyone else does it doesn't mean there's no point in being different.

I am letting @morevnaproject handle this.

synfig-studio/src/gui/workarea.cpp
mohamedAdhamc: Agreed.

mohamedAdhamc:
> 
> It's combining menu and lock-guides into one button is something I am conflicted about. It's not intuitive now at all, who would think that in order to bring menu you need to right-click a lock button? You need to know this prior. Perhaps we should insert it along the grid buttons?

I think this makes sense as well.

Svarov-RZM:
@BobSynfig I am not approving this until some agreement is reached on whether we should combine two actions in one icon (like it is in this PR) or separate the two. Maybe discuss it on forums?

BobSynfig:
No problem, it can be removed, I was simply keeping it for compatibility with previous versions (GIMP still implements it like before on Synfig).
It is indeed not needed.

Svarov-RZM:
Hey, @BobSynfig 
So I made a poll here: https://forums.synfig.org/t/where-the-lock-guides-icon-should-be-placed/17064 and according to it people are in favor of separating the Menu and Lock Guides icons. Can you move the Lock Guides icon to right toolbar where all other grid-related icons are located? I think it's better to do it here rather than in separate PR. I do plan on squashing commits since it's all related to one feature.

zaynopenapp (approved):

PR #3677, opened 2026-01-26
feat: add online update checker

036006:
Add asynchronous online update checker with stable/dev endpoints and menu notification entry highlighted for visibility; clicking opens landing URL.

Startup launches update checker only when enable_update_check (defaults to on) is true; new System setting toggle persists the preference.

Mark development snapshot builds via DEVELOPMENT_SNAPSHOT definition (CMake/autotools) and wire update checker into startup when enabled.

### DIFF
--- a/synfig-studio/src/gui/mainwindow.cpp
+++ b/synfig-studio/src/gui/mainwindow.cpp
@@ -51,7 +51,9 @@
 #include <gui/docks/dockmanager.h>
 #include <gui/exception_guard.h>
 #include <gui/localization.h>
+#ifdef ENABLE_UPDATE_CHECKER
 #include <gui/updatechecker.h>
+#endif
 #include <gui/widgets/widget_link.h>
 #include <gui/widgets/widget_time.h>
 #include <gui/widgets/widget_vector.h>
@@ -91,7 +93,10 @@ escape_underline(const std::string& raw)
 
 MainWindow::MainWindow(const Glib::RefPtr<Gtk::Application>& application)
 	: Gtk::ApplicationWindow(application),
-	  save_workspace_merge_id(0), custom_workspaces_merge_id(0), update_menu_item_(nullptr)
+	  save_workspace_merge_id(0), custom_workspaces_merge_id(0)
+#ifdef ENABLE_UPDATE_CHECKER
+	  , update_menu_item_(nullptr)
+#endif
 {
 	register_custom_widget_types();
 
@@ -301,6 +306,7 @@ MainWindow::toggle_show_toolbar()
 	}
 }
 
+#ifdef ENABLE_UPDATE_CHECKER
 void
 MainWindow::on_update_notification_clicked()
 {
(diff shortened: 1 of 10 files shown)


BobSynfig:
I'm not sure that there is really an added value for such a gadget according the frequency of updates ...
Let's hope also that is is not blocking in case of no access to the internet while activated like it was the case for some other software

036006:
The update check runs asynchronously in the background and does not block the UI or startup, even if there is no internet connection. It is triggered only once at application startup.

While updates may not be very frequent, this feature helps users stay informed without manually checking releases.

synfig-studio/src/gui/mainwindow.cpp
```diff
@@ -35,6 +35,9 @@
 #include <gui/mainwindow.h>
 
 #include <gtkmm/box.h>
+#include <gtkmm/stylecontext.h>
```
rodolforg: please move to below `#include <gtkmm/stock.h>` to trying to keep it sorted alphabetically.

ice0:
I think this option is only needed for Windows build.

Debian/RedHat/Arch/Brew (macOS)/Flatpak/Snap: it should be updated through repository

AppImage: I am not sure if this is needed.

Also I suggest to add an options to skip update and disable showing notifications.

morevnaproject:
This feature is requested by me. ) 

@ice0 There is an option to disable notifications (in Preferences).

ice0:
Proposed improvements following discussion with @morevnaproject:

Implement a build flag for the update notification mechanism to give **packagers** more control.
This is specifically useful for standalone Windows builds, macOS .app (excluding Homebrew/MacPorts), and AppImages.
Improve UX for update notifications:

- Add a "Never show this again" option directly in the Notification window, allowing users to disable update checks without searching through Preferences
- Add a "Skip this update" button to the notification window

036006:
Add ENABLE_UPDATE_CHECKER build flag for CMake and Autotools

Add a compile-time option to conditionally include the online update
checker functionality. This option is disabled by default in both
build systems.

CMake: option(ENABLE_UPDATE_CHECKER) in root CMakeLists.txt.
  When enabled, adds -DENABLE_UPDATE_CHECKER definition and compiles
  updatechecker.cpp into the synfigstudio target.

Autotools: --enable-update-checker flag in configure.ac.
  When enabled, defines ENABLE_UPDATE_CHECKER via AC_DEFINE and
  conditionally includes updatechecker sources via AM_CONDITIONAL.

All update checker integration points in the source code are wrapped
with #ifdef ENABLE_UPDATE_CHECKER:
- app.h/cpp: static members, settings get/set, preferences keys,
  start_async() call
- mainwindow.h/cpp: notification UI members, methods
- dialog_setup.h/cpp: preferences toggle widget, save/refresh logic

036006:
https://github.com/user-attachments/assets/0fee2912-ed95-41a9-835b-5d964fb9009b




On first launch, the user is prompted to choose whether update checks are enabled, instead of silently checking for updates or requiring manual preference configuration.

When the user clicks on an update notification, a dialog is shown with the following options:
Skip this update, Disable update check, or Download new version.

PR #3682, opened 2026-02-05
fix: gaussian blur vertical pass

### LINKED ISSUE #3631: The object gets cut off when Gaussian blur is applied.
If you create a blur and apply "Gaussian Blur", the object will be cut off at the frame boundaries.

Image

Bug reproduction:

   - Create an object. The object must extend beyond the frame boundaries.
   - Create a blur effect.
   - In the blur settings, select "Gaussian Blur".
   - Set the X-axis blur to a value between 0.1 and 20.9, and leave the Y-axis blur at 0.
   - The object will be cut off at the frame boundaries.

Additional information: 

The object will be cut off proportionally to the value set for X, from 0.1 to 20.9. If the X value is >= 21, the bug disappears. 

Image

Similarly, if you set Y = X, the bug also disappears.

Image

 If Y < X, the clipping will increase based on the difference between the X and Y values.

X=20, Y=5
Image

X=20, Y=15

Image

The bug occurs with all objects, images, and gradients.

Furthermore, if you leave the X-axis at 0 and increase the Y-axis, the blur will only become visible after the Y value is >= 21.2 . For some reason, if Y < 21.2, the blur is not applied to the object at all.

Y=5

Image

Y=21.1

Image

Y=21.2

Image

Tested on versions 1.5.1, 1.5.2, and 1.5.3 on the test branch. This bug is present in all these versions. Testing was conducted on both Linux and Windows systems.

omda777:
Fixes #3631

The object was being cut off because the vertical blur pass reused the row-based blur logic instead of applying the correct vertical behavior.

When  X = 20 and Y = 5

### before :
image

### After : 
image
<br><br><br><br>

when X =0 and  Y = 15 
### befor :  
The blur effect only becomes visible when Y ≥ 21.2; values below this produce no blur.
image

### After :
image

### DIFF
--- a/synfig-core/src/synfig/rendering/software/function/blur.cpp
+++ b/synfig-core/src/synfig/rendering/software/function/blur.cpp
@@ -288,7 +288,7 @@ software::Blur::blur_pattern(const Params &params)
 
 		for(Array<ColorReal, 3>::Iterator src_channel(arr_src_surface_cols), dst_channel(arr_dst_surface_cols); dst_channel; ++src_channel, ++dst_channel)
 			for(Array<ColorReal, 2>::Iterator sr(*src_channel), dr(*dst_channel); dr; ++sr, ++dr)
-				BlurTemplates::blur_pattern(*dr, *sr, arr_row_pattern);
+				BlurTemplates::blur_pattern(*dr, *sr, arr_col_pattern);
 	}
 
 	// copy result surface and restore alpha


Svarov-RZM:
Yeah, I'll take a look. Changes like that are quite scary because they can fix one thing but introduce a regression somewhere else. I will handle it on weekends, most likely, you know me.

ice0:
Yes, looks like copy-paste error.

Svarov-RZM:
Alright, the PR fixes blur getting cut off. Other tests show that blur computes a bit different but in a better way, results are more precise, the blur is distributed more even:
BlurTest

You can spot it on star layers specifically.

The PR's merged, thank you!

PR #3690, opened 2026-02-23
fix: improve preview dialog default quality and FPS

### LINKED ISSUE #3584: Correcting default parameters in Preview dialog
### Description

I started working on my next animation and the thing that bothers me is default parameters for Preview dialog don't make sense and thus I have to change them on every preview of my animation.

Let's talk Quality first. It is set to 0.5 which produces a very pixelated result, all the outlines get fuzzy, small details are lost. I do understand the point of preview but quality 0.5 generates output much worse than in canvas view. In my opinion it is just too low to be useful for previewing animation.

Now, let's talk FPS. It's set to half of the project rate, so if your project is 24 FPS, preview will default to 12 FPS. Which is an odd decision since this way preview can skip frames which will result in a different more rugged feel of the animation. Especially if there's some manual frame-by-frame animation, we definitely don't want to skip that.

### Proposal

Overall, my proposal is to change default settings in Preview dialog as follows:
* Quality = 1.0  -  a good balance between rendering time and quality of the image
* FPS = [project's rate]  -  there shouldn't be any frame skipping, it alters intended feel of animation

P.S. If you agree with my points, please add "easy" and "good first issue" labels.

omda777:
fix #3584 
updates the default values of the Preview dialog to provide a more accurate and visually consistent animation preview

image

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -3331,7 +3331,7 @@ CanvasView::on_preview_option()
 			if(!po)
 			{
 				po = Dialog_PreviewOptions::create();
-				po->set_fps(r.get_frame_rate()/2);
+				po->set_fps(r.get_frame_rate());
 				set_ext_widget("prevoptions",po);
 			}
 
--- a/synfig-studio/src/gui/resources/ui/preview_options.glade
+++ b/synfig-studio/src/gui/resources/ui/preview_options.glade
@@ -13,7 +13,7 @@
   <object class="GtkAdjustment" id="adj_zoom">
     <property name="lower">0.10000000000000001</property>
     <property name="upper">5</property>
-    <property name="value">0.5</property>
+    <property name="value">1</property>
     <property name="step_increment">0.10000000000000001</property>
     <property name="page_increment">0.20000000000000001</property>
   </object>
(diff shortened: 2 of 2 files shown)


Svarov-RZM:
Hey, thank you for addressing my issue! Changing those parameters always bugged me...
I would merge it right away, but we're talking about changing defaults, not everyone may agree with it. Let me think, maybe I should make a poll on Synfig forums first...

BobSynfig:
The question is what is the point to do this ?
A preview at (Quality 1 / FPS 24) is a full size and full quality render in cache (Quality 1 means size 100%). 
The preview default settings are on purpose low quality to speed up the process in order to have a rough idea (you can see that it is a render with the progression bar).
In the other hand, you can obtain the same type of preview in the workarea activating "Render future and past frames in background when enabled" when you want to preview and if you set bounds it pre-renders everything in this interval only.
And at least in the workarea, you can set "Low Res" and select a rendering mode to make it lighter!

omda777:
my  goal is simply to improve the user experience without negatively affecting performance or existing workflows.
I understand that not everyone may agree with changing the defaults, so I’m completely open to suggestions :)

Svarov-RZM:
OK, according to my poll here: https://forums.synfig.org/t/changing-preview-defaults-issue-3584/16967 the change in defaults is a bit problematic, there are some users who prefer the old ones. I am closing this PR, we'll approach this in a different way by making preview defaults configurable in GUI preferences.

PR #3697, opened 2026-02-28
feat: Add support for additional audio file extensions

### LINKED ISSUE #3689: feature request: import FLAC audio

I always use lossless compression for my audio, more specific FLAC

https://xiph.org/flac/

It would be great if I could use my FLAC files directly in Synfig by importing them using "File" -> "Import". Right now, I need to convert them to WAV since FLAC is not supported. There is no other way to use a lossless audio format in Synfig at the moment.

**Additional comments**

FLAC is an open source format available for macOS, Windows and Linux so it's very suitable to include in Synfig.

License:

https://xiph.org/flac/license.html

Image

omda777:
fixed #3689 
adds support for additional audio file extensions
now this extensions aviable in synfig (*.ogg, *.mp3, *.wav, *.flac, *.aac, *.m4a, *.opus).

### DIFF
--- a/synfig-studio/src/gui/app.cpp
+++ b/synfig-studio/src/gui/app.cpp
@@ -2339,9 +2339,18 @@ App::dialog_open_file_ext(const std::string& title, std::vector<synfig::filesyst
 	filter_supported->add_mime_type("audio/x-vorbis+ogg");
 	filter_supported->add_mime_type("audio/mpeg");
 	filter_supported->add_mime_type("audio/x-wav");
+	filter_supported->add_mime_type("audio/flac");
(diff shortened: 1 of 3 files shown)


Svarov-RZM:
The original code here is quite a mess , but refactoring it is out of scope of this PR.
I am currently trying to build Synfig with MLT support on my MSYS2 environment but it doesn't work... I am currently investigating what's wrong, hold on.

Svarov-RZM:
OK, I tested it on a very surface level since new MLT version (7.36.1) I built Synfig against has some problems (audio gets accelerated) but it should work fine with 6.6.0 version we're currently using.

Issue #3674, opened 2026-01-20
Feature request: Palette tab in Color picker dialog window


yoSachinkr:
**Is your feature request related to a problem? Please describe.**
Right now, if you've created a shape and want to change the color of the layer to a specific color that you have saved in the palette is, by using the fill tool or the otherwise long process of going to the color property in the properties panel, then open the color picker dialog windows, then select the color picker and use that color picker to select the color in the palette.

**Describe the solution you'd like**
The second process can be shortened and streamlined by adding and color palette tab in the color picker dialog as well.

<img width="665" height="452" alt="Image" src="https://github.com/user-attachments/assets/342c88a5-9651-4f65-8056-db50d9548d70" />

BobSynfig:
It would be more simple not to go in Properties at all to select the color from this dialog.
Simply by adding an entry to the context menu of the Color Palette panel with "Apply to the current layer"
It even would require less clicks

Otherwise, as the Color Palette already has the possibility to modify its colors, we risk to fall in a recursive loop and crash.

menna1261:
Do you mean adding a fourth entry here ? 
I'd like to work on this issue and I need to make sure I'm getting it right @BobSynfig 

<img width="544" height="268" alt="Image" src="https://github.com/user-attachments/assets/0d7a1d2c-d0fb-495e-b3f8-5da2e4ac9175" />

BobSynfig:
This is what I meant, it would be more logical.
Note that the first option itself opens a Color picker dialog!

As always, I would prefer people to discuss about it first on the forums instead of opening lots of feature request while GitHub Issues is first for Issues, then for (validated/pertinent) feature requests.
I suggest, I bring arguments but I'm not the one taking decisions for the direction to take.

Link to the discussion in the forums: https://forums.synfig.org/t/16974

menna1261:
@BobSynfig @rodolforg 
Can I start implementing the 4th entry in the context menu ?

Svarov-RZM:
Implemented in https://github.com/synfig/synfig/pull/3707

PR #3704, opened 2026-03-06
feat(palette): add "Apply to selected layer" option in palette context menu

menna1261:
#3674 

This suggestion was proposed here : https://forums.synfig.org/t/16974

I Added a new "Apply to selected layer" entry to the color palette context menu. Right-clicking any color in the palette now shows this option, which directly sets the selected layer's color parameter to the chosen palette color in a single click.

**How it looks now :** 

Note : I didn't now what icon to add so I just put the color icon :)

https://github.com/user-attachments/assets/9201f8b5-f938-48a0-a26d-f9b0333b8248

### DIFF
--- a/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
+++ b/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
@@ -33,7 +33,7 @@
 #ifdef HAVE_CONFIG_H
 #	include <config.h>
 #endif
-
+#include <canvasview.h>
 #include "dock_paledit.h"
 
 #include <errno.h>
@@ -380,7 +380,14 @@ Dock_PalEdit::show_menu(int i)
 			i ));
 	menu->append(*item);
 
-	menu->popup(3,gtk_get_current_event_time());
+	item = image_menu_item("type_color_icon", _("Apply to selected layer"), true);
+	item->signal_activate().connect(
+		sigc::bind(
+			sigc::mem_fun(*this,&studio::Dock_PalEdit::apply_color_to_selected_layer),
+			i ));
+	menu->append(*item);
+
+	menu->popup(4,gtk_get_current_event_time());
 }
 
 int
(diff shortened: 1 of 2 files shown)


BobSynfig:
What if there are several selected layers?
What if it (or one/some of them) doesn't have the color property to be applied to?

menna1261:
> What if there are several selected layers? What if it (or one/some of them) doesn't have the color property to be applied to?

if a layer doesn't contain a color property like a skeleton layer , it shows this message : 

image



and if there are several selected layers, it only colors the top one in the layer stack (because in the code `get_selected_layer() `returns `layers.front()` ) should I make the color get applied to all selected layers at once?



if some of the layers don't have a color property, they don't get affected unless they are at the top, it show the dialog " layer didn't accept parameter"

what do you think should be the correct behavior ?

menna1261:
I updated the code to get all selected layers and iterate over them to apply the same color :

https://github.com/user-attachments/assets/5ad8cc71-4504-4895-b7ff-38c7c0004874


and if (one / some) of the layers don't have a color property it shows the dialog first and then colors the rest :

https://github.com/user-attachments/assets/813369c7-f732-4cf1-8046-b0373045b569



@BobSynfig

BobSynfig:
I would say that it should iterate on all the selected layers
check if there is a Color property for each of them before to apply the action
or at least
avoid to return an error message that seems not so much useful and only make worse the user experience (better fail silently)
but still apply it on Layers that are compatible

menna1261:
> I would say that it should iterate on all the selected layers check if there is a Color property for each of them before to apply the action or at least avoid to return an error message that seems not so much useful and only make worse the user experience (better fail silently) but still apply it on Layers that are compatible

I added a condition to check if the layer has a color property, if it doesn't -> skip 
it looks like this now : 

https://github.com/user-attachments/assets/98ec0c03-7353-47e7-8e4f-a5845c5a08a2

menna1261:
Can you please review it now ? @rodolforg @ice0 @BobSynfig

menna1261:
Can you please check this out ? @Svarov-RZM

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -370,7 +370,14 @@ Dock_PalEdit::show_menu(int i)
...
 
-	menu->popup(3,gtk_get_current_event_time());
+	item = image_menu_item("type_color_icon", _("Apply to selected layer"), true);
+	item->signal_activate().connect(
+		sigc::bind(
+			sigc::mem_fun(*this,&studio::Dock_PalEdit::apply_color_to_selected_layer),
+			i ));
+	menu->append(*item);
+
+	menu->popup(4,gtk_get_current_event_time());
```
rodolforg: It doesn't make any difference, but the change from 3 to 4 on this method call is wrong :(
The first parameter here should be button index (3 means right mouse button).
Anyway, this method signature is deprecated since 3.22, so maybe we should fix it too:
```c++
#if GTK_CHECK_VERSION(3, 22, 0)
	menu->popup_at_pointer(nullptr);
#else
	menu->popup(3, gtk_get_current_event_time());
#endif
```

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -529,6 +536,50 @@ Dock_PalEdit::select_outline_color(int i)
...
+		return;
+
+    // Get ALL selected layers 
+    synfigapp::SelectionManager::LayerList layers = canvas_view->canvas_interface()->get_selection_manager()->get_selected_layers();
+    if (layers.empty()) 
+		return;
+
+    synfigapp::Action::PassiveGrouper group(canvas_view->canvas_interface()->get_instance().get(), _("Apply palette color"));
+
+    for (synfig::Layer::Handle layer : layers)
```
rodolforg: For (minor) performance reasons, use reference instead of value here:
```c++
for (synfig::Layer::Handle& layer : layers) {
```
And, as you see, prefer open curly bracket `{` inline for `for`, `while`, `if`, `do`, `switch` states and lambda.

zaynopenapp (approved):

rodolforg (approved):

rodolforg:
Merged. Thank you.

PR #3707, opened 2026-03-08
feat(palette): add palette tab to color picker dialog

menna1261:
#3674

This suggestion was proposed here : https://forums.synfig.org/t/16974

- I Added a new Palette tab to the Color Picker dialog (`Widget_ColorEdit`) that displays the current palette colors inline.
- The tab reuses the same `Widget_Color` cells used by the Palette panel (`Dock_PalEdit`). 
- A static instance pointer was added to `Dock_PalEdit` so that `Widget_ColorEdit` can access the currently loaded palette directly. This means any changes made in the Palette panel (loading a new palette, adding or deleting colors) are automatically reflected in the Palette tab the next time it is opened.
- Left-clicking a color sets it as the current color in the Color Picker and updates the fill color in real time. Right-clicking shows a small context menu with Set as Fill Color option. 
- Editing and deleting palette colors is intentionally not available in this tab to keep it in simple selection mode, as suggested by @rodolforg 

**How it looks now :** 

https://github.com/user-attachments/assets/7adbca35-3c7b-4223-8f01-b00af7fa2307

### DIFF
--- a/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
+++ b/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
@@ -127,7 +127,9 @@ Dock_PalEdit::Dock_PalEdit():
 	//palette_settings(new PaletteSettings(this,"colors")),
 	table(2,2,false)
 {
-	instance = this;
+	assert(!instance && "Only one instance of Dock_PalEdit should exist.");
+    instance = this;
+	
 	// Make Palette Editor toolbar buttons small for space efficiency
 	get_style_context()->add_class("synfigstudio-efficient-workspace");
 
@@ -231,7 +233,9 @@ Dock_PalEdit::Dock_PalEdit():
 
 Dock_PalEdit::~Dock_PalEdit()
 {
+	instance = nullptr;
 	//delete palette_settings;
+
 }
 
 void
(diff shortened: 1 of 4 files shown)


synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -125,6 +127,9 @@ Dock_PalEdit::Dock_PalEdit():
 	//palette_settings(new PaletteSettings(this,"colors")),
 	table(2,2,false)
 {
+	assert(!instance && "Only one instance of Dock_PalEdit should exist.");
+    instance = this;
```
ice0: I suggest using the Singleton approach here.

ice0:
@menna1261 

Looks good, please check other comments

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
menna1261: Please check the latest commit, I added a get_instance() function to access the instance pointer (made it private), is that what you mean ? or should i make it a whole singleton pattern with a private constructor ?

menna1261:
Can you please check this out @Svarov-RZM

Svarov-RZM:
> Can you please check this out

Yeah, I can take a look. I am not good at C++ though, I am more of a tester than a developer.
So let me test it, if I find no issues, I guess it would be safe to merge,

Svarov-RZM (approved):

Svarov-RZM:
Alright, I tested the PR and everything works as expected.
I see that ice's points were addressed, so I think it's safe to merge.

Thank you for contribution!

Issue #259, opened 2016-10-21
Ability to change playing speed


caryoscelus:
It's sometimes useful to quickly change playing speed to see how it looks before making actual changes. It would be nice to be able to change the speed, at least in preview window.

PR #3723, opened 2026-03-26
feat: add playback speed option for preview

veermetri05:
Added the ability to set playback speed in the preview window, allowing playing the rendered animation at different speed.

Issue: https://github.com/synfig/synfig/issues/259

image

### DIFF
--- a/synfig-studio/src/gui/preview.cpp
+++ b/synfig-studio/src/gui/preview.cpp
@@ -655,7 +655,7 @@ bool Widget_Preview::on_playback_input_focus_out(GdkEventFocus* event)
 void studio::Widget_Preview::update_playback() {
 	Glib::ustring text = playback_input->get_text();
 	try {
-		float value;
+		double value;
 		value = std::stod(text);
 		if(value <= 0) {
 			playback_speed = 1;
@@ -663,6 +663,7 @@ void studio::Widget_Preview::update_playback() {
 			playback_speed = value;
 		}
 	} catch(const std::exception& e) {
+		// when input can't be converted to double, revert to previous value.
 	}
 	char buf[16];
 	snprintf(buf, sizeof(buf), "%.2fx", playback_speed);
--- a/synfig-studio/src/gui/preview.cpp
+++ b/synfig-studio/src/gui/preview.cpp
@@ -344,6 +344,7 @@ Widget_Preview::Widget_Preview():
 	scr_time_scrub(adj_time_scrub),
 	b_loop(/*_("Loop")*/),
 	currentindex(-100000),//TODO get the value from canvas setting or preview option
+	playback_speed(1.00),
 	timedisp(-1),
 	//audiotime(0),
 	adj_sound(Gtk::Adjustment::create(0, 0, 4)),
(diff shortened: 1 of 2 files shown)


synfig-studio/src/gui/preview.cpp
```diff
@@ -631,6 +645,33 @@ void studio::Widget_Preview::update()
...
+		}
+	} catch(const std::exception& e) {
+	}
+	char buf[16];
+	snprintf(buf, sizeof(buf), "%.2fx", playback_speed);
+	playback_label->set_text(buf);
+	playback_input->set_text(Glib::ustring(buf));
+	play_button->grab_focus();
+	update();
+}
```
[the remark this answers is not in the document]
veermetri05: Adding x in the input field itself is inteneded behaviour as there is no other visual hint for the user to understand that it is related to playback speed.
Type consistency should be followed, made changes.
Reverting back to previous value is intended behaviour, adding comment to make it clear.

Svarov-RZM:
Hey, this is nice feature! I have some feedback though:

* First, we don't need to display current speed in status bar, we already have it in the text box, it doesn't serve purpose
* Second, I don't like the speed parameter location. Why putting it before play button? I think it should be somewhere near the zoom level since changing speed is not something user would often do
* And lastly, I think it would be better to make text box into a GtkEntry like Zoom level. We can use the following presets: 0.25x, 0.5x, 1.0x (our default), 1.5x, 2.0x. For the icon we can use the one for animation mode

Here's a mock-up:
SpeedC

@veermetri05 What do you think? The size of the GtkEntry box can be of course made smaller than in mock-up.

veermetri05:
Thanks for the feedback.
I agree with your suggestions, and will make changes to the code soon.

synfig-studio/src/gui/preview.cpp
```diff
@@ -502,6 +504,26 @@ Widget_Preview::Widget_Preview():
...
+	speed_preview.append("0.25x");
+	speed_preview.append("0.5x");
+	speed_preview.append("1.0x");
+	speed_preview.append("1.5x");
+	speed_preview.append("2.0x");
+	Gtk::Entry* speed_entry = speed_preview.get_entry();
+	speed_entry->set_text("1.0x"); //default playback speed
+	speed_entry->set_icon_from_icon_name("animate_mode_on_icon");
+	speed_entry->signal_activate().connect(sigc::mem_fun(*this, &Widget_Preview::update_playback));
+	speed_entry->signal_focus_out_event().connect(sigc::mem_fun(*this, &Widget_Preview::on_playback_speed_focus_out), false);
```
rodolforg, line 517: Does this event really need to be caught?

synfig-studio/src/gui/preview.cpp
```diff
@@ -502,6 +504,26 @@ Widget_Preview::Widget_Preview():
 
 	toolbar->pack_end(zoom_preview, Gtk::PACK_SHRINK, 0);
 
+	//playback speed
+	speed_preview.append("0.25x");
+	speed_preview.append("0.5x");
+	speed_preview.append("1.0x");
+	speed_preview.append("1.5x");
+	speed_preview.append("2.0x");
```
rodolforg, line 512: What about adding a tooltip for the combobox?

synfig-studio/src/gui/preview.cpp
```diff
@@ -631,6 +653,34 @@ void studio::Widget_Preview::update()
 	l_currenttime.set_text(timecode);
 
 }
+
+bool Widget_Preview::on_playback_speed_focus_out(GdkEventFocus* event)
+{
+    update_playback();
+    return false;
```
rodolforg, line 660: style: `tab` for indentation

synfig-studio/src/gui/preview.cpp
```diff
@@ -631,6 +653,34 @@ void studio::Widget_Preview::update()
...
+}
+
+void studio::Widget_Preview::update_playback() {
```
rodolforg, line 663: style: for functions, procedures, methods and classes, please let the opening curly bracket `{` in the next line.
And the return type in the previous line.

synfig-studio/src/gui/preview.cpp
```diff
@@ -631,6 +653,34 @@ void studio::Widget_Preview::update()
...
+	Gtk::Entry* entry = speed_preview.get_entry();
+	Glib::ustring text = entry->get_text();
+	try {
+		double value;
+		value = std::stod(text);
+		if(value <= 0) {
```
rodolforg, line 669: style: space between `if`, `while`, `for` and the opening parenthesis `(`

synfig-studio/src/gui/preview.cpp
```diff
@@ -631,6 +653,34 @@ void studio::Widget_Preview::update()
...
+			playback_speed = 1;
+		} else {
+			playback_speed = value;
+		}
+	} catch(const std::exception& e) {
+		// when input can't be converted to double, revert to previous value.
+	}
+	char buf[16];
+	snprintf(buf, sizeof(buf), "%.2fx", playback_speed);
```
rodolforg, line 678: you could use our `synfig::strprintf()` ;)

rodolforg (changes requested):

PR #3724, opened 2026-03-26
feat: [GSoC'26]  Initial MVP for 2D Free-Form Deformation Layer

### LINKED ISSUE #3792: Feature: FFD Project
### Architecture and ownership

| File | Responsibility |
| --- | --- |
| `synfig-studio/src/gui/states/state_ffd.cpp` | Tool state, options, previews, creation, editing, actions, and coordinate conversion |
| `synfig-studio/src/gui/states/state_ffd.h` | FFD tool state declaration |
| `synfig-studio/src/gui/duckmatic.cpp` | Persistent FFD point and edge overlay outside structural editing |
| `synfig-core/src/synfig/layers/layer_freeformdeform.*` | Parameters, image-analysis helpers, interpolation, triangulation, and render-mesh construction |
| `synfig-core/src/synfig/layers/layer_meshtransform.*` | Generic source-to-destination triangle mapping and render task |

The Studio layer decides *what the user is editing*. The core layer decides
*how persisted data becomes a renderable mesh*. Geometry needed by both preview
and rendering lives on `Layer_FreeFormDeform` as static helpers.

### Developer phase 1: Data model and invariants

| Parameter | Type and role |
| --- | --- |
| `control_points` | List of destination control points; connected to `ValueNode_DynamicList` for animation |
| `grid_size_x`, `grid_size_y` | Grid dimensions; each must be at least two |
| `smoothness` | Grid interpolation blend, clamped to `[0, 1]` during mesh preparation |
| `source_tl`, `source_br` | Opposite corners of the undeformed Grid source rectangle |
| `source_angle` | Rotation of the Grid source rectangle |
| `source_points` | Undeformed Custom Mesh points |
| `mesh_mode` | `0` for Grid and `1` for Custom Mesh |
| `triangles` | Flat integer list; each consecutive triple is one Custom Mesh triangle |
| `cull_threshold` | Circumradius threshold used by the alpha-shape fallback/render path |
| `auto_mesh_margin` | Circular alpha-mask dilation radius in raster pixels |
| `auto_mesh_edge_length` | UI mesh-density value despite the historical parameter name |
| `auto_mesh_dpi` | Controls retracing resolution for an existing FFD layer |

Required invariants:

- Grid mode requires at least `grid_size_x * grid_size_y` destination points in
  row-major order.
- Custom Mesh mode requires `source_points.size() == control_points.size()` and at
  least three points.
- Every persisted triangle index must be non-negative and smaller than both
  Custom Mesh point lists.
- Source points describe texture coordinates; destination points describe
  world positions after deformation.
- A topology operation must update points and every affected triangle index as
  one logical change.

`set_param()` calls `prepare_mesh()` for parameters that affect geometry.
Auto-mesh settings are metadata until Studio regenerates points and triangles.

### Developer phase 2: Target validation and group creation

`group_has_image()` recursively searches nested `Layer_PasteCanvas` layers for
a `Layer_Bitmap`. `is_valid_group_for_ffd()` accepts only layers named `group`
or `switch` that pass that search.

`on_make_ffd_pressed()` groups its actions with `Action::PassiveGrouper`. It
creates a wrapper Group, reparents the selected target, transfers dynamic or
static `origin` and `transformation` values to the wrapper, resets those values
on the child, and inserts `free_form_deform` at depth zero in the child canvas.
This keeps the artwork visually stationary while placing FFD in the correct
render context.

### Developer phase 3: Bounds and oriented Grid fitting

There are two related bounds paths. Having dual implementations is an architectural necessity driven by the separation of concerns between `synfig-core` and `synfig-studio`:

- **Core backend:** Requires a fast, mathematical fallback (axis-aligned bounds) to initialize layers headlessly (e.g., when loading `.sifz` files), as it lacks access to UI rendering pipelines.
- **Studio frontend:** Leverages viewport access to run expensive pixel-analysis (PCA-based oriented bounds), providing artists with a high-quality, tight grid fit during interactive creation.

#### Core context bounds

`get_context_bounds()` walks to the context below the FFD layer and unions all
finite, non-infinite layer bounds with area below `1e10`. PasteCanvas layers use
their context-dependent bounding rectangles. If no finite layer is found, the
canvas render rectangle is used.

`compute_grid_for_bounds()` distributes points uniformly over the resulting
axis-aligned rectangle. `on_canvas_set()` uses this path when a newly inserted
layer still needs its initial grid.

#### Studio oriented preview fit

`update_grid_preview()` analyzes a `128 x 128` alpha render of the selected
group. Pixels with alpha at most `0.05` are ignored. For accepted pixels it
computes an alpha-weighted centroid:

```text
cx = sum(x * alpha) / sum(alpha)
cy = sum(y * alpha) / sum(alpha)
```

It then computes the weighted covariance terms `xx`, `yy`, and `xy`. The
principal eigenvector of the `2 x 2` covariance matrix gives the dominant image
axis. This is equivalent to a small principal-component analysis (PCA).

Every accepted pixel is projected onto the principal axis `u` and its
perpendicular axis `v`. Minimum and maximum projections define an oriented
rectangle. The rectangle is expanded by 5 percent, then by Grid Margin, rotated
back to canvas orientation, and transformed through the selected group's full
ancestor transform stack.

This fit follows elongated or rotated artwork better than an axis-aligned
bounding box. Symmetric or nearly square artwork has no strong principal axis,
so its chosen orientation can be less meaningful.

### Developer phase 4: Alpha-mask image processing

Custom Mesh creation begins in `cache_auto_mesh_surface()` or
`render_context_below()`:

1. Determine finite content bounds.
2. Expand each side by 10 percent to avoid clipping the silhouette.
3. Preserve aspect ratio and rasterize with a longest side of 512 pixels during
   creation. Existing-layer regeneration derives a resolution from
   `auto_mesh_dpi`, clamped to `[64, 1024]`.
4. Render with antialias level one into `synfig::Surface`.

`generate_contour_polygon()` then performs these image-processing steps.

#### Binary alpha mask

Each pixel becomes foreground when `alpha > 0.05`; color channels are ignored.
The result is a binary mask.

#### Circular morphological dilation

For every foreground pixel, all pixels inside radius `margin` are marked. The
kernel test is:

```text
dx^2 + dy^2 <= margin^2
```

This is binary dilation with a circular structuring element. It expands the
mesh away from the artwork and closes transparent gaps smaller than the chosen
radius. The straightforward implementation is proportional to mask size times
the kernel area, so large margins cost more.

#### Moore-neighborhood contour tracing

The first foreground pixel in row-major order becomes the start. The tracer
walks its eight neighboring pixels, changing the search direction according to
the previous movement, until it returns to the start or reaches the safety cap
`width * height * 4`.

Pixel contour coordinates are mapped to canvas coordinates with:

```text
canvas_x = bounds.minx + (pixel_x / width) * bounds.width
canvas_y = bounds.maxy - (pixel_y / height) * bounds.height
```

The Y term is inverted because raster Y increases downward while canvas Y
increases upward.

The current tracer returns one connected outer contour. Separate alpha islands
and holes are not represented as independent contour loops.

### Developer phase 5: Boundary and interior point generation

`generate_edge_points()` samples the dense traced contour. Its target spacing
is based on the largest bounds dimension:

```text
target_spacing = bounds_size / (mesh_strength * 8)
minimum_spacing = bounds_size / 300
```

Walking distance accumulates along the contour until it reaches the target,
then that contour point is emitted. Higher Mesh Strength therefore means a
smaller target spacing and more boundary points, capped at roughly 300
divisions by the minimum spacing.

`generate_interior_points()` adds Steiner points so the interior is not formed
only from long triangles between boundary vertices. It computes:

```text
step = contour_bounds_size / (mesh_strength * 2)
minimum_step = contour_bounds_size / 25
```

Candidates are generated on a staggered row grid: odd rows shift horizontally
by half a step. A ray-casting even-odd test keeps only candidates inside the
contour. The stagger reduces directional bias compared with a square-aligned
grid. The minimum step caps density at roughly 25 divisions per axis.

### Developer phase 6: Delaunay triangulation and filtering

`triangulate()` implements incremental Bowyer-Watson Delaunay triangulation:

1. Compute the point bounding box and create a super-triangle much larger than
   the point set.
2. Insert each user point in order.
3. Find all existing triangles whose circumcircle contains the new point.
4. Remove those triangles. Their non-shared edges form the polygonal cavity.
5. Connect every cavity boundary edge to the new point.
6. Remove triangles touching a super-triangle vertex.
7. Normalize triangle winding for rendering.

The circumcenter determinant uses a `1e-6` degeneracy guard. The containment
test includes a `1e-10` relative tolerance to reduce floating-point boundary
errors.

Delaunay triangulation fills the convex hull, which is wrong for a concave
image silhouette. `filter_triangles_by_polygon()` addresses this by computing
each triangle centroid and retaining it only when the even-odd ray-casting test
places that centroid inside the traced contour. This centroid-in-polygon (CIP)
filter is inexpensive, but a triangle can still cross a narrow concavity even
when its centroid lies inside.

When no contour is available during editing, `cull_triangles()` is the fallback.
It removes triangles whose circumradius is greater than a threshold derived
from half the point bounding-box diagonal. This alpha-shape-style heuristic
primarily rejects long sliver triangles.

### Developer phase 7: Grid interpolation

Each logical Grid cell is subdivided into a fixed `10 x 10` render mesh. Every
subcell vertex has local coordinates `(u, v)` in `[0, 1]`.

#### Bilinear interpolation

For cell corners `P00`, `P10`, `P01`, and `P11`, destination position is:

```text
B(u,v) = P00(1-u)(1-v) + P10*u(1-v) + P01(1-u)*v + P11*u*v
```

This interpolation stays within the four-corner influence of one cell and is
the complete result when `smoothness == 0`.

#### Bicubic Catmull-Rom interpolation

For smooth deformation, `catmull_rom_1d()` evaluates four scalar samples:

```text
C(t) = 0.5 * [
  (-t^3 + 2t^2 - t)P0 +
  (3t^3 - 5t^2 + 2)P1 +
  (-3t^3 + 4t^2 + t)P2 +
  (t^3 - t^2)P3
]
```

The implementation evaluates four neighboring control points along `u` for
each of four rows, then evaluates those four row results along `v`. This
separable evaluation produces a bicubic surface.

Cells at the lattice boundary do not have all 16 neighbors. Instead of simply
duplicating edge points, `get_clamped_ctrl_point()` linearly extrapolates them:
`Poutside = 2 * Pedge - Padjacent`. This keeps undeformed outer edges straight.

The final destination is:

```text
destination = B * (1 - smoothness) + C * smoothness
```

Source texture positions come from uniform points between `source_tl` and
`source_br`, rotated by `source_angle`. Thus moved control points affect only
the destination side of the source-to-destination mapping.

`get_interpolated_grid()` uses the same bilinear/Catmull-Rom blend to sample an
existing control lattice when its dimensions change. If source data is invalid,
it falls back to a fresh uniform grid built from the saved rectangle and angle.

### Developer phase 8: Custom Mesh rendering and stabilization

In Custom Mesh mode, `prepare_mesh()` pairs each destination `control_points[i]`
with source texture coordinate `source_points[i]`. It uses persisted topology
when available and retriangulates only when no valid flat triangle list exists.

`prepare_mesh()` also appends fixed identity points intended to stabilize the
area outside the artist's contour:

1. Compute the source-point axis-aligned bounding box.
2. Add four fence points at 50 percent padding, with a minimum padding of
   `0.05` canvas units.
3. Add eight outer anchors around `source_tl` and `source_br`, expanded by one
   canvas unit.
4. Add each anchor to both source and destination lists at the same position,
   so those vertices do not deform.

When no explicit triangle list exists, fallback triangulation includes these
anchors and connects the user mesh to the stable exterior. When explicit
topology is persisted, that topology is used as-is; appended anchors not
referenced by its triangles remain disconnected. Preserve this distinction
when changing clipping or stabilization behavior.

If `cull_threshold` rejects a user-only triangle in the core render path, the
triangle is not left as a hole. Its three source vertices are duplicated as
fixed destination vertices, freezing that triangle while allowing adjacent
accepted triangles to deform.

Grid mode uses a related stabilization scheme. It expands bounds to include
all deformed vertices, adds at least 50 percent padding with a one-unit minimum,
then creates identity border strips and corner triangles around all four Grid
edges. Both modes create a contour mask covering the stabilized mesh region.

### Developer phase 9: Mesh mapping and render task

`Layer_MeshTransform` interprets each mesh vertex as:

- `position`: destination/world coordinate.
- `tex_coords`: source coordinate sampled from the context below.

`transform_coord_world_to_texture()` finds the destination triangle containing
a rendered pixel and maps it barycentrically to the corresponding source
triangle. `transform_coord_texture_to_world()` provides the inverse direction
for transforms and tool interaction.

The render task clones the context below, combines it with the mask, and passes
it through `TaskMesh`. `get_color()` follows the same world-to-texture mapping
for the software color path and blends the sampled source according to layer
amount and blend method.

### Developer phase 10: Coordinate spaces and nested groups

FFD creation uses three spaces:

- **Raster space:** temporary surface pixels used for alpha analysis.
- **Canvas/world space:** preview ducks and mouse events.
- **Selected-group local space:** persisted source and destination points.

`find_transform_stack_to_layer()` recursively walks nested PasteCanvas layers
and accumulates their summary transformation matrices. Preview generation
applies the selected group's origin, scale, skew, rotation, and offset, then the
ancestor stack. Persistence performs the inverse operations in reverse order:
ancestor unperform, subtract offset, inverse rotate, inverse skew, inverse
scale, then restore origin.

Changing this order changes the result. New code must preserve the exact
transform composition order and guard zero scale before division.

### Developer phase 11: Topology-editing algorithms

`edit_triangles_` is the temporary explicit topology while structural editing
is active.

#### Add point

`insert_point_locally()` first uses a barycentric sign test to find a containing
triangle. If found, it replaces that triangle with three triangles connected to
the new point.

If the point is outside the mesh, the function builds an edge-to-triangle map.
Edges referenced by one triangle are boundary edges. It finds boundary edges
whose outward normal faces the new point, chooses the closest, then walks
adjacent facing edges while their distance stays within twice the closest
squared distance. Connecting this contiguous boundary run to the new point
forms a local tent without retriangulating the entire mesh.

#### Remove point

Right-clicking a point erases it from the temporary point list and calls
`recompute_edit_triangles()`. Full Delaunay triangulation is filtered by the
current image contour. If filtering leaves orphan points, triangles touching
those orphans are added back until the points participate in topology.

#### Split edge

The clicked interpolation value is clamped to `[0.1, 0.9]` to avoid placing a
new vertex almost on an endpoint. Every triangle sharing the edge is replaced
by two triangles.

#### Swap edge

Swap is valid only when exactly two triangles share the edge. Their opposite
vertices become the new diagonal, creating two replacement triangles.

#### Collapse edge

The lower index is retained and moved to the endpoint midpoint. References to
the removed index are rewritten, triangles with repeated vertices are removed,
the point is erased, and all greater indices are decremented.

#### Subdivide adjacent triangles

For each triangle sharing the selected edge, a new centroid point is appended
and the triangle is replaced by three triangles. If two triangles share the
edge, each receives its own centroid.

### Developer phase 12: Tool state, actions, and overlays

Entering `StateFFD_Context` installs Tool Options, enables layer clicks, changes
the cursor, connects parameter signals, and builds preview ducks. Selection
changes update controls and previews only while FFD Tool is active.

Selecting an FFD layer from another tool must not activate FFD Tool. Creation
explicitly sends `EVENT_STOP`, returns to Transform Tool, and selects the new
layer asynchronously after the layer actions complete.

Creation and structural updates use Synfig actions and `PassiveGrouper` so they
can participate in undo. Preview-only point removal and restoration use
`polygon_point_list` and `redo_point_list`; that local history is separate from
the document undo stack until changes are persisted.

Outside structural editing, `Duckmatic::get_ffd_overlay_config()` reads
`mesh_mode`, Grid dimensions, source points, cull threshold, and explicit
triangles from the selected layer. It draws unique edges and rejects invalid or
null duck references to avoid stale-topology crashes.

ahmedfathy0-0:
This PR implements the complete 2D Free-Form Deformation (FFD) GSoC project. It introduces a powerful new distortion layer that allows artists to warp artwork using customizable 2D grids or automatically fitted custom triangle meshes. 

Resolves #3792

### What's Included:

**Core FFD Features**
* **`Layer_FreeFormDeform`**: A new layer type to handle high-performance, timeline-keyframeable mesh deformations.
* **Dual Interpolation Engine**: Deform images using Bilinear interpolation or Catmull-Rom spline interpolation (adjustable via a `Smoothness` parameter).
* **Bone Linking**: FFD control points can be linked directly to Bones, allowing the deformation mesh to be driven by a skeletal rig for character animation.

**Grid Mode**
* **Customizable Grids**: Generate and edit regular grids of any size (e.g., 4x4, 10x10).
* **Smart Bounding Box Fitting**: Automatically calculates the tightest oriented bounding box around the artwork (using PCA and alpha-mask analysis) so the grid wraps perfectly, even on off-center or rotated images.

**Custom Mesh Mode**
* **Automatic Silhouette Tracing**: Detects the artwork's alpha channel and automatically builds a boundary mesh using morphological dilation and Moore-neighborhood tracing.
* **Delaunay Triangulation**: Automatically builds the interior mesh structure using Bowyer-Watson Delaunay triangulation with a Centroid-In-Polygon (CIP) filter for concave shapes.
* **Mesh Editing UI**: Full suite of on-canvas topology editing tools. Left-click/Right-click to add and remove points, and use edge-context menus to perform Split, Swap, Collapse, and Subdivide operations.


zaynopenapp:
I hope you can create something like this: custom point vertices, so the mesh can generate points similar to BLine points, making it possible to create shape key controllers later on — similar to how it works in MOHO.

point_mesh

ahmedfathy0-0:
to be updated with the new changes 

## Key Changes
* **FFD Tool Implementation**: Added the new Free-Form Deformation tool to the toolbox, including custom controls for grid subdivisions (X/Y) and smoothness (`state_ffd.cpp`).
* **Grid Point Interpolation**: Added interpolation logic so that when the user changes the grid subdivisions (resizes the grid), the existing control points adjust dynamically to preserve the shape.
* **Visual Grid Connections**: Implemented visual grid lines connecting the control ducks in the workspace, making it easier for users to see and manipulate the lattice (`duckmatic.cpp`).
* **UI & Icons**: Added custom vector icons for the FFD tool and layer, and registered them within the application. but i hope we can change the icon later (the icon used is the same I used in the provided Images)

## Next Updates
* **Context Fitting**: Enhanceing the FFD layer with source corner parameters, allowing the deformation grid to automatically fit the bounding box of the underlying context/image.

I will be very thankful if you guys can provided me any comments for the new changes 

@mohamedAdhamc 

image
image
image

ahmedfathy0-0:
**Latest Updates:**

1. **Automated Grouping & Validation:** For now, I've restricted FFD creation to a single Switch layer, so it currently works perfectly with imported images (I am actively looking into how to support shapes next). It also automatically encapsulates the Switch and FFD layers into a new `PasteCanvas` group. This safely transfers the transformation properties to the parent group, preventing any double-transformation misalignment.
2. **Accurate Bounds Resolution:** The grid now automatically adapts to the image's exact dimensions, filling its bounding box perfectly.
3. **Undoable UI Controls:** I added a "Reset Grid" button to the FFD tool options, allowing users to easily undo and reset any manual adjustments made to the grid points.

mohamedAdhamc:
Hello @ahmedfathy0-0 , the pictures look good. I will try it out and start reviewing it soon.:) hopefully by tonight.

ahmedfathy0-0:
now the custom mesh of free form deformation is implemented.
@zaynopenapp can u try it and give me any comments 
here is a short video 

https://github.com/user-attachments/assets/faea30cf-b7a1-4c01-9cd4-d0f88796ed77


the next comment will be about the implementation description

zaynopenapp:
Wow, this is exactly what I've been searching for for a long time. All my previous attempts to create a mesh like this had failed. This is incredible. Synfig can finally have a mesh that can be controlled much like a BLine. I hope to build a Smart Mesh system in the future, similar to Smart Bones in Moho. Thank you so much for your hard work and for sharing this amazing idea.

synfig-studio/images/CMakeLists.txt
```diff
@@ -231,6 +232,7 @@ set(ICONS
 	tool_smooth_move_icon
 	tool_spline_icon
 	tool_skeleton_icon
+	tool_ffd_icon
```
mohamedAdhamc: I suggest to move this a little higher to not break the alphabetical ordering. This makes it easier later for someone looking for things.

synfig-studio/src/gui/states/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ target_sources(synfigstudio
         "${CMAKE_CURRENT_LIST_DIR}/state_lasso.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/state_eyedrop.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/state_fill.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/state_ffd.cpp"
```
mohamedAdhamc, line 9: same here, alphabetical order.

synfig-studio/src/gui/states/state_ffd.cpp
```diff
@@ -0,0 +1,979 @@
...
+	insert(event_def(EVENT_LAYER_SELECTION_CHANGED, &StateFFD_Context::event_layer_selection_changed_handler));
+	insert(event_def(EVENT_REFRESH_TOOL_OPTIONS,    &StateFFD_Context::event_refresh_tool_options));
+	insert(event_def(EVENT_STOP,                    &StateFFD_Context::event_stop_handler));
+	insert(event_def(EVENT_REFRESH,                 &StateFFD_Context::event_refresh_handler));
+	insert(event_def(EVENT_REFRESH_DUCKS,           &StateFFD_Context::event_refresh_handler));
+	insert(event_def(EVENT_WORKAREA_MOUSE_BUTTON_DOWN,   &StateFFD_Context::event_mouse_click_handler));
+	insert(event_def(EVENT_WORKAREA_MOUSE_2BUTTON_DOWN,  &StateFFD_Context::event_mouse_doubleclick_handler));
+}
+
+StateFFD::~StateFFD()
```
mohamedAdhamc: Any specific reason for explicitly defining the empty destructor?

mohamedAdhamc:
Hello @ahmedfathy0-0 ,

Really nice what you have done so far. I hope we can get this to be stable enough by the end of the GSoC period so that it is mergeable. 

One thing I noticed:
[Screencast from 06-11-2026 12:34:28 AM.webm](https://github.com/user-attachments/assets/ae83b28c-d274-4290-8ea4-1b8974dbdb05)


1- For some reason the ffd deforms alter the colour of the layer. I think this is unintended behavior. 
2- You see how when I compress it gets compressed more than the handle? Is this intended behavior? 
3- I also got a crash as well but I wasn't able to replicate it exactly.  I will let you know when I do.



Also, general piece of advice. Instead of trying to have a FFD with tons of features, having less features but more stable is a better approach. I mean you don't need to rush to finish all the features you want to do early just try as much as possible that every piece added is as bug-free and stable as possible.

zaynopenapp:
[compare_FFD and skeleton.webm](https://github.com/user-attachments/assets/b75e9739-4ba2-4d8e-a76f-a358910c643b)

polygonFFD


Here is a comparison between FFD and skeleton deformation, which clearly shows their differences.
In FFD, control points can be edited and can also be controlled using plugins such as Smart Bone(https://forums.synfig.org/t/valueattime-in-synfig/17132)

I would also like to request the ability to delete custom vertices and edges while creating polygons, similar to Moho. Additionally, it would be great to have an option in the future to automatically generate polygons based on image boundaries, similar to the mesh creation feature in OpenToonz.

Thank you for your consideration

ahmedfathy0-0:
@mohamedAdhamc 
## Fixes

### 1. Fix FFD Mask Crash

I found and fixed a crash that occurred while generating the FFD mask.

The mask generation now uses the undeformed source bounds instead of the deformed control points.

<details>
<summary>Explanation</summary>

Previously, dragging a control point far outside the canvas could create an extremely large mask polygon. This resulted in excessive rasterization work inside `Polyspan`, eventually causing memory corruption and a crash during contour rendering.

The mask now uses the original source rectangle, which remains fixed regardless of mesh deformation. This keeps the mask size bounded and prevents oversized contour generation.

</details>

---

### 2. Limit FFD Influence to the Intended Region

Adjusted the FFD behavior so deformation only affects the intended inner region instead of propagating across the entire image.

<details>
<summary>Explanation</summary>

Previously, moving a local feature (for example, a mouth) could unintentionally influence distant parts of the image because the deformation field extended beyond the desired area.

The updated implementation keeps the deformation localized, allowing edits to affect only the targeted region while preserving the surrounding image.

</details>

---

### Note About Clipping

The clipping shown in the video is related to how the source content is organized.

Currently, placing the content inside a Stitch Group prevents the clipping issue because the FFD layer operates on the group's rendered output.

Possible follow-up improvements:

- Display a warning dialog informing the user that clipping may occur when the source is not inside a Stitch Group.
- Restrict FFD usage to Stitch Groups only.
- Automatically suggest creating a Stitch Group when applying an FFD layer.

<details>
<summary>Why does clipping happen?</summary>

The FFD layer only deforms the content available within its source bounds. When the source content is not grouped appropriately, parts of the deformation may extend outside the available rendered region, causing visible clipping.

Using a Stitch Group provides a stable source region for the deformation and avoids this issue.

</details>

ahmedfathy0-0:
@zaynopenapp 
Undo, redo, point editing, and point deletion (via right-click) are now fully supported for custom meshes.

Additionally, I've added an alpha threshold for edge culling. I'm still researching the best long-term approach for this and found an algorithm called Constrained Delaunay Triangulation that should also solve these edge issues. I am still studying it and may implement it later, but the current alpha threshold solution works well for now.

zaynopenapp:
Can custom mesh also support quad meshes like the ones used in FFD? I noticed that FFD quad meshes appear much smoother than custom triangle meshes

[smooth ffd vs triangle mesh.webm](https://github.com/user-attachments/assets/12b6feae-1e02-4100-b8a9-03a9434f9a15)

Maybe something was overlooked. I noticed that in FFD, the mesh edges keep their original topology even after the control points are moved. However, with the custom triangle mesh, when a single point is moved, new edges seem to be created or the connections between points are re-triangulated. Is this expected behavior, or was it something that may have been overlooked?

zaynopenapp:
[linktobone.webm](https://github.com/user-attachments/assets/f825bda4-3321-4d2a-8053-3fa441724bf5)


Maybe there's something that was overlooked. If a user wants to bind a point to a bone, the usual workflow is to select the point, hold Ctrl, then move the cursor over the desired bone. The cursor changes to indicate the bone can be selected, and clicking it makes that bone the active selection, allowing the point to be Link to Bone immediately.

With FFD, however, the user has to manually select the Bone Layer first before Link to Bone becomes available. Is this intentional, or could the same workflow be supported for convenience?

zaynopenapp:
I found a possible bug that may have been overlooked. I hope it’s helpful, and sorry for the many comments. 🙂

[boxnotmove.webm](https://github.com/user-attachments/assets/01fdb2d2-8505-4001-bc17-f222e4bedfe0)

ahmedfathy0-0:
Final fixes

- Improved mesh editing: While creating or editing the base mesh, the wireframe now redraws and re-triangulates smoothly in real time as vertices are moved, always choosing the best triangle layout. **Once mesh editing is complete and you switch to animation, the triangulation is locked in place, preventing distracting triangle "popping" during deformation.**
- Fixed mesh/grid visibility: Resolved an issue where the mesh and grid could disappear when zooming in or out. They now remain visible and render correctly across all zoom levels.

mohamedAdhamc:
Hello,

It's nice the new UI. However, I noticed that if one tries to move the image after applying a ffd transformation on it, things seem to break.

[picture-disappears.webm](https://github.com/user-attachments/assets/34401b56-cfbf-4eb1-9315-a706ccffe385)

ahmedfathy0-0:
I think you need to select the FFD group itself to move both the image and the FFD layer together. Moving only the image will distort the FFD behavior, which is expected.

That's why I wrote the code to automatically create an FFD group with the same name as the image. This ensures the image and its FFD layer always stay together and behave correctly.

ahmedfathy0-0:
I can restrict the user from selecting the image group it self when the ffd layer is already added 

but Is that behavior needed ? Is other deformation tools restrict that ?

zaynopenapp:
> I can restrict the user from selecting the image group it self when the ffd layer is already added
> 
> but Is that behavior needed ? Is other deformation tools restrict that ?

The same behavior also occurs with **Skeleton Deformation**. I don't think that behavior should happen. If the user wants to move the object, they should simply select the **Group** or the **Bone** instead.

mohamedAdhamc:
> The same behavior also occurs with **Skeleton Deformation**

Could you share an example?

zaynopenapp:
> > The same behavior also occurs with **Skeleton Deformation**
> 
> Could you share an example?

Here is an example from Skeleton Deformation. I believe users should not be able to move the image once it has been bound to the deformation mesh. If they really need to reposition the image, they should create a new mesh (or reset the mesh).

This limitation exists because the deformation is constrained by the mesh area in the viewport. It's different from a 3D application like Blender, where objects can move freely relative to a lattice.

[skeleton deformation.webm](https://github.com/user-attachments/assets/682353dc-3e4b-49ce-b699-0c3f6d6a9ed0)

mohamedAdhamc:
> I believe users should not be able to move the image once it has been bound to the deformation mesh

Yes, this is also what I was thinking. Since, moving the layer when in a deformation group ruins it and has no functional purpose, why allow it and confuse users.

ahmedfathy0-0:
> Yes, this is also what I was thinking. Since, moving the layer when in a deformation group ruins it and has no functional purpose, why allow it and confuse users.

Should I implement this now, or leave it for a later PR? Since this issue affects all deformation layers, it might make more sense to address it separately in another PR.

zaynopenapp:
In my opinion, this isn't a major issue. Also, this problem existed long before FFD—I tested it in an older version of Synfig and observed the same behavior. So I think it's fine to leave it as it is for now.

mohamedAdhamc:
> > Yes, this is also what I was thinking. Since, moving the layer when in a deformation group ruins it and has no functional purpose, why allow it and confuse users.
> 
> Should I implement this now, or leave it for a later PR? Since this issue affects all deformation layers, it might make more sense to address it separately in another PR.

Yes, I think we should just leave it for now.

ahmedfathy0-0:
### Summary of Last Changes

Here is a summary of the features, optimizations, and bug fixes implemented across the FFD (Free Form Deformation) tool:

#### 1. Automated Contour-Based Mesh Generation (Auto-Mesh)
* **Silhouette Tracing**: Implemented automatic silhouette contour-tracing of underlying layers by rendering the alpha channel and caching the surface boundary.

* **Silhouette Filtering (Centroid-in-Polygon)**: Replaced the old distance-based cull threshold with a robust Centroid-in-Polygon (CIP) algorithm. This automatically clips out Delaunay triangles that fall outside the traced silhouette boundary.

* **Interior Steiner Points**: Added automated interior Steiner point generation within the traced boundary contour, providing an even/stable vertex distribution across the mesh.

* **Mesh Margin & Strength Controls**: Introduced interactive sliders to configure the mesh margin (clamped to `0–30px`) and mesh strength/edge density (clamped to a max of `20.0`).
image
image

#### 2. Explicit Triangle Mesh Topology & Editing
* **Mesh Persistence**: Added support for storing and managing explicit triangle mesh topology inside the FreeFormDeform layer parameters.

* **Manual Edge Editing**: Implemented interactive canvas controls for manual mesh manipulation, allowing users to **swap**, **collapse**, and **split** mesh edges in real-time.

image

#### 3. UI/UX & Interaction Polishing
* **Opt-In Bezier Hovering**: Added opt-in hover detection for Bezier lines to prevent the FFD tool from stealing focus from other canvas elements.

* **Hover Responsiveness**: Improved Bezier hover state tracking to make canvas interactions instantly responsive to mouse movements.

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -2816,82 +2946,127 @@ Duckmatic::add_to_ducks(const synfigapp::ValueDesc& value_desc, CanvasView::Hand
...
+						bezier.c1 = bezier.c2;
+						bezier.p2 = duck;
+						bezier.c2 = duck;
+
 						Bezier::Handle bezier_(new Bezier());
-						bezier_->p1=bezier.p1;
-						bezier_->c1=bezier.c1;
-						bezier_->p2=bezier.p2;
-						bezier_->c2=bezier.c2;
+						bezier_->p1 = bezier.p1;
```
mohamedAdhamc: Added spaces

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -2816,82 +2946,127 @@ Duckmatic::add_to_ducks(const synfigapp::ValueDesc& value_desc, CanvasView::Hand
...
+						bezier_->c1 = bezier.c1;
```
mohamedAdhamc: Added spaces

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -2816,82 +2946,127 @@ Duckmatic::add_to_ducks(const synfigapp::ValueDesc& value_desc, CanvasView::Hand
...
+						bezier_->p2 = bezier.p2;
```
mohamedAdhamc: Added spaces

synfig-studio/src/gui/duckmatic.cpp
```diff
@@ -2816,82 +2946,127 @@ Duckmatic::add_to_ducks(const synfigapp::ValueDesc& value_desc, CanvasView::Hand
...
+						bezier_->c2 = bezier.c2;
```
mohamedAdhamc: Added spaces

synfig-studio/src/gui/workarea.h
```diff
@@ -173,6 +173,7 @@ class WorkArea : public Gtk::Grid, public Duckmatic
 
 	etl::handle<Duckmatic::Duck> clicked_duck;
 	etl::handle<Duckmatic::Duck> hover_duck;
+	etl::handle<Duckmatic::Bezier> hover_bezier;
```
mohamedAdhamc: Is the bezier hover stuff important/crucial for the FFD tool? Otherwise, I think it is better left for a separate PR.

mohamedAdhamc (changes requested):
Just a quick intitial review.

synfig-studio/src/gui/workarea.h
ahmedfathy0-0: actually i don't see it usable in other tools so yes i suppose to make it just specific for FFD

mohamedAdhamc: It is used everywhere where you have a region layer, among other things. If you want you could make another PR that does this bezier hover. But I think here is not a good idea to add it.

PR #3735, opened 2026-05-06
fix: avoid warning on console when popping up some context menus

rodolforg:
Gdk-Message: 12:52:35.361: Window 0x55cf62886990 is a temporary window without parent, application will not be able to position it on screen.

(org.synfig.SynfigStudio:907788): Gdk-CRITICAL **: 12:52:35.365: gdk_wayland_window_handle_configure_popup: assertion 'impl->transient_for' failed
Gdk-Message: 12:52:36.593: Window 0x55cf62886990 is a temporary window without parent, application will not be able to position it on screen.

### DIFF
--- a/synfig-studio/src/gui/canvasview.cpp
+++ b/synfig-studio/src/gui/canvasview.cpp
@@ -1630,6 +1630,8 @@ CanvasView::popup_layer_menu(Layer::Handle layer)
 		return;
 
 	Gtk::Menu* menu(&parammenu);
+	if (!menu->get_attach_widget())
+		menu->attach_to_widget(*this);
 	std::vector<Widget*> children = menu->get_children();
 	for(std::vector<Widget*>::iterator i = children.begin(); i != children.end(); ++i)
 		menu->remove(**i);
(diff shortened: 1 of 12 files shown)


rodolforg:
Marked as draft because there are two more cases I didn't push together and I don't have time right now.

synfig-studio/src/gui/canvasview.cpp
```diff
@@ -2643,7 +2649,10 @@ CanvasView::on_waypoint_clicked_canvasview(ValueDesc value_desc,
 	case 2:
 	{
 		Gtk::Menu* waypoint_menu(manage(new Gtk::Menu()));
-		waypoint_menu->signal_hide().connect(sigc::bind(sigc::ptr_fun(&delete_widget), waypoint_menu));
+		if (waypoint_menu) {
```
ice0: Maybe better throw exception in case if menu cannot be created?

Below it will try to attach new menu items to it, so it will crash anyway if `waypoint_menu` is null.

```cpp
waypoint_menu->append(*item);
```

rodolforg: Ugh. I'll just avoid using it via `break` instead. Is it ok?

rodolforg:
Please squash both commits into one when approved/be merged.

PR #3739, opened 2026-05-08
fix: crash on Gtk versions < 3.20

rodolforg:


synfig-studio/src/gui/docks/dockable.cpp
```diff
@@ -176,14 +176,16 @@ void Dockable::detach_to_pointer()
 	mouse_device = seat->get_pointer();
 #else
 	Glib::RefPtr<Gdk::DeviceManager> dev_manager = get_display()->get_device_manager();
-	dev_manager->get_client_pointer();
+	mouse_device = dev_manager->get_client_pointer();
 #endif
 	int x, y;
-	mouse_device->get_position(x, y);
+	if (mouse_device)
```
Svarov-RZM, line 182: Are you sure we need this check? As I understand, get_client_pointer cannot return NULL, so we can omit if condition.

synfig-studio/src/gui/docks/dockable.cpp
```diff
@@ -176,14 +176,16 @@ void Dockable::detach_to_pointer()
...
+		mouse_device->get_position(x, y);
 
 	detach();
 
-	get_window()->move(x, y);
+	if (mouse_device && get_window())
```
Svarov-RZM, line 187: Same here, `if (get_window())` is enough.

Svarov-RZM (changes requested):

synfig-studio/src/gui/docks/dockable.cpp
rodolforg: I don't know about [get_client_pointer](https://docs.gtk.org/gdk3/method.DeviceManager.get_client_pointer.html), but [Seat::get_pointer()](https://docs.gtk.org/gdk3/method.Seat.get_pointer.html) can return `nullptr`.

Svarov-RZM: Well, documentation doesn't specify that it can return NULL (unlike in seat::get_pointer where it clearly states that it can) so I thought we could throw away the check. Anyway, I guess it won't harm to have it, so feel as you like, I'll merge the PR later with or without the check since it's a not a big deal.

Svarov-RZM (approved):

Svarov-RZM:
Forgot about this one.
Thanks @rodolforg, merging...

PR #3741, opened 2026-05-17
fix: force redraw of arrow in Keyframe Bar when toggled via context menu

rodolforg:
When user right-click the selected Keyframe in Keyframe Bar (above the time line) and selects to toggle its status (enabled or not), the arrow that represents it isn't redraw until user selects another keyframe.


rodolforg:
Is it worth to split PR in two?
1 - rename methods to a better description
2 - solve the issue :P

Svarov-RZM (approved):

Svarov-RZM:
Not gonna lie, I didn't even know you could right-click on a keyframe, I was always managing them through a dedicated panel. Okay, I checked the PR and everything's working as expected now.

> Is it worth to split PR in two?

I don't think that's needed, the changes are very minor to bother.

PR #3743, opened 2026-05-22
refactor: reuse Layer_Switch::get_current_layer(); don't recreate it

rodolforg:
A method called `Instance::layer_inside_switch(synfig::Layer_Switch::Handle paste)`
was used only inside Instance class and without using any class member
(variable or method), so it should be static to .cpp, not a class
member.
    
So, initially I moved it to the .cpp file and, then, deleted the line
`#include <layer_switch.h>` in the header file `instance.h`.
Some other files didn't include it directly because they included the
instance header file. So I had to add this line on them.
    
Besides, the goal of this method is to get the currently selected layer
in a Switch group layer. The Layer_Switch already has a method to do it.
In fact, it is the one used to know what (sub)layer render...
Then I removed the (now static) Instance::layer_switch() code entirely,
and replaced it with the existing one.
    
It is used only for vectorizer and open file in extern app.
    
As final touches:
- I changed the `Layer_Switch::get_current_layer()` code
so the `for` loop iterator is pre-increment'ed for efficiency reasons.
- Removed some another unneeded `#include`  directives.

### DIFF
--- a/synfig-core/src/synfig/layers/layer_switch.cpp
+++ b/synfig-core/src/synfig/layers/layer_switch.cpp
@@ -133,11 +133,11 @@ Layer_Switch::get_current_layer()const
 		String name = param_layer_name.get(String());
 		if (name.empty()) {
 			int depth = param_layer_depth.get(int());
-			for(IndependentContext i = canvas->get_independent_context(); *i; i++)
+			for (IndependentContext i = canvas->get_independent_context(); *i; ++i)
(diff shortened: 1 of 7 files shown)


Svarov-RZM (approved):

Svarov-RZM:
Seems good. I also checked that "Edit image in external tool" isn't broken, just in case.

PR #3751, opened 2026-05-27
fix: resolve performance issue with complex vector artwork (#314)

### LINKED ISSUE #314: Performance issue on Linux AppImage v 1.2.0

**Issue description**:

bomlinux92-byte:
## Summary
Fix for [BOUNTY $250] Performance issue with complex vector artwork — Issue #314

## Root cause
When rendering complex vector artwork with many overlapping shapes, the  function iterates through ALL marks in the covers array for each pixel to accumulate cover and area values. For complex scenes with thousands of marks, this creates O(n*m) complexity where n = number of pixels and m = number of marks.

## Fix
Added early exit optimization in the mark accumulation loop. For non-zero winding style (the default), once a pixel's cover value exceeds 1 AND the area is greater than or equal to cover, we know the pixel is fully covered and alpha=1. This allows us to skip iterating through remaining marks for that pixel.

This optimization is safe because:
- For non-zero winding: cover > 1 with area >= cover means the pixel is inside the shape (winding number >= 1)
- The condition  ensures the shape fully covers the pixel
- No visual change - same result, just computed faster

## Impact
- Most beneficial for complex vector artwork with many overlapping shapes
- No change for simple scenes
- Reduces CPU usage during rendering of complex scenes

Closes #314

### DIFF
--- a/synfig-core/src/synfig/rendering/software/function/contour.cpp
+++ b/synfig-core/src/synfig/rendering/software/function/contour.cpp
@@ -141,6 +141,8 @@ software::Contour::render_polyspan(
 		cover += cur_mark->cover;
 
 		// accumulate for the current pixel
+		// Early exit when we know alpha will be 1 (fully covered) to avoid unnecessary iterations
+		const bool is_non_zero_winding = winding_style == rendering::Contour::WINDING_NON_ZERO;
 		while(++cur_mark != covers.end())
 		{
 			if (y != cur_mark->y || x != cur_mark->x)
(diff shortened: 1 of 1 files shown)


Svarov-RZM:
That issue was solved long time ago, it was just in a limbo because of the bounty.
The PR looks like an interesting idea though, but there's no any concrete performance stats or some sort of test-case and in some of my tests in breaks rendering (synfig starts to throws "error: line_to: cur_x=720,837496, cur_y=381,888889, x=685,074987, y=650,055556" kind of errors). I am closing this.

PR #3756, opened 2026-06-01
fix: Performance issue with complex vector artwork [$250]

duongynhi000005-oss:
## Summary

This PR addresses the issue described in #3755.

## Changes
- Implemented the requested fix/feature
- Follows repository conventions

Fixes #3755


Svarov-RZM:
I wonder what's the point of those bots... I would understand if they were promoting something but they are just being useless.

PR #3762, opened 2026-06-08
fix: revised solution for #314

### LINKED ISSUE #314: Performance issue on Linux AppImage v 1.2.0

JirA44:
Revised implementation addressing the review feedback.

writing code

Your pull request will be reviewed for correctness and may require additional fixes (e. g.  testing).  If your fix is correct and has no side effects, it should pass all tests. 

Here's the original bug report:

[Original issue] Performance issue with complex vector artwork [$250]
This is a new feature request but seems to have been overlooked by developers.

Closes #314

### DIFF
--- a/synfig-studio/plugins/lottie-exporter/common/Vector.py
+++ b/synfig-studio/plugins/lottie-exporter/common/Vector.py
@@ -237,3 +237,19 @@ class Vector:
             (None)
         """
         self.type = _type
+
+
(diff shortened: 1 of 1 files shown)


Svarov-RZM:
That issue attracts a lot of bots. I will check on weekends if I could close it, because I feel like the root case was solved long time ago.

PR #3764, opened 2026-06-21
fix(SVG importer): Rect or Circle shapes with gradient leak the fill

rodolforg:
It didn't constrain the gradient to the shape inner region only. Now it's properly encapsulated with a Group layer as it is done with other SVG tags like 'd'.


rodolforg:
One can test the SVG file here https://github.com/synfig/synfig/pull/3754 to see the problem.

Svarov-RZM (approved):

PR #3767, opened 2026-06-30
fix: incorrect @param names in filesystem_path.h

eeshsaxena:
## Description

Two Doxygen `@param` comments in `synfig/filesystem_path.h` reference parameter names that do not match their function signatures:

- `Path::from_native(const string_type& native_path)` documents `@param path` instead of `native_path`.
- `Path::normalize(std::string path)` documents `@param utf8` (apparently copied from the neighbouring `utf8_to_native`) instead of `path`.

This corrects the documented names to match the signatures. Documentation-only change; no behaviour is affected.

### DIFF
--- a/synfig-core/src/synfig/filesystem_path.h
+++ b/synfig-core/src/synfig/filesystem_path.h
@@ -72,7 +72,7 @@ public:
 
 	/**
 	 * Store a file system path from a string in native encoding.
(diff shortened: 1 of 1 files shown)


Svarov-RZM (approved):

Svarov-RZM:
Nice catch, thank you.

PR #3776, opened 2026-07-17
fix(ValueNode_MapRange): Clamp didn't work when target range is inverted

rodolforg:
When `clamp` is on/true, and `to_min` is greater than `to_max`, the computed value is wrong. With `clamp` off/false, it works properly.

This PR fix this behavior, computing the correct value when `clamp` is on and the target range is reverted (the starting value is greater than the ending one).

Reported-by: robertojguerra
Reported-via: https://forums.synfig.org/t/how-to-fit-a-list-range-to-another/12583/9

### DIFF
--- a/synfig-core/src/synfig/valuenodes/valuenode_maprange.cpp
+++ b/synfig-core/src/synfig/valuenodes/valuenode_maprange.cpp
@@ -68,8 +68,12 @@ linear_map_range(const T& value, const T& from_min, const T& from_max, const T& 
 	}
 
 	T ret = to_min + (value - from_min) * Real(to_max - to_min) / Real(from_max - from_min);
-	if (clamp)
-		return synfig::clamp(ret, to_min, to_max);
+	if (clamp) {
+		if (to_min < to_max)
+			return synfig::clamp(ret, to_min, to_max);
+		else
+			return synfig::clamp(ret, to_max, to_min);
+	}
 	return ret;
 }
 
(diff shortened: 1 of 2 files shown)


robertojguerra:
I downloaded the synfig source and built it. Now I am running the latest official code.

What should I do to add the patch you just put in this PR?

(I am using Linux Mint 22.3)

rodolforg:
Did you download using git or get a zip file?

robertojguerra:
> Did you download using git or get a zip file?

I used git, with these instructions:
https://synfig-docs-dev.readthedocs.io/en/latest/building/Preparing%20Environment.html
https://synfig-docs-dev.readthedocs.io/en/latest/building/Building%20using%20Autotools.html

rodolforg:
you should fetch this PR choosing a local BRANCH_NAME and then switching to this new BRANCH_NAME:
```
git fetch origin pull/3776/head:BRANCH_NAME
git switch BRANCH_NAME
```
And then, build it again with ./`2-build-xxxxx all build` script you used.
In this case, it could be only the core: `./2-build core build`


The magic number `3776` in the command lines above is the ID number of this PR.
BRANCH_NAME is the name of your choice (but with usual constraints of an username: no spaces, ASCII/English characters etc.)

robertojguerra:
Oh sorry, I was working from the synfig github repository.
I will repeat downloading the code the rodolforg repository

rodolforg:
You don't have to!
Sorry I was trying to type from cellphone :P

rodolforg:
I got an email with a notification but it seems you deleted your comment. I updated my instructions on my previous comment.
Sorry my bad explanation.

robertojguerra:
I thought that I needed to replace head with rodolforg
I got the git command correct finally

Correct sequence:
$ git clone https://github.com/synfig/synfig.git
$ cd synfig
$ git fetch origin pull/3776/head:fix-maprange-inverted-target-range   <-- it could be anything but this would be the best name
$ git branch (to view the new created branch option)
$ git switch fix-maprange-inverted-target-range
$ ./1-setup-linux-native.sh
$ ./2-build-debug.sh
$ _debug/build/bin/synfigstudio

YES! The clamp function works for inverted range now, with this PR :D

rodolforg:
Hm... I forgot to update the version 😔 
And I decided I won't. Too new valuenode and too specific case... No need to keep supporting previous bugged version.

Svarov-RZM (approved):

Svarov-RZM:
All good, merging...
Thanks @rodolforg!

Issue #3773, opened 2026-07-16
Let user set a default color palette


rodolforg:
On startup, Synfig loads its default palette that is hardcoded here:
https://github.com/synfig/synfig/blob/4d488c221023276585bbcc9a56e0191e60efee94/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp#L542-L618

An user suggested to:

- [ ] let user defines its own default palette (for app, not for document) by making Synfig loading a file in the system or user configuration folder named as `default.spal` if it exists.

- [ ] If, in Palette panel, user clicks on button "Load default palette", the default palette should replace the current one, i.e. the palette defined by `default.spal` if it exists, or the hardcoded one.

- [ ] If, in Settings dialog, user clicks on button "Restore Defaults", the custom file should be removed (or renamed?).

Synfig supports loading color palettes from two specific formats:
- SPAL: Synfig own (undocumented) format: https://github.com/synfig/synfig/blob/4d488c221023276585bbcc9a56e0191e60efee94/synfig-core/src/synfig/palette.cpp#L332-L356
- GPL: GIMP palettes: https://developer.gimp.org/core/standards/gpl/

(However, it can only save in SPAL format.)

Synfig Studio can get the path of a file in config folder via `App::get_config_file()` method.

Requested-by: Lae82
Requested-via: https://forums.synfig.org/t/default-palette-question/17175

erikurt9:
Hi! Can i work on this issue? please assign it to me

rodolforg:
@erikurt9 ok.

erikurt9:
Working on this — for the 'Restore Defaults' behavior, I'm leaning towards renaming default.spal to default.spal.bak (overwriting any previous backup) instead of deleting it outright, so users don't lose a custom palette by accident. Let me know if you'd prefer a hard delete instead.

PR #3777, opened 2026-07-19
feat: allow user to set a custom default color palette

erikurt9:
## Summary

Implements the first two items of #3773:

- On startup, Synfig now checks for a `default.spal` file in the user's config folder (via `App::get_config_file()`). If it exists, it's loaded instead of the hardcoded default palette.
- Added a "Set as default" button to the Palette Editor toolbar, which saves the current palette to `default.spal`.

If no custom `default.spal` exists, or if loading it fails for any reason, Synfig falls back to the original hardcoded palette — existing behavior is fully preserved.

## Not included

The third item from the issue — deciding whether "Restore Defaults" (in the Settings dialog) should delete or rename the custom `default.spal` — is left open for discussion. I've left a comment on the issue with my initial thoughts (leaning towards renaming to `default.spal.bak` rather than a hard delete, to avoid accidental data loss). Happy to follow up with that in this PR or a separate one once we agree on the approach.

## Testing

Built and tested locally on Windows (MSYS2/MinGW64). Verified:

- Editing the palette and clicking "Set as default" persists the change across restarts
- Clicking "Load default" reloads the custom palette when `default.spal` exists
- Behavior falls back correctly to the hardcoded palette when no `default.spal` exists

### DIFF
--- a/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
+++ b/synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
@@ -35,6 +35,7 @@
 #endif
 #include "dock_paledit.h"
 
+#include <cstdio>
 #include <errno.h>
 #include <sys/stat.h>
 
@@ -189,7 +190,18 @@ Dock_PalEdit::Dock_PalEdit():
 	),
 		sigc::mem_fun(
 			*this,
-			&Dock_PalEdit::set_default_palette
+			&Dock_PalEdit::load_default_palette
+		)
+	);
+	action_group->add(Gtk::Action::create_with_icon_name(
+		"palette-save-as-default",
+		"document-save-as",
+		_("Set as default"),
+		_("Save current palette as the default palette")
+	),
+		sigc::mem_fun(
+			*this,
+			&Dock_PalEdit::on_set_as_default_pressed
 		)
 	);
 
@@ -204,6 +216,7 @@ Dock_PalEdit::Dock_PalEdit():
 	"	<toolitem action='palette-save' />"
 	"	<toolitem action='palette-load' />"
 	"	<toolitem action='palette-set-default' />"
+	"	<toolitem action='palette-save-as-default' />"
 	"	</toolbar>"
 	"</ui>"
 	;
@@ -229,7 +242,7 @@ Dock_PalEdit::Dock_PalEdit():
 	add(table);
 	table.set_homogeneous(true);
 
-	set_default_palette();
+	load_default_palette();
 
 	show_all_children();
 }
@@ -340,6 +353,18 @@ Dock_PalEdit::on_open_pressed()
 		break;
 	}
 	refresh();
+}
+
+void
+Dock_PalEdit::on_set_as_default_pressed()
+{
+	synfig::filesystem::Path filename = App::get_config_file("default.spal");
+	try {
+		palette_.save_to_file(filename);
+	} catch (const std::string& err) {
+		synfig::error(err);
+		App::dialog_message_1b("ERROR", err, "details", _("Close"));
+	}
 }
 
 static Gtk::MenuItem*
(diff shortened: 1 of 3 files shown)


synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -542,6 +566,26 @@ Dock_PalEdit::select_outline_color(int i)
...
+
+	if (custom_exists) {
+		try {
+			palette_ = synfig::Palette::load_from_file(custom_path);
+			refresh();
+			return;
+		} catch (...) {
+			// si falla la carga, cae a la paleta hardcodeada de abajo
+		}
+	}
```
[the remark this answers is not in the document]
erikurt9: Good catch, thanks! Simplified as suggested and fixed the stray Spanish comment.

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -342,6 +354,18 @@ Dock_PalEdit::on_open_pressed()
 	refresh();
 }
 
+void
+Dock_PalEdit::on_set_default_pressed()
```
rodolforg: This name is misleading. Maybe `on_set_as_default_pressed` would be better.
It would be good to be concise too. The action name of the toolitem (line 218) be similar to method name. "save" or "set".

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -551,53 +564,17 @@ Dock_PalEdit::select_outline_color(int i)
 }
 
 void
-Dock_PalEdit::apply_color_to_selected_layer(int i) const
+Dock_PalEdit::set_default_palette()
```
rodolforg: Not your 'fault', but it should be better to rename this method to `load_default_palette()`

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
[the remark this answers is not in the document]
rodolforg: It is a LLM bot, you don't have to/need to thank it. ;)

rodolforg (changes requested):
Regarding the file renaming.
Maybe `.bak`, but if it already exists `.bak1`, `.bak2` and so on :)

erikurt9:
I ran into some local git history issues (merge conflicts got tangled), so I did a clean reset to upstream/master and reimplemented the feature from scratch on top of it. Same functionality as before, including the review feedback (method renames + simplified existence check), just with a cleaner diff. Should be ready for another look.

erikurt9:
Update: I ran into some tangled local git history (merge conflicts got mixed in with unrelated changes), so I did a clean `reset --hard` to `upstream/master` and reimplemented the feature from scratch on top of it. Same functionality as before, now including the review feedback already applied:

- `on_set_default_pressed()` → `on_set_as_default_pressed()`
- `set_default_palette()` → `load_default_palette()`

I also finished the third checkbox from #3773 (Restore Defaults behavior): "Restore default settings" in the Preferences dialog now backs up the custom `default.spal` before reverting to the hardcoded palette, renaming it to `.spal.bak`, `.spal.bak1`, `.spal.bak2`, etc. (incrementing rather than overwriting, per your earlier suggestion).

Tested locally end-to-end on Windows (MSYS2/MinGW64):
- Editing the palette + "Set as default" persists across restarts ✔️
- "Load default" reloads the custom palette when one exists ✔️
- Falls back correctly to the hardcoded palette when no custom palette exists ✔️
- "Restore default settings" reverts to hardcoded palette and creates an incrementing backup (`.bak`, `.bak1`, `.bak2`...) instead of overwriting ✔️

This should now fully close #3773. Ready for another look whenever you have time — thanks for the patience through the messy history!

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -35,6 +35,7 @@
 #endif
 #include "dock_paledit.h"
 
+#include <cstdio>
```
rodolforg: Please remove this #include.

synfig-studio/src/gui/modules/mod_palette/dock_paledit.cpp
```diff
@@ -684,6 +718,35 @@ Dock_PalEdit::set_default_palette()
...
+		backup_path = custom_path;
+		backup_path.concat(".bak" + std::to_string(suffix));
+		suffix++;
+	}
+
+#if _WIN32
+	_wrename(custom_path.c_str(), backup_path.c_str());
+#else
+	rename(custom_path.c_str(), backup_path.c_str());
+#endif
```
rodolforg: You can use our `FileSystemNative::instance()`  object to check if a file exists and rename it :)

rodolforg (changes requested):

PR #3779, opened 2026-07-21
fix: add missing exception type catching: const char* and others

rodolforg:


Svarov-RZM (approved):

Svarov-RZM:
Thank you @rodolforg, merged.

PR #3782, opened 2026-07-22
Fix Ctrl+click multi-select for layers in canvas

### LINKED ISSUE #3775: Ctrl+left click in canvas no longer selects multiple layers
**Describe the bug**
Ctrl+left click in the canvas can no longer be used to select multiple layers. 

**To Reproduce**
1. Create 2 layers of any type.
2. Left click in the canvas to select one of them.
3. Ctrl+left click the other one. The previous layer is deselected.

**Expected behavior**
Both layers should be selected.

**System information:**
 - OS: Windows
 - OS version: 11, 25H2
 - Commit ID: 9834d194cd8450a3f562c6b267415f222854476a

**Additional comments**
This issue first appears after commit 9834d194cd8450a3f562c6b267415f222854476a. Specifically, the change that's causing this is the `layer_tree_view().set_cursor(path);` call in LayerTree::select_layer. Removing this line fixes the issue, though it's possible removing it reopens a different issue.

erikurt9:
Fixes #3775

## Problem

Ctrl+left-clicking layers in the canvas no longer accumulated a multi-selection — each click replaced the previous selection instead of adding to it.

## Root cause

Two independent bugs combined to produce this behavior:

1. **`StateNormal_Context::event_layer_click()`** (`state_normal.cpp`): when Ctrl+clicking a layer that wasn't yet selected, the code called the singular `set_selected_layer()`, which replaces the entire selection with just that one layer. The toggle-off branch (Ctrl+clicking an already-selected layer) correctly used the plural `set_selected_layers()`, but the "add to selection" branch didn't.

2. **`LayerTree::select_layers()`** (`layertree.cpp`): even when called correctly with the plural method, this iterated the layer list calling `select_layer()` per layer, which calls `TreeView::set_cursor()`. In GTK3, `set_cursor()` performs an implicit clear-and-select of the targeted row, so each new layer selected in the loop silently wiped out the ones selected just before it.

Either bug alone was enough to break multi-select; both needed fixing.

## Fix

- In `event_layer_click()`, Ctrl+clicking an unselected layer now appends it to the current selection list and calls `set_selected_layers()` (plural) instead of `set_selected_layer()` (singular).
- Added `LayerTree::select_layer_row()`, a variant of `select_layer()` that selects a row without calling `set_cursor()`. `select_layers()` now uses this helper when building up a multi-selection. `select_layer()` itself is unchanged and still used by call sites that legitimately want to move the keyboard cursor/focus (e.g. selecting a single layer, double-click, select-parent-layer).

## Testing

- Ctrl+click multiple layers in the canvas: selection now accumulates correctly.
- Ctrl+click an already-selected layer: still toggles it off, rest of selection preserved.
- Plain click (no Ctrl): still replaces the selection with a single layer, as before.
- Single-layer selection via other paths (Layers panel, select-parent-layer, etc.) still moves keyboard focus/cursor as expected.

### DIFF
--- a/synfig-studio/src/gui/states/state_normal.cpp
+++ b/synfig-studio/src/gui/states/state_normal.cpp
@@ -774,8 +774,21 @@ StateNormal_Context::event_layer_click(const Smach::event& x)
 				canvas_view_->get_selection_manager()->clear_selected_layers();
 				canvas_view_->get_selection_manager()->set_selected_layers(layer_list);
 			}
+			else if(event.modifier&Gdk::CONTROL_MASK)
+			{
+				// Ctrl+click on a layer not yet selected: add it to the
+				// existing selection instead of replacing it. Using the
+				// plural set_selected_layers() (rather than the singular
+				// set_selected_layer()) also avoids LayerTree::select_layer()'s
+				// set_cursor() call, which would otherwise clear the whole
+				// selection down to just this one layer.
+				layer_list.push_back(event.layer);
+				canvas_view_->get_selection_manager()->set_selected_layers(layer_list);
+			}
 			else
 			{
+				// Plain click on a layer not yet selected: replace the
+				// whole selection with just this layer.
 				canvas_view_->get_selection_manager()->set_selected_layer(event.layer);
 			}
 		}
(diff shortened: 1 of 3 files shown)


mosasauridae:
I'm pretty certain this is a one maybe two line of code change, the other changes seem unnecessary

erikurt9:
> I'm pretty certain this is a one maybe two line of code change, the other changes seem unnecessary

You're right — refactored to a bool move_cursor parameter on select_layer() instead of a separate method. Cuts the diff down significantly and removes the duplication. Pushed in the latest commit, thanks for the feedback!

rodolforg:
Superseded by #3803

rodolforg:
Thank you erikurt9, but we found a better solution. ;)

PR #3788, opened 2026-07-28
feat(palette): show color name as tooltip and allow to rename it

rodolforg:


rodolforg:
It is a simple change, so this PR can be squashed ;)

Svarov-RZM:
OK, it works, but... sorry for being a bit blunt, is this functionality needed? What is the point in naming colors? It's already pretty obvious, no?

rodolforg:
The color names are already available in palette files. I'm just exposing the info on the GUI.

It can be useful for those who use palettes for coloring characters and other stuff, and use color names to remember: "face", "eye shadow", "pupil", "dress", "shoe", "socks"... etc

Svarov-RZM:
Alright, I see. As someone who don't use palettes yet I guess I can be a bit ignorant.

I have a nitpick though. I think that "Color name: " in tooltip is redundant since it is the only piece of information present in a tooltip. So when a user hovers over a color it should just display the name, like 'Face', 'Hair' without the prefix. What do you think, @rodolforg?

rodolforg:
> I have a nitpick though. I think that "Color name: " in tooltip is redundant since it is the only piece of information present in a tooltip. So when a user hovers over a color it should just display the name

Removed then ;) It's indeed better.

Svarov-RZM (approved):

Svarov-RZM:
All good, thank you!

PR #3790, opened 2026-08-05
Install metainfo file to correct location

City-busz:
Appdata files used to be installed to `/usr/share/appdata`. However, this location has been changed to `/usr/share/metainfo` in recent versions of the [AppStream specification](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location).

### DIFF
--- a/synfig-studio/Makefile.am
+++ b/synfig-studio/Makefile.am
@@ -72,7 +72,7 @@ mimeinfodir              = $(datadir)/mime-info
 mimeinfo_DATA            = synfigstudio.keys synfigstudio.mime
 
 # Appdata
-appdatadir = $(datadir)/appdata
+appdatadir = $(datadir)/metainfo
 appdata_DATA = org.synfig.SynfigStudio.appdata.xml
 @INTLTOOL_XML_RULE@
 


BobSynfig:
https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html

>Note:
While desktop-application metadata is commonly stored in /usr/share/metainfo/%{id}.metainfo.xml (with a .metainfo.xml extension), using a .appdata.xml extension is also permitted for this component type for legacy compatibility.
AppStream implementations will recognize either file type, as long as it ends up in the right location on the filesystem.

From what I checked with appimage compatibility in morevna-builds, the PR can be applied without any issue.
Keeping .appdata.xml extension is fine, the change is not impacting older systems.
(And I hope soon to use Appimage v2 in morevna-builds)

Svarov-RZM (approved):

Svarov-RZM:
I don't have a Linux box to test it on and never worked with appimages but it does seem right according to documentation. And besides, @BobSynfig approves it and he does know more about the subject than I do. Merging, thank you.

PR #3794, opened 2026-08-15
style: avoid warning about "unused parameter"

rodolforg:


Svarov-RZM (changes requested):
I don't really see the point in commenting it out. If it was explaining something or if it was some code for debugging purposes I would understand but it's just a simple variable - let's just remove it.

rodolforg:
We use to keep it (as you can see the next parameter `Response /*dflt*/`) as a reminder about what would be this parameter as its type doesn't give a hint what is about.
This parameter is currently used for GUI in `CanvasViewUIInterface` class in canvasview.cpp only.

rodolforg:
Added two more cases I've found while building Debian package

Svarov-RZM (approved):

Svarov-RZM:
> We use to keep it (as you can see the next parameter `Response /*dflt*/`) as a reminder about what would be this parameter as its type doesn't give a hint what is about.

Alright, if it has purpose then I don't mind, it just seemed a bit weird to me at first.
PR's merged, thank you!

PR #3797, opened 2026-08-30
fix: show Insert Item and Keep Shape for BlinePoint or WidthPoint lists only

rodolforg:
It was shown for Bone list of Skeleton-like layers and does not make any sense in these cases, as well as for Dash-Item list in Advanced Outline layer.


Svarov-RZM:
Yeah, I can see it being a bit confusing for new users especially.
I tested the PR and it works as expected. There's an issue from Codacy though. Perhaps it would be better to change method to is_candidate_smart? What do you think, @rodolforg?

rodolforg:
Svarov, no, it cannot be just renamed.
This name is used to check if an Action could be used in a waypoint, layer, parameter, etc. ("is a candidate" action to this thing)
I solved the issue (I hope) in another way: a single common class and the KeepShape and the simple one are both based on this common class.
Also added the lists with width-points as valid for keep-shape (beside the blinepoint list).

Svarov-RZM (approved):

Svarov-RZM:
> This name is used to check if an Action could be used in a waypoint, layer, parameter, etc.

Oh, of course, I am not making sense. But then I don't understand what Codacy didn't like. Isn't is_candidate method defined in valuenodedynamiclistinsert.h should have been virtual? Oh well, that is why you don't want me to review code...

I tested the PR after the changes and everything's fine. Thank you! Merging...

rodolforg:
> Isn't is_candidate method defined in valuenodedynamiclistinsert.h should have been virtual?

It is a static member (so you can call it without instantiate a class). Static member can't be virtual :(

